PR #159, opened 2015-02-26
Per test results on submissions tab

IlyaKornakov:
Per test results in submission list. Very preliminary (and trivial) version, displays smallish per-test results on "submissions" jury tab.

### DIFF
--- a/lib/www/common.php
+++ b/lib/www/common.php
@@ -57,7 +57,7 @@ function parseRunDiff($difftext){
  * match <key> = <value>. Output is always limited to the
  * current or last contest.
  */
-function putSubmissions($cdatas, $restrictions, $limit = 0, $highlight = null)
+function putSubmissions($cdatas, $restrictions, $limit = 0, $highlight = null, $testcases = false)
 {
 	global $DB, $username;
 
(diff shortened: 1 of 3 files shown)


eldering:
Hi Ilya,

The code looks very clean. Only a few minor comments:
- In lib/www/common.php on line 115, in the SQL query, the field
'j.judgingid' should be added for later use.
- In lib/www/common.php on line 276, '$test' should be '$text'.

If you're changing this, could you create a new single commit, to remove
the clutter from the merges and undo's? (Sorry that we were just merging
changes from the master branch into there that conflicted with your code.)

Best,
Jaap

eldering:
And I forgot: you may (later) want to restyle/resize things a bit,
because at the WFs, there are typically about 50 testcases per problem.

lib/www/common.php
```diff
@@ -253,6 +254,37 @@ function putSubmissions($cdatas, $restrictions, $limit = 0, $highlight = null)
...
+			if ( $testcases ) {
+				$judgingid = $row['judgingid'];
+				$probid = $row['probid'];
+				$runs = $DB->q('SELECT r.runresult, t.rank
+						FROM testcase t
+						LEFT JOIN judging_run r ON ( r.testcaseid = t.testcaseid
+							AND r.judgingid = %i )
+						WHERE t.probid = %i ORDER BY rank',
+						$judgingid, $probid);
+				$runinfo = $runs->gettable();
```
thijskh: you could use $DB->q('TABLE SELECT' above instead of SELECT + gettable (a feature of our db lib)

IlyaKornakov:
Hi,

Thanks for the comments!

As for test/text and j.judgingid - fixed. Is there any way to test changes in the branch locally? I've ended up applying the same modifications to source code and installed copy of domjudge, as a result, getting these typos.

As for merging commits into a single one: I'm not familiar with github workflow, but I've tried to do it. Hopefully it worked correctly (I've ended up with rebase + push -f).

As for TABLE SELECT - changed.

Regarding restyling: I've already changed it to use custom style with smaller boxes. I'm afraid making them even smaller would make these results unreadable. This shouldn't break UI for 50 testcases, since corresponding table cell has max-width, and extending it would just make testcase results multiline (which also isn't very good, but at least UI isn't broken).

eldering:
Thanks, merged. Or actually: manually rebased and pushed on top of
current ICPC-live branch to keep history simple.

PR #170, opened 2015-07-13
Fix a bug that "solved" value is always true.

nya3jp:
Currently, scoreboard API says every team solved all problems.


thijskh:
Great catch, thanks!

PR #171, opened 2015-07-15
Make APIs return correctly typed values.

nya3jp:
Since our DB returns all retrieved values as strings, our API
also returns string values for most of entries. It is okay for
PHP clients because of weak typing, but accessing from other
strongly typed languages is difficult and unpredictable.

### DIFF
--- a/lib/www/scoreboard.php
+++ b/lib/www/scoreboard.php
@@ -134,7 +134,7 @@ function genScoreBoard($cdata, $jury = FALSE, $filter = NULL) {
 			// provide default scores when nothing submitted for this team,problem yet
 			if ( ! isset ( $MATRIX[$team][$prob] ) ) {
 				$MATRIX[$team][$prob] = array('num_submissions' => 0, 'num_pending' => 0,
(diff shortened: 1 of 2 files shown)


thijskh:
Yes, useful :+1:

eldering:
Looks useful and good code to me.
Just some minor suggestions, see: https://github.com/eldering/domjudge/commits/typed-api
In particular, I think the variable $jid returned from the SQL query in judgings_POST() still needs to be cast to int, right?

nya3jp:
Thanks Jaap for reviewing!

I merged your update to my pull request. And about $jid, since $jid is used only to fill $row, it is okay to leave as a string. But maybe:

     $row['judgingid']   = safe_int($jid);

is more clearer as I did in the latest patch?

eldering:
I overlooked the line
$row['judgingid']   = safe_int($row['judgingid']);
in your original patch. Sorry for the confusion.

Merged it.

nya3jp:
Thanks!

Issue #173, opened 2015-09-17
MySQL improved unicode support


bertptrs:
Currently, the database driver uses `utf8_general_ci` as the collation and `utf8` as the character set. This character set does not support characters more than two bytes wide. Since teams generally attempt to use weird characters in their name to show they know they exist, the newer `utf8mb4` could be used.

You can change the database on your own and then change the `DJ_CHARACTER_SET_MYSQL` to the proper value, but in exactly two queries a `COLLATE` statement is used with a hard coded `utf8_general_ci` collation.

So the issue is, could this be changed to a more configurable option, with the default being `utf8mb4`?

thijskh:
Yes, agreed. It would be nice if it's indeed configurable with that default. We can change the recommended version of MySQL to 5.5, while still supporting earlier MySQL's (the user would just need to change that config parameter).

Would you mind making a pull request?

bertptrs:
Sure, I just have to clean up what I've done to my own installation and make a patch.

eldering:
A few remarks/questions:
- Are we actually trying to support arbitrary character sets or is this just a transition to proper 4-byte UTF-8 support? I'd vote for the latter.
- I think the specific "COLLATE utf8_general_ci" should be replaced by "utf8_unicode_ci" (or the utfmb4 version), since utf8_general_ci is not precisely adhering to the UTF-8 collation rules.
- There should either be a configure time option to set this, or we should drop support for MySQL < 5.5.3 because with the proposed patch, things will break for older versions.

bertptrs:
To answer a your questions:

* I think we should improve the UTF-8 support, and not arbitrary character sets. MySQL's utf8 character set unfortunately is a mistake and only implements a small subset of utf8.
* You may be correct, but I don't think it will make that much of a difference in sorting, at least not for characters for which users expect an odering.
* You are right. I'm not sure on which platforms Domjudge usually runs, but Ubuntu ships with at least 5.5.44 for all supported versions, and I'm sure CentOS can upgrade to it as well. A version bump may have very little compatibility issues.

eldering:
Ok, RHEL6, Debian wheezy (oldstable) and Ubuntu 14.04 (LTS) have MySQL 5.5.44, so then I'd say we simply drop support for older stuff and get rid of the cruft.

thijskh:
:+1:

thijskh:
[...]

PR #174, opened 2015-09-18
Improve unicode handling

bertptrs:
This pull requests changes the database to use the `utf8mb4` character set rather than the `utf8`. This refers to #173.

I was unsure how to structure an update sql file, so there is none.

### DIFF
--- a/lib/www/clarification.php
+++ b/lib/www/clarification.php
@@ -329,7 +329,7 @@ function confirmClar() {
 		if ( ! $respid ) {
 			$teams = $DB->q('KEYVALUETABLE SELECT teamid, name
 			                 FROM team
-			                 ORDER BY categoryid ASC, team.name COLLATE utf8_general_ci ASC');
+			                 ORDER BY categoryid ASC, team.name COLLATE '. DJ_CHARACTER_SET_MYSQL . '_general_ci ASC');
 			$options += $teams;
 		} else {
 			if ( $clar['sender'] ) {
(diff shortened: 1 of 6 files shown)


lib/www/clarification.php
```diff
@@ -329,7 +329,7 @@ function confirmClar() {
 		if ( ! $respid ) {
 			$teams = $DB->q('KEYVALUETABLE SELECT teamid, name
 			                 FROM team
-			                 ORDER BY categoryid ASC, team.name COLLATE utf8_general_ci ASC');
+			                 ORDER BY categoryid ASC, team.name COLLATE '. DJ_CHARACTER_SET_MYSQL . '_general_ci ASC');
```
thijskh: I would make the entire collation a new constant, defined next to `DJ_CHARACTER_SET_MYSQL` and not be 'collating' the collation-string deep in the code.

www/jury/ajax_teams.php
```diff
@@ -5,7 +5,7 @@
 
 $teams = $DB->q("TABLE SELECT teamid AS id, name,
                  CONCAT(name, ' (t', teamid, ')') AS search FROM team
-                 WHERE (name COLLATE utf8_general_ci LIKE %c OR
-                 CONCAT('t', teamid) LIKE %c)", $_GET['q'], $_GET['q']);
+                 WHERE (name COLLATE %s_general_ci LIKE %c OR
+                 CONCAT('t', teamid) LIKE %c)", DJ_CHARACTER_SET_MYSQL, $_GET['q'], $_GET['q']);
```
thijskh: Does this work? I think %s will create unwanted `"` around it.

eldering: Indeed, and even if it works, since this is not user-provided content, there's no reason to escape it with '%s' operators from the libdb library. So I'd suggest as Thijs and simply string-join the new PHP constant.

PR #199, opened 2016-02-02
d3 charts for stats instead of flot

TPolzer:
I replaced the flot charts with [d3](http://d3js.org/), a javascript visualization framework under BSD license.

The main advantage is that it is possible to get the graph (with legend) as a named download with a single click (at least with Firefox and Chrome, it does not seem possible with IE).  As a side benefit picture quality is a lot better and file size a lot smaller.

As there is no such thing as a free lunch, there are downsides, imho the biggest one is that d3 is quite low level compared to flot, but we also lose compatibility with OLD browsers (that do not support svg).

Feel free to suggest improvements to the code, I'm not very experienced with javascript...

An example of the output:
stats of FAU Wintercontest


ubergeek42:
Interesting and I like it; but the code does seem to be waaaay more complex than with flot. I did a quick search and see that there are wrappers around D3 for graphs(like this: http://nvd3.org/examples/multiBar.html ). Do you have any thoughts on that or know of any other libraries that would do something similar?

TPolzer:
My knowledge here is quite limited. nvd3 could be a possible alternative (didn't know about it before), although it looks like keeping the visual style could be difficult (is that important?).
One motivation for me to do it in d3 was that it actually has decent documentation — the same cannot be said of nvd3 and I do not want to brute force my way through their API.

meisterT:
what's the status of this PR? Did anyone have a deeper look in nvd3 (or other similar libs)?

TPolzer:
I updated this so that it should
- work without hard coding the public participant category
- properly show frozen submissions (with optional unfreeze via ?unfrozen=1)

I should probably also update the script, although completely automating it in this form looks difficult.
Also that whole SQL query looks pretty fragile, any better way of doing this?

While developing I also noticed that I don't get SQL errors logged/displayed anymore without inserting try/catch/echo in the code. Is that intentional? Shouldn't those be logged (maybe my setup is broken)?

www/jury/statistics.php
```diff
@@ -2,12 +2,9 @@
 
 require('init.php');
 
-// Include flot javascript library
 $extrahead = '';
-$extrahead .= '<!--[if lte IE 8]><script language="javascript" type="text/javascript" src="../js/flot/excanvas.js"></script><![endif]-->';
 $extrahead .= '<script language="javascript" type="text/javascript" src="../js/flot/jquery.js"></script>';
```
thijskh: Still including some flot component here, that doesn't seem right.

webapp/web/js/d3.v3.min.js
```diff
@@ -0,0 +1,5 @@
...on Kt(){function n(n,r){var u=n-t,i=r-e,a=Math.sqrt(u*u+i*i);Lo+=a*(t+n)/2,qo+=a*(e+r)/2,To+=a,Gt(t=n,e=r)}var t,e;Bo.point=function(r,u){Bo.point=n,Gt(t=r,e=u)}}function Qt(){Bo.point=Gt}function ne(){function n(n,t){var e=n-r,i=t-u,a=Math.sqrt(e*e+i*i);Lo+=a*(r+n)/2,qo+=a*(u+t)/2,To+=a,a=u*n-r*t,Ro+=a*(r+n),Do+=a*(u+t),Po+=3*a,Gt(r=n,u=t)}var t,e,r,u;Bo.point=function(i,a){Bo.point=n,Gt(t=r=i,e=u=a)},Bo.lineEnd=function(){n(t,e)}}function te(n){function t(t,e){n.moveTo(t+a,e),n.arc(t,e,a,0,ja)}function e(t,e){n.moveTo(t,e),o.point=r}function r(t,e){n.lineTo(t,e)}function u(){o.point=t}function i(){n.closePath()}var a=4.5,o={point:t,lineStart:function(){o.point=e},lineEnd:u,polygonStart:function(){o.lineEnd=i},polygonEnd:function(){o.lineEnd=u,o.point=t},pointRadius:function(n){return a=n,o},result:b};return o}function ee(n){function t(n){return(o?r:e)(n)}function e(t){return ie(t,function(e,r){e=n(e,r),t.point(e[0],e[1])})}function r(t){function e(e,r){e=n(e,r),t.point(e[0],e[1])}function r(){M=NaN,S.point=i,t.lineStart()}function i(e,r){var i=dt([e,r]),a=n(e,r);u(M,x,y,b,_,w,M=a[0],x=a[1],y=e,b=i[0],_=i[1],w=i[2],o,t),t.point(M,x)}function a(){S.point=e,t.lineEnd()}function l(){
```
thijskh: Would I think be better to `composer require mbostock/d3`, and symlink the file to the webroot (like twbs/bootstrap).

www/jury/statistics.php
```diff
@@ -74,37 +85,202 @@
...
+	placeholder.append("br");
+	placeholder.append("a")
+		.attr("download", "<?=isset($shortname) ? $shortname . "_" : ""?>stats.svg")
+		.attr("href", "data:image/svg+xml;base64," +
+			utf8_to_b64("<svg xmlns=\"http://www.w3.org/2000/svg\" width='" + $("#placeholder").width() +
+			"' height='" + $("#placeholder").height()+ "'>" +
+			new XMLSerializer().serializeToString(document.getElementById('stats_graph')) + "</svg>"))
+		.text("download graph");
+	placeholder.append("p")
+		.text("If the above link does not work for you, please try another browser.");
```
thijskh: Perhaps helpful to state what browser would suffice/in which browsers it works? Do we even support browsers where it doesn't?

www/jury/statistics.php
```diff
@@ -18,6 +15,10 @@
 	$probid = (int)$_GET['probid'];
 }
 
+if ( !empty($_GET['unfrozen']) ) {
```
thijskh: This option is not currently 'reachable' from the interface, is it?

www/jury/statistics.php
TPolzer: Fixed.

www/jury/statistics.php
TPolzer: Fixed.

www/jury/statistics.php
TPolzer: I don't remember exactly what I tested. Will have to retest and clarify.

www/jury/statistics.php
```diff
@@ -63,13 +63,24 @@
 // All problems
 $problems = $DB->q('SELECT p.probid,p.name FROM problem p
                     INNER JOIN contestproblem USING (probid)
-                    WHERE cid = %i ORDER by shortname', $cid);
+					WHERE cid = %i ORDER by shortname', $cid);
```
eldering: Minor issue: whitespace changed into tabs.

webapp/web/js/d3.v3.min.js
TPolzer: Hm, I tried, but I have no idea what I did wrong: https://pastebin.com/WzkzzXhr

www/jury/statistics.php
TPolzer: That was not the only part where indentation was mixed. Should be better now.

webapp/web/js/d3.v3.min.js
meisterT: Your parameters.yml is out of date. I guess a 
make dist 
will do the trick. Otherwise just copy the missing line from the stub file.

TPolzer: `make dist` failed too, but `git clean -dxf; make dist` did the trick.
Should work now.
Do I need the v=<?=ASSET_TAG?> for the js include, what is that for?

www/jury/statistics.php
TPolzer: I did some quick testing (on up to date browsers only):
- Internet Explorer does not download anything at all.
- (desktop) Safari navigates to the image instead of downloading.
- Chrome and Firefox look good.

PR #238, opened 2016-10-28
permanent settings for judgehost consistency

thuvh:
As the documents, all settings will be lost if you reboot the server. This commit give some instructions to permanents settings.

### DIFF
--- a/doc/admin/admin-manual.sgml
+++ b/doc/admin/admin-manual.sgml
@@ -1182,6 +1182,15 @@ memory addressing bugs give more reproducible results:
 <code>
 echo 0 &gt; /proc/sys/kernel/randomize_va_space
 </code>
+or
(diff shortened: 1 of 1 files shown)


thijskh:
Thanks, very useful. But is it an idea to make the instructions so that it's permanent from the start? That is, replace all with: add line to sysctl.conf; run `sysctl -p`?

thuvh:
Yes, it make permanent from start. After add the line to sysctl.conf, you can use sysctl -p to reload from config file.

thijskh:
Could you change the PR to reflect that?

thijskh:
I've merged it but squashed the two commits together: 97333e71b04dcfea2fc8bbcab43bedfb5980542b.
Thanks for your contribution!

Issue #237, opened 2016-10-14
Current password hashing is not secure


bertptrs:
The current hashing technique used for hashing password is `md5($username . '#' . $password)`. While this does prevent users with the same password showing in the database, it is not a secure password hashing scheme as it can be brute forced quite easily.

Would it be possible to implement a secure hashing scheme, like PHP's own `password_hash` and `password_verify`? For backwards compatibility, a package such as [password_compat](https://github.com/ircmaxell/password_compat) can be used. That way the PHP requirement does not have to be bumped.

thijskh:
Agreed that this is the case. We shied away from changing it earlier because it would mean that when upgrading DOMjudge everyone's password would need to be reset (or a complicated migration scheme implemented). We could consider to accept the breakage and just change it in the next major release.

bertptrs:
If I were to create a patch and send a PR, would you consider it?

thijskh:
Yes!

eldering:
I would suggest to create `dj_*` wrappers (see `lib/lib.wrappers.php`) around the functions `password_verify` and `password_needs_rehash` to detect our old-style `md5($username . '#' . $password)` hashes and be able to update these when a user logs in.

eldering:
@bertptrs: I see you already pushed a PR, great! Do you think implementing my suggestion would be easy/sensible?

PR #239, opened 2016-10-28
Implement a secure hashing algorithm

bertptrs:
Implement the feature suggested in #237.

The password hashing algorithm now used is `PASSWORD_DEFAULT` (which currently is PASSWORD_BCRYPT) for future compatibility. The `ircmaxell/password-compat` library is used to remain backwards compatible with PHP 5.4.

The cost for the hashing function is configurable in `common-config.php`.

Note that this patch invalidates all current passwords. The default data has been updated to reflect the new hash style.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,6 +13,7 @@
 		"jumbojett/openid-connect-php": "0.1.*",
 		"mustangostang/spyc": "0.5.*",
 		"php": ">= 5.4",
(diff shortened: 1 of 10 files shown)


eldering:
I made some small fixes, additions, and squashed those commits. Furthermore, I added `dj_password_*` wrappers supporting migrating from MD5 to the new hashes.

PR #252, opened 2017-02-04
allows filter for room/affiliation in balloon page, fixes #251

### LINKED ISSUE #251: Add option to filter balloons per room
Not only but especially for multi site contests.

DerBaer0:
Allows filtering the balloon page by affiliation and/or room for multi site contests.
More or less the same code as the scoreboard filter.

### DIFF
--- a/www/jury/balloons.php
+++ b/www/jury/balloons.php
@@ -7,9 +7,42 @@
  * under the GNU GPL. See README and COPYING for details.
  */
 
+/**
+ * Reads balloon filter settings from a cookie and explicit POST of
+ * filter settings. Also sets the cookie, so must be called before
+ * headers are sent. Returns the balloon filter settings array.
+ */
+function initBalloonfilter()
+{
+	$balloonfilter = array();
+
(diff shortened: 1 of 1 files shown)


www/jury/balloons.php
```diff
@@ -52,6 +85,62 @@
     addSubmit($viewall ? 'view unsent only' : 'view all') . "</p>\n" .
     addEndForm();
 
+
+// Filtering by affiliation or room
+$affils = $DB->q('TABLE SELECT affilid,
```
eldering: I think this query currently does not return what it should: the lists of all affiliations and rooms of teams that are part of selected list of contests (which can be one, or all active contests). Some issues:
- the array variable `$contestids` (currently defined down in the code) should be used instead of `$cdata['cid']`
- I think the where clause `c.cid = %i` is superfluous
- the generated filter list contains duplicate universities (not sure why)

I'd suggest to select from `team` and then LEFT JOIN on `team_affiliation`; the only point is then that empty affiliations have to be filtered out, or that would show up in the filter list and be selected by default.

PR #259, opened 2017-04-18
Allow problems to be hidden until another problem is solved.

SolalPirelli:
As discussed on the mailing list, here's the PR.

I'm not a PHP expert by any means, nor do I know the domjudge codebase, so there may be a better way to do it. Also, I guess I should add something about this feature to the admin manual?

### DIFF
--- a/lib/lib.misc.php
+++ b/lib/lib.misc.php
@@ -91,20 +91,31 @@ function getRequestID($numeric = TRUE)
 /**
  * Returns whether the problem with probid is visible to teams and the
  * public. That is, it is in the selected (active) contest, which has
- * started and it is submittable.
+ * started and it is submittable; if it depends on another problem,
+ * it should be visible only to teams who have solved that problem
+ * until the contest has ended.
  */
 function problemVisible($probid)
 {
-	global $DB, $cdata;
+	global $DB, $cdata, $userdata;
 
 	if ( empty($probid) ) return FALSE;
 
 	$fdata = calcFreezeData($cdata);
 	if ( !$fdata['cstarted'] ) return FALSE;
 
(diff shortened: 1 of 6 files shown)


meisterT:
Am I right that the (hidden) problem is still submittable, shows up in the scoreboard and teams can view clarification answers? Isn't that confusing for the teams?

SolalPirelli:
IMHO the scoreboard should show all problems so teams can know who did what; the point is not to have a secret number of problems, or to hide the existence of certain problems, only to disable them until teams have solved the previous one.

Making hidden problems unsubmittable and hiding clarifications for them would probably be better indeed, I'll work on that.

SolalPirelli:
Clarifications for hidden problems are now hidden, hidden problems don't appear in the submit list any more, and the branch has been rebased on top of master. Sorry for the delay.

eldering:
Sorry for not getting back on this sooner.
Ok, I reviewed the pull request, and there are still some issues. In particular, in https://github.com/DOMjudge/domjudge/blob/master/www/team/index.php#L25, https://github.com/DOMjudge/domjudge/blob/master/www/team/upload.php#L64 and https://github.com/DOMjudge/domjudge/blob/master/lib/lib.misc.php#L680 (I may have missed somethin), the query should also incorporate the logic of restricting to problems that are already solved if there is a problem dependency. The best is probably to abstract this away into a function, but we have to think about how to best do this, since at least in `www/team/index.php` the SQL query is retrieving a list of problems instead of checking a single one.
I'll try to look into this soon. Futhermore, I added some changes on top of your pull, see https://github.com/eldering/domjudge/commits/dependent_problems.

eldering:
Closing this because of lack of activity. Feel free to reopen.

PR #291, opened 2017-10-18
Added download submission on the submission details page

rjalfa:
With all the checks.


meisterT:
That opens up an attack surface. So far, getting access to a team account only meant that you can submit in disguise (which then can be ignored later). But now teams can access other source code.

Do you run DOMjudge in an online judge setting or what's your use case?

I would at least argue that this should be behind an configuration option which is off by default.

rjalfa:
Ya an option would be good.
Download submissions is provided by almost all online judges and is a great option in case we want to see our code or there is a system failure during the contest and migration is not an option. But you rightly said that this should be a disabled-by-default option.
I will try adding that. Thanks

rjalfa:
Also this is good for teams when multiple computers are allowed.

thijskh:
This would be good to have if configurable. But the current documentation is not sufficient. Besides from being configurable, it only offers one file for download. And the team interface has been changed completely since.

We'd accept a PR to this effect but the current code is not mergeable.

PR #298, opened 2017-10-21
Allow Jury members to Override the Validator's Decision on a Given Submission

matt-hermes:
This creates buttons on the submission page which show whenever the submission is not marked as verified, to allow the jury to override the validator's decision as to whether an answer is correct or not.

### DIFF
--- a/www/jury/override.php
+++ b/www/jury/override.php
@@ -1,0 +1,60 @@
+<?php
+/**
+ * Change the judging result of a given judging.
+ *
+ * Part of the DOMjudge Programming Contest Jury System and licenced
+ * under the GNU GPL. See README and COPYING for details.
+ */
+
+require('init.php');
+$id      = @$_POST['id'];
+$val     = @$_POST['val'];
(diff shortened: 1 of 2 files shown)


eldering:
I saw this was already closed before we had time to comment. I don't know what your reasons were for closing it, but let me just mention that we preferin principle not to include features that allow manually overriding a judgement. We think that by design a contest control system should not allow that. Instead if you find that something is wrong, with either a validator, test data, or something else, then that should be fixed and then a rejudging performed.
If you strongly believe that this would be a good feature, then please start a discussion on the `domjudge-devel` emaillist. We're happy to discuss such issues there.

matt-hermes:
I closed it because I found a fault in my code when used in certain
situations. I had it requested by a jury member in one of the contests I
administer because there is occasionally a submission which fails the
comparator for no apparent reason (we believe encoding) but we usually do
not have time to investigate during the contest.

In the past, this has been done with a database edit, but I am attempting
to eliminate any need for those.

PR #309, opened 2017-11-02
Should update the length of `action` column in `auditlog` table due t…

khhan1993:
…o this error.

error: SQL error, Error#1406: Data too long for column 'action' at row 1, query: 'INSERT INTO auditlog (logtime, cid, user, datatype, dataid, action, extrainfo) VALUES("1509614985.217912", null, "admin", "configuration", null, "update show_balloons_postfreeze", "1")'

This error occurs when I trying to change "Configuration settings" in web interface.


matt-hermes:
FYI, this also occurs with openid registration, as that overflows the same
field.

eldering:
Fixed in a slightly broader patch that uniformizes varchar lengths a bit more.

PR #330, opened 2018-02-07
runguard: Kill all children on exit(EXIT_FAILURE)

dpaulus2:
When encountering an error, e.g. I/O error on child_redirfds, make sure
that all children are killed before exit()ing runguard.

This was causing problems with an interactive problem in our contest.
The validator exited once encountering the first invalid response line from
the submission, and runguard failed to forward the submission output
to it, with the write() failing with EPIPE. exit()ing in this case left behind the
submission, causing testcase_run.sh to fail when checking if there were still
any processes by RUNUSER left over.

### DIFF
--- a/judge/runguard.c
+++ b/judge/runguard.c
@@ -152,7 +152,7 @@ rlim_t filesize;
 rlim_t nproc;
 size_t streamsize;
 
-pid_t child_pid;
+pid_t child_pid = -1;
 
 static volatile sig_atomic_t received_SIGCHLD = 0;
 static volatile sig_atomic_t received_signal = -1;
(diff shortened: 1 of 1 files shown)


judge/runguard.c
```diff
@@ -254,6 +263,24 @@ void error(int errnum, const char *format, ...)
 
 	write_meta("internal-error","%s: %d - %s","runguard error", errnum, format);
 
+	/* Make sure that all children are killed before terminating */
+	if ( child_pid != -1) {
```
eldering: I think this check should be `child_pid > 0` since `child_pid == 0` when in the child fork, but there an `error()` call would work fine.

judge/runguard.c
```diff
@@ -226,6 +226,15 @@ void error(int errnum, const char *format, ...)
 {
 	va_list ap;
 	va_start(ap,format);
+	sigset_t sigs;
```
eldering: Do you think it's better to handle this here, than simply call `terminate()` which already does signal resetting and the same sending SIG{TERM,KILL} below? (This is a genuine question.)

judge/runguard.c
dpaulus2: Good point, that's right

judge/runguard.c
dpaulus2: I've decided to duplicate this for two reasons:
 - terminate() itself might call error(), which is unlikely, but possible. A failed kill()-call could then cause infinite recursion
 - terminate() messes with signal-handlers: It resets SIGALRM and SIGTERM to DFL. Normally, this is fine, as both are blocked anyway when terminate() is invoked during signal-handling. When calling it from the main control flow, however, there would be a short window where SIGTERM would terminate runguard while still leaving children behind, which I intended to avoid

PR #338, opened 2018-02-24
Debug travis check-api.sh issue

eldering:


travis.sh
```diff
@@ -170,5 +170,8 @@ if [ $NUMNOTVERIFIED -ne 2 ] || [ $NUMNOMAGIC -ne 0 ]; then
 	exit -1;
 fi
 
+mysql domjudge -e 'SELECT max(eventid) FROM event'
+mysql domjudge -e 'SELECT * FROM event WHERE eventid=112\G'
```
meisterT: what is \G here? a typo?

eldering: No, \G prints SQL rows per line instead of column.

PR #398, opened 2018-07-28
Allow eventlog and API to have more than one primary key

nickygerritsen:
The previous, hacky version was so hacky because there were two code paths in both the `eventlog` function as well as in the only API endpoint supporting it (`runs`).

I have now modified all V4 API endpoints to allow passing more than one ID in __primary_key and changed `restapi.php` to turn `__primary_key` into an array. Although this still results in quite some changes in the API, there are almost no extra code paths anymore. Only the `eventlog` function has a check whether one or more ID's are passed, as that results in getting back an array of objects or only one object from the API.

I have locally called all API endpoints I modified to see if the correct data is still returned and that seems to be the case.

There is one endpoint which is a little bit more work to fix: the `contest` endpoint. This one is already in the Symfony `APIController` and it will always return a single contest.
I think we should accept this for now and make sure we do not call `eventlog` for contests with more than one ID.
In the future, when we have all endpoints in Symfony, I think we can come up with a better way to log events anyway; one that does not go through the API, but uses the Doctrine entities directly.

### DIFF
--- a/lib/lib.misc.php
+++ b/lib/lib.misc.php
@@ -1373,19 +1373,28 @@ function eventlog($type, $dataids, $action, $cid = null, $json = null, $ids = nu
 	$dataids = array_values($dataids);
 	$ids = array_values($ids);
 
-	$cids = array();
+	$cids = [];
 	if ( $cid!==null ) {
 		$cids[] = $cid;
+		$expectedEvents = count($dataids);
 	} else {
 		if ( $type==='problems' ) {
-			$cids = $DB->q('COLUMN SELECT DISTINCT cid FROM contestproblem WHERE probid IN (%Ai)', $dataids);
-		} elseif( $type==='teams' ) {
-			$cids = [];
+			$expectedEvents = 0;
 			foreach ($dataids as $dataid) {
-				$cids = array_unique(array_merge($cids, getCurContests(FALSE, $dataid)));
+				$cidsForId = $DB->q('COLUMN SELECT DISTINCT cid FROM contestproblem WHERE probid = %Ai', $dataid);
+				$expectedEvents += count($cidsForId);
+				$cids = array_unique(array_merge($cids, $cidsForId));
+			}
+		} elseif( $type==='teams' ) {
+			$expectedEvents = 0;
+			foreach ($dataids as $dataid) {
+				$cidsForId = getCurContests(FALSE, $dataid);
+				$expectedEvents += count($cidsForId);
+				$cids = array_unique(array_merge($cids, $cidsForId));
 			}
 		} else {
 			$cids = getCurContests();
+			$expectedEvents = count($dataids) * count($cids);
 		}
 	}
 	if ( count($cids)==0 ) {
@@ -1474,9 +1483,9 @@ function eventlog($type, $dataids, $action, $cid = null, $json = null, $ids = nu
 		error("eventlog: failed to release lock");
 	}
 
-	if ( count($ids) !== count($cids) * count($dataids) ) {
+	if ( count($eventids) !== $expectedEvents ) {
 		error("eventlog: failed to $action $type/$idsCombined " .
-		      '('.count($ids).'/'.(count($cids) * count($dataids)).' contests done)');
+		      '('.count($eventids).'/'.$expectedEvents.' events done)');
 	}
 
 	logmsg(LOG_DEBUG,"eventlog: ${action}d $type/$idsCombined " .
--- a/lib/lib.misc.php
+++ b/lib/lib.misc.php
@@ -998,7 +998,10 @@ function rejudging_finish($rejudgingid, $request, $userid = NULL, $show_progress
 	auditlog('rejudging', $rejudgingid, $request.'ing rejudge', '(start)');
 
 	while ( $row = $res->next() ) {
-		if ( $show_progress ) echo "s" . specialchars($row['submitid']) . ", ";
+		if ( $show_progress ) {
+			echo "s" . specialchars($row['submitid']) . ", ";
+			sleep(1);
+		}
 		if ( $request=='apply' ) {
 			$DB->q('START TRANSACTION');
 			// first invalidate old judging, maybe different from prevjudgingid!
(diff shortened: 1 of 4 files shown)


nickygerritsen:
Ah yes and of course I have applied it to the rejudging runs, as that was where the problem was.

lib/lib.misc.php
```diff
@@ -998,7 +998,12 @@ function rejudging_finish($rejudgingid, $request, $userid = NULL, $show_progress
 	auditlog('rejudging', $rejudgingid, $request.'ing rejudge', '(start)');
 
 	while ( $row = $res->next() ) {
-		if ( $show_progress ) echo "s" . specialchars($row['submitid']) . ", ";
+		if ( $show_progress ) {
+			echo "s" . specialchars($row['submitid']) . ", ";
+			// Also flush the output, so the browser displays updates
+			ob_implicit_flush(true);
```
meisterT: Perhaps I don't understand the documentation correctly but why do we do this in every loop iteration?

lib/lib.misc.php
```diff
@@ -998,7 +998,12 @@ function rejudging_finish($rejudgingid, $request, $userid = NULL, $show_progress
...
+			ob_flush();
```
meisterT: Do we need this at all? See the first user contributed example here: http://php.net/manual/en/function.ob-implicit-flush.php

lib/lib.misc.php
```diff
@@ -1319,26 +1344,41 @@ function eventlog($type, $dataid, $action, $cid = null, $json = null, $id = null
 	}
 
 	// Look up external/API ID from various sources.
-	if ( $id===null ) $id = rest_extid($type, $dataid);
+	if ( $ids===null ) {
+		$ids = array_map(function($dataid) use ($type) {
+			return rest_extid($type, $dataid);
```
meisterT: Please add a TODO to also add multiple id functionality to rest_extid. Otherwise this does n DB queries which could be a single one.

lib/lib.misc.php
```diff
@@ -1319,26 +1344,41 @@ function eventlog($type, $dataid, $action, $cid = null, $json = null, $id = null
...
+		}, $dataids);
+	}
 
-	if ( $id===null && $json!==null ) {
+	if ( $ids===null && $json!==null ) {
```
meisterT: can $ids be null at this point? it's rather that single entries in the array are null, right? Do we have to do the check here for every id instead?

lib/lib.misc.php
```diff
@@ -1319,26 +1344,41 @@ function eventlog($type, $dataid, $action, $cid = null, $json = null, $id = null
...
-		if ( !empty($data['id']) ) $id = $data['id'];
+		if ( !empty($data['id']) ) $ids = [$data['id']];
+	}
+
+	if ($ids !== null && !is_array($ids)) {
+		$ids = [$ids];
 	}
 
-	if ( $id===null ) {
+	if ($type !== 'state' && count(array_filter($ids)) !== count($dataids) ) {
```
meisterT: Please add a comment why 'state' is special.

lib/lib.misc.php
```diff
@@ -1348,14 +1388,28 @@ function eventlog($type, $dataid, $action, $cid = null, $json = null, $id = null
...
 
+	$idsCombined = implode(',', $ids);
+
+	// We should pass multiple ID's if instructed to do so
+	$multiple = count($dataids) > 1 || count($ids) > 1;
+
 	// Generate JSON content if not set, for deletes this is only the ID.
 	if ( $action === 'delete' ) {
-		$json = "{\"id\":\"$id\"}";
+		$json = array_values(array_map(function($id) {
```
meisterT: my PHP foo is weak, what are you trying to do here?

lib/lib.misc.php
```diff
@@ -1386,28 +1445,35 @@ function eventlog($type, $dataid, $action, $cid = null, $json = null, $id = null
...
-		                  $now, $cid, $type, $id,
-		                  $table, $dataid, $action, $json);
-		$eventids[] = $eventid;
+		foreach ($dataids as $idx => $dataid) {
+			if ($multiple) {
+				$jsonElement = dj_json_encode($json[$idx]);
+			} else {
+				$jsonElement = dj_json_encode($json);
+			}
+			$eventid = $DB->q('RETURNID INSERT INTO event
```
meisterT: Can we do one bigger DB query here instead?

lib/lib.misc.php
```diff
@@ -1348,14 +1388,28 @@ function eventlog($type, $dataid, $action, $cid = null, $json = null, $id = null
 		return;
 	}
 
+	$idsCombined = implode(',', $ids);
```
meisterT: this assumes that no primary ID contains a comma - is that true for all primary IDs?

lib/www/restapi.php
```diff
@@ -265,15 +267,21 @@ public function showDocs()
 
 	private function createResponse($response)
 	{
-		header('Content-Type: application/json');
+		// Only send headers if not done already. Headers might be sent if calling the API internally
```
meisterT: I would split out this and the change below into a separate commit.

www/api/v4/index.php
```diff
@@ -249,7 +249,11 @@ function problems($args)
 			$api->createError("You cannot specify a primary ID both via /{id} and ?probid={id}");
 			return '';
 		}
-		$args['probid'] = rest_intid('problems', $args['__primary_key'], $cid);
+		$args['probids'] = array_map(function($probId) use ($cid) {
+			return rest_intid('problems', $probId, $cid);
```
meisterT: Please add same TODO comment as above to support multiple ids in rest_intid. Or at the function definition level.

lib/lib.misc.php
nickygerritsen: I think you are right that we can move this to above the loop

lib/lib.misc.php
nickygerritsen: Let me test this :)

lib/lib.misc.php
nickygerritsen: When `$json` is not null, it can only be `[null]`. So I have changed this.

lib/lib.misc.php
nickygerritsen: It maps every `$id` to `['id'=>$id]` and then makes sure the array is 0-indexed.

lib/lib.misc.php
nickygerritsen: There is already a TODO above here.
It might be possible (although I don't know how we are getting the RETURNID then), but I think this is not for 6.0 for now.

lib/lib.misc.php
nickygerritsen: For external primary ID's this might not be generally true, but in practice I think it is.

I have no easy solution if we do allow comma's, so we should think about this later on maybe?

meisterT: I would add a TODO so we don't forget about it.

PR #399, opened 2018-07-28
Make entry point requirement configurable per language.

thijskh:
Also allow to specify the description of the field in the UI.

Still missing: submit client changes for this.

### DIFF
--- a/webapp/web/js/domjudge.js
+++ b/webapp/web/js/domjudge.js
@@ -253,19 +253,6 @@ function hideTcSample(tcid, str)
 	node.parentNode.appendChild(span);
 }
 
-// Construct base part of Kotlin entry point from filename base
-function kotlinBaseEntryPoint(filebase)
-{
-	if ( filebase === '' ) return '_';
-
-	filebase = filebase.replace(/[^a-zA-Z0-9]/, '_');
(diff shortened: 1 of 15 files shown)


webapp/web/js/domjudge.js
```diff
@@ -403,26 +378,55 @@ function initReload(refreshtime)
...
+			case 'kt':
+				entry_point.value = entryPointDetectKt(filename);
+				break;
+			default:
+				entry_point.value = filename;
 		}
 	}
 }
 
+function entryPointDetectJava(filename)
```
eldering: What about moving this into a per language code snippet in the DB or as a new type of script?

meisterT: if we make this a per language code snippet that we could make the same snippet available via the API as well and use the API endpoint from both the web site and the submit client

nickygerritsen: THen we do have to move it to PHP I guess, but that is possible by using some AJAX of course

thijskh: Moving the detection to API calls both from commandline and web is possible but seems a big enough change vs how it works now that I'd like to move that to a separate feature request.

meisterT: sounds reasonable to do this after the 6.0 release

PR #401, opened 2018-07-28
Preview of how the code changes with PSR-2.

meisterT:
I don't want to merge in this stage, but that should give you an impression of the changes.

### DIFF
--- a/lib/lib.error.php
+++ b/lib/lib.error.php
@@ -5,8 +5,8 @@
  * Part of the DOMjudge Programming Contest Jury System and licenced
  * under the GNU GPL. See README and COPYING for details.
  */
-if ( ! defined('SCRIPT_ID') ) {
-	define('SCRIPT_ID', basename($_SERVER['PHP_SELF'], '.php'));
+if (! defined('SCRIPT_ID')) {
+    define('SCRIPT_ID', basename($_SERVER['PHP_SELF'], '.php'));
 }
 
 // Default verbosity and loglevels:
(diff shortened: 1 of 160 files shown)


lib/lib.error.php
```diff
@@ -45,93 +45,102 @@
...
+            // normal interactive web interface: output with markup
+            } else {
+                // string 'ERROR' parsed by submit client, don't modify!
+                if ($msglevel == LOG_ERR) {
+                    echo "<fieldset class=\"error\"><legend>ERROR</legend> " .
+                         $msg . "</fieldset><!-- trigger HTML validator error: --><b>\n";
+                } elseif ($msglevel == LOG_WARNING) {
+                    echo "<fieldset class=\"warning\"><legend>Warning</legend> " .
+                        $msg . "</fieldset><!-- trigger HTML validator error: --><b>\n";
+                } else {
```
eldering: How much is enforced by PSR-2? For example I see that the line continuations of the echo commands do not match up.

nickygerritsen: I think this specific case is "left open" ("Inter-line alignment"), but there are other cases where they explicitly say that things must be indented "once" (i.e. with 4 spaces extra), for example when having function arguments on multiple lines.

I'd say that it would make sense to then also use this same rule for these multi-line echo statements, as this feels more consistent. But if others think we should match their alignment I won't fight it too hard.

eldering: I'd prefer to leave it open. If we wanted to strictly enforce the same policy as for multiline function arguments, then we should write (with implicit parentheses for the `echo` function):
```
echo
    "<fieldset class=\"error\"><legend>ERROR</legend> " .
    $msg . "</fieldset><!-- trigger HTML validator error: --><b>\n";
```
which I personally dislike a lot. So I prefer to keep the flexibility to format this as the person coding sees fit.

PR #406, opened 2018-09-06
Use cgroup base directory to detect cgroup support instead of grub commandline

nickygerritsen:
I'm not committing this directly as I'd like input from you guys.

Both Docker on my macOS machine as well as my Ubuntu 18.04 machine do have cgroups enabled, even for memory and including swap accounting, but don't have it in the grub commandline.

Although I can easily add it on my Ubuntu 18.04 machine, it is way harder to add this on Docker on my Mac.

I *think* this should also detect whether cgroups are enabled while not looking at the grub commandline.

Thoughts?

### DIFF
--- a/misc-tools/create_cgroups.in
+++ b/misc-tools/create_cgroups.in
@@ -9,21 +9,31 @@
 JUDGEHOSTUSER=@DOMJUDGE_USER@
 CGROUPBASE=@judgehost_cgroupdir@
 
-if ! grep -Eq 'cgroup_enable.+memory' /proc/cmdline || ! grep -Eq 'swapaccount.+1' /proc/cmdline; then
-        echo "Error: cgroup support missing from running kernel. Unable to continue.
-
-        To fix this, please make the following changes:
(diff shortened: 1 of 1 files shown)


thijskh:
I don't think this change is right because the dir you check for is created by the script after the check.

nickygerritsen:
Yes @eldering also pointed that out. However the original code is also not correct, as I have two different cases where the cgroups we use are enabled, but nothing is added to `cmdline`.

Maybe we can move this check to below the `mount` command (adding `|| exit 1` to that `mount` command), assuming the mount command will exit with a non-zero exitcode if cgroups are not enabled at all? Or do you have another idea?

thijskh:
The idea of the check is that it's informative and checks early on if everything is in place. Before the check, you would only found out when judging the first submission. So I agree with checking the mount errors but the errors should be made explicit as to what the problem is.

nickygerritsen:
Just checking the mount result didn't work because the directories might already be mounted (which is the case at least when running in Docker) so I also added a check for that.

eldering:
The new version looks good to me and works here.

PR #407, opened 2018-09-08
Update Symfony to 3.4

nickygerritsen:
As we were running a version of Symfony that does not receive any security updates anymore, we should upgrade.
3.4 is the latest LTS. It is also the last version to support PHP 7.0, so I think it makes sense to upgrade to this version.

At some point we should upgrade to 4.x, but that will increase the required PHP version to 7.1+, so I don't think we should do that right now.
Also the Symfony people recommend using [Symfony Flex](https://symfony.com/doc/current/setup/flex.html) for 4.0+ applications and it is definitely non-trivial to get that working if Symfony is not 'installed' in the directory where `composer.json` is located, like we have where it is installed in `webapp`. I tried this with Symfony 3.4 (as Flex is supported there) but couldn't get it to work. We should investigate later if we can get this to work or only start using Flex when we have `composer.json` in the same directory.

Note that Symfony 3.4 [is slower](http://www.phpbenchmarks.com/en/comparator/compare.html?benchmarkTools=apache-bench&phpVersions=php-7.0%7Ephp-7.1%7Ephp-7.2&groups=symfony-3.x%7Esymfony-4.x&components=symfony-4.1) compared to 3.2 but I do think we should upgrade.


nickygerritsen:
Will take a look at the tests now

meisterT:
The 4.x series is not an option for us? We currently don't have much symfony code so I expect the upgrade to 4.x to be much easier now than say in a year.

nickygerritsen:
If we move to 4.x, we update our PHP requirement to 7.1+. I am not opposed to that for 6.1 or something, but I do want opinions from others.
Upgrading to 4.0 from this 3.4 branch should be almost trivial, as I have fixed all (unless I missed some) deprecation warnings for 3.4, which should give a clean upgrade path.

nickygerritsen:
Oh and we can probably not use Symfony Flex yet then, which *is* a bigger change than the actual upgrade.

meisterT:
Ok, makes sense to pick 3.4 then.

nickygerritsen:
Found the Travis issue, it is because we do not copy composer.json. For now I have reverted back from `%kernel.project_dir%` to `%kernel.root_dir%/..` again (basically) but we might consider copying the `composer.json` when installing.

PR #410, opened 2018-09-22
Analysis/Reporting Pages(to replace statistics.php) + bootstrap UI for jury

ubergeek42:
This adds a new "Analysis" section to the domjudge jury interface. This new section is written entirely within symfony, taking advantage of twig to do the page rendering. I find this much cleaner than our spaghetti mess elsewhere.

The new bootstrappy template I made for this mostly works, but I'm not 100% certain the header is fully functional(there's no contest switcher for one; it probably doesn't do alerts/notifications right, etc)

The analysis section consists of 3 views. The first is the contest view, which gives an overall summary of the contest. Then there are pages for each problem, and for each team that summarize the relevant data for each.

I deleted the old statistics page and updated the link for it to point to these fancy new pages.

---
This also adds some graphs to the submission page in the jury interface that might be useful when determining timelimits(or to see how a submission does compared to the timelimit). One graph shows the max runtime of the previous rejudgings for this submission(and doesn't appear if it has never been rejudged). The other graph shows the runtimes for each testcase of this judging.

To make this work, a few key changes were made:
* Add nvd3 library for creating charts much easier than just messing directly with d3
* Exposing the symfony twig renderer to the legacy code, and using twig to construct the html for the graphs. I find this a bit cleaner than all the echo stuff we do elsewhere. Maybe it'd be useful to move some other rendering components over to twig as well.

## Some screenshots below:
Submission page:
image

Analysis Contest Page
domjudge k8s ubergeek42 com_jury_analysis_

Analysis Problem Page
domjudge k8s ubergeek42 com_jury_analysis_problem_8

Analysis Team Page
domjudge k8s ubergeek42 com_jury_analysis_team_212


webapp/src/LegacyBundle/Controller/FallbackController.php
```diff
@@ -83,6 +83,7 @@ public function fallback(Request $request, $path)
         ob_start();
         global $G_SYMFONY;
         $G_SYMFONY = $this->container->get('domjudge.domjudge');
+        $G_SYMFONY_RENDER = array($this, 'renderView');
```
nickygerritsen: Isn't it better, especially after merging 3.4 which makes more in the container private by default, to inject `Twig_Environment` in the constructor of the FallbackController and setting that as a variable here? That is what `renderView` uses to render the views

ubergeek42: Yes! that does indeed sound wiser; I was taking the easy way since searching the docs didn't find me that information quickly and I wanted to skip to just having it working.

meisterT:
I guess it would be helpful to show in the right chart (testcase runtimes) also one bar per rejudging. And perhaps include the result in the label (at least if it differs).

Can we also (in a separate change) make these graphs available per submission (instead of per rejudging) for a problem? Also you could show the expected result(s) there. I guess that would be even more helpful when determining the time limit.

nickygerritsen:
I really like these graphs 😄

webapp/src/DOMJudgeBundle/Controller/AnalysisController.php
```diff
@@ -0,0 +1,393 @@
...
+    /**
+     *@Route("/", name="analysis_index")
+     */
+    public function indexAction(Request $request)
+    {
+        $em = $this->getDoctrine()->getManager();
+        $contest = $this->DOMJudgeService->getCurrentContest();
+
+        // First collect information about problems in this contest
+        $problems = $contest->getProblems(); // TODO: this gets ContestProblem, but we need Problem as well
```
nickygerritsen: I was thinking about this actually, as for the API it would also result in N+1 queries. I think for contestproblem it makes sense to eagerly load the problems as well in the entity?

webapp/src/DOMJudgeBundle/Controller/AnalysisController.php
```diff
@@ -0,0 +1,393 @@
...
+
+        // Need to query directly the count, otherwise symfony memory explodes
```
nickygerritsen: I think we can also solve this by marking the testcases as extra lazy, then we can do testcase count.
I was also thinking about maybe linking Problem two two testcase entities: a `SlimTestcase` and a normal `Testcasez, where the first one doesn't contain the big fields. We often don't need them anyway. I'm not sure if we can do this with an `extend` from one to the other, but that can be tested

ubergeek42:
"Misery index" is something Ryan Stansifer mentioned to me. It's how many people minutes are spent unsuccessfully trying to solve a problem at the end. It's calculated per team as:
`3(people) * (contest end time - team last solved time)`

If a team solved no problems, their misery index would be 900(5 hour contest is 300 minutes long, times 3 people). If a team solves a problem at the last minute, it'd be ~0.

Then the contest misery index is just the average of that value across all teams. I tried to put an explanation in a box on the page when you hover over it. Looking at the screenshots I don't think I'm calculating it correctly, as I don't think our average last submit time was ~30minutes(88/3) from the end of the contest(and the graphs tend to support that conclusion)

The graphs are svg, they're drawn by d3 using the nvd3 wrapper.

Filters on teams are currently "teams in a category that is visible". So that's all that will show on the contest page.

The team view page is just /jury/analysis/team/<teamid>, so there's nothing stopping it from working for the jury teams. I just don't have links for it. Maybe I'll add all categories to the teams list on the contest page, and color the rows accordingly(and then also include non-visible categories). It just gets weird to graph them because a lot of the jury solutions happen way before the contest, so I have to be careful with some calculations)

The graph of max runtimes against timelimit has fancy hover tooltips, and clicking on one of those bars will take you to the corresponding submission judging page. I wasn't able to take a screenshot of that feature though.

thijskh:
Really nice!

PR #426, opened 2018-10-18
[EXPERIMENT] Add new script to build chroot from a docker image

ubergeek42:
This is an experiment to see if we can make the chroot build process a bit better. My thinking was that docker images are much simpler to make, so maybe we could just take the root filesystem from them, extract it, and use it as our chroot. Turns out it isn't super difficult to do just that, so I wrote a new make_chroot script to do so.

I then built and uploaded a simple docker image that should approximate what `dj_make_chroot` gives you(See dockerfile at the bottom). I configured travis to use it, and it seems to work ok.

I'm also thinking it might be nice to have per language chroots; so maybe our chroot dir ends up looking like:
```
/chroot/domjudge/c++/<a chroot for running c++ programs>
/chroot/domjudge/java/<a chroot for running java programs>
```

I think we could switch to using it for compilation as well, and then things get a lot simpler. E.g. how do you support weird/random versions of kotlin? Just use the official go docker image + do whatever modifications we need for domjudge.

What do you all think? Is this something you'd like to see done/continued/polished?

```Dockerfile
FROM ubuntu:18.04
ENV DEBIAN_FRONTEND=noninteractive
RUN apt-get update && apt-get -y install \
  ca-certificates default-jre-headless pypy locales \
  software-properties-common \
  && rm -rf /var/lib/apt/lists/*

RUN chmod a-s \
  /usr/bin/wall \
  /usr/bin/newgrp \
  /usr/bin/chage \
  /usr/bin/chfn \
  /usr/bin/chsh \
  /usr/bin/expiry \
  /usr/bin/gpasswd \
  /usr/bin/passwd \
  /bin/su \
  /bin/mount \
  /bin/umount \
  /sbin/unix_chkpwd \
  || true

```

### DIFF
--- a/misc-tools/Makefile
+++ b/misc-tools/Makefile
@@ -11,7 +11,7 @@ SUBST_DOMSERVER = save_sources2file restore_sources2db \
                   simulate_contest create_accounts multi_rejudge \
                   eventdaemon
 
-SUBST_JUDGEHOST = create_cgroups dj_make_chroot dj_run_chroot
+SUBST_JUDGEHOST = create_cgroups dj_make_chroot dj_run_chroot dj_make_chroot_docker
 
 SUBST_FILES = $(SUBST_DOMSERVER) $(SUBST_JUDGEHOST)
 
--- a/misc-tools/dj_make_chroot_docker.in
+++ b/misc-tools/dj_make_chroot_docker.in
@@ -1,0 +1,147 @@
+#!/bin/bash
+# @configure_input@
+#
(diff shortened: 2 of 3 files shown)


nickygerritsen:
Really cool idea! This should also make Travis faster I assume?

nickygerritsen:
I restarted Travis and it ran in 9:36, whereas it justed to run in around 11-12 minutes :) (yesterday this PR ran in 14 minutes but probably it was crowded or something)

PR #443, opened 2018-10-28
Use FontAwesome icons on new jury pages

nickygerritsen:
As a 'trial' I replaced the edit/delete/etc icons in tables in the new interface with FontAwesome glyphs. FontAwesome is GPL-compliant and works in IE 10/11 and modern browsers, so that shouldn't be a problem. They have not registered a 5.x version of their package for Composer, so I use a git URL directly.

See this screenshot:
screenshot 2018-10-28 at 16 52 39
(We might consider making the icons a bit lighter. Also we might want to change the blue + button to black or something; it now uses the default link color; or even change it to a `<button>`?)

Any thoughts?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,6 +3,20 @@
 	"description": "DOMjudge Programming Contest Control System",
 	"homepage": "https://www.domjudge.org",
 	"license": "GPL-2.0+",
+	"repositories": [
+		{
+			"type": "package",
+			"package": {
+				"name": "fortawesome/font-awesome",
+				"version": "5.4.2",
+				"source": {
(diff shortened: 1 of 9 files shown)


meisterT:
I like them much more than the old icons.

nickygerritsen:
Btw we could, at some point, also replace the octicons in the menu with these

ubergeek42:
:+1: I agree, it'd be nice to have a more feature complete icon library we can use. I also think the `+` icon should be a <button>, and then styled like usual bootstrap buttons.

nickygerritsen:
I’ll modify the + button to look like that later today and update the screenshot

thijskh:
Excellent plan

PR #444, opened 2018-11-01
[WIP] Delay and batch API calls from judgedaemon

meisterT:
This significantly speeds up judging for problems with many test cases.

Current state is 165s -> ~65s for a problem with 277 test cases.

### DIFF
--- a/webapp/src/DOMJudgeBundle/Controller/API/JudgehostController.php
+++ b/webapp/src/DOMJudgeBundle/Controller/API/JudgehostController.php
@@ -627,7 +627,7 @@ class JudgehostController extends FOSRestController
     }
 
     /**
-     * Add a new JudgingRun. When relevant, finalize the judging.
+     * Add an array of JudgingRuns. When relevant, finalize the judging.
      * @Rest\Post("/add-judging-run/{hostname}/{judgingId}")
      * @Security("has_role('ROLE_JUDGEHOST')")
      * @SWG\Response(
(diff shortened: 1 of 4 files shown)


webapp/src/DOMJudgeBundle/Controller/API/JudgehostController.php
```diff
@@ -495,6 +495,18 @@ public function getNextJudgingAction(string $hostname)
             $result['judgingid'] = $judging->getJudgingid();
         });
 
+        $testcases = $submission->getProblem()->getTestcases();
```
nickygerritsen: This should still be added to the Swagger doc at some point

webapp/src/DOMJudgeBundle/Controller/API/JudgehostController.php
```diff
@@ -689,185 +701,91 @@ public function addJudgingRunAction(
         string $hostname,
         int $judgingId
     ) {
-        $required = [
```
nickygerritsen: Maybe it's better to have a separate endpoint for single / multiple runs? I think it now is pretty hard to denote this in Swagger?

webapp/src/DOMJudgeBundle/Controller/API/JudgehostController.php
```diff
@@ -689,185 +701,91 @@ public function addJudgingRunAction(
...
+    private function addMultipleJudgingRuns(Request $request, string $hostname, int $judgingId) {
+        $judgingRuns = json_decode($request->request->get('batch'), TRUE);
+        if (!is_array($judgingRuns)) {
+            throw new BadRequestHttpException(
+                sprintf("Argument 'batch' is not an array"));
+        }
+        /** @var Judgehost $judgehost */ // <--- FIXME: do this in the other commit
         $judgehost = $this->entityManager->getRepository(Judgehost::class)->find($hostname);
         if (!$judgehost) {
+            // FIXME: silent return??? also at the original place
```
nickygerritsen: The old v4 code did the same, but we might want to throw an exception instead?

webapp/src/DOMJudgeBundle/Controller/API/JudgehostController.php
```diff
@@ -689,185 +701,91 @@ public function addJudgingRunAction(
...
-                $problem    = $submission->getProblem();
-                $this->scoreboardService->calculateScoreRow($contest, $team, $problem);
-
-                // We call alert here before possible validation. Note that this means that these
-                // alert messages should be treated as confidential information.
-                $this->DOMJudgeService->alert($result === 'correct' ? 'accept' : 'reject',
-                                              sprintf("submission %s, judging %s: %s",
-                                                      $submission->getSubmitid(),
-                                                      $judging->getJudgingid(), $result));
+        // FIXME: why don't we do a similar check as three lines above?
```
nickygerritsen: Assuming you mean checking whether the judging exists: we definitely should

webapp/src/DOMJudgeBundle/Controller/API/JudgehostController.php
```diff
@@ -689,185 +701,91 @@ public function addJudgingRunAction(
...
-                // Log to event table if no verification required
-                // (case of verification required is handled in www/jury/verify.php)
-                if (!$this->DOMJudgeService->dbconfig_get('verification_required', false)) {
-                    if ($judging->getRejudgingid() === null) {
-                        $this->eventLogService->log('judging', $judging->getJudgingid(),
-                                                    EventLogService::ACTION_UPDATE,
-                                                    $judging->getCid());
-                        $this->balloonService->updateBalloons($contest, $submission, $judging);
-                    }
+            // FIXME: this check seems to be broken...
```
nickygerritsen: What is broken here?

webapp/src/DOMJudgeBundle/Controller/API/JudgehostController.php
```diff
@@ -1012,4 +930,169 @@ protected function giveBackJudging(int $judgingId)
                                              $judging->getCid());
         }
     }
+
+    /**
+     * @param string $hostname
+     * @param int $judgingId
+     * @param $testCaseId
```
nickygerritsen: At some point add more type hints here :)

webapp/src/DOMJudgeBundle/Controller/API/JudgehostController.php
meisterT: Can I leave this for you? :-)

webapp/src/DOMJudgeBundle/Controller/API/JudgehostController.php
meisterT: Well, on the one hand we don't really need the old endpoint anymore. On the other hand, it's still nice to have it around especially for swagger.
With swagger you'll never to do a batch update, so I think it's fine to just not document it in the Swagger API. What do you think?

webapp/src/DOMJudgeBundle/Controller/API/JudgehostController.php
meisterT: That was just a stale comment.

webapp/src/DOMJudgeBundle/Controller/API/JudgehostController.php
meisterT: I blame phpstorm :-)

webapp/src/DOMJudgeBundle/Controller/API/JudgehostController.php
meisterT: Done.

webapp/src/DOMJudgeBundle/Controller/API/JudgehostController.php
meisterT: Done.

meisterT:
some numbers in a working state:
* it's 72s with 10s update rate
* 79s with 1s update rate
* 104s with 0s update rate (after each test case)

webapp/src/DOMJudgeBundle/Controller/API/JudgehostController.php
nickygerritsen: OK this sounds reasonable. If at some point we want to get rid of the old style of doing this, we can fix it then.

PR #445, opened 2018-11-01
Setup curl handle once and reuse it between API requests.

thijskh:
This saves us on setup time and cURL will reuse the HTTP connection.

Not yet fully tested, do not merge.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -14,8 +14,7 @@ require(LIBVENDORDIR . '/autoload.php');
 
 require(ETCDIR . '/judgehost-config.php');
 
-$resturl = $restuser = $restpass = null;
-$endpoints = array();
+$endpoints = [];
 
 function read_credentials()
 {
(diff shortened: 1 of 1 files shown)


meisterT:
I also tried this once and the difference wasn't that big but I think we should do it anyway. I'll measure with my setup and post numbers here.

thijskh:
This is currently not complete since we can change api credentials between calls. Can fix that if we decide to go through with it.

judge/judgedaemon.main.php
```diff
@@ -46,6 +46,28 @@ function read_credentials()
...
+        curl_setopt($curl_handle, CURLOPT_RETURNTRANSFER, true);
+    }
+}
+
+function close_curl_handle()
+{
+    global $curl_handle;
+    if ( ! empty($curl_handle) ) {
+        curl_close($curl_handle);
+    }
```
meisterT: currently it's not necessary from the calls that I can see, but we should reset the handle here

judge/judgedaemon.main.php
```diff
@@ -46,6 +46,28 @@ function read_credentials()
     }
 }
 
+
+function setup_curl_handle()
+{
+    global $curl_handle, $restuser, $restpass;
```
meisterT: one curl_handle is not enough. We need one per endpoint, right?

judge/judgedaemon.main.php
thijskh: agreed

judge/judgedaemon.main.php
thijskh: yes, could create one per endpoint, or move the basic auth credential setting out of this function

judge/judgedaemon.main.php
```diff
@@ -70,39 +92,38 @@ function request(string $url, string $verb = 'GET', string $data = '', bool $fai
...
-    $ch = curl_init($url);
-    curl_setopt($ch, CURLOPT_USERAGENT, "DOMjudge/" . DOMJUDGE_VERSION);
-    curl_setopt($ch, CURLOPT_HTTPAUTH, CURLAUTH_BASIC);
-    curl_setopt($ch, CURLOPT_USERPWD, $restuser . ":" . $restpass);
-    curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+    curl_setopt($curl_handle, CURLOPT_URL, $url);
+
     if ($verb == 'POST') {
-        curl_setopt($ch, CURLOPT_POST, true);
+        curl_setopt($curl_handle, CURLOPT_POST, true);
```
meisterT: Adding 
`curl_setopt($curl_handle, CURLOPT_POST, false);`
two lines above gets past the second request. It still ends up with 404 later, but I'll keep digging what else we have to reset.

judge/judgedaemon.main.php
```diff
@@ -70,39 +92,38 @@ function request(string $url, string $verb = 'GET', string $data = '', bool $fai
...
         if (is_array($data)) {
-            curl_setopt($ch, CURLOPT_HTTPHEADER, array('Content-Type: multipart/form-data'));
+            curl_setopt($curl_handle, CURLOPT_HTTPHEADER, ['Content-Type: multipart/form-data']);
         }
     } elseif ($verb == 'PUT' || $verb == 'DELETE') {
-        curl_setopt($ch, CURLOPT_CUSTOMREQUEST, $verb);
+        curl_setopt($curl_handle, CURLOPT_CUSTOMREQUEST, $verb);
```
meisterT: this is the other option which has to be reset between requests

judge/judgedaemon.main.php
thijskh: Agreed, `curl_close()` promises to "delete the handle" but that's not fully resetting `$curl_handle` apparently.

thijskh:
My dev environment is a bit borked so I made this PR more as a sketch. Have added the things above except for the multi-endpoint support.

PR #447, opened 2018-11-04
[WIP] Declutter config options by adding a category.

meisterT:
We just keep adding options at random places in the defaultdata file and I figured it would be useful to group them together.
This is still missing a proper CSS style to look good, but I think it's already an improvement.


lib/lib.dbconfig.php
```diff
@@ -68,6 +68,7 @@ function dbconfig_init()
         $LIBDBCONFIG[$key] = array('value' => $val,
                                    'type' => $row['type'],
                                    'public' => $row['public'],
+                                   'category' => $row['category'],
```
nickygerritsen: This should also be done in the `DOMJudgeBundle\Entity\Configuration` class, but you will probably notice this when you move this page to Symfony 😄 .

meisterT:
The code doesn't allow edits yet, but it looks like this:
config

webapp/src/DOMJudgeBundle/Resources/views/jury/config.html.twig
```diff
@@ -0,0 +1,62 @@
+{% extends "@DOMJudge/jury/base.html.twig" %}
+{% import _self as self %}
+{% import "@DOMJudge/jury/jury_macros.twig" as macros %}
+
+{% block title %}Configuration - {{ parent() }}{% endblock %}
+
+{% block extrahead %}
+    {{ parent() }}
+    {{ macros.table_extrahead() }}
```
nickygerritsen: I am guessing we don't use datatables here. In that case the whole block and macro import can be removed.

webapp/src/DOMJudgeBundle/Resources/views/jury/config.html.twig
```diff
@@ -0,0 +1,62 @@
...
+{% endblock %}
+
+{% block content %}
+    <h1>Configuration</h1>
+    <div class="alert alert-danger" role="alert">View only page, edit values <a href="config.php">here</a></div>
+    <form>
+    {% for category in options %}
+        <div class="card">
```
nickygerritsen: Indenting looks a bit weird IMHO

lib/lib.dbconfig.php
meisterT: Done.

webapp/src/DOMJudgeBundle/Resources/views/jury/config.html.twig
meisterT: We need it, that's how I import the eyecandy toggles.

webapp/src/DOMJudgeBundle/Resources/views/jury/config.html.twig
meisterT: Is there an easy way to autoformat this?

nickygerritsen: Code menu in PHPStorm and then `Reformat code` should do the trick.

webapp/src/DOMJudgeBundle/Controller/Jury/ConfigController.php
```diff
@@ -0,0 +1,118 @@
...
+                    'description' => $option->getDescription()
+                );
+            }
+            $all_data[] = array(
+                'name' => $category,
+                'data' => $data
+            );
+        }
+        return $this->render('@DOMJudge/jury/config.html.twig', [
+            'options' => $all_data,
```
nickygerritsen: If you set a value like `edited` to true and pass it to Twig when settings have been saved, you can do something like this in the twig:

```twig
    {% if edited %}
        {% include '@DOMJudge/jury/refresh_scoreboard.html.twig' with {message: 'After changing specific settings, you might need to refresh the scoreboard.'} %}
    {% endif %}
```

webapp/src/DOMJudgeBundle/Resources/views/jury/jury_macros.twig
```diff
@@ -1,5 +1,7 @@
 {% macro table_extrahead() %}
+    <link rel="stylesheet" href="{{ asset('css/bootstrap-toggle.min.css') }}" type="text/css"/>
```
nickygerritsen: I think the original idea of @ubergeek42 was to put here the extra CSS/JS we need for tables (hence `table_extrahead`). You might want to define a `toggle_extrahead` or something.

www/jury/config.php
```diff
@@ -84,56 +84,69 @@
     "<tr class=\"thleft\"><th>Option</th><th>Value(s)</th><th>Description</th></tr>\n" .
     "</thead>\n<tbody>\n";
 
-$extra = ' class="config_input"';
```
nickygerritsen: This whole file can be removed now right?

webapp/src/DOMJudgeBundle/Controller/Jury/ConfigController.php
meisterT: Done.

webapp/src/DOMJudgeBundle/Resources/views/jury/config.html.twig
meisterT: Done.

webapp/src/DOMJudgeBundle/Resources/views/jury/jury_macros.twig
meisterT: Done.

www/jury/config.php
meisterT: Done.

PR #464, opened 2018-11-25
Prevent spurious cases where not all program output was captured.

eldering:
This was likely a race condition where after reading data from
the pipes, some more data was pushed and then the program exited
and a SIGCHLD was raised and captured, leaving unread data in the
pipes. To fix this, read from the pipes once more after SIGCHILD.

Also rename variable sigmask_empty and only initialize it once.

### DIFF
--- a/judge/runguard.c
+++ b/judge/runguard.c
@@ -151,6 +151,7 @@ int64_t memsize;
 rlim_t filesize;
 rlim_t nproc;
 size_t streamsize;
+int use_splice;
 
 pid_t child_pid = -1;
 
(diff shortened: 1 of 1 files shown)


judge/runguard.c
```diff
@@ -1239,72 +1311,38 @@ int main(int argc, char **argv)
...
+				nfds = max(nfds,child_pipefd[i][PIPE_OUT]);
 			}
 		}
 
+		errno = 0;
+		verbose("final call pselect, nfds = %d", nfds);
+		r = pselect(nfds+1, &readfds, NULL, NULL, &pipedelay, &emptymask);
+		verbose("pselect done, r = %d, errno = %d = %s", r, errno, strerror(errno));
+
+		if ( r>0 ) pump_pipes(readfds, data_read, data_passed);
```
meisterT: There could be more data in the pipe(s) after this.

judge/runguard.c
```diff
@@ -1239,72 +1311,38 @@ int main(int argc, char **argv)
...
+		for(i=1; i<=2; i++) {
+			if ( child_pipefd[i][PIPE_OUT]>=0 ) {
+				FD_SET(child_pipefd[i][PIPE_OUT],&readfds);
+				nfds = max(nfds,child_pipefd[i][PIPE_OUT]);
 			}
 		}
 
+		errno = 0;
+		verbose("final call pselect, nfds = %d", nfds);
+		r = pselect(nfds+1, &readfds, NULL, NULL, &pipedelay, &emptymask);
```
meisterT: I haven't tried but when rereading the code, I thought it would be possible to *know* when to read further, by remembering on which pipes we reached EOF. That would make this pselect call unnecessary.

Am I mistaken there?

judge/runguard.c
```diff
@@ -852,6 +860,71 @@ void setrestrictions()
...
+			if ( nread==-1 ) {
+				if (errno == EINTR || errno == EAGAIN || errno == EWOULDBLOCK) continue;
+				error(errno,"copying data fd %d",i);
+			}
+			if ( nread==0 ) {
+				/* EOF detected: close fd and indicate this with -1 */
+				if ( close(child_pipefd[i][PIPE_OUT])!=0 ) {
+					error(errno,"closing pipe for fd %d",i);
+				}
+				child_pipefd[i][PIPE_OUT] = -1;
```
meisterT: This is where we detect EOF. After pumping all data, these should be -1, right?
(See next comment.)

judge/runguard.c
meisterT: I would assume that even if data is not available right away, the read/ splice is blocking. So we could read until eof.

PR #468, opened 2018-11-26
Fix race condition in runguard.

meisterT:
When the child process (i.e. the team submission) is finished, we still
may have to process more data from the pipes even though we received
sigchild. Otherwise the program output (or stderr) could be truncated.

We saw this only happening *very* rarely for very fast submissions that
output either more than line or more than 4k (the buffer size).

### DIFF
--- a/judge/runguard.c
+++ b/judge/runguard.c
@@ -878,6 +878,9 @@ int main(int argc, char **argv)
 	struct itimerval itimer;
 	struct sigaction sigact;
 
+	int total_data_read = 0;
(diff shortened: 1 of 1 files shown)


meisterT:
I can't reply to the first comment directly so doing it here.

I tried to implement a fix with the least amount of code changes, as I find that cleaner than to couple it with a refactoring.

I also tried to read until there's no new data. I tried this approach, it's basically the first commit in this PR. Later commits each fix one of our tests.

PR #470, opened 2018-11-29
Provoke truncation bug in runguard and test it in travis.

meisterT:
Test how often the runguard bug appears on travis and how much time it takes to test for it.


meisterT:
Doesn't seem reproducible on travis. I'll probably create a script that you can run locally instead.

PR #471, opened 2018-12-06
Add JS snippet to remember last scroll location in local storage.

### LINKED ISSUE #461: Scoreboard refresh shouldn't scroll to top
When the scoreboard refreshes, it always scrolls to top, which is rather annoying if you are currently spending time looking at some part further down on the scoreboard, e.g. statistics or just some range of teams.

### Steps to reproduce

- Go to scoreboard.
- Scroll down.
- Wait for refresh.

### Expected behaviour

Refresh of scoreboard doesn't change what part of the scoreboard is currently viewed.

### Actual behaviour

Refresh of scoreboard scrolls to top.

meisterT:
Tested in Chrome and FF.

Fixes #461.

### DIFF
--- a/webapp/web/js/domjudge.js
+++ b/webapp/web/js/domjudge.js
@@ -825,6 +825,20 @@ function enableRefresh($url, $after) {
     }, $after*1000);
     refreshEnabled = true;
     setCookie('domjudge_refresh', 1);
+
+    if(window.location.href == localStorage.getItem('lastUrl')) {
+    	window.scrollTo(0, localStorage.getItem('scrollTop'));
+    } else {
+    	localStorage.setItem('lastUrl', window.location.href);
+    	localStorage.setItem('scrollTop', 42);
+    }
+    var scrollTimeout;
+    document.addEventListener('scroll', function() {
+      clearTimeout(scrollTimeout);
+      scrollTimeout = setTimeout(function() {
+        localStorage.setItem('scrollTop', $(window).scrollTop());
+      }, 100)
+    });
 }
 
 function disableRefresh() {


webapp/web/js/domjudge.js
```diff
@@ -825,6 +825,20 @@ function enableRefresh($url, $after) {
...
+
+    if(window.location.href == localStorage.getItem('lastUrl')) {
+    	window.scrollTo(0, localStorage.getItem('scrollTop'));
+    } else {
+    	localStorage.setItem('lastUrl', window.location.href);
+    	localStorage.setItem('scrollTop', 42);
+    }
+    var scrollTimeout;
+    document.addEventListener('scroll', function() {
+      clearTimeout(scrollTimeout);
```
nickygerritsen: Isn't doing stuff on scroll going to impact the performance of the browser quite a bit? I don't know, just wondering...

meisterT: nope, it doesn't.

meisterT:
> However this does not work for 'legacy' pages.
> So I am wondering whether we should use your approach for legacy pages or just rewrite those pages to use Symfony and twig?

I'm inclined to apply my fix now for all pages for the upcoming release.
Even if you do the AJAX refresh (which is preferred of course), then the JS snippet doesn't hurt (or break anything). So need to special case any page.

nickygerritsen:
> > However this does not work for 'legacy' pages.
> > So I am wondering whether we should use your approach for legacy pages or just rewrite those pages to use Symfony and twig?
> 
> I'm inclined to apply my fix now for all pages for the upcoming release.
> Even if you do the AJAX refresh (which is preferred of course), then the JS snippet doesn't hurt (or break anything). Then we don't need to special case any page.

I can agree to that :)

PR #474, opened 2018-12-07
Fix exception and undefined behavior in sort function.

meisterT:
If anyone wants to repro, I have a database dump where the exception is triggered.


webapp/src/DOMJudgeBundle/Utils/Scoreboard/Scoreboard.php
```diff
@@ -377,10 +377,14 @@ public static function scoreTiebreaker(TeamScore $a, TeamScore $b): int
...
+        if (isset($atimes[0]) && isset($btimes[0])) {
             if ($atimes[0] != $btimes[0]) {
                 return $atimes[0] <=> $btimes[0];
             }
+        } else if (!isset($atimes[0])) {
+            return -1;
+        } else {
+            return 1;
         }
         return 0;
```
thijskh: This is now dead code?

thijskh:
If both a and b times are empty, function should return 0, not -1.

Function no longer returns an integer when a[0] and b[0] are set but equal (because first if is true but second isn't). last return 0 should be outside of else.

meisterT:
Thanks for the review and the tests. I pushed an update that passes the tests.

webapp/src/DOMJudgeBundle/Utils/Scoreboard/Scoreboard.php
```diff
@@ -377,10 +377,16 @@ public static function scoreTiebreaker(TeamScore $a, TeamScore $b): int
         $btimes = $b->getSolveTimes();
         rsort($atimes);
         rsort($btimes);
-        if (isset($atimes[0])) {
+        if (isset($atimes[0]) && isset($btimes[0])) {
             if ($atimes[0] != $btimes[0]) {
```
thijskh: this condition can be removed: `<=>` will take care of that case.
then also the final `return 0`; can be removed since it's then dead code.

PR #478, opened 2018-12-20
Report freeze duration of "00:00:00" if not set.

### LINKED ISSUE #477: If no freeze time is set, API reports a huge scoreboard_freeze_duration
### Description of the problem / feature request

In the //contests/<cid> endpoint, give "00:00:00" for the freeze duration instead of some nonsense number in the case that the contest doesn't have a freeze.

It could be fixed by checking "!empty()" before calculating freeze duration [here in Entity/Contest.php](https://github.com/DOMjudge/domjudge/blob/2dc45f9ecb8368e0b284716afcb103d3e36a2cbd/webapp/src/DOMJudgeBundle/Entity/Contest.php#L1108), like how it already does in a few other places in this file): 

### Your environment
DOMjudge  6.0.2DEV (https://domjudge.bath.ac.uk/)
PHP 7.0.30-0+deb9u1.

### Steps to reproduce
- Create a contest with a Start time and an End time.
- Leave "Scoreboard freeze time" empty.
- Verify that the scoreboard is not frozen.

So far, so good

- Check the CLICS API endpoint at //api/contests/<cid>. Specifically look at scoreboard_freeze_duration

### Expected behaviour
Ideally "scoreboard_freeze_duration" wouldn't even show up if the scoreboard doesn't ever freeze, but it looks like the API spec requires that this is either always given as something non-null, or never given. The next-best behaviour would be to return "00:00:00" for this.

### Actual behaviour
The scoreboard freeze is given as number of minutes between the end time and the UNIX epoch because it subtracts "freeze time" from "end time" without checking if either is null.

For example:

{
"id": "6",
"name": "Bath 2018 December Contest 2",
"formal_name": "Bath 2018 December Contest 2",
"start_time": "2018-12-19T20:00:00.000+00:00",
"duration": "328:00:00.000",
"scoreboard_freeze_duration": "429564:00:00.000",
"penalty_time": 20,
"external_id": null,
"shortname": "bath18dec2",
"end_time": "2019-01-02T12:00:00.000+00:00"
}

This confuses scoreboard clients which go into resolve mode after the end of the contest.

hex539:
Leaves Contest::getDuration alone because it isn't as clear-cut what to do there.

Fixes #477

### DIFF
--- a/webapp/src/DOMJudgeBundle/Entity/Contest.php
+++ b/webapp/src/DOMJudgeBundle/Entity/Contest.php
@@ -1105,7 +1105,7 @@ class Contest
      */
     public function getScoreboardFreezeDuration()
     {
-        if (!empty($this->getFreezetime()) {
+        if (!empty($this->getFreezetime())) {
             return Utils::relTime($this->getEndtime() - $this->getFreezetime());
         } else {
             return Utils::relTime(0);
--- a/webapp/src/DOMJudgeBundle/Entity/Contest.php
+++ b/webapp/src/DOMJudgeBundle/Entity/Contest.php
@@ -1105,7 +1105,11 @@ class Contest
      */
     public function getScoreboardFreezeDuration()
     {
-        return Utils::relTime($this->getEndtime() - $this->getFreezetime());
+        if (!empty($this->getFreezetime()) {
+            return Utils::relTime($this->getEndtime() - $this->getFreezetime());
+        } else {
+            return Utils::relTime(0);
+        }
     }
 
     /**


thijskh:
This feels a bit hackish to me. Would prefer to improve the API specification.

hex539:
I would also prefer to improve the API specification (like I said on the bug), but I'd also like DOMjudge to work for the version of the API that exists now.

eldering:
Manually merged with a small tweak from "00:00:00" to null in 25dcb9cb540edceb3c00bdc56a9a8130b779dfb6

PR #480, opened 2019-01-02
Fix typo in runguard.

meisterT:
Also add some error checking.


eldering:
Looks good to me. The i was indeed a typo.

PR #487, opened 2019-01-14
Introduce separate roles for API reading and writing and getting source code using the API

nickygerritsen:


sql/mysql_db_defaultdata.sql
```diff
@@ -124,13 +124,16 @@ INSERT INTO `language` (`langid`, `externalid`, `name`, `extensions`, `require_e
...
+(1,  'admin',             'Administrative User'),
+(2,  'jury',              'Jury User'),
+(3,  'team',              'Team Member'),
+(4,  'balloon',           'Balloon runner'),
+(6,  'judgehost',         '(Internal/System) Judgehost'),
+(7,  'event_reader',      '(Internal/System) event_reader'),
+(8,  'full_event_reader', '(Internal/System) full_event_reader'),
+(9,  'api_reader',        'API reader'),
+(10, 'api_writer',        'API writer'),
+(11, 'source_reader',     'Source code reader');
```
eldering: I would rename this to `api_source_reader` to clarify that this applies to API access only.

Also, are we still using `{full_,}event_reader` anywhere?

eldering: Replying to myself: `full_event_reader` is still used in https://github.com/DOMjudge/domjudge/blob/master/www/feed/init.php#L22. Shouldn't we just drop that code? Or at least make it use `api_reader` or another common role.

nickygerritsen: I’d say let’s use `api_reader` and drop both old roles. Also I agree renaming to `api_source_reader`. Will fix these things on Wednesday

nickygerritsen: What do the others think? Using `reader` and `writer` also is strange: the writer should not be a general write user (although that might indicate the name is wrong in general)

eldering: I think the whole problem here is that we do not have a clearly defined and small set of real roles that we'd like to use. While `judgehost`, `team`, `jury` are all clear roles with clear permissions, the other API-access only roles are not (really). So for now I think it's ok to leave these names as is.

PR #490, opened 2019-01-16
Fixing crashes with compile.meta

DerBaer0:
During judging process, the following compile.meta file was created:
[compile.meta.txt](https://github.com/DOMjudge/domjudge/files/2764465/compile.meta.txt)

Preparing the compile.meta file does not work as intended and results
in crashes of the judgedaemon. By default, preg_replace does not match
newlines with $-sign. And therefore only the last line was surrounded with quotes. The regex option /m is required for multiline support.

(This happened on an Ubuntu 16.04 machine with PHP 7.0.30)


meisterT:
Looks good to me, thanks! Once travis is also happy, I'll merge it.

Issue #300, opened 2017-10-25
Implement awards endpoint


meisterT:
https://clics.ecs.baylor.edu/index.php/Contest_API#Awards

nickygerritsen:
#493 fixed this right? i.e. I think this can be closed.

PR #493, opened 2019-01-20
Implement awards endpoint

thijskh:
This is an attempt to implement the awards endpoint as requested in #300.

It currently only does the `/contests/<id>/awards` endpoint, not yet  `/contests/<id>/awards/<id>` but that is easy to add if the general approach is deemed OK.

The spec gives as an example the `organization-winner` award, since we do not have that concept in DOMjudge, and it's an example, I think it's fine not to output that.

### DIFF
--- a/webapp/src/DOMJudgeBundle/Controller/API/AwardsController.php
+++ b/webapp/src/DOMJudgeBundle/Controller/API/AwardsController.php
@@ -14,6 +14,7 @@ use FOS\RestBundle\Controller\Annotations as Rest;
 use Swagger\Annotations as SWG;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
 /**
  * @Rest\Route("/api/v4/contests/{cid}/awards", defaults={ "_format" = "json" })
@@ -53,13 +54,14 @@ class AwardsController extends AbstractRestController
      * @param Request $request
      * @return array
      * @Rest\Get("")
+     * @Rest\Get("/{requestedType}")
      * @SWG\Response(
      *     response="200",
      *     description="Returns the current teams qualifying for each award"
      * )
      * @throws \Exception
      */
-    public function getAwardsAction(Request $request)
+    public function getAwardsAction(Request $request, string $requestedType = null)
     {
         $public   = false;
         if ($this->DOMJudgeService->checkrole('jury') && $request->query->has('public')) {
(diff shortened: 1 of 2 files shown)


webapp/src/DOMJudgeBundle/Controller/API/AwardsController.php
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        $contest       = $this->entityManager->getRepository(Contest::class)->find($this->getContestId($request));
+        $isJury        = $this->isGranted('ROLE_JURY');
+        $accessAllowed = ($isJury && $contest->getEnabled()) || (!$isJury && $contest->isActive());
+        if (!$accessAllowed) {
+            throw new AccessDeniedHttpException();
+        }
+
+        $b = $contest->getB() ?? 0;
```
meisterT: nit: although we know what b means, I would prefer a variable name like $additionalBronzeMedals.

webapp/src/DOMJudgeBundle/Controller/API/AwardsController.php
```diff
@@ -0,0 +1,152 @@
...
+
+        $scoreboard = $this->scoreboardService->getScoreboard($contest, !$public, null, true);
+
+        $group_winners = $problem_winners = [];
+        foreach ($scoreboard->getTeams() as $team) {
+            if ( $scoreboard->isBestInCategory($team) ) {
+                $group_winners[$team->getCategoryId()][] = $team->getTeamid();
+            }
+            foreach($scoreboard->getProblems() as $problem) {
+                if ($scoreboard->solvedFirst($team, $problem)) {
```
meisterT: The only problem I have with that is that this is not guaranteed to be stable. If we still are judging the submission of another team that submitted earlier but takes longer to judge we will do it wrong (see last year's WF the first correct submission).

thijskh: I think it makes sense if the awards endpoint is consistent with what you see on the scoreboard. And I think we now paint the currently first already judged correct submission datkgreen. If we want to change that, I think it needs to happen in the ScoreboardService so all output formats do it the same.

webapp/src/DOMJudgeBundle/Controller/API/AwardsController.php
thijskh: Apropos medals, I was already a bit hestitant to encode their names and numers here, but I think this is the only place in DOMjudge that does something with the concept, correct?

webapp/src/DOMJudgeBundle/Controller/API/AwardsController.php
eldering: Agreed with both.
Also there is an issue already raised (not sure by whom) that the winner may need to be temporarily undefined when a submission is being judged that could bring a new team to first place. Although we may add an exception for that in the specification.

thijskh: I think those are different things. The concept of 'winner' (probably leader is more appropriate) is expected to change continuously during the contest. I think it makes sense that we report as 'winner' at any given time whoever is the first on the scoreboard. On the scoreboard, we don't remove ranks while something is being judged (because that would essentially be always since submissions are continuously flowing in - and why would indeed rank = 3  be differnt from rank =1). I think it's sufficiently clear that all ranks are provisional until the contest has been finalized.

This is different from first-to-solve. Those awards are expected to be final as soon as they are issued, that is, if a problem has been deemed as solved first by team A, it will not later be awarded to team B (excluding rejudgings). So here it makes more sense here not to award it before it is certain that there cannot be an earlier correct submission for this problem.

thijskh: In any case this discussion is not specific to this PR, as I believe the output of the API should match what you see on the scoreboard and get from the scoreboard API.

PR #507, opened 2019-02-06
Allow to rejudge a submission that has no judging

nickygerritsen:
So during the system test we had the case where we did a `DELETE FROM submission WHERE cid=x` to clean up an old contest and at the same time a judgehost would request a new judging.
We assign the judgehost to a submission and afterwards create a judging for it. If, however, there is a transaction lock on the `judging` table (which happens during the `DELETE FROM`), creating this judging fails.

The only way to fix this before was to manually remove the judgehost from the submission, because when you pressed the rejudge button we couldn't find any judgings to 'invalidate'.
With this PR we can actually press the button, because we now just remove the judgehost from the submission in this edge case.

### DIFF
--- a/webapp/src/DOMJudgeBundle/Controller/Jury/RejudgingController.php
+++ b/webapp/src/DOMJudgeBundle/Controller/Jury/RejudgingController.php
@@ -21,7 +21,6 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
-use Symfony\Component\HttpKernel\KernelInterface;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\Routing\Annotation\Route;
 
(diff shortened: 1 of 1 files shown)


eldering:
I had a quick glance. Nothing specific about the code, but do we really want to include this very special case in the code path? This is pretty exceptional, and only happens if you start deleting stuff. I think that it might be better to keep the code simpler and deal with this by hand when it arises.

nickygerritsen:
I can live with that if others agree

thijskh:
If you start issuing database queries outside DOMjudge, I think the onus is in you to issue the correct queries. So in the given example, also update the judings table with a query. Or stop the judgehosts first before you do it (which you can already do inside DOMjudge).

nickygerritsen:
> Or stop the judgehosts first before you do it (which you can already do inside DOMjudge).

I like this idea. I think we should expose it in the API, so that for next year we can use this during the system test. Agreed? If so, I'll create an issue to track this (and close this PR).

nickygerritsen:
> I think we should expose it in the API

We already do for a single judgehost. Does it make sense to also add an API action to update all judgehosts like we have in the UI? Or do we say that if people want that they should look over all judgehosts?

meisterT:
looping is fine

PR #508, opened 2019-02-07
Consider earlier submitted but still pending commits in solvedFirst()

thijskh:
It has been noted that it would be nice if we only marked something as 'first to solve' (darkgreen color in scoreboard and in the awards api) if there are no earlier submitted but still pending submissions for this problem.

This is one (first) attempt at solving that. It has an additonal cost of keeping the sortorder in the score matrix and looping over the matrix rows so we can check for each other team whether they indeed have earlier pending submissions. This in contrast to the earlier solution that used the scoreboard summary to determine whether this team was the one submitting the fast solution.

Other solutions may be possible. This PR is to see what we think of this, the added cost and whether that's worth it, or that there are better solutions to the problem.

### DIFF
--- a/webapp/src/DOMJudgeBundle/Utils/Scoreboard/Scoreboard.php
+++ b/webapp/src/DOMJudgeBundle/Utils/Scoreboard/Scoreboard.php
@@ -291,7 +291,7 @@ class Scoreboard
                 $problemId = $contestProblem->getProbid();
                 // Provide default scores when nothing submitted for this team + problem yet
                 if (!isset($this->matrix[$teamId][$problemId])) {
-                    $this->matrix[$teamId][$problemId] = new ScoreboardMatrixItem(false, 0, 0, 0, 0);
+                    $this->matrix[$teamId][$problemId] = new ScoreboardMatrixItem(false, 0, 0, 0, 0, 0);
                 }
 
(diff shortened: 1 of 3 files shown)


meisterT:
Did you measure how much that slows down the scoreboard rendering?

thijskh:
Not yet.

meisterT:
Can we just compute a new number (e.g. earliestNotIncorrectTimes) per problem where we currently compute the best times?

earliestNotIncorrectTimes is the earliest submission of any team which is either not judged yet or correct. That would save the loop here.

eldering:
> Can we just compute a new number (e.g. earliestNotIncorrectTimes) per problem where we currently compute the best times?
> 
> earliestNotIncorrectTimes is the earliest submission of any team which is either not judged yet or correct. That would save the loop here.

Completely agree. Only to nitpick: I would call it `earliestPotentialCorrect` (if only to prevent too ugly camel-casing).

thijskh:
This is not so trivial since we don't store that information in the scorecache. I've created a separate PR that does store it in the scorecache and then uses the besttime summary row to determine this: #509

eldering:
This PR can be closed since #509 is a better solution, right? (Irrespective of whether we implement the FTS logic.)

thijskh:
I think in any case this is not the desired approach, yes.

PR #509, opened 2019-02-09
Only show problem as First To Solve if there are no earlier pending submissions

thijskh:
The time in scorecache is now set to the time of the earliest pending
or correct submission, while before it was only set for the earliest.

We can use this to determine if we are ready to award the first-to-solve
prize yet. We update the summary of the minimum time for any cell which
has a correct or pending submission. We only show that summary line
when a dark green cell has actually been awarded.

### DIFF
--- a/webapp/src/DOMJudgeBundle/Resources/views/partials/scoreboard_table.html.twig
+++ b/webapp/src/DOMJudgeBundle/Resources/views/partials/scoreboard_table.html.twig
@@ -86,6 +86,7 @@
     <tbody>
     {% set previousSortOrder = -1 %}
     {% set previousTeam = null %}
+    {% set firsttosolveknown = [] %}
     {% for score in scoreboard.scores if (limitToTeams is null or score.team.teamid in limitToTeamIds) %}
         {% set classes = [] %}
         {% if score.team.category.sortorder != previousSortOrder %}
@@ -201,6 +202,7 @@
                         {% set scoreCssClass = 'score_correct' %}
                         {% if scoreboard.solvedFirst(score.team, problem) %}
                             {% set scoreCssClass = scoreCssClass ~ ' score_first' %}
+                            {% set firsttosolveknown = firsttosolveknown|merge([problem.probid]) %}
                         {% endif %}
                     {% elseif showPending and scoreboard.matrix[score.team.teamid][problem.probid].numberOfPendingSubmissions > 0 %}
                         {% set scoreCssClass = 'score_pending' %}
@@ -302,7 +304,7 @@
 
                             <i class="fas fa-clock fa-fw"></i>
                             <span style="font-size:90%;" title="first solved">
-                                {% if scoreboard.summary.problem(problem.probid).bestTimeInMinutes(0) is not null %}
+                                {% if problem.probid in firsttosolveknown %}
                                     {{ scoreboard.summary.problem(problem.probid).bestTimeInMinutes(0) }}min
                                 {% else %}
                                     n/a
--- a/webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
+++ b/webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
@@ -323,6 +323,10 @@ class ScoreboardService
         $correctJury     = false;
         $correctPubl     = false;
 
+        // Loop over the submissions to find whether it is correct
+        // and update the times/scores. We stop the loop at the
+        // first correct solution, as we don't want to count any
+        // submissions after that.
         foreach ($submissions as $submission) {
             /** @var Judging|null $judging */
             $judging    = $submission->getJudgings()->first() ?: null;
--- a/webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
+++ b/webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
@@ -328,6 +328,10 @@ class ScoreboardService
             $judging    = $submission->getJudgings()->first() ?: null;
             $submitTime = $contest->getContestTime((float)$submission->getSubmittime());
 
+            $timeJury    = $submitTime;
+            if (!$submission->isAfterFreeze()) {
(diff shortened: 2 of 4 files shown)


webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
```diff
@@ -347,10 +351,8 @@ public function calculateScoreRow(Contest $contest, Team $team, Problem $problem
             // if correct, don't look at any more submissions after this one
             if ($judging->getResult() == Judging::RESULT_CORRECT) {
                 $correctJury = true;
-                $timeJury    = $submitTime;
                 if (!$submission->isAfterFreeze()) {
                     $correctPubl = true;
-                    $timePubl    = $submitTime;
                 }
                 // stop counting after a first correct submission
```
meisterT: It took me a while to find this comment/break to understand why it's correct (which is obvious in hindsight). Perhaps it's easier to grok if that is announced at the loop header that we only count until the first correct submission.

webapp/src/DOMJudgeBundle/Resources/views/partials/scoreboard_table.html.twig
```diff
@@ -302,7 +304,7 @@
 
                             <i class="fas fa-clock fa-fw"></i>
                             <span style="font-size:90%;" title="first solved">
-                                {% if scoreboard.summary.problem(problem.probid).bestTimeInMinutes(0) is not null %}
+                                {% if problem.probid in firsttosolveknown %}
```
nickygerritsen: We still use the non-Symfony scoreboard for the public and team pages (as these parts are not in Symfony yet at all). Shouldn't we change the logic there too?

thijskh: Unfortunately, yes.

webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
eldering: Moving the code that sets `$time{Jury,Publ}` outside of the `if ($judging->getResult() == Judging::RESULT_CORRECT)` condition means that it is triggered by *all* submissions. Unless I misunderstand something, we want this to be set only for potentially correct submissions, so those that are either correct or still pending.

eldering:
I think we should also rename the columns `scorecache.solvetime_{jury,public}` to reflect the slightly changed meaning, or at least update the comments on those columns.

webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
thijskh: It does not really matter. This function is just about filling the cache. This column only has relevance when there is a pending or correct submission.

Before this change, it would be set to int `0` when a problem was not yet solved.

thijskh:
Yes, would be nice to change that, would prefer to do this at the earliest after the wf2019 branch is obsolete so that branch and master do not become mutually incompatible.

webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
eldering: I still don't quite agree. What if there are 2 submissions, say the first pending and the second correct (to take an extreme case). In that case the time recorded using `$timeJury` should be that of the first, pending submission if it is to be interpreted as the earliest potential solve time. But in the current loop (assuming it's ordered by submission id/time) it would be that of the second submission.
How is `$timeJury` that is stored in `scorecache.solvetime_restricted` used later?

thijskh: This case indeed futher complicates the issue.

Would it also mean that if there's indeed a submission judged correct received at 20 but a pending one received at 18, we should also not show the score value of "20" on the scoreboard at all?

eldering:
Indeed, we should not. We should only show the first solve time once there are *no* pending earlier submissions than the first correct one.

Basically, a client should be guaranteed that FTS once set, never changes, unless there is a rejudging.

On February 13, 2019 9:03:55 AM GMT+01:00, Thijs Kinkhorst <notifications@github.com> wrote:
>thijskh commented on this pull request.
>
>
>
>>                  if (!$submission->isAfterFreeze()) {
>                     $correctPubl = true;
>-                    $timePubl    = $submitTime;
>                 }
>                 // stop counting after a first correct submission
>
>This case indeed futher complicates the issue.
>
>Would it also mean that if there's indeed a submission judged correct
>received at 20 but a pending one received at 18, we should also not
>show the score value of "20" on the scoreboard at all?
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/DOMjudge/domjudge/pull/509#discussion_r256280773

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

thijskh:
So but not only FTS, but even just a S

eldering:
> So yes, but my point is that this is now broader than _first_ to solve. That we need to change the scoreboard to not show a problem as _solved_ in general (or not give any score for it, at least?) until we are sure that there are no pending earlier submissions for the same team.

Ah, you mean the solve time in the scoreboard cell for that team,problem? I agree that strictly speaking probably we should not show it. I'm not sure thought it is worth the effort. The reason is that that value is not really viewed as much as the FTS.

But I'll raise this point with the others and see what they think.

nickygerritsen:
> But I'll raise this point with the others and see what they think.

I put this, together with the issue that FTS is never final until the scoreboard is final (because you can rejudge something that happened before) on the agenda for the next CLICCS meeting already :)

webapp/src/DOMJudgeBundle/Resources/views/partials/scoreboard_table.html.twig
thijskh: Fortunately, no.

Issue #513, opened 2019-03-01
FHS compliant build is broken


GugolPlex:
### Description of the problem

````configure --enable-fhs```` fails to create a valid configuration and the frontend fails.

  I'm aware of #280 but it seems the problem was not completely solved.

### Environment

- DOMjudge 6.0.3 (also tested with github version)
- Debian 9.8, 64 bits
- Apache

### Steps to reproduce

After install all prerequisites and untar the source code from https://www.domjudge.org/releases/domjudge-6.0.3.tar.gz :

    ./configure --enable-fhs
    [ ... ]
    Summary:
     * project.............: DOMjudge 6.0.3
     * prefix..............: /usr/local
    [ ... ]

     * documentation.......: ${datarootdir}/doc/${PACKAGE_TARNAME}

     * domserver...........: 
        - bin..............: ${exec_prefix}/bin
        - etc..............: ${prefix}/etc/domjudge
        - lib..............: ${exec_prefix}/lib/domjudge
        - libvendor........: ${exec_prefix}/lib/domjudge/vendor
        - libwww...........: ${exec_prefix}/lib/domjudge/www
        - libsubmit........: ${exec_prefix}/lib/domjudge/submit
        - log..............: ${prefix}/var/log/domjudge
        - run..............: ${prefix}/var/run/domjudge
        - sql..............: ${datarootdir}/domjudge/sql
        - submit...........: ${prefix}/var/lib/domjudge/submissions
        - tmp..............: /tmp
        - www..............: ${datarootdir}/domjudge/www
        - webapp...........: 
    [ ... ]

  The webapp variable is left uninitialized and ````make domserver; make install-domserver```` ends creating files in ````/web````  (root directory). Apache is unable to serve DOMjudge.

  This behaviour also occurs when using the source code from GitHub (doing ````make dist```` first, as said in the admin manual and suggested in #280).

  I've been looking for the problem and ````configure[.ac]```` seems to forget configuring ````webapp```` when FHS compliant is enabled:

    [ configure.ac, near line 140 ]
    if test "x$FHS_ENABLED" = xyes ; then
    	AC_SUBST(domserver_root,        '')
    	AC_SUBST(domserver_bindir,      $bindir)
    	AC_SUBST(domserver_etcdir,      $sysconfdir/${PACKAGE_TARNAME})
    	AC_SUBST(domserver_sqldir,      $datadir/${PACKAGE_TARNAME}/sql)
[...]

PR #514, opened 2019-03-01
Fix FHS setup

nickygerritsen:
This is the first part of fixing the FHS setup, which was broken (#513).

However there is still a problem that I can not seem to solve: composer generates some files in `vendor/composer` called `autoload_*` that contain paths that expect the relative location of the `vendor` directory to not change with respect to `webapp`.
But during an FHS install we do change these, making it so that the autoloader can not find `AppKernel`, `AppCache` and any classes in `webapp/src`.

What basically needs to happen:
* In `autoload_{psr4,classmap,files,namespaces}.php` replace the line that reads `$baseDir = dirname(dirname($vendorDir));` with `$baseDir = dirname('@domserver_webappdir@');`
* In `autoload_static.php` replace all (three) occurences of `__DIR__ . '/../../..'` with `'@domserver_webappdir@' . '/..'`

where `@domserver_webappdir@` is the webappdir as we have configured it.

I don't know what is the best way to do this? Should we just add this to `lib/Makefile` under `install-domserver` and `install-judgehost` with two `sed` commands? That feels ugly...


nickygerritsen:
Our symlinks in `webapp/web`, `webapp/web/js` and `webapp/web/css` are also using relative paths that are broken. These symlinks are as they are committed on git. How do we fix these?

nickygerritsen:
The webapp symlinks are now fixed with some `for`, `if`, `grep`, `sed`, `readlink` and `ln` magic in the Makefile. It feels to me like we can improve this still but it's a start.

webapp/Makefile
```diff
@@ -44,6 +50,30 @@ install-domserver:
...
+			target=$$(readlink $$link) ; \
+			if echo $${target} | grep -q '^..\/..\/..\/lib\/vendor' ; then \
+				rm $$link ; \
+				realtarget=$(DESTDIR)$(domserver_libvendordir)$$(echo $${target} | sed 's/..\/..\/..\/lib\/vendor//') ; \
+				ln -s $$realtarget $$link ; \
+			fi \
+		done ; \
+	done
+	for d in app bin src web tests ; do \
 		chmod -R a+rX      $(DESTDIR)$(domserver_webappdir)/$$d ; \
```
eldering: Is there a reason to split this `chmod` from the `cp -R` above? Keeping them in the same for loop would simplify the code.

webapp/Makefile
```diff
@@ -44,6 +50,30 @@ install-domserver:
...
+# Now change all relative symlinks in webapp/web to their correct paths
+	for link in $$(find $(DESTDIR)$(domserver_webappdir)/web -maxdepth 1 -type l); do \
+		target=$$(readlink $$link) ; \
+		if echo $${target} | grep -q '^..\/..\/lib\/vendor' ; then \
+			rm $$link ; \
+			realtarget=$(DESTDIR)$(domserver_libvendordir)$$(echo $${target} | sed 's/..\/..\/lib\/vendor//') ; \
+			ln -s $$realtarget $$link ; \
+		fi \
+	done
+	for dir in css js ; do \
```
eldering: What about merging this second set of for loops over links into the first one with a `-maxdepth 2` option? Replacing the sed command with `sed 's!^.*\.\./\.\./lib/vendor/!/!'`. Note escaping the dots to only let them match a a literal `.` and using a different separator `!` to make the sed expression more readable since then the `/` does not have to be escaped anymore.

webapp/Makefile
```diff
@@ -1,8 +1,14 @@
 ifndef TOPDIR
 TOPDIR=..
 endif
+
+REC_TARGETS = domserver
+
 include $(TOPDIR)/Makefile.global
 
+# Subdirectories to recurse into for REC_TARGETS
+SUBDIRS = app
```
eldering: Given that we now have a `Makefile` under `webapp/app` I think it makes sense to move all things rules that target files under there from `webapp/Makefile` into `webapp/app/Makefile`.

lib/Makefile
```diff
@@ -12,17 +12,27 @@ $(OBJECTS): %$(OBJEXT): %.c %.h
 clean-l:
 	rm -f $(OBJECTS)
 
+# Change baseDir in composer autogenerated files
+define fix_composer_paths
+	for file in autoload_psr4.php autoload_classmap.php autoload_files.php autoload_namespaces.php ; do \
+		sed -i "s#^\$$baseDir = .*#\$$baseDir = dirname('$(DESTDIR)$(domserver_webappdir)');#" $(1)/composer/$$file ; \
+	done
+	sed -i "s#__DIR__ . '/../../..'#'$(DESTDIR)$(domserver_webappdir)' . '/..'#" $(1)/composer/autoload_static.php
```
eldering: I would make this a bit more robust by matching and replacing up to `/webapp/`. Then things won't break if the number of relative directories change or if some other file gets loaded from `/../../..`.

eldering:
Rebased and committed. I will still address the last two points.

PR #516, opened 2019-03-02
Support R language

vicrac:
As in title, I would like to add R language support to domjudge. Is there anything else besides what I already did I need to especially take care of? Any help/advices/hints very much welcome :smile:


nickygerritsen:
There is an sql upgrade script for every version where you also need to add the R language. I’m on my phone but if you can’t find it I’ll point it out later.

I think that is all that is needed still

sql/files/defaultdata/r/run
```diff
@@ -0,0 +1,55 @@
...
+# Check if entry point is valid
+if [ ! -r "$MAINSOURCE" ]; then
+    echo "Error: main source file '$MAINSOURCE' is not readable" >&2
+    exit 1
+fi
+
+# Write executing script:
+cat > "$DEST" <<EOF
+#!/bin/sh
+# Generated shell-script to execute python interpreter on source.
```
aginiewicz: comment about python left after copy-paste

sql/files/defaultdata/r/run
```diff
@@ -0,0 +1,55 @@
...
+MEMLIMIT="$1" ; shift
+MAINSOURCE="${ENTRY_POINT:-$1}"
+
+# Report the entry point, so it can be saved, e.g. for later replay:
+if [ -z "$ENTRY_POINT" ]; then
+    echo "Info: detected entry_point: $MAINCLASS"
+fi
+
+# Check syntax
+Rscript "$@"
```
aginiewicz: won't it hang and wait for input if there is reading from stdin inside R script? I don't know of any pycompile alternative on R - there is lintr but it's external package not available in default install. Maybe skip this like in for example awk language?

vicrac: I was thinking of something like:
```
R -e "install.packages('lintr')"
lint_command="library(lintr); lint('$@')"
R -e "$lint_command"
```
but for a reason I don't know it's not working as expected, in particular it doesn't return non-zero exit code (but only prints result to output) while error is detected. Yes, this check should be removed.

sql/upgrade/upgrade_6.0.0_6.1.0DEV.sql
```diff
@@ -68,6 +68,7 @@ DELETE FROM `role` WHERE `role` IN ('print', 'event_reader', 'full_event_reader'
 
 UPDATE `language` SET `externalid` = 'ada' WHERE langid = 'adb';
 UPDATE `language` SET `externalid` = 'ruby' WHERE langid = 'rb';
+UPDATE `language` SET `externalid` = 'R' WHERE langid = 'r';
```
nickygerritsen: Instead of this, you need an
```
INSERT INTO `language` ...
```

as when the upgrade from 6.0 to 6.1 happens R will not be in the DB yet (because R is introduced in 6.1)

vicrac: I don't fully understand - it is not enough to add r to db in mysql_db_defaultdata.sql script?

nickygerritsen: There are two paths for DOMjudge:

* A fresh installation. This will use `mysql_db_defaultdata.sql`.
* An upgrade from a previous version. This will have used an old version of `mysql_db_defaultdata.sql` which didn't contain R and then apply all `sql/upgrade/upgrade_*` files that are not applied yet in order to get to the latest schema
That is why it is needed in two places.

vicrac: I see, I pushed a fix

nickygerritsen: Could you move it two lines down with an extra newline, so it is its separate block?
Sorry for nitpicking 🙈

vicrac: Sure

vicrac:
Oh, I also wanted to ask how does this change go with the web interface, I saw a bunch of `webapp/*.js` files which referenced language names (e.g. `webappweb/js/ace/ext-settings_menu.js`), but thet looked generated automatically, do I need to do something about it?

nickygerritsen:
> Oh, I also wanted to ask how does this change go with the web interface, I saw a bunch of `webapp/*.js` files which referenced language names (e.g. `webappweb/js/ace/ext-settings_menu.js`), but thet looked generated automatically, do I need to do something about it?

We have some automatic language detection in the ACE editor, but that should automatically work.
However, we also have some automatic entry point detection code which might need to be updated.

See https://github.com/DOMjudge/domjudge/blob/master/webapp/src/DOMJudgeBundle/Resources/views/team/partials/submit_scripts.html.twig
But if the entrypoint is just the filename, nothing has to be done.

nickygerritsen:
It seems the submit client doesn't detect the R submission correctly. From the Travis log:

```
[2019-03-02 15:42:44] Submitting 'test-hello.R' [Mar 02 15:42:44.863] ../submit/submit[5583]: error: no language specified or detected
[2019-03-02 15:42:44] Type '../submit/submit --help' to get help.
```

I think because `allow_submit` = 0, which is fine for Travis...
Did you test locally whether the submit client understands it if you submit the file? I think it should be ok 😄

vicrac:
No, I didn't test it locally. I actually wanted to ask for it on the mailing list, but if we're here already :smile:  Is there any local development workflow that woul allow me to setup whole application without need to installing db and all stuff separetely? To run domjudge on server I use docker and docker-compose (you can check out my repo with this configuration, called `nomad`, name after a competition this is prepared for :smile: ), it would be convenient for me if I can possibly rebuild image and run a fresh instance of app with my local changes included, is there something like that? I saw your packaging repo and Dockerfiles used to build images, but I am not sure how can I possibly use them now, could you tell me how do you personally manage this kind of things?

nickygerritsen:
You should be able to use the `docker-compose.yml` file from the `domjudge` repo itself. That will set up a MariaDB server in Docker and set up a container with nginx, php-fpm, the judgedaemons and a basic install in dev-mode, which will work from the repo files, so any changes you make will immediately be visible. It uses this docker image: https://hub.docker.com/r/domjudge/domjudge-contributor

vicrac:
Unfortunately, something is wrong (and it looks non-trivial to me) with docker-compose, as output from `docker-compose up` command (last few lines) for me is as follows:
```
domjudge_1  | checking for gcc... gcc
domjudge_1  | checking whether the C compiler works... no
domjudge_1  | configure: error: in `/domjudge':
domjudge_1  | configure: error: C compiler cannot create executables
domjudge_1  | See `config.log' for more details
domjudge_1  | make: *** [maintainer-conf] Error 77
domjudge_1  | Makefile:176: recipe for target 'maintainer-conf' failed
```

nickygerritsen:
Are you on Linux? I think it is a permission problem. We run the commands inside the container as a domjudge user, which probably can’t write in the current dir

vicrac:
Yes, I am on Linux, re-running with sudo didn't change anything, here's my config.log:
[config.log](https://github.com/DOMjudge/domjudge/files/2922307/config.log)

nickygerritsen:
It seems our version detection in https://github.com/DOMjudge/domjudge/blob/master/configure.ac#L3 somehow adds a newline (maybe a `\r`?) to the version number. Could you try replacing the `tr -d '\n'` with `tr -d '\n\r'` and see if that would help? If I remove the `tr` locally I seem to get the same kind of output as you.

vicrac:
I am sorry to say it, but I did it, but different error occured :cry: Output:
```
domjudge_1  | make[1]: Entering directory '/domjudge/etc'
domjudge_1  | touch initial_admin_password.secret
domjudge_1  | chmod go= initial_admin_password.secret
domjudge_1  | ./genadminpassword >> initial_admin_password.secret
domjudge_1  | /bin/sh: 1: ./genadminpassword: not found
domjudge_1  | make[1]: *** [initial_admin_password.secret] Error 127
domjudge_1  | Makefile:40: recipe for target 'initial_admin_password.secret' failed
domjudge_1  | make[1]: Leaving directory '/domjudge/etc'
domjudge_1  | /domjudge/Makefile.global:73: recipe for target 'domserver' failed
domjudge_1  | make: *** [domserver] Error 1
```

nickygerritsen:
Can you check if `etc/genadminpassword` is actually there? It should as it is on git(see https://github.com/DOMjudge/domjudge/blob/master/etc/genadminpassword ).
If not or if the Docker container can't run it, something strange is wrong...

vicrac:
Yes, it's there, moreover, when I rerun it also failed, but this time, it could not find `./genresatpicredentials` file, which is also there :thinking:

nickygerritsen:
I am lost... maybe your git `core.autocrlf` is set to true? Which might break line endings on Linux...
Otherwise I don’t know

vicrac:
Still cannot run it locally, anyway, could you test it for me then and possibly tell me what's wrong with ci for this test so I can possibly fix it?

vicrac:
After quick research I found a possible explanation for the second issue you mentioned:
https://stackoverflow.com/questions/6307836/fatal-error-cannot-mkdir-r-tempdir
It seems that user executing R script has not permissions to write to /tmp folder, I don't know if it's docker's or domjudge's itself issue.

But I don't know why do you use litter? Rscript is the default frontend for R and it doesn't require separate installation.

sql/files/defaultdata/r/run
```diff
@@ -0,0 +1,52 @@
...
+# NOTICE: when this compiler script is used with USE_CHROOT on, the
+# chroot requires r to be installed
+
+DEST="$1" ; shift
+MEMLIMIT="$1" ; shift
+MAINSOURCE="${ENTRY_POINT:-$1}"
+
+# Report the entry point, so it can be saved, e.g. for later replay:
+if [ -z "$ENTRY_POINT" ]; then
+    echo "Info: detected entry_point: $MAINCLASS"
```
eldering: `MAINCLASS` is undefined here. I see that this is a copy-paste from a mistake that was already there in other compile scripts for scripted languages, though, so actually my mistake ;-)
It should probably be `MAINSOURCE` instead, or in case of a multifile submissiosn, any other file that the compile script can detect as being the primary (I have no experience with R, so don't know whether this makes sense).

eldering:
> I added SQL to create the R executable, as that was still missing. Hope you don't mind.
> 
> There are still two issues:
> 
>     * We somehow need to update the executable zipfile contents in the DB when running the upgrade script. @eldering any idea how to do this?

A bit dirty, but you can copy the `UPDATE executable SET zipfile = ...` line from `sql/mysql_db_files_defaultdata.sql` into the SQL upgrade script after you ran `make dist`.

>     * When I try to submit your default `test-hello.R` inside my setup (after `sudo apt install littler` inside the chroot), I get the following error:
> 
> 
> ```
> Fatal error: cannot create 'R_TempDir'
> ```
> 
> I'm not sure why it does that and how to fix it, as I have no skills in R

Replying to your chat on IRC: I'm a little hesitant to add extra mountpoints and especially submission-writable ones or directories. This creates possible ways to circumvent the sandbox, store state between judging runs, etc. The best I can think of is to create a writable subdirectory of the judging run working directory, so  something like `output/judgings/$JUDGEHOST/endpoint-default/cX-sYYY-jZZZ/testcase001/write_tmp`. I would prefer to keep this out of the main branch, or at least out of default configuration though.

nickygerritsen:
Yeah perhaps we can add a config option (disabled by default) that creates this dir? And then somewhere (in the run script of R?) note that the option is required for R?

sql/files/defaultdata/r/run
vicrac: Why do we actually test in here wheter ENTRY_POINT is not set before we log it is detected?

vicrac: @aginiewicz How about multifile submissions? Is it possible/desired with R?

aginiewicz: I think multifile submissions should be possible with R, because you can use source to load other script.

nickygerritsen: Does R in that case require you to specify the main file first? Or *only* specify the first file? Or does it not matter

aginiewicz:
As for R, it might be useful to set TMPDIR environment variable, then R tries to create unique directory with the name $TMPDIR/RtmpXXXXXX where XXXXXX is a unique id

sql/files/defaultdata/r/run
aginiewicz: Only the first file, other files are sourced on demand

nickygerritsen:
> As for R, it might be useful to set TMPDIR environment variable, then R tries to create unique directory with the name $TMPDIR/RtmpXXXXXX where XXXXXX is a unique id

Yeah that is my current idea: if we add the config option to create a temp writable dir, we can use that in R.

I will try to come up with an implementation for that

eldering:
I would suggest adding that config option to `etc/judgehost-config.php` so that it sits near `USE_CHROOT`, not visible by default (I don't want people to accidentally turn this on without understanding the implications), and closer to the code using it (not loaded from DB, via API, ...)

nickygerritsen:
> I would suggest adding that config option to `etc/judgehost-config.php` so that it sits near `USE_CHROOT`, not visible by default (I don't want people to accidentally turn this on without understanding the implications), and closer to the code using it (not loaded from DB, via API, ...)

That makes sense 😄

nickygerritsen:
Woohoo after some fighting it now works. If you set `CREATE_WRITABLE_TEMP_DIR` to true it will create a writable directory and set `TEMPDIR` to that location and I can correctly run the sample R program inside a chroot.

sql/files/defaultdata/r/run
nickygerritsen: @vicrac I think that that means line 46 of runscript needs to change to only contain `$MAINSOURCE` and not `\$@` but I might be wrong. cc @eldering

nickygerritsen: I think I *am* mistaken as we do the same for Python and others so I think this is already fine but Jaap will (hopefully) confirm.

eldering: The reason to log ENTRY_POINT (which for scripted languages normally is the main source file) is a bit of a technical DOMjudge/Contest API specification issue. Even when the user didn't specify an entry point, we still want to make sure that if we forward the submission to another system (or e.g. rejudge later) that we deterministically use the same entry point as the first time.

eldering: > @vicrac I think that that means line 46 of runscript needs to change to only contain `$MAINSOURCE` and not `\$@` but I might be wrong. cc @eldering

Indeed. Also if you have a multifile submission and no entry point was specified then it would be good to add some automagic code to detect the main source, that is, the entry point, and log that on this line above.

nickygerritsen: But why don't we do these things (both only passing the main file as well as detecting entry point using a "smarter" way than only picking the first file) for Python2/3? As in this is a direct copy from the python2/3 run-script so I think doing more "smart" things should be done for all these scripts at the same time?

Also regarding `\$@`: it is escaped so these are not the *other* files passed to the runscript, but the arguments passed to the actual program, so that should be fine already

eldering: > But why don't we do these things (both only passing the main file as well as detecting entry point using a "smarter" way than only picking the first file) for Python2/3? As in this is a direct copy from the python2/3 run-script so I think doing more "smart" things should be done for all these scripts at the same time?

Agreed, we should (besides fixing the bug `MAINCLASS` -> `MAINSOURCE`).

> Also regarding `\$@`: it is escaped so these are not the _other_ files passed to the runscript, but the arguments passed to the actual program, so that should be fine already

Indeed.

nickygerritsen: > > [...]
> 
> Agreed, we should (besides fixing the bug `MAINCLASS` -> `MAINSOURCE`).
> 
> > [...]
> 
> Indeed.

So then I'd say this PR is complete for now, unless of course you or someone finds other things that should be fixed, right? I'll add you as a reviewer and if you approve, let's merge this and then do the changes to the scripts. I think we can copy these changes from the WF branch, where we do more logic: we filter out files with wrong extensions.

sql/files/defaultdata/r/run
```diff
@@ -0,0 +1,57 @@
...
+# Check if entry point is valid
+if [ ! -r "$MAINSOURCE" ]; then
+    echo "Error: main source file '$MAINSOURCE' is not readable" >&2
+    exit 1
+fi
+
+# Enable creating writable directory - needed by R to store temporary files
+# via `tempfile` function - if this is ommited, we get error like:
+# Fatal error: cannot create 'R_TempDir'
+export CREATE_WRITABLE_TEMP_DIR=1
```
nickygerritsen: Sorry, I was not entirely clear: this should not be needed as we do this in testcase_run.sh automatically 😄 .

vicrac: Oh, I do not completely understand (sorry :sweat_smile: ), but does it mean I have to set this variable as env variable inside my container with application instance? I am a little lost in here

nickygerritsen: No so if you want this feature, change `etc/judgehost-config.php` to have `define('CREATE_WRITABLE_TEMP_DIR', true)` instead of `false`. For in a container you can probably mount a modified version of that file in the correct spot.

vicrac: Could possibly this config file read this value from environment? Otherwise I actually need to care about source code, mounting and so on, which scares me little bit more than simply putting one more env variable inside my docker-compose.yml and rely on docker image entirely :sweat_smile:

nickygerritsen: Let me create a separate issue in `domjudge-packaging` for this, ok? I think that makes the most sense.

vicrac: I mean inside my docker-compose.yml in my own configuration :smile:  here I would only like something like (sorry, this is rather pseudocode, I don't know php):

`define('CREATE_WRITABLE_TEMP_DIR', getenv('CREATE_WRITABLE_TEMP_DIR') ? true : false);`

in `etc/judgehost-config.php`, would it make sense?

nickygerritsen: @eldering what do you think about this change? I think it's fine but you might disagree.

vicrac: Actually, do we need to have this in config at all? I realized with this change included we would read env only to put it back again :see_no_evil:

nickygerritsen: But not everyone can/will set up env variables, for example when not running inside docker but using a shell script or systemd, so it's better to have it.

eldering: > @eldering what do you think about this change? I think it's fine but you might disagree.

Sounds ok to me. I would prefix the environment variable with `DJ_` or `DOMJUDGE_` to prevent any possible name clash.

nickygerritsen: Sounds good. I can change it on wednesday but if @vicrac wants do to it before go ahead

vicrac: Is this ok now?

nickygerritsen: I’d say only the one in judgehost-config should read DOMJUDGE_, as the other one is in a controlled environment already

nickygerritsen:
I think there are two small things left:
* Update the SQL upgrade script with the new executable content
* Test that the environment variable works

I will do both tomorrow evening and I think we can merge afterwards.

nickygerritsen:
Just a small FYI: a meeting ran pretty long so I didn't have time yet for these two points. I'm not sure if I can do tomorrow, so it might be Friday.

nickygerritsen:
Did a test for the environment variable and it indeed works 🎉 .
I will wait for the Travis build to complete (which should be fine) and then merge.

nickygerritsen:
I also pushed a commit to the `domjudge-packaging` repo to add support for the environment variable to the judgehost docker container: https://github.com/DOMjudge/domjudge-packaging/commit/91a59b008672867566b75d27d8dbb0b660500da5

As stated there, of course this will only work starting with the next release of DOMjudge.

PR #526, opened 2019-03-17
Add "new rejudging" page.

meisterT:


webapp/src/DOMJudgeBundle/Controller/Jury/RejudgingController.php
```diff
@@ -419,6 +420,82 @@ public function finishAction(Request $request, RejudgingService $rejudgingServic
         }
     }
 
+    /**
+     * @Route("/rejudging/add", name="jury_rejudging_add")
```
nickygerritsen: Nitpick: the other endpoints are under `/rejudgings` (and we have that for other things likes teams and users as well), so I'd do that here too.

webapp/src/DOMJudgeBundle/Controller/Jury/RejudgingController.php
```diff
@@ -419,6 +420,82 @@ public function finishAction(Request $request, RejudgingService $rejudgingServic
...
+                $queryBuilder
+                    ->andWhere('j.result IN (:verdicts)')
+                    ->setParameter(':verdicts', $verdicts);
+            }
+
+            /** @var array[] $judgings */
+            $judgings = $queryBuilder
+                ->getQuery()
+                ->getResult(Query::HYDRATE_ARRAY);
+            dump($judgings);
```
nickygerritsen: Woops 😄

webapp/src/DOMJudgeBundle/Resources/views/jury/rejudging_form.html.twig
```diff
@@ -0,0 +1,23 @@
+{% extends "@DOMJudge/jury/base.html.twig" %}
+
```
nickygerritsen: If you add

```twig
{% import "@DOMJudge/jury/jury_macros.twig" as macros %}

{% block extrahead %}
    {{ parent() }}
    {{ macros.select2_extrahead() }}
{% endblock %}
```

you get the Select2-based multi-selects, which IMHO look nicer (and we also have in some other places like judgehost restrictions)

webapp/src/DOMJudgeBundle/Resources/views/jury/rejudging_form.html.twig
```diff
@@ -0,0 +1,23 @@
...
+{% block title %}Add rejudging - {{ parent() }}{% endblock %}
+
+{% block content %}
+
+    <h1>Add rejudging</h1>
+
+    <div class="row">
+        <div class="col-lg-4">
+            {{ form_start(form) }}
```
nickygerritsen: I think you output the whole form? If so, you can replace all this by
```twig
{{ form(form) }}
```

webapp/src/DOMJudgeBundle/Resources/views/jury/rejudging_form.html.twig
meisterT: beautiful, thanks!

webapp/src/DOMJudgeBundle/Resources/views/jury/rejudging_form.html.twig
meisterT: That's easy :-D

PR #527, opened 2019-03-17
Consider earlier pendings in first to solve

thijskh:
This is a new approach, because the previous PR had become quite stale given all the source changes in between, and I opted to solve it in yet another way now.

This moves all logic to determine what constitutes a first to solve to one place, calculation of the score row. When refreshing a row, and the judgement is correct, determine whether there are any earlier
submissions for this contest, problem, sort order that are either correct or still pending. If this is the case, this submission is not (yet) the first to solve.

Advantage is that this extra complexity and time is spent in updating the cache. Every _display_ of the scoreboard or team row is now even simpler than before so remains fast. Drawback is that a change of
sortorder information now also requires recalculating of the cache. It should likely be in the release notes that after upgrading you need to recalculate the scores.

### DIFF
--- a/sql/mysql_db_structure.sql
+++ b/sql/mysql_db_structure.sql
@@ -399,6 +399,7 @@ CREATE TABLE `scorecache` (
   `pending_public` int(4) unsigned NOT NULL DEFAULT '0' COMMENT 'Number of submissions pending judgement (public)',
   `solvetime_public` decimal(32,9) NOT NULL DEFAULT '0.000000000' COMMENT 'Seconds into contest when problem solved (public)',
   `is_correct_public` tinyint(1) unsigned NOT NULL DEFAULT '0' COMMENT 'Has there been a correct submission? (public)',
+  `is_first_to_solve` tinyint(1) unsigned NOT NULL DEFAULT '0' COMMENT 'Is this the first solution to this problem?',
   PRIMARY KEY (`cid`,`teamid`,`probid`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci COMMENT='Scoreboard cache';
 
--- a/webapp/src/DOMJudgeBundle/Entity/ScoreCache.php
+++ b/webapp/src/DOMJudgeBundle/Entity/ScoreCache.php
@@ -65,7 +65,12 @@ class ScoreCache
      */
     private $solvetime_public;
 
-
+    /**
+     * @var boolean
+     *
+     * @ORM\Column(type="boolean", name="is_first_to_solve", options={"comment"="Is this team the first to solve for this problem,sortorder?"}, nullable=false)
+     */
+    private $is_first_to_solve = false;
 
     /**
      * @ORM\Id
@@ -278,6 +283,30 @@ class ScoreCache
     public function getSolvetimePublic()
     {
         return $this->solvetime_public;
+    }
+
+    /**
+     * Set isFirstToSolve
+     *
+     * @param boolean $isFirstToSolve
+     *
+     * @return ScoreCache
+     */
+    public function setIsFirstToSolve(bool $isFirstToSolve)
+    {
+        $this->is_first_to_solve = $isFirstToSolve;
+
+        return $this;
+    }
+
+    /**
+     * Get isFirstToSolve
+     *
+     * @return boolean
+     */
+    public function getIsFirstToSolve() : bool
+    {
+        return $this->is_first_to_solve;
     }
 
     /**
--- a/webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
+++ b/webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
@@ -326,7 +326,8 @@ class ScoreboardService
         foreach ($submissions as $submission) {
             /** @var Judging|null $judging */
             $judging    = $submission->getJudgings()->first() ?: null;
-            $submitTime = $contest->getContestTime((float)$submission->getSubmittime());
+            $absSubmitTime = (float)$submission->getSubmittime();
+            $submitTime = $contest->getContestTime($absSubmitTime);
 
             // Check if this submission has a publicly visible judging result:
             if ($judging === null || ($verificationRequired && !$judging->getVerified()) || empty($judging->getResult())) {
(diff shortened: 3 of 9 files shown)


thijskh:
It does not consider the 'current' team in any case special when determining FTS for a scoreboard cell. So if you yourself have earlier pendings for this problem, you will also not yet be awarded FTS.

webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
```diff
@@ -357,6 +358,35 @@ public function calculateScoreRow(Contest $contest, Team $team, Problem $problem
...
+            // Find out how many valid submissions were submitted earlier
+            // that have a valid judging that is correct, or are awaiting judgement.
+            // Only if there are 0 found, we are definitely the first to solve this problem.
+            $solvedFirst = 0 == $this->entityManager->getConnection()->fetchColumn('
+                SELECT count(*) FROM submission s
+                    LEFT JOIN judging j USING (submitid)
+                    LEFT JOIN team t USING(teamid)
+                    LEFT JOIN team_category tc USING (categoryid)
+                WHERE s.valid = 1 AND
+                    ((j.valid = 1 AND ( j.rejudgingid IS NULL AND (j.result IS NULL OR j.result = :correctResult))) OR
```
meisterT: Wow, this is quite complex. I wonder if it makes sense to split this up even further and add comments for:
- in progress judging
- correct result
- queued judging
or similar.

webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
```diff
@@ -357,6 +358,35 @@ public function calculateScoreRow(Contest $contest, Team $team, Problem $problem
...
+                      s.judgehost IS NULL) AND
+                    s.cid = :cid AND s.probid = :probid AND
+                    tc.sortorder = :teamSortOrder AND
+                    round(s.submittime,4) < :submitTime',
```
meisterT: why not use the submitid instead?

nickygerritsen:
> Drawback is that a change of sortorder information now also requires recalculating of the cache.

We already display a message about this when editing a team category, see https://github.com/DOMjudge/domjudge/blob/master/webapp/src/DOMJudgeBundle/Resources/views/jury/team_categories.html.twig#L15

(we might want to change that `edited` logic into a flash message btw, I will do that at some point)

webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
nickygerritsen: When we import submissions we don’t have any guarantees that submission ID’s and times have any correlation, right? Or am I forgetting something?

thijskh: I reasoned based on the semantics. But indeed if we can keep the guarantee that submitid is ordered it would make it simpler.

webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
thijskh: Yes, there’s quite some things to account for. Will see if I can make it more transparent.

thijskh:
Which suggests that sortorder was already relevant for the cache. But was it really?

nickygerritsen:
I don’t think it really was relevantly indeed.

eldering:
The previous PR is #509 right? So we can probably close that one too?

sql/mysql_db_structure.sql
```diff
@@ -399,6 +399,7 @@ CREATE TABLE `scorecache` (
   `pending_public` int(4) unsigned NOT NULL DEFAULT '0' COMMENT 'Number of submissions pending judgement (public)',
   `solvetime_public` decimal(32,9) NOT NULL DEFAULT '0.000000000' COMMENT 'Seconds into contest when problem solved (public)',
   `is_correct_public` tinyint(1) unsigned NOT NULL DEFAULT '0' COMMENT 'Has there been a correct submission? (public)',
+  `is_solved_first` tinyint(1) unsigned NOT NULL DEFAULT '0' COMMENT 'Is this the first solution to this problem?',
```
eldering: There should be a `restricted` and `public` version of this, since a first solve may happen in the freeze.

webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
eldering: I haven't read the code in full, but is this still keeping track of the first pending or correct submission? Then we'd not need this complex query, but just check whether the currrent submission is the first pending/correct one, and if so, it's the FTS, right?

sql/mysql_db_structure.sql
thijskh: Why would you need restricted and public versions? If the fts happens after freeze, is_correct_public will be 0 so there's no problem?

webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
thijskh: We don't keep any other administration of whether this is the first one for any problem. Because the scorecache is only updated one cell (problem,team) at a time, any solution would always need to do some kind of query on all other team's submissions for this problem.

Indeed the previous pr #509 kept some track of it, namely it changed the definition of the 'solved'-time cell, then stored this information in the cache, and then on scoreboard display check these values to see if it's lowest. This approach does not rely on any other part doing further calculations.

nickygerritsen:
> (we might want to change that `edited` logic into a flash message btw, I will do that at some point)

Done on master

webapp/src/DOMJudgeBundle/Entity/ScoreCache.php
```diff
@@ -65,7 +65,12 @@ class ScoreCache
      */
     private $solvetime_public;
 
-
+    /**
+     * @var boolean
+     *
+     * @ORM\Column(type="boolean", name="is_first_to_solve", options={"comment"="Is this team the first to solve for this problem,sortorder?"}, nullable=false)
+     */
+    private $is_first_to_solve = false;
```
meisterT: why do we have to set this to `false` explicitly?

sql/mysql_db_structure.sql
eldering: Ah, ok. Makes sense, so the use of `is_solved_first` is dependent on `is_correct_*`.

webapp/src/DOMJudgeBundle/Entity/ScoreCache.php
thijskh: Seems a reasonable default; something is only first when it has been explicitly been confirmed to be that.

PR #536, opened 2019-04-05
Implement IP authentication autologin

tuupke:
In the old implementation, when IP authentication was enabled, you still needed to click a button to actually log in. This commit makes a configuration option available `ip_autologin` (default off) which immediately logs in the user.

Furthermore, after login; you can now visit `/login` to log in with a different user.

### DIFF
--- a/sql/mysql_db_defaultdata.sql
+++ b/sql/mysql_db_defaultdata.sql
@@ -58,7 +58,8 @@ INSERT INTO `configuration` (`name`, `value`, `type`, `public`, `category`, `des
 ('openid_provider', '"https://accounts.google.com"', 'string', '1', 'Misc', 'OpenID Provider URL'),
 ('openid_clientid', '""', 'string', '0','Misc', 'OpenID Connect client id'),
 ('openid_clientsecret', '""', 'string', '0', 'Misc', 'OpenID Connect client secret'),
-('data_source', '0', 'int', '0', 'Misc', 'Source of data. Choices: 0 = all local, 1 = configuration data external, 2 = configuration and live data external');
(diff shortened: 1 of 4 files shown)


webapp/src/DOMJudgeBundle/Security/DOMJudgeIPAuthenticator.php
```diff
@@ -23,13 +23,16 @@ class DOMJudgeIPAuthenticator extends AbstractGuardAuthenticator
     private $security;
     private $container;
     private $em;
+    private $dj;
```
meisterT: nit: we typically name this variable $DOMJudgeService, we should do so here too for consistency. We could in a later step rename all of this to $dj, which is much shorter and nicer.

sql/mysql_db_defaultdata.sql
```diff
@@ -58,7 +58,8 @@ INSERT INTO `configuration` (`name`, `value`, `type`, `public`, `category`, `des
 ('openid_provider', '"https://accounts.google.com"', 'string', '1', 'Misc', 'OpenID Provider URL'),
 ('openid_clientid', '""', 'string', '0','Misc', 'OpenID Connect client id'),
 ('openid_clientsecret', '""', 'string', '0', 'Misc', 'OpenID Connect client secret'),
-('data_source', '0', 'int', '0', 'Misc', 'Source of data. Choices: 0 = all local, 1 = configuration data external, 2 = configuration and live data external');
+('data_source', '0', 'int', '0', 'Misc', 'Source of data. Choices: 0 = all local, 1 = configuration data external, 2 = configuration and live data external'),
+('ip_autologin', '0', 'bool', '0', 'Misc', 'IP authentication still requires a button to be pressed, enable this option to skip this step');
```
meisterT: what about "Enable to skip the login page when using IP authentication."?

PR #542, opened 2019-04-08
Implemented max-width for team-, affiliate-names

### LINKED ISSUE #431: Truncate team name on scoreboard
Currently a long team name get displayed in full length on the scoreboard. This creates a suboptimal layout wasting a lot of space on the team column and potentially pushing the problem columns out of the window. We should probably restrict the team column to a maximum width, and truncate the team name (don't wrap to a new line, since it would create rows with uneven height).
Attached a screenshot from UKIEP. Note also that the category leader button adds to the problem.
screen

tuupke:
Added a configuration option to set the max-width for the affiliation and team names. When set to 0, it does not apply a maximum width. Some JS adds a class when the contents are too wide, which adds a slot that houses the gradient, this gradient overlaps with the rightmost 30% of the field. It has been tested in Webkit browsers (opera, safari) and Firefox

Old:
old

New:
new

Fixes #431

### DIFF
--- a/sql/mysql_db_defaultdata.sql
+++ b/sql/mysql_db_defaultdata.sql
@@ -59,7 +59,8 @@ INSERT INTO `configuration` (`name`, `value`, `type`, `public`, `category`, `des
 ('openid_clientid', '""', 'string', '0','Misc', 'OpenID Connect client id'),
 ('openid_clientsecret', '""', 'string', '0', 'Misc', 'OpenID Connect client secret'),
 ('data_source', '0', 'int', '0', 'Misc', 'Source of data. Choices: 0 = all local, 1 = configuration data external, 2 = configuration and live data external'),
-('ip_autologin', '0', 'bool', '0', 'Misc', 'Enable to skip the login page when using IP authentication.');
+('ip_autologin', '0', 'bool', '0', 'Misc', 'Enable to skip the login page when using IP authentication.'),
+('team_column_width', '0', 'int', '0', 'Misc', 'Maximum width of team column on scoreboard. Leave 0 for no maximum.');
 
 
 --
--- a/sql/mysql_db_defaultdata.sql
+++ b/sql/mysql_db_defaultdata.sql
@@ -51,6 +51,7 @@ INSERT INTO `configuration` (`name`, `value`, `type`, `public`, `category`, `des
 ('time_format', '"%H:%M"', 'string', '0', 'Display', 'The format used to print times. For formatting options see the PHP \'strftime\' function.'),
 ('thumbnail_size', '128', 'int', '0', 'Display', 'Maximum width/height of a thumbnail for uploaded testcase images.'),
 ('show_limits_on_team_page', '0', 'bool', '1', 'Display', 'Show time and memory limit on the team problems page'),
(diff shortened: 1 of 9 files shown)


sql/mysql_db_defaultdata.sql
```diff
@@ -59,7 +59,8 @@ INSERT INTO `configuration` (`name`, `value`, `type`, `public`, `category`, `des
 ('openid_clientid', '""', 'string', '0','Misc', 'OpenID Connect client id'),
 ('openid_clientsecret', '""', 'string', '0', 'Misc', 'OpenID Connect client secret'),
 ('data_source', '0', 'int', '0', 'Misc', 'Source of data. Choices: 0 = all local, 1 = configuration data external, 2 = configuration and live data external'),
-('ip_autologin', '0', 'bool', '0', 'Misc', 'Enable to skip the login page when using IP authentication.');
+('ip_autologin', '0', 'bool', '0', 'Misc', 'Enable to skip the login page when using IP authentication.'),
+('team_column_width', '0', 'int', '0', 'Misc', 'Maximum width of team column on scoreboard. Leave 0 for no maximum.');
```
nickygerritsen: How about putting it in display options?

webapp/src/DOMJudgeBundle/Twig/TwigExtension.php
```diff
@@ -893,4 +894,32 @@ public function wrapUnquoted(string $text, int $width = 75, string $quote = '>')
...
+    public function hexColorToRGBA(string $text, float $opacity = 1): string
+    {
+        $col = Utils::convertToHex($text);
+        preg_match_all("/[0-9A-Fa-f]{2}/", $col,$m);
+        if (!count($m)) {
+            return $text;
+        }
+
+        $m = current($m);
+        switch (count($m)){
```
nickygerritsen: Some formatting here :)

webapp/src/DOMJudgeBundle/Resources/views/partials/scoreboard_table.html.twig
```diff
@@ -323,3 +337,49 @@
         </table>
     {% endif %}
 {% endif %}
+
+<style>
+    {% for color,i in backgroundColors %}
+    {% set cssColor = "#"~color %}
```
nickygerritsen: Should we indent this one more time?

nickygerritsen:
Looks good to me in general, some small comments

sql/mysql_db_defaultdata.sql
tuupke: Changed in 2a5408eb58df7c7126c411d447e4a3578d0cc0d8

webapp/src/DOMJudgeBundle/Twig/TwigExtension.php
tuupke: Resolved in f7ccd52cd87e43d97000bd0aca80af5607a28160

webapp/src/DOMJudgeBundle/Resources/views/partials/scoreboard_table.html.twig
tuupke: Resolved in f7ccd52cd87e43d97000bd0aca80af5607a28160

meisterT:
Does it show the original name if you hover over it?

PR #545, opened 2019-04-09
Rename and move teampage_first_visited.

nickygerritsen:
I did this as a pull request so we can still discuss the naming and related.
I also dropped `hostname` from team as we already have `last_ip_address` on user. Do we actually want the *hostname*? If so, should we rename `last_ip_address` to `last_hostname` on user?


meisterT:
LGTM

PR #559, opened 2019-04-17
Do not use setfacl anymore to change webapp/var permissions.

### LINKED ISSUE #558: Problem with setfacl
### Description of the problem / feature request
In latest version, the install script calls [setfacl](https://github.com/DOMjudge/domjudge/blob/master/webapp/Makefile#L44) to set permission. Which is making packaging DOMjudge difficult in Debian/Ubuntu/Arch linux in my knowledge. Because while packaging it will call `fakeroot` to run install commands while pretending as root, but `fakeroot` does not support `setfacl`.
The permission in 6.0.3 is set by owners and groups, which seems working great for me. Can we change it back into the old way?

### Your environment
commit: e5e8f3c0a2c2a7e5935f9033a8059f1d5867f6ba
Distro: Arch linux

### Steps to reproduce
`./configure --with-webserver-group=http`
`make install-domserver` while packaging

### Expected behaviour
No errors.

### Actual behaviour

`setfacl:/home/username/domjudge/pkg/domjudge-domserver//opt/domjudge/domserver/webapp/var: Operation not supported`

nickygerritsen:
This was needed for the eventdaemon to work, but we dont have it anymore. Fixes #558.


lib/Makefile
```diff
@@ -12,12 +12,18 @@ $(OBJECTS): %$(OBJEXT): %.c %.h
 clean-l:
 	rm -f $(OBJECTS)
 
+ifeq "$(FHS_ENABLED)" "yes"
```
eldering: `DESTDIR` should never be prepended here, since WEBAPP_DIR is later used only as contents inside files, not for actual paths that things are installed to. See for an explanation https://www.gnu.org/prep/standards/html_node/DESTDIR.html

PR #544, opened 2019-04-08
Add country validation

tuupke:
The old version of the `TeamAffiliation` did not do any validation on the `$country` property, this resulted in a database error when random text was entered, to prevent this, I implemented a validator which checks whether the entered value is either empty or a valid ISO3166-1 alpha-3 value. A phpleague library provides the list of the countrycodes; this might be overkill. Another option is to simply add them (to Utils perhaps) as a const.

PR #563, opened 2019-04-21
Allow to specify team name and affiliation details during sign up.

### LINKED ISSUE #561: Add option to specify team and association during self-registration
When self-registration is enabled, it would be nice if users could specify the name of the team they want and the affiliation they want to belong to.

For the team name we can just add an input field, which checks if the team doesn't exist yet. If it does, give an error.

For affiliation, we can show a toggle to either select an existing one or create a new one. In the latter case users can provide a name and optional country (if enabled). Again, when creating a new one we need to check if it already exists.

nickygerritsen:
This will fix #561.

Note that we want to merge this after 7.0 is released, but I've opened the PR already 🙂  .

Also note that when #544 is merged, I will first improve the country selection.


nickygerritsen:
Should be done now 😄 . This is how it looks:
image

webapp/src/DOMJudgeBundle/Resources/views/jury/partials/country_select.html.twig
```diff
@@ -0,0 +1,34 @@
+<style>
+    img.countryflag {
+        margin-top: -3px;
+    }
+</style>
```
thijskh: While we're at it, any reason we cannot move this to our global css?

nickygerritsen: @tuupke I think we can?

tuupke: Go for it!

sql/mysql_db_defaultdata.sql
```diff
@@ -54,6 +54,7 @@ INSERT INTO `configuration` (`name`, `value`, `type`, `public`, `category`, `des
 ('team_column_width', '0', 'int', '0', 'Display', 'Maximum width of team column on scoreboard. Leave 0 for no maximum.'),
 ('enable_printing', '0', 'bool', '1', 'Misc', 'Enable teams and jury to send source code to a printer via the DOMjudge web interface.'),
 ('registration_category_name', '""', 'string', '1', 'Misc', 'Team category for users that register themselves with the system. Self-registration is disabled if this field is left empty.'),
+('registration_allow_affiliation', '0', 'bool', '1', 'Misc', 'Whether to allow to select / add an affiliation during registration.'),
```
meisterT: Do we want to have a separate config flag for this or rather use the existing show_affiliations? I am a bit torn, but we have a recent increase in flags...

webapp/src/DOMJudgeBundle/Form/Type/UserRegistrationType.php
```diff
@@ -41,20 +148,49 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+                /** @var Form $form */
+                $form = $context->getRoot();
+                switch ($form->get('affiliation')->getData()) {
+                    case 'new':
+                        if (empty($form->get('affiliationName')->getData())) {
+                            $context->buildViolation('This value should not be blank.')
+                                ->atPath('affiliationName')
+                                ->addViolation();
+                        }
+                        // TODO: country validation after PR #544 is merged
```
meisterT: this is merged now

sql/mysql_db_defaultdata.sql
nickygerritsen: Hmmmm I’m also torn. I can’t think of a reason where you would show affiliations but not ask them during registration so I am leaning towards using the existing flag.

webapp/src/DOMJudgeBundle/Form/Type/UserRegistrationType.php
nickygerritsen: Whoops forgot this part

nickygerritsen:
Should all be fixed now 😄

PR #565, opened 2019-04-22
Fix syntax of linear-gradient() and clean up all others.

thijskh:
We really need not support century old browsers. Just use the W3C
standard. Also so we know whether we use it correctly.


thijskh:
Mart please review

webapp/src/DOMJudgeBundle/Resources/views/partials/scoreboard_table.html.twig
```diff
@@ -345,35 +345,16 @@
         {% set hexColor = color | colorToHex | trim('#') %}
```
nickygerritsen: This variable is not used anymore now, I think?

thijskh: Indeed, good catch. Can clean up even more.

PR #567, opened 2019-04-25
Feature/make affiliate countries fancier

tuupke:
This implements the dropdown for the selection of the country for an affiliation.
So we do not need to look up the country codes anymore. Most if not all occurrences of the old ISO3166-A3 codes should be replaced with the full version.

### DIFF
--- a/webapp/src/DOMJudgeBundle/Resources/views/jury/partials/team_affiliation_form.html.twig
+++ b/webapp/src/DOMJudgeBundle/Resources/views/jury/partials/team_affiliation_form.html.twig
@@ -1,19 +1,38 @@
-<div class="row">
-    <div class="col-lg-4">
-        {{ form_start(form) }}
-        {% if form.offsetExists('externalid') %}
-            {{ form_row(form.externalid) }}
-        {% endif %}
-        {{ form_row(form.shortname) }}
-        {{ form_row(form.name) }}
-        <div class="form-group">
-            {{ form_label(form.country) }}
-            {{ form_widget(form.country) }}
-            <small class="form-text text-muted">
-                <a target="_blank" href="http://en.wikipedia.org/wiki/ISO_3166-1_alpha-3#Current_codes"><i class="fas fa-question-circle"></i></a>
-            </small>
-        </div>
-        {{ form_row(form.comments) }}
-        {{ form_end(form) }}
-    </div>
+<div class="row col-lg-4">
+    {{ form(form) }}
 </div>
+
+<style>
+    img.countryflag {
+        margin-top: -3px;
+    }
+</style>
+
+<script>
+    var template = function(element) {
+        var imgBase = '';
+        {% set countryFlag = 'images/countries/NLD.png' %}
+        {% if countryFlag | assetExists %}
+            // Use a country flag we know exists to automatically determine base path and asset version
+            imgBase = "{{ asset(countryFlag) }}";
+        {% endif %}
+
+        var img = element.id == undefined || element.id == "" || imgBase == ""
+            ? ''
+            : '<img src="' + imgBase.replace("NLD.png", element.id + ".png", 1) + '" class="countryflag">&nbsp;';
+
+        return img + element.text;
+    };
+
+    $('#team_affiliation_country').select2({
+        templateResult: template,
+        templateSelection: template,
+
+        // Prevent escaping
+        escapeMarkup: function(t) {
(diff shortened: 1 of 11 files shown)


nickygerritsen:
Now I'm lost 😛 Whats the difference with #544 ? Also I think we should still split the formatting change of Util into a separate commit or is that not what @meisterT meant?

webapp/src/DOMJudgeBundle/Utils/Utils.php
```diff
@@ -1087,10 +1120,17 @@ public static function generatePassword()
...
+            case 'm':
+                return (int)$size_str * 1048576;
+            case 'K':
+            case 'k':
+                return (int)$size_str * 1024;
+            case 'G':
+            case 'g':
+                return (int)$size_str * 1073741824;
+            default:
+                return $size_str;
```
thijskh: Not cleaner to just strtolower?

webapp/src/DOMJudgeBundle/Utils/Utils.php
```diff
@@ -1068,8 +1069,40 @@ public static function wrap_unquoted(string $text, int $width = 75, string $quot
...
+            'z',
+            '2',
+            '3',
+            '4',
+            '5',
+            '6',
+            '7',
+            '8',
+            '9'
+        ];
```
thijskh: Not sure this formatting change is an improvement.

thijskh:
I understand it.

I do think indeed that the numerous whitespace changes in Utils, although generally a good idea, should be a separate commit. We can accept a change here or there but an entire reformat of a file should be separate in the history.

nickygerritsen:
OK I'm not lost anymore, I just thought the other PR already contained this stuff :)

webapp/src/DOMJudgeBundle/Resources/views/jury/partials/team_affiliation_form.html.twig
```diff
@@ -7,13 +7,44 @@
         {{ form_row(form.shortname) }}
         {{ form_row(form.name) }}
         <div class="form-group">
-            {{ form_label(form.country) }}
```
thijskh: Is it intentional that this label is dropped? Form looks a bit stranger now.

nickygerritsen: Also I think we can now replace the whole `<div>` with just a `{{ form_row(form.country) }}` or even replace the whole form stuff with `{{ form(form) }}` since we do no custom stuff anymore

webapp/src/DOMJudgeBundle/Utils/Utils.php
eldering: If removing the nested case on a single line, then `strtolower` is cleaner indeed. Also, if changing this, then I'd prefer to sort by modifier size, so k,m,g (or reverse).

tuupke:
With regards to the changes in Util, these were unintended; will revert them in a bit

webapp/src/DOMJudgeBundle/Resources/views/jury/partials/team_affiliation_form.html.twig
tuupke: Done!

tuupke:
Util changes should be reverted now

webapp/src/DOMJudgeBundle/Utils/Utils.php
tuupke: Utils changes have been removed.

webapp/src/DOMJudgeBundle/Resources/views/jury/partials/team_affiliation_form.html.twig
```diff
@@ -1,19 +1,38 @@
...
+            : '<img src="' + imgBase.replace("NLD.png", element.id + ".png", 1) + '" class="countryflag">&nbsp;';
+
+        return img + element.text;
+    };
+
+    $('#team_affiliation_country').select2({
+        templateResult: template,
+        templateSelection: template,
+
+        // Prevent escaping
```
meisterT: why is this necessary?

nickygerritsen: Otherwise you see `<img...` instead of an image

PR #575, opened 2019-04-28
For 'make dist' run composer in prod mode.

thijskh:
This will save us from shipping stuff in our tarball that's
only "require-dev". Also allows to optimize the autoloader for
performance.

The maintainer-conf target will then install the dev dependencies.

May not be the best solution Makefile-wise, maybe a Makefile guru can comment on thjat.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,10 @@ endif
 # because at this point the autoload.php file is not generated yet, which
 # is needed to run the post-install scripts.
 	$(MAKE) -C webapp/app params-from-stub static-from-stub
-	composer $(subst 1,-q,$(QUIET)) install --no-scripts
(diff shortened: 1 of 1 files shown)


nickygerritsen:
This does mean we can't run with `app_dev.php` in non-maintainer mode. If that is something we are OK with, then this PR looks good.

meisterT:
> This does mean we can't run with `app_dev.php` in non-maintainer mode. If that is something we are OK with, then this PR looks good.

hmmm, I am not sure I like that. The toolbar is just too useful to give it away for a smaller tarball.

thijskh:
A subset of this PR would be to set "-o" for "dist" and leave it out for maintainer-conf.

thijskh:
Reduced the change to only the autoloader optimization.

PR #576, opened 2019-04-29
Documentation fix

mvr320:


nickygerritsen:
If I'm correct you are saying that RedHat nowadays ships with Java 1.8, right? If that is the case this looks good.

However there is a strange merge commit in the PR, could you try to remove it? We can also squash the commits into one if that is easier.

PR #582, opened 2019-05-03
Take into account pending submissions in the freeze.

nickygerritsen:
After a discussion with @deboer-tim we came to the conclusion that we leaked some info to the public during the freeze.

Please please double-triple check my code.

To explain what this tries to fix, here are two scenario's:

# Scenario 1

Imagine the following: we have a team, say team `X`, and a problem, say `A`. Now imagine the team does two submissions, both after the freeze. The first one is a `compiler-error` and the second one is `correct`. `compile_penalty` is set to `false`.
Our previous code would show `1 try` for this team to the public, while the team actually did try two times. Showing `1 try` gives something away, namely that one of the submissions was not 'counted'. It doesn't necessarily mean it was a compiler-error, because of the second scenario below.
Moreover, until the first submission has actually been judged, we would show `1 try`. Then after the judging is done, we show no tries. This *clearly* gives away the fact that this submission was a `compiler-error` (or it was ignored by the jury).
The new code will show `2 tries` until we unfreeze, then we show `1 try` with the score of the last try.

# Scenario 2

Imagine the following: we have a team, say team `X`, and a problem, say `A`. Now imagine the team does two submissions, both after the freeze. The first one is `correct` and the second one is `wrong-answer`.
Our previous code would show `1 try` for this team to the public, because we stopped checking submissions after the first correct one. Again, this gives away info, because we again ignore one of the submissions.
The same holds for the state *during* judging. We would show `2 tries` until the second submission is done and then it would jump back to `1 try`.
The new code will also show `2 tries` until we unfreeze, then we show `1 try` with the score of the first try, ignoring the second one.

# Rationale

The rationale to change this is that the [Contest API spec](https://clics.ecs.baylor.edu/index.php?title=Contest_API_2019#Access_restrictions_at_WF_11) regulates that judgements *during* the freeze will not be shown to the public. Hence we should not leak information from these judgements during the freeze to the public.

# Testing

I tested this by using a copy of the DOMjudge online contest, where team DerBaer had two during-freeze-submissions on problem G, the first of which was correct and the second was compiler-error. I played a bit with these submissions to get different scenario's. Let's call these submissions `S1` and `S2`.

The results after this change are:
* `S1 = correct, S2 = compiler-error, S1.time < S2.time`: frozen scoreboard shows `2 tries`, jury shows `1 try` and correct with time of S1.
* `S1 = correct, S2 = compiler-error, S1.time > S2.time` (i.e. swapping them) with `compile_penalty` set to `false`: frozen scoreboard shows `2 tries`, jury shows `1 try` and correct with time of S2.
* `S1 = correct, S2 = compiler-error, S1.time > S2.time` (i.e. swapping them) with `compile_penalty` set to `true`: frozen scoreboard shows `2 tries`, jury shows `2 tries` and correct with time of S2 + 20.
* `S1 = correct, S2 = wrong-answer, S1.time > S2.time` (i.e. swapping them and changing the outcome of S2): frozen scoreboard shows `2 tries`, jury shows `2 tries` and correct with time of S2 + 20.
* `S1 = correct, S2 = correct, S1.time < S2.time` )i.e. changing the outcome of S2 to correct0: frozen scoreboard shows `2 tries`, jury shows `1 try` and correct with time of S1.

### DIFF
--- a/webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
+++ b/webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
@@ -306,18 +306,15 @@ class ScoreboardService
             ->setParameter(':cid', $contest->getCid())
             ->orderBy('s.submittime');
 
-        if (!$this->dj->dbconfig_get('compile_penalty', true)) {
-            $queryBuilder
-                ->andWhere('j.result IS NULL or j.result != :compileError')
-                ->setParameter(':compileError', Judging::RESULT_COMPILER_ERROR);
-        }
+        // Check if we need to count compile error as a penalty.
+        $compilePenalty = $this->dj->dbconfig_get('compile_penalty', true);
 
         /** @var Submission[] $submissions */
         $submissions = $queryBuilder->getQuery()->getResult();
 
         $verificationRequired = $this->dj->dbconfig_get('verification_required', false);
 
-        // Initialize variables
+        // Initialize variables.
         $submissionsJury = $pendingJury = $timeJury = 0;
         $submissionsPubl = $pendingPubl = $timePubl = 0;
         $correctJury     = false;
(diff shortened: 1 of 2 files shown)


webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
```diff
@@ -325,36 +322,61 @@ public function calculateScoreRow(Contest $contest, Team $team, Problem $problem
...
             }
 
+            // If we encountered a correct submission during the whole contest, do not consider
+            // the submissions after that one for correctness
+            if ($correctJury) {
+                continue;
+            }
+
+            $absSubmitTime = (float)$submission->getSubmittime();
+            $submitTime    = $contest->getContestTime($absSubmitTime);
```
meisterT: $submission->getRelativeSubmittime?

webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
```diff
@@ -325,36 +322,61 @@ public function calculateScoreRow(Contest $contest, Team $team, Problem $problem
...
             /** @var Judging|null $judging */
-            $judging    = $submission->getJudgings()->first() ?: null;
-            $absSubmitTime = (float)$submission->getSubmittime();
-            $submitTime = $contest->getContestTime($absSubmitTime);
+            $judging = $submission->getJudgings()->first() ?: null;
 
             // Check if this submission has a publicly visible judging result:
             if ($judging === null || ($verificationRequired && !$judging->getVerified()) || empty($judging->getResult())) {
-                $pendingJury++;
+                // For the jury: only consider it pending if we don't have a correct one yet
```
meisterT: This is pedantic, but can you please finish these sentences with a . ? Also for the other comments.

webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
nickygerritsen: Do you think that is better than this? Imho doing it on the contest feels more logical, as it is relative to the contest start?

meisterT: A submission always belongs to exactly one contest and it removes one line from the code here, so I would be in favor of using the convenience function (or delete if we decide not to use it in general).

nickygerritsen: Ah yeah didn't know the function was there already. I will use it.

nickygerritsen: As discussed on Slack: we need both the relative time (to store in the scorecache) and the absolute time (to calculate first-to-solve) so we kept it this way.

We also still use the function in Submission, as that is used in the API. However, it was wrong (when the contest was paused or we had removed intervals), so I have fixed it (by making use of the function on Contest which already did the correct thing).

PR #584, opened 2019-05-03
Use environment variable to pass allowed language extensions to compile scripts. Fixes #566.

### LINKED ISSUE #566: Generic way to limit compile scripts to specific extensions
Introduce a generic way to bind compile scripts to specific file extensions.

This was done ad hoc in 1fb7566 and 6550377.

nickygerritsen:
I have modified the compile scripts of the languages used at the world finals to have this check, but in the future we can/might change more.

I tried adding the `LANGUAGE_EXTENSIONS` to the `ENVIRONMENT_VARS` variable a few lines higher in `compile.sh`, but because it contains spaces and my bash/sh magic is very limited, I couldn't get it to work using that.


nickygerritsen:
@eldering didn’t want that, see #566 for his reasoning, which I agree with.

The integration is broken now because some submissions are now not compiler-error anymore

meisterT:
Ah, I haven't read that, thanks for pointing me to it.

I still disagree, though :-)
Doing it in each of the compile scripts is error prone and repeated work. I would much rather optimize for ease of use than for a hypothetical use case.

However, if you and @eldering are in agreement, feel free to ignore my concerns.

nickygerritsen:
I do like the fact that we are not copy pasting the same thing every time if we move it... can we somehow find a nice middle ground?

nickygerritsen:
Also currently I can’t think of a compiler script that would need other files to be passed on the cli, so although it is nice to have this feature I wonder whether we should make it complicated / cumbersome to write this in every script...

Issue #241, opened 2016-11-02
Allow to Set Team Group on Contest


jGleitz:
To manage multiple, parallel running contests, it would be really nice to be able to assign some sort of groups (e.g. Team Category) of users to a contest. Right now, each team allowed to see a non-public contest needs to be specified on its own, which is a tedious (and error prone) process.

meisterT:
Good idea! A patch to implement this feature would be very welcome.

jGleitz:
@meisterT I already had a look at the code and tried to figure out what needs to be done for this.

Do I see it correctly that there are no data classes in DOMJudge but every code accesses the database directly? If yes, could you give a estimation which parts would be affected by this change?

ubergeek42:
There's no ORM so to speak, but there are some helper functions that try to centralize some bits. But for the most part things get touched all over the place.

You'd probably want to start somewhere around here, as this determines what contest a team has access to: https://github.com/DOMjudge/domjudge/blob/master/lib/lib.misc.php#L35

Right now access is done by the contestteam table which is a mapping of contestid to teamid. Searching the codebase for references to that table will show the places you'd need to modify.

I'm not sure what the right way to implement it would be, maybe a contestcategoryteam table or something like that.

eldering:
That is right, we do not (yet) an ORM. There are long-term plans to migrate to Symfony and Doctrine for this, see #198. All database access is done through the home-grown `lib/lib.database.php` library with the `$DB->q()` function.
As ubergeek42 said, I think a good option would be to copy the `contestteam` table and its use into a new `contestteamcategory` table.

jGleitz:
@ubergeek42, @eldering Thanks for the guidance! I will look into this during the next days.

PR #586, opened 2019-05-04
Split up contest.public in contest.public and contest.open_to_all_teams.

nickygerritsen:
This should do the trick of splitting up public visibility and whether all teams are part of a contest.

I tested a little around with public-not-all-teams, private-all-teams and the two modes we already supported and it seems to work, but of course I might have missed some edge case.

### DIFF
--- a/lib/lib.misc.php
+++ b/lib/lib.misc.php
@@ -32,7 +32,7 @@ function getCurContests(
     if ($onlyofteam !== null && $onlyofteam > 0) {
         $contests = $DB->q("SELECT * FROM contest
                             LEFT JOIN contestteam USING (cid)
-                            WHERE (contestteam.teamid = %i OR contest.public = 1)
+                            WHERE (contestteam.teamid = %i OR contest.open_to_all_teams = 1)
                             AND enabled = 1 ${extra}
                             AND ( deactivatetime IS NULL OR
                                   deactivatetime > UNIX_TIMESTAMP() )
--- a/webapp/src/Form/Type/RejudgingType.php
+++ b/webapp/src/Form/Type/RejudgingType.php
@@ -151,15 +151,15 @@ class RejudgingType extends AbstractType
                 ->andWhere('t.enabled = 1')
                 ->addOrderBy('t.name');
 
-            $anyPublic = false;
+            $selectAllTeams = false;
             foreach ($contests as $contest) {
-                if ($contest->getPublic()) {
-                    $anyPublic = true;
+                if ($contest->isOpenToAllTeams()) {
+                    $selectAllTeams = true;
                     break;
                 }
             }
 
-            if (!$anyPublic) {
+            if (!$selectAllTeams) {
                 $teamsQueryBuilder
                     ->join('t.contests', 'c')
                     ->andWhere('c IN (:contests)')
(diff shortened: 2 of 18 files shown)


nickygerritsen:
That is #241 . Do you have a better name for it? In the end it still adds teams, but then via categories.

meisterT:
perhaps @ubergeek42 can have a look - I don't know if he started on #241 already

lib/lib.misc.php
```diff
@@ -32,7 +32,7 @@ function getCurContests(
     if ($onlyofteam !== null && $onlyofteam > 0) {
         $contests = $DB->q("SELECT * FROM contest
                             LEFT JOIN contestteam USING (cid)
-                            WHERE (contestteam.teamid = %i OR contest.public = 1)
+                            WHERE (contestteam.teamid = %i OR contest.allTeams = 1)
```
eldering: I think this is a typo and should be `all_teams`.

webapp/src/DOMJudgeBundle/Command/ImportEventFeedCommand.php
```diff
@@ -931,7 +931,7 @@ protected function importTeam(array $event)
         // Also check if this is a private contest. If so, we need to add the team to the contest
         /** @var Contest $contest */
         $contest = $this->em->getRepository(Contest::class)->find($this->contestId);
-        if (!$contest->getPublic()) {
+        if (!$contest->getAllTeams()) {
```
eldering: I think this is confusing: the name `getAllTeams` suggests that the method returns all teams in the contest, not whether all teams are part of the contest.

webapp/src/DOMJudgeBundle/Controller/API/AbstractRestController.php
```diff
@@ -165,7 +165,7 @@ protected function getContestQueryBuilder(): QueryBuilder
         if (!$this->dj->checkrole('api_reader') && !$this->dj->checkrole('judgehost')) {
             if ($this->dj->checkrole('team') && $this->dj->getUser()->getTeamid()) {
                 $qb->leftJoin('c.teams', 'ct')
-                    ->andWhere('ct.teamid = :teamid OR c.public = 1')
+                    ->andWhere('ct.teamid = :teamid OR c.allTeams = 1')
```
eldering: As above: `all_teams`?

nickygerritsen: This one uses the doctrine entity so I think this one is correct. The above one should be fixed yes

webapp/src/DOMJudgeBundle/Command/ImportEventFeedCommand.php
nickygerritsen: Way better

webapp/src/DOMJudgeBundle/Controller/API/AbstractRestController.php
eldering: Hmmm, ok. I would still argue to write it using the SQL column name for consistency and (from my point of view) less confusion.

nickygerritsen: Well my IDE autocompletes to this way of writing it currently. I think it even breaks if I use the SQL column name, because that field doesn't exist on the entity.

We could rename the property in the entity from `openToAllTeams` to `open_to_all_teams`, then it does what you want. Shall I do that?

webapp/src/Form/Type/RejudgingType.php
```diff
@@ -151,15 +151,15 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 ->andWhere('t.enabled = 1')
                 ->addOrderBy('t.name');
 
-            $anyPublic = false;
+            $anyAllTeams = false;
```
eldering: I think this variable name is not so clear. Better would be e.g. `$selectAllTeams`

PR #588, opened 2019-05-05
Use external judgements for scoreboard calculation when shadowing.

nickygerritsen:
I think @meisterT came up with this idea: when we are shadowing, we should use the external judgements to calculate our scoreboard, instead of using our own judgings.

I tested this by shadowing our online contest and comparing scoreboards afterwards. They are 100% the same.

### DIFF
--- a/webapp/src/Service/ScoreboardService.php
+++ b/webapp/src/Service/ScoreboardService.php
@@ -15,11 +15,12 @@ use App\Entity\TeamCategory;
 use App\Utils\FreezeData;
 use App\Utils\Scoreboard\Filter;
 use App\Utils\Scoreboard\Scoreboard;
-use App\Utils\Scoreboard\SingleTeamScoreboard;
 use App\Utils\Scoreboard\TeamScore;
 use App\Utils\Utils;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Query\Expr\Join;
+use App\Entity\ExternalJudgement;
+use App\Utils\Scoreboard\SingleTeamScoreboard;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -293,6 +294,11 @@ class ScoreboardService
                                          $lockString));
         }
 
+        // Determine whether we will use external judgements instead of judgings
+        $localSource           = DOMJudgeService::DATA_SOURCE_LOCAL;
+        $shadow                = DOMJudgeService::DATA_SOURCE_CONFIGURATION_AND_LIVE_EXTERNAL;
+        $useExternalJudgements = $this->dj->dbconfig_get('data_source', $localSource) == $shadow;
+
         // Note the clause 's.submittime < c.endtime': this is used to
         // filter out TOO-LATE submissions from pending, but it also means
         // that these will not count as solved. Correct submissions with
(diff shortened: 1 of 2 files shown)


webapp/src/DOMJudgeBundle/Command/ImportEventFeedCommand.php
```diff
@@ -1418,6 +1418,13 @@ protected function importJudgement(array $event)
         }
         $this->em->flush();
 
+        // Now we need to update the scoreboard cache to get this judgement result in
```
meisterT: nit: scoreboard cache of this particular scoreboard cell

webapp/src/DOMJudgeBundle/Service/ScoreboardService.php
```diff
@@ -306,6 +311,22 @@ public function calculateScoreRow(Contest $contest, Team $team, Problem $problem
             ->setParameter(':cid', $contest->getCid())
             ->orderBy('s.submittime');
 
+        if ($useExternalJudgements) {
+            $queryBuilder
+                ->addSelect('ej')
+                ->leftJoin('s.external_judgements', 'ej')
+                // We want the external judgement that was started the latest, because that is the
+                // current valid one. We do this by left-joining on external judgements again, bot
```
meisterT: What about marking the external judgements as valid/invalid during import? This would make this query easier and closer to the original.

nickygerritsen: I thought about this, but how are we going to do that properly? The feed could expose runs in an order that differs from sorting by `starttime`. Do we then, every time we encounter a judgement, update all judgements to `valid=false` except the one with the highest `starttime`? This would be kinda similar to what we do here, but then we only do it in one spot.

meisterT: Yes, I think that would be nicer to do it during import. It should be rare anyway, so it's not costly.

nickygerritsen: I will come up with a way to do that and then remove the ugliness here.

thijskh:
Why do we want this?

nickygerritsen:
> Why do we want this?

Basically, to make Doug happy 🙈 
The reason is that in shadow mode it is interesting to see if there are differences, but it is way better to use the primary results for the scoreboard, so we only check if our scoreboard calculation is correct, not our determination of the result.

meisterT:
Well, it also makes me more happy :-)

So let's assume we have a bug in judging that we cannot easily fix - but identify. Then we still would want to be able to compare scoreboards. This PR separates the individual comparison on submissions from the global comparison on the scoreboard.

webapp/src/Service/ScoreboardService.php
```diff
@@ -412,17 +433,34 @@ public function calculateScoreRow(Contest $contest, Team $team, Problem $problem
             //   - either it's still ongoing (pending judgement, could be correct)
             //   - or already judged to be correct (if it's judged but != correct, it's not a first to solve)
             // - or the submission is still queued for judgement (judgehost is NULL).
-            $firstToSolve = 0 == $this->em->getConnection()->fetchColumn('
+            if ($useExternalJudgements) {
```
meisterT: I think this would be more readable by just repeating the SQL copy instead of avoiding repetition in an complex way.
Feel free to leave it like it is now since it is a matter of taste.

nickygerritsen: After trying it I have to agree so I will do it

PR #594, opened 2019-05-12
Do not check dependent events while applying rejudgings

nickygerritsen:
We should test this on a large rejudging, but on my test with ~20 submissions it went from ~22 seconds to ~1 second in production mode, where there is also an initial overhead for fetching the submissions to apply.

It should now be as fast again as it was when we ported this code to Symfony, after we replaced Doctrine calls with direct SQL queries and I think we benchmarked that and concluded it was good enough (for now).

But I don't have a database where I have a rejudging ready with that many submissions currently, so hopefully someone else can create one to test the speed.

### DIFF
--- a/webapp/src/DOMJudgeBundle/Service/RejudgingService.php
+++ b/webapp/src/DOMJudgeBundle/Service/RejudgingService.php
@@ -138,6 +138,18 @@ class RejudgingService
 
         $this->dj->auditlog('rejudging', $rejudgingId, $action . 'ing rejudge', '(start)');
 
+        // Add missing state events for this contest. We do this here and disable doing it in the
+        // loop when calling EventLogService::log, because then we would do the same check a lot of
+        // times, which is really inefficient and slow. Note that it might be the case that a state
+        // change event will happen exactly during applying a rejudging *and* that no client is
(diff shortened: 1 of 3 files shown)


webapp/src/DOMJudgeBundle/Service/RejudgingService.php
```diff
@@ -138,6 +138,18 @@ public function finishRejudging(Rejudging $rejudging, string $action, callable $
...
 
+        // Add missing state events for this contest. We do this here and disable doing it in the
+        // loop when calling EventLogService::log, because then we would do the same check a lot of
+        // times, which is really inefficient and slow. Note that it might be the case that a state
+        // change event will happen exactly during applying a rejudging *and* that no client is
+        // listening. Given that applying a rejudging will only create judgement and run events and
+        // that for these events contest state change events don't really matter, we will only check
+        // it once, here.
+        // We will also not check dependant object events in the loop, because if we apply a
+        // rejudging, the original judgings will already have triggered all dependant events.
```
eldering: There's a typo: it should be "dependent".

nickygerritsen: Thanks, seems I fail with this more often, as I found many places in the code that I now have fixed.

PR #595, opened 2019-05-19
Add screen to show shadow differences in a matrix view. Fixes #210.

### LINKED ISSUE #210: Use matrix view for external diff.
Add a page/view to see the external diff results in a matrix style like rejudgings.

nickygerritsen:
This is the first version of a matrix-style view for shadow differences.

Some screenshots:

All submissions with a differences between local and external:
Screenshot 2019-05-19 at 16 13 09
Note that this will *not* show submissions that are still pending either locally or externally. I think this makes the most sense. There are separate buttons to view them

After clicking on the `3` in the `JU/AC` cell, I get all submissions that are accepted externally but still judging locally:
Screenshot 2019-05-19 at 16 15 07

The view might be improved. I am open for (simple) UI improvements in this PR, but if we want bigger changes I'd propose to merge this first and let someone like @thijskh do the UI changes.

Technically it is very similar to the rejudging view, where I build up a matrix, a set of buttons and a submission list.
Notable changes I made:
* The shadow matrix also contains a column/row for 'judging' submissions, to see them quickly.
* I added some restrictions to the submission retrieval function to filter on external judgements.
* The submission list twig partial has gotten an extra option to show external testcases. When enabled, it will use two columns for some of the table fields, such that it can display the local and external result and testcase results below each other.

### DIFF
--- a/webapp/src/Controller/Jury/ShadowDifferencesController.php
+++ b/webapp/src/Controller/Jury/ShadowDifferencesController.php
@@ -1,0 +1,207 @@
+<?php declare(strict_types=1);
+
+namespace App\Controller\Jury;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Query\Expr\Join;
+use App\Controller\BaseController;
+use App\Entity\ExternalJudgement;
+use App\Entity\Judging;
+use App\Entity\Submission;
+use App\Service\DOMJudgeService;
+use App\Service\SubmissionService;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
+use Symfony\Component\Routing\Annotation\Route;
+
+/**
+ * @Route("/jury/shadow-differences")
+ * @IsGranted("ROLE_ADMIN")
+ */
+class ShadowDifferencesController extends BaseController
+{
+    /**
+     * @var DOMJudgeService
+     */
+    protected $dj;
+
+    /**
+     * @var SubmissionService
+     */
+    protected $submissions;
+
+    /**
+     * @var SessionInterface
+     */
+    protected $session;
+
+    /**
+     * @var EntityManagerInterface
+     */
+    protected $em;
+
+    public function __construct(
+        DOMJudgeService $dj,
+        SubmissionService $submissions,
+        SessionInterface $session,
+        EntityManagerInterface $em
+    ) {
+        $this->dj          = $dj;
+        $this->submissions = $submissions;
+        $this->session     = $session;
+        $this->em          = $em;
+    }
+
+    /**
+     * @Route("", name="jury_shadow_differences")
+     */
+    public function indexAction(Request $request)
+    {
+        $shadowMode = DOMJudgeService::DATA_SOURCE_CONFIGURATION_AND_LIVE_EXTERNAL;
+        $dataSource = $this->dj->dbconfig_get('data_source', DOMJudgeService::DATA_SOURCE_LOCAL);
+        if ($dataSource != $shadowMode) {
+            $this->addFlash('danger', sprintf(
+                'Shadow differences only supported when data_source is %d',
+                $shadowMode
+            ));
+            return $this->redirectToRoute('jury_index');
+        }
+
+        // Close the session, as this might take a while and we don't need the session below
+        $this->session->save();
+
+        $contest        = $this->dj->getCurrentContest();
+        $verdictsConfig = $this->dj->getDomjudgeEtcDir() . '/verdicts.php';
+        $verdicts       = array_merge(['judging' => 'JU'], include $verdictsConfig);
+
+        $used         = [];
+        $verdictTable = [];
+        // pre-fill $verdictTable to get a consistent ordering
+        foreach ($verdicts as $verdict => $abbrev) {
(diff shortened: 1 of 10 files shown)


webapp/src/Controller/Jury/ShadowDifferencesController.php
```diff
@@ -0,0 +1,207 @@
...
+    /**
+     * @Route("", name="jury_shadow_differences")
+     */
+    public function indexAction(Request $request)
+    {
+        $shadowMode = DOMJudgeService::DATA_SOURCE_CONFIGURATION_AND_LIVE_EXTERNAL;
+        $dataSource = $this->dj->dbconfig_get('data_source', DOMJudgeService::DATA_SOURCE_LOCAL);
+        if ($dataSource != $shadowMode) {
+            $this->addFlash('danger', sprintf(
+                'Shadow differences only supported when data_source is %d',
```
meisterT: This is not an actionable comment: in the long run, I would like to even import results from the shadow when we're primary.

webapp/src/Controller/Jury/ShadowDifferencesController.php
```diff
@@ -0,0 +1,207 @@
...
+            ));
+            return $this->redirectToRoute('jury_index');
+        }
+
+        // Close the session, as this might take a while and we don't need the session below
+        $this->session->save();
+
+        $contest        = $this->dj->getCurrentContest();
+        $verdictsConfig = $this->dj->getDomjudgeEtcDir() . '/verdicts.php';
+        $verdicts       = array_merge(['judging' => 'JU'], include $verdictsConfig);
```
meisterT: Kattis also has JE/judge error - how would we process/display that?

webapp/src/Controller/Jury/ShadowDifferencesController.php
```diff
@@ -0,0 +1,207 @@
...
+            // add row
+            $verdictTable[$unknownVerdict] = [];
+            foreach ($verdicts as $verdict => $abbreviation) {
+                $verdictTable[$unknownVerdict][$verdict] = [];
+            }
+        };
+
+        // Build up the verdict matrix
+        foreach ($submissions as $submitid => $submission) {
+            /** @var ExternalJudgement|null $externalJudgement */
```
meisterT: More a nit: why is the one table name ...judgement and the other one ...judging?

webapp/templates/jury/index.html.twig
```diff
@@ -60,6 +60,9 @@
                         <li><a href="{{ path('jury_scoreboard') }}">Scoreboard</a></li>
                         <li><a href="{{ path('analysis_index') }}">Statistics/Analysis</a></li>
                         <li><a href="{{ path('jury_submissions') }}">Submissions</a></li>
+                        {% if is_granted('ROLE_ADMIN') and data_source == constant('App\\Service\\DOMJudgeService::DATA_SOURCE_CONFIGURATION_AND_LIVE_EXTERNAL') %}
```
meisterT: this kind of looks ugly - perhaps do it in PHP instead of twig?

meisterT: same below

webapp/templates/jury/shadow_differences.html.twig
```diff
@@ -0,0 +1,96 @@
...
+        {% include 'jury/partials/shadow_submissions.html.twig' %}
+    </div>
+
+{% endblock %}
+
+{% block extrafooter %}
+    <script>
+        $(function () {
+            $('input[name=viewtype]').on('change', function () {
+                window.location = '{{ path('jury_shadow_differences', {view: 'REPLACE_ME', external: external, local: local}) }}'.replace('REPLACE_ME', $(this).val());
```
meisterT: why is this replace me necessary?

webapp/src/Controller/Jury/ShadowDifferencesController.php
nickygerritsen: What do you suggest, that we remove this limitation in full?

webapp/src/Controller/Jury/ShadowDifferencesController.php
nickygerritsen: Currently we don't, because the import command only accepts stuff from `verdicts.php`. We should fix this at a later point I guess, but not sure how yet.

webapp/src/Controller/Jury/ShadowDifferencesController.php
nickygerritsen: I picked judgement because that is how CLICS calls it.

webapp/src/Controller/Jury/ShadowDifferencesController.php
meisterT: As I said: my comment is not actionable. I would not do anything right now but keep in mind that even in primary mode we might want to show diffs to shadow on our side (and don't add anything that makes this harder).

webapp/src/Controller/Jury/ShadowDifferencesController.php
meisterT: Perhaps leave a TODO.

webapp/src/Controller/Jury/ShadowDifferencesController.php
meisterT: ACK.

webapp/src/Controller/Jury/ShadowDifferencesController.php
nickygerritsen: I have added it to the import command

PR #602, opened 2019-06-20
Various makefile changes

### LINKED ISSUE #601: wrong permission at webapp/app/config/parameters.yml
webapp/app/config/parameters.yml is installed with 644 with root as owner. But this file contains database's password. It should be installed as the same permission as dbpasswords.secret does.

### Your environment
DOMjudge 7.0.2

eldering:
These fix #601 and clean up the special ownership installation a bit: by consolidating those in the root Makefile, we easily which commands fail and need to be rerun when not installing as root.

Also reverts a dirty hack to install `restapi.secret` with webserver group, from when we hadn't fully moved to Symfony yet. @nickygerritsen can you double check this is ok to revert?

### DIFF
--- a/webapp/Makefile
+++ b/webapp/Makefile
@@ -44,8 +44,7 @@ install-domserver:
 			ln -s $$realtarget $$link ; \
 		fi \
 	done
-	cp        -t $(DESTDIR)$(domserver_webappdir) phpunit.xml.dist
-	chmod a+r    $(DESTDIR)$(domserver_webappdir)/phpunit.xml.dist
+	$(INSTALL_DATA) -t $(DESTDIR)$(domserver_webappdir) phpunit.xml.dist
 
 install-docs:
 	$(INSTALL_DATA) -T web/images/countries/README    $(DESTDIR)$(domjudge_docdir)/README.country_flags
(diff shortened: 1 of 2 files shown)


webapp/Makefile
```diff
@@ -44,12 +45,7 @@ install-domserver:
 			ln -s $$realtarget $$link ; \
 		fi \
 	done
-	-for d in cache logs sessions ; do \
-		$(INSTALL_WEBSITE) -m 0775 -d $(DESTDIR)$(domserver_webappdir)/var/$$d ; \
-	done
-	-$(INSTALL_WEBSITE) -t $(DESTDIR)$(domserver_webappdir)/var var/*.php var/*.cache
-	cp        -t $(DESTDIR)$(domserver_webappdir) phpunit.xml.dist
-	chmod a+r    $(DESTDIR)$(domserver_webappdir)/phpunit.xml.dist
+	$(INSTALL_DATA) -t $(DESTDIR)$(domserver_webappdir) phpunit.xml.dist
```
thijskh: Do we need to install phpunit.xml at all? Is it not only for dev?

eldering: It's used in Gitlab CI, which does a plain install (not maintainer-mode), so we can't simply delete it, or only install in maintainer-mode. Any alternatives to leaving this as is?

thijskh: It does not seem worth it to jump many hoops for not installing this small file, does not really hurt to keep it

Issue #581, opened 2019-05-03
Balloons status page not showing correct information


ss1h2a3tw:
### Description of the problem / feature request
In the page `domjudge/jury/balloons`, it should show the balloon status. But this page changes every time I reload it. There are even submissions not in the current contest.  

### Your environment
Arch Linux
https://github.com/DOMjudge/domjudge/commit/84b4f56b5e014e1d3c3f164134fc8e8509b8d1e9

### Steps to reproduce
The current environment have held two contests, so maybe the case here is quite complicated. I have my database dumps and logs here. 
prod.log: https://drive.google.com/file/d/1ZV0BsCl66JCPQSJwNVOwymzhtHZ8F438/view?usp=sharing
sql: https://drive.google.com/file/d/1sd2DYfV8CbCrxOz46AW4zsgfk3aMLLNs/view?usp=sharing

thijskh:
The balloons page currently is not scoped to the selected contest.

I believe indeed that this is not nice per se (but a legacy from the time we were single contest).

The approach is now different per page. Some will only show data for the selected contest. Others (e.g. clarifications) will show all, but with an indication of the contest it pertains to. Some are hybrid, like problems. Definitely something we need to be more consistent in.

Probably a good idea if we decide how we want to go forward in general and then clean up all pages to match.

ss1h2a3tw:
But the page info I got here does not match the scenario if it should list all of the activated contests. They even list deactivated contests but not all the submission from that contest. And sometimes it shows nothing.

thijskh:
Thanks. In that case we need to investigate.

thijskh:
Small data point: the (arguably simple) cosmic ray contest had a correct list of balloons.

thijskh:
The database dump is not complete so I cannot import it to reproduce the problem. Looking at the dump, there's a number of balloons listed for contest 6 and a small number for contest 7. Is that not what you see?

ss1h2a3tw:
Yes, but there are not all of the submission of either contest. Here are some screenshots of the scoreboard and the balloon pages. https://imgur.com/a/46KgHi0

thijskh:
Were you by any chance also running the `misc-tools/balloons` daemon?

ss1h2a3tw:
No, it is not even installed in the domserver host. Should it be installed?

ss1h2a3tw:
Ah, sorry. It **is** installed.

thijskh:
Installed but not used?

[...]

PR #607, opened 2019-07-09
Balloonpage Filtering

DerBaer0:
Slightly reworked the Balloon Page using a custom table instead of the macro and allowing it to be filtered by affiliation. The filter is now persistent across refreshes.
Also fixes a bug that might be the reason for #581

### DIFF
--- a/webapp/src/DOMJudgeBundle/Controller/Jury/BalloonController.php
+++ b/webapp/src/DOMJudgeBundle/Controller/Jury/BalloonController.php
@@ -168,11 +168,10 @@ class BalloonController extends Controller
             if ( $balloon->getDone() ) {
                 $cssclass = 'disabled';
                 $balloonactions = [[]];
-                $balloondata['done'] = True;
+                $balloondata['done'] = true;
             } else {
                 $cssclass = null;
-                $balloondata['done'] = False;
-                $balloondata['done'];
+                $balloondata['done'] = false;
                 $balloonactions = [[
                     'icon' => 'running',
                     'title' => 'mark balloon as done',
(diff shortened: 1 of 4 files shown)


webapp/src/DOMJudgeBundle/Controller/Jury/BalloonController.php
```diff
@@ -165,17 +163,16 @@ public function indexAction(Request $request, Packages $assetPackage, KernelInte
...
 
-            $balloondata['awards']['value'] = implode('; ', $comments);
+            $balloondata['awards'] = implode('; ', $comments);
 
             if ( $balloon->getDone() ) {
                 $cssclass = 'disabled';
                 $balloonactions = [[]];
-                $balloondata['status']['value'] = '<i class="far fa-check-circle"></i>';
-                $balloondata['status']['sortvalue'] = '1';
+                $balloondata['done'] = True;
```
nickygerritsen: Nitpick: I think our current coding standard says these should be `true`/`false` without a capital first letter.

webapp/src/DOMJudgeBundle/Controller/Jury/BalloonController.php
```diff
@@ -184,19 +180,35 @@ public function indexAction(Request $request, Packages $assetPackage, KernelInte
...
+                ->from('DOMJudgeBundle:TeamAffiliation', 'a')
+                ->select('a')
+                ->where('a.affilid IN (:affilIds)')
+                ->setParameter(':affilIds', $filters['affiliation-id'])
+                ->getQuery()
+                ->getResult();
+        }
+
         return $this->render('@DOMJudge/jury/balloons.html.twig', [
             'refresh' => ['after' => 60, 'url' => $this->generateUrl('jury_balloons')],
```
nickygerritsen: This now needs to also have `'ajax' => true` to let the page refresh using AJAX.

PR #611, opened 2019-07-28
Make database SQL consistent with Doctrine entities

nickygerritsen:
(Wow that was a lot of work).

This PR makes it so that if we run `webapp/bin/console doctrine:migrations:diff` it will now tell us there are no differences 🎉 .

The most interesting to check is the actual SQL files. The upgrade one *should* match with the DB structure. I will comment on some changes in the DB structure to explain what I did and why I did it.


sql/mysql_db_structure.sql
```diff
@@ -37,7 +37,7 @@ CREATE TABLE `balloon` (
 -- Table structure for table `clarification`
 --
 CREATE TABLE `clarification` (
-  `clarid` int(4) unsigned NOT NULL AUTO_INCREMENT COMMENT 'Unique ID',
+  `clarid` int(4) unsigned NOT NULL AUTO_INCREMENT COMMENT 'Unique clarification ID',
```
nickygerritsen: I changed the comment of this, and other primary keys, to not be "Unique ID", because if they are used as a foreign key in another table, Doctrine will reuse the comment. Sometimes it is possible to add an `@ORM\Column` to that table to overwrite this behaviour, but not in the case of ManyToMany associations or when the foreign key is part of the primary key.

We could even think about changing the comment from "Unique X ID" to "X ID", as the ID is not really unique in the table where it is a foreign key. Any thoughts?

p.s.: I never ever have even read the MySQL field comments. Especially now that we have an ORM, IMHO it is not that useful, but others might think otherwise.

sql/mysql_db_structure.sql
```diff
@@ -56,9 +56,13 @@ CREATE TABLE `clarification` (
   KEY `probid` (`probid`),
   KEY `cid` (`cid`),
   KEY `cid_2` (`cid`,`answered`,`submittime`),
+  KEY `sender` (`sender`),
```
nickygerritsen: Doctrine assumes we have keys for all foreign keys, so I have added them.

sql/mysql_db_structure.sql
```diff
@@ -56,9 +56,13 @@ CREATE TABLE `clarification` (
...
+  KEY `recipient` (`recipient`),
   CONSTRAINT `clarification_ibfk_1` FOREIGN KEY (`cid`) REFERENCES `contest` (`cid`) ON DELETE CASCADE,
   CONSTRAINT `clarification_ibfk_2` FOREIGN KEY (`respid`) REFERENCES `clarification` (`clarid`) ON DELETE SET NULL,
-  CONSTRAINT `clarification_ibfk_3` FOREIGN KEY (`probid`) REFERENCES `problem` (`probid`) ON DELETE SET NULL
+  CONSTRAINT `clarification_ibfk_3` FOREIGN KEY (`probid`) REFERENCES `problem` (`probid`) ON DELETE SET NULL,
```
nickygerritsen: Doctrine assumes we have foreign keys for associations, so these are added.

sql/mysql_db_structure.sql
```diff
@@ -68,7 +72,7 @@ CREATE TABLE `clarification` (
 CREATE TABLE `configuration` (
   `configid` int(4) unsigned NOT NULL AUTO_INCREMENT COMMENT 'Unique ID',
   `name` varchar(32) NOT NULL COMMENT 'Name of the configuration variable',
-  `value` longtext NOT NULL COMMENT 'Content of the configuration variable (JSON encoded)',
+  `value` longtext NOT NULL COMMENT 'Content of the configuration variable (JSON encoded)(DC2Type:json)',
```
nickygerritsen: The `DC2Type` thing is needed by Doctrine when one uses a non-built-in type, so it knows what to do with the column.

sql/mysql_db_structure.sql
```diff
@@ -99,13 +103,13 @@ CREATE TABLE `contest` (
   `endtime_string` varchar(64) NOT NULL COMMENT 'Authoritative absolute or relative string representation of endtime',
   `unfreezetime_string` varchar(64) DEFAULT NULL COMMENT 'Authoritative absolute or relative string representation of unfreezetime',
   `deactivatetime_string` varchar(64) DEFAULT NULL COMMENT 'Authoritative absolute or relative string representation of deactivatetime',
-  `finalizetime` decimal(32,9) DEFAULT NULL COMMENT 'Time when contest was finalized, null if not yet',
+  `finalizetime` decimal(32,9) UNSIGNED DEFAULT NULL COMMENT 'Time when contest was finalized, null if not yet',
```
nickygerritsen: IMHO this should be unsigned

sql/mysql_db_structure.sql
```diff
@@ -99,13 +103,13 @@ CREATE TABLE `contest` (
...
   `finalizecomment` text DEFAULT NULL COMMENT 'Comments by the finalizer',
   `b` smallint(3) unsigned NOT NULL default 0 COMMENT 'Number of extra bronze medals',
   `enabled` tinyint(1) unsigned NOT NULL DEFAULT 1 COMMENT 'Whether this contest can be active',
   `starttime_enabled` tinyint(1) unsigned NOT NULL DEFAULT 1 COMMENT 'If disabled, starttime is not used, e.g. to delay contest start',
-  `process_balloons` tinyint(1) UNSIGNED DEFAULT 1 COMMENT 'Will balloons be processed for this contest?',
-  `public` tinyint(1) UNSIGNED DEFAULT 1 COMMENT 'Is this contest visible for the public and non-associated teams?',
+  `process_balloons` tinyint(1) UNSIGNED DEFAULT 1 NOT NULL COMMENT 'Will balloons be processed for this contest?',
```
nickygerritsen: I don't know why it was nullable, did that mean something? Same for public below.

sql/mysql_db_structure.sql
```diff
@@ -138,11 +142,11 @@ CREATE TABLE `contestproblem` (
...
 
 CREATE TABLE `contestteam` (
-  `cid` int(4) unsigned NOT NULL COMMENT 'Contest ID',
-  `teamid` int(4) unsigned NOT NULL COMMENT 'Team ID',
+  `cid` int(4) unsigned NOT NULL COMMENT 'Unique contest ID',
+  `teamid` int(4) unsigned NOT NULL COMMENT 'Unique team ID',
   PRIMARY KEY (`cid`,`teamid`),
-  KEY `cid` (`cid`),
-  KEY `teamid` (`teamid`),
+  KEY `IDX_8328F8554B30D9C4` (`cid`),
```
nickygerritsen: For ManyToMany join tables we can not really influence the name of keys. Doctrine uses a hash of the fields for this.

sql/mysql_db_structure.sql
```diff
@@ -184,7 +188,7 @@ CREATE TABLE `executable` (
 
 CREATE TABLE `external_judgement` (
   `extjudgementid` int(4) unsigned NOT NULL AUTO_INCREMENT COMMENT 'Unique ID',
-  `externalid` varchar(255) DEFAULT NULL COMMENT 'Judgement ID in external system, should be unique inside a single contest',
+  `externalid` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin DEFAULT NULL COMMENT 'Judgement ID in external system, should be unique inside a single contest',
```
nickygerritsen: The external ID `a` should be different from `A`, so we need utf8mb4_bin

sql/mysql_db_structure.sql
```diff
@@ -319,12 +324,13 @@ CREATE TABLE `judging_run` (
 
 CREATE TABLE `judging_run_output` (
   `runid` int(4) unsigned NOT NULL COMMENT 'Run ID',
-  `output_run` longblob DEFAULT NULL COMMENT 'Output of running the program',
-  `output_diff` longblob DEFAULT NULL COMMENT 'Diffing the program output and testcase output',
-  `output_error` longblob DEFAULT NULL COMMENT 'Standard error output of the program',
-  `output_system` longblob DEFAULT NULL COMMENT 'Judging system output',
+  `output_run` longblob DEFAULT NULL COMMENT 'Output of running the program(DC2Type:blobtext)',
```
nickygerritsen: I have created `blobtext` for fields that are blobs in the DB but should not be a `resource` in PHP.

sql/mysql_db_structure.sql
```diff
@@ -319,12 +324,13 @@ CREATE TABLE `judging_run` (
...
+  `output_diff` longblob DEFAULT NULL COMMENT 'Diffing the program output and testcase output(DC2Type:blobtext)',
+  `output_error` longblob DEFAULT NULL COMMENT 'Standard error output of the program(DC2Type:blobtext)',
+  `output_system` longblob DEFAULT NULL COMMENT 'Judging system output(DC2Type:blobtext)',
   PRIMARY KEY  (`runid`),
-  CONSTRAINT `judging_run_output_ibfk_1` FOREIGN KEY (`runid`) REFERENCES `judging_run` (`runid`)
+  KEY `runid` (`runid`),
+  CONSTRAINT `judging_run_output_ibfk_1` FOREIGN KEY (`runid`) REFERENCES `judging_run` (`runid`) ON DELETE CASCADE
```
nickygerritsen: Cascading the delete makes sense to me

sql/mysql_db_structure.sql
```diff
@@ -364,23 +372,31 @@ CREATE TABLE `problem` (
...
+  `cid` int(4) unsigned NOT NULL COMMENT 'Unique contest ID',
+  `teamid` int(4) unsigned NOT NULL COMMENT 'Unique team ID',
   `points_restricted` int(4) unsigned NOT NULL DEFAULT 0 COMMENT 'Total correctness points (restricted audience)',
   `totaltime_restricted` int(4) NOT NULL DEFAULT 0 COMMENT 'Total penalty time in minutes (restricted audience)',
   `points_public` int(4) unsigned NOT NULL DEFAULT 0 COMMENT 'Total correctness points (public)',
   `totaltime_public` int(4) NOT NULL DEFAULT 0 COMMENT 'Total penalty time in minutes (public)',
   PRIMARY KEY (`cid`,`teamid`),
   KEY `order_restricted` (`cid`,`points_restricted`,`totaltime_restricted`) USING BTREE,
-  KEY `order_public` (`cid`,`points_public`,`totaltime_public`) USING BTREE
+  KEY `order_public` (`cid`,`points_public`,`totaltime_public`) USING BTREE,
```
nickygerritsen: Unrelated, but is there any reason we have `USING BTREE` here (and only for these two lines)?

sql/mysql_db_structure.sql
```diff
@@ -477,13 +499,15 @@ CREATE TABLE `submission` (
   KEY `judgehost_2` (`judgehost`),
   KEY `origsubmitid` (`origsubmitid`),
   KEY `rejudgingid` (`rejudgingid`),
+  KEY `probid_2` (`cid`,`probid`),
```
nickygerritsen: This one is needed because we link ContestProblem to Submission

sql/mysql_db_structure.sql
eldering: I would really prefer to keep the SQL comments, and find it a bit silly that Doctrine reuses those comments blindly in FKs. But if it does so, I'm fine with renaming those to "X ID".

sql/mysql_db_structure.sql
eldering: We should indeed!

sql/mysql_db_structure.sql
eldering: Yes! The fact that we can compare on this key is used here: https://github.com/DOMjudge/domjudge/blob/master/webapp/src/Service/ScoreboardService.php#L181

sql/mysql_db_structure.sql
eldering: I doubt that was on purpose.

sql/mysql_db_structure.sql
nickygerritsen: But does that only work when we have btree explicitly set?

eldering: No, but with BTREE that query is better optimized. Also BTREE is the default index type nowadays (instead of HASH), but still better to be explicit about it.

PR #613, opened 2019-08-04
Doctrine migrations

nickygerritsen:
This replaces our custom SQL install/upgrade scripts with Doctrine migrations and Doctrine fixtures.

Some remarks:

* I have used Fixtures for the example data but the normal migrations for the default data. The reason is that if we have new default data, we need to insert it during upgrades, while for example data we only provide a set of data that you can install after a fresh install. I think this gives the best solution for both.
* All these version files are basically the upgrade scripts.
* For the version files that upgrade the structure up to and including 7.0.0 I have created one file per version upgrade and have not provided `down()` migrations. Also I have added a `skipIf` at the beginning of each, so people upgrading from older versions won't get errors that it can not apply the migration.
* For the migrations after 7.0 I have created one migration per logical part to upgrade. Here I *have* provided `down` migrations, which I also tested. Also no `skipIf`'s here, because people will never have them, besides people using master and those people can manually indicate they have the version already. For future upgrades to master this will be way easier, because we can split the migrations per logical unit.
* We have lost the functionality of passing `-r` or reading `~/.my.cnf` in `dj_setup_database` for the migrations. I see no easy way to get it back but if somebody has an idea I am open to it.
* We might want to consider to move the stuff that is in `sql` somewhere else, because there are no actual SQL files in that directory anymore...
* I have compared the DB dumps of the old script and this new one, both with and without example data, and they match besides some expected things and an inconsistency we had with contest times in the example data.
* We should come up with a way to check demoweb, although we could say we just need to check if all migrations are applied. Another option would be to rerun the whole migration on a separate database and compare the structure of both databases.


thijskh:
Instead of including all the historic migrations, we could also require that you upgrade to 7.0 first before you upgrade to 7.1. How many people will be directly upgrading from 3.0 to 7.1?

nickygerritsen:
> Instead of including all the historic migrations, we could also require that you upgrade to 7.0 first before you upgrade to 7.1. How many people will be directly upgrading from 3.0 to 7.1?

But we do need the SQL for everything up to 7.0 anyway. The only thing we could do is put everything in one big .sql, i.e. take the `mysql_db_structure.sql`, `mysql_db_defaultdata.sql` and `mysql_db_files_defaultdata.sql` and put them all in one `Versionxxxxx` file. I'm not sure if that is better or has any advantages, besides having one file instead of 14 (but a very big one).

thijskh:
So in one release you need to keep the legacy SQL files and introduce the doctrine migrations. And in the release afterwards you can drop the SQL files.

nickygerritsen:
But then what is the difference with this? I now drop the SQL files immediately. I mean also for upcoming releases we need either the SQL files or the migrations of all previous versions (possibly combined) for setting up new installations.

The only thing we could do is merge all the pre-7.0 migrations in one, resulting in less files and maybe a few less lines (but not that many, only tables that we create and later drop or change will result in less lines).

I can see why it would make sense to only have one .sql file for everything <= 6.0 so if the majority thinks that is better I can merge them.

eldering:
We don't need all the historic SQL update scripts in versions >= 7.1. If you're on a version below 7.0, you first install 7.0 and use the old SQL scripts to upgrade the DB to 7.0. Then you install 7.1+ and use the new Doctrine scripts to upgrade from 7.0 to the current version.
This means we can drop the old SQL scripts in master at the cost of people upgrading from below 7.0 having to make a two step upgrade. I think that's acceptable.


On August 6, 2019 2:09:22 PM GMT+02:00, Nicky Gerritsen <notifications@github.com> wrote:
>But then what is the difference with this? I now drop the SQL files
>immediately. I mean also for upcoming releases we need either the SQL
>files or the migrations of all previous versions (possibly combined)
>for setting up new installations.
>
>The only thing we could do is merge all the pre-7.0 migrations in one,
>resulting in less files and maybe a few less lines (but not that many,
>only tables that we create and later drop or change will result in less
>lines).
>
>I can see why it would make sense to only have one .sql file for
>everything <= 6.0 so if the majority thinks that is better I can merge
>them.
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/DOMjudge/domjudge/pull/613#issuecomment-518641855

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

nickygerritsen:
> We don't need all the historic SQL update scripts in versions >= 7.1. If you're on a version below 7.0, you first install 7.0 and use the old SQL scripts to upgrade the DB to 7.0. Then you install 7.1+ and use the new Doctrine scripts to upgrade from 7.0 to the current version. This means we can drop the old SQL scripts in master at the cost of people upgrading from below 7.0 having to make a two step upgrade. I think that's acceptable.
> […](#)

My point is that we *also* dropped the non-update scripts, i.e. the file previously called `mysql_db_structure.sql`, since with Doctrine migrations upgrading and doing a fresh install uses the exact same process. So for people wo have no DOMjudge installation at all yet, we need an initial DB structure.

eldering:
> > We don't need all the historic SQL update scripts in versions >= 7.1. If you're on a version below 7.0, you first install 7.0 and use the old SQL scripts to upgrade the DB to 7.0. Then you install 7.1+ and use the new Doctrine scripts to upgrade from 7.0 to the current version. This means we can drop the old SQL scripts in master at the cost of people upgrading from below 7.0 having to make a two step upgrade. I think that's acceptable.
> > […](#)
> 
> My point is that we _also_ dropped the non-update scripts, i.e. the file previously called `mysql_db_structure.sql`, since with Doctrine migrations upgrading and doing a fresh install uses the exact same process. So for people wo have no DOMjudge installation at all yet, we need an initial DB structure.
> 
> That's why I suggusted that we could put all migrations <= 7.0 in one big migration file, making that basically the `mysql_db_structure.sql` for 7.0 and only having separate migration files for everything after that.

Hmmm. Just having a single file that is basically `mysql_db_structure.sql` but then written in "Doctrine script" sounds better to me than trying to install the database through this long list of upgrade scripts translated into "Doctrine script". Is that the normal way of installing the DB with Doctrine?

nickygerritsen:
Yes but normally the upgrade scripts are way smaller, since you create one for every change you do, like I did for 7.0->7.1.

PR #615, opened 2019-08-21
Improve logging in runpipe

eldering:


judge/runpipe.c
```diff
@@ -121,6 +121,20 @@ void verb(const char *format, ...)
...
 
+void join(char *dest, size_t size, char separator, const char **sources, size_t nsources)
+{
+	size_t i, len;
+
+	memset(dest, 0, size);
+	strncpy(dest, sources[0], size-1);
+	for(i=1; i<nsources; i++) {
+		len = strlen(dest);
+		if ( len>=size-1 ) return;
```
meisterT: Can we leave a bit room here to mark the string as abbreviated, e.g. with `...`?

eldering:
> I think this is fine. In general I think we could rewrite our whole c code in c++ for improved readability and safety.

I guess this is a left-over from making this a rewrite/reimplementation of various sources of C-code. I agree that it would be better to reimplement things in C++, especially the string/memory handling.

judge/runpipe.c
eldering: Hmm, then we're getting into nitty-gritty details: what if the string actually exactly fits? That would be difficult to distinguish from truncation; do we still print the `...` then? Although I agree in principle, I think that in practice the buffer of 1024 bytes should be sufficient 99.9% of the time, and I would leave it as is for simplicity.

eldering:
Ok, I changed `runpipe` to be C++ (only changing the string handling). This also got rid of the whole truncation question.

PR #618, opened 2019-09-06
Small fixes2

DerBaer0:


webapp/src/Migrations/Version20190803145008.php
```diff
@@ -41,7 +41,7 @@ public function up(Schema $schema): void
 SQL
         );
 
-        $this->addSql("INSERT INTO testcase_content (testcaseid, input, output, image, image_thumb) SELECT testcaseid, input, output, image, image_thumb FROM testcase`");
```
nickygerritsen: How did this never crash 🙈 . Maybe 'our' MySQL is less strict...

PR #623, opened 2019-09-21
Support selecting a distribution release when making the chroot

ankon:
This is needed when trying to set up a Ubuntu chroot, and works up-to 'xenial'.


meisterT:
Thanks!

PR #624, opened 2019-09-21
Set the PATH for running debootstrap

ankon:
This is required when running debootstrap on CentOS and other systems where the PATH
is different from what debootstrap needs.

### DIFF
--- a/misc-tools/dj_make_chroot.in
+++ b/misc-tools/dj_make_chroot.in
@@ -249,7 +249,7 @@ fi
 echo "Running debootstrap to install base system, this may take a while..."
 
 # shellcheck disable=SC2086
-eval $BOOTSTRAP_COMMAND $INCLUDEOPT $EXCLUDEOPT \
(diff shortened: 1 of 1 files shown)


ankon:
For reference: Solution came from https://www.linuxquestions.org/questions/debian-26/error-apt-get-not-found-during-installing-debian-4175602890/.

Host system is 
~~~~
# cat /etc/redhat-release 
CentOS Linux release 7.7.1908 (Core)
~~~~

ankon:
> First, the lefthand side $PATH should be without $.

Oops! Will fix.

> $BOOTSTRAP_COMMAND is already an absolute path, and from the linked issue, it seems only /bin and /sbin might be missing on the chrooted Debian/Ubuntu system.

The path to debootstrap is fine, the problem seems to be that stuff executing _as part of debootstrap_ assume these path entries -- in particular it fails extracting anything because `tar` invoked through `dpkg` isn't findable.

> Also, what's your target system? Because from CentOs 7.4 to Debian Stretch it works fine.

As per the added comment: CentOS 7.7.

eldering:
Merged in ff281bc7cc56f3bb88c50bdaa28be040378c2df5 with an extra comment.

ankon:
Thank you!

PR #630, opened 2019-09-23
When uploading judging runs fails, do not crash judgedaemon but add an internal error on the problem

nickygerritsen:
Crashing judgedaemons are never what we want. It is better to not judge the problem anymore and notify the admins.
This would happen in two instances:
* When PHP's post_max_size is too low.
* When using nginx and client_max_body_size is too low.

Screenshots of the internal error in these two cases:

Screenshot 2019-09-23 at 20 29 53
Screenshot 2019-09-23 at 20 30 08


eldering:
LGTM

PR #649, opened 2019-11-05
Fix link to "Developer information" documentation

vors:
https://www.domjudge.org/docs/admin-manual-11.html is the current place for "Developer information"


nickygerritsen:
Thanks!

PR #652, opened 2019-11-09
Don't use number of times updateClock function is called when calculating clock

nickygerritsen:
As reported by @ubergeek42 (and I also noticed this before), the clock on the upper right corner gets skewed quite a bit if you wait long enough, especially when the browser window/tab is not active.

I think this is becasue we use `setInterval` and add `1` to a variable each time. However if the interval does not get called exactly every second, you will have a skew.

The new approach determines the client offset to the server, by initially comparing the local time and server time and then uses this offset and the current time to determine server time.

Please double-triple check my logic....

### DIFF
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -128,8 +128,7 @@ function togglelastruns()
 // start (and end?).
 function updateClock()
 {
-	var curtime = initial+offset;
-	date.setTime(curtime*1000);
+	var curtime = Math.round((new Date().getTime() - clientOffset) / 1000);
 
 	var fmt = "";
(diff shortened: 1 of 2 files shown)


nickygerritsen:
> I think this is better, but there are still some (slightly more theoretical) issues:
> 
> * Do we fetch updates in contest start time?
Nope. But we didn’t do that before either... perhaps we add this info as some data- attributes somewhere? Although “somewhere” heavily depends on the page. Another option is to add a json endpoint that returns this and use that?
> * This will break if the client/server offset changes, e.g. due to DST change, or clock reset.
>   So I would propose to fetch both the server time and the contest times at each reload.
DST still has the same timestamp right? The other point is valid and we can take into account returning server time as well.

nickygerritsen:
As discussed on Slack, we will merge this and do the improved version later.

PR #653, opened 2019-11-11
Delete old float compare program.

meisterT:
We have
https://github.com/DOMjudge/domjudge/blob/master/sql/files/defaultdata/compare/compare.cc
for a while now which does support the default output validator options
as specified in
https://www.problemarchive.org/wiki/index.php/Problem_Format#Output_Validators
and also use it for our own fltcmp example problem.


meisterT:
This is merged in 481408e2ff53242ef55424790a6df72968196668

PR #658, opened 2019-11-17
Allow jury role to view and download executables.

### LINKED ISSUE #657: 403 for jury to view compile script source
A found by Pehr, if you have jury (but not admin) access, then you can view an individual (compile) script executable page, but from there the link to view executable content gives a 403 since you need admin rights.

We should allow jury to view (but not edit) executable script contents.

eldering:
Fixes #657.

### DIFF
--- a/webapp/src/Controller/Jury/ExecutableController.php
+++ b/webapp/src/Controller/Jury/ExecutableController.php
@@ -174,12 +174,12 @@ class ExecutableController extends BaseController
                     ]),
                     'ajaxModal' => true,
                 ];
-                $execactions[] = [
-                    'icon' => 'file-download',
-                    'title' => 'download this executable',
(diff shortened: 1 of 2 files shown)


webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -226,7 +226,6 @@ public function viewAction(Request $request, string $execId)
 
     /**
      * @Route("/{execId}/content", name="jury_executable_content")
-     * @IsGranted("ROLE_ADMIN")
```
meisterT: Don't we need to keep ROLE_JURY here (and in the other places)?

nickygerritsen: It has this on the whole controller level already

meisterT: Oh, I see now that this is on top of this file.

PR #668, opened 2019-11-21
Do not show FTS on team row scoreboard on team page during freeze.

nickygerritsen:
We always had `$jury = true` for the team row, so I removed that and replaced it with a flag that indicates whether we should show FTS in the freeze, since we only want that on the jury pages.

I tested the following:

* Submissions before freeze on both jury and team page show FTS
* Submissions during freeze on team page don't show FTS, but on jury it does
* After unfreeze it also shows FTS on team page

I think that is what we want, but please doublecheck...

### DIFF
--- a/webapp/src/Service/ScoreboardService.php
+++ b/webapp/src/Service/ScoreboardService.php
@@ -115,21 +115,16 @@ class ScoreboardService
      * Get scoreboard data for a single team based on the cached data in the
      * scorecache table.
      *
-     * @param Contest $contest The contest to get the scoreboard for.
-     * @param int     $teamId  The ID of the team to get the scoreboard for.
-     * @param bool    $jury    If true, the scoreboard will always be current.
-     *                         If false, frozen results will not be returned.
+     * @param Contest $contest         The contest to get the scoreboard for.
+     * @param int     $teamId          The ID of the team to get the scoreboard for.
+     * @param bool    $showFtsInFreeze If false, the scoreboard will hide first
+     *                                 to solve for submissions after contest freeze.
      * @return Scoreboard|null
      * @throws \Exception
      */
-    public function getTeamScoreboard(Contest $contest, int $teamId, bool $jury = false)
+    public function getTeamScoreboard(Contest $contest, int $teamId, bool $showFtsInFreeze = true)
     {
         $freezeData = new FreezeData($contest);
-
-        // Don't leak information before start of contest
-        if (!$jury && !$freezeData->started()) {
-            return null;
-        }
 
         $teams = $this->getTeams($contest, true, new Filter([], [], [], [$teamId]));
         if (empty($teams)) {
@@ -139,15 +134,11 @@ class ScoreboardService
         $problems   = $this->getProblems($contest);
         $rankCache  = $this->getRankcache($contest, $team);
         $scoreCache = $this->getScorecache($contest, $team);
-        if ($jury || !$freezeData->showFrozen()) {
-            $teamRank = $this->calculateTeamRank($contest, $team, $rankCache, $freezeData, $jury);
-        } else {
-            $teamRank = 0;
-        }
+        $teamRank   = $this->calculateTeamRank($contest, $team, $rankCache, $freezeData, true);
 
         return new SingleTeamScoreboard(
             $contest, $team, $teamRank, $problems,
-            $rankCache, $scoreCache, $freezeData, $jury,
+            $rankCache, $scoreCache, $freezeData, $showFtsInFreeze,
             (int)$this->dj->dbconfig_get('penalty_time', 20),
             (bool)$this->dj->dbconfig_get('score_in_seconds', false)
         );
--- a/webapp/src/Utils/Scoreboard/SingleTeamScoreboard.php
+++ b/webapp/src/Utils/Scoreboard/SingleTeamScoreboard.php
@@ -38,7 +38,7 @@ class SingleTeamScoreboard extends Scoreboard
     /**
      * @var bool
      */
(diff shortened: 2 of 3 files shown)


webapp/src/Utils/Scoreboard/SingleTeamScoreboard.php
```diff
@@ -100,7 +100,7 @@ protected function calculateScoreboard()
 
             $this->matrix[$scoreRow->getTeam()->getTeamid()][$scoreRow->getProblem()->getProbid()] = new ScoreboardMatrixItem(
                 $scoreRow->getIsCorrect($this->restricted),
-                $scoreRow->getIsFirstToSolve(),
+                $scoreRow->getIsCorrect($this->showFts) ? $scoreRow->getIsFirstToSolve() : false,
```
meisterT: do you really want getIsCorrect here?

nickygerritsen: I believe so. If it is not correct, we never want to show FTS (obviously). If it is correct, we want to show FTS if:

* It was before freeze
* It was during freeze but now is after unfreeeze for teams
* You are on the jury pages

The first and only parameter to `getIsCorrect` is whether we are restricted. For jury pages we set `showFts` to `true`, which corresponds to "not restricted". For team pages we set it to true only if we want to show frozen.

Or am I missing something?

meisterT: Can't it be simpler written as something along those lines:
```
$isCorrect = $scoreRow->getIsCorrect($this->restricted);
$this->matrix[$scoreRow->getTeam()->getTeamid()][$scoreRow->getProblem()->getProbid()] = new ScoreboardMatrixItem(
                $isCorrect,
                $isCorrect && $scoreRow->getIsFirstToSolve() && $this->showFts,
...
```

nickygerritsen: That logic is incorrect. It will hide *all* first to solves, also those that appear before the freeze. I do want to show those....

eldering: Half-related: I think it would be more natural to directly pass the `$showFts` argument to `$scoreRow->getIsFirstToSolve()`.

webapp/src/Service/ScoreboardService.php
```diff
@@ -115,22 +115,17 @@ public function getScoreboard(
...
-    public function getTeamScoreboard(Contest $contest, int $teamId, bool $jury = false)
+    public function getTeamScoreboard(Contest $contest, int $teamId, bool $showFtsInFreeze = true)
     {
         $freezeData = new FreezeData($contest);
 
-        // Don't leak information before start of contest
-        if (!$jury && !$freezeData->started()) {
-            return null;
-        }
-
```
eldering: Why are you removing this code? Was it irrelevant to begin with?

webapp/src/Service/ScoreboardService.php
```diff
@@ -139,15 +134,11 @@ public function getTeamScoreboard(Contest $contest, int $teamId, bool $jury = fa
         $problems   = $this->getProblems($contest);
         $rankCache  = $this->getRankcache($contest, $team);
         $scoreCache = $this->getScorecache($contest, $team);
-        if ($jury || !$freezeData->showFrozen()) {
-            $teamRank = $this->calculateTeamRank($contest, $team, $rankCache, $freezeData, $jury);
-        } else {
-            $teamRank = 0;
-        }
+        $teamRank   = $this->calculateTeamRank($contest, $team, $rankCache, $freezeData, true);
```
eldering: Are we now showing the team its rank during the freeze?

thijskh:
> We always had $jury = true for the team row, 

I think in the past we might also have shown this row or perhaps wanted to show this on the team’s public page.

webapp/src/Service/ScoreboardService.php
nickygerritsen: Jury is always true. At least I could not find any place in our code where it was false.
So this if always evaluated to false...

webapp/src/Service/ScoreboardService.php
nickygerritsen: No. We show a ‘?’ But that is handled elsewhere

nickygerritsen:
Ah yeah that might be. I can put back the unused var so we can do it later if we want

webapp/src/Utils/Scoreboard/SingleTeamScoreboard.php
nickygerritsen: I like that

nickygerritsen:
And yes the code is non trivial. Any ways to make it more trivial are highly appreciated

webapp/src/Service/ScoreboardService.php
eldering: Hmmm... I think that it's better to still handle it here in backend, so that we have all the logic in a single place. Frontend can do the rendering logic of showing the actual `?`, but I think it shouldn't be responsible for what we're allowed to show.

eldering:
> Ah yeah that might be. I can put back the unused var so we can do it later if we want

That might have been the case, but I'd argue that we there's no need to accomodate that potential use case if it complicates the current code.

webapp/src/Service/ScoreboardService.php
nickygerritsen: So you want to move the code here? That code that was here was never used.
However if we change `showFtsInFreeze` to something like `showFreezeInfo` we can basically reuse that variable to output a 0?

PR #677, opened 2019-12-18
Problemarchive

mvr320:
After discussion about if a feature should be implemented according to specs. I'm not sure if its not smarter to refer to the CLICS and have as sidenote the other (the superset) as DOMjudge is mainly developed for ICPC style contests.


eldering:
I've pushed a580a0f980b55849159fb4056c64a7fed9ef8745 to fix this.

PR #678, opened 2019-12-20
Allow multiple team categories for self-registration

tom93:
Replace the configuration option `registration_category_name` with a boolean property `TeamCategory::$allow_self_registration`. When it is set to true on multiple categories, they are offered as choices during registration.

---
_Original version_

Change the configuration option `registration_category_name` from a string to a list. When more than one category name is specified, users are offered a choice during registration.

An empty list disables self-registration (like an empty string did before).

---
_Cover letter_

Hi,

I'm involved with the New Zealand Programming Contest, which uses DOMjudge. We want self-registered teams to be able to choose their team category during registration instead of being assigned to a fixed category (we use categories such as "High School", "Tertiary", "Open" rather than "Participants", "Self-Registered").

I've written a patch that implements this, are you interested in merging it in?

Disclaimer: I am not familiar with PHP/Symfony. In particular, please check the way I've hidden the category field when only one category name is specified in the configuration (I used an HTML attribute to hide it, which I thought was simpler than conditionally building the Symfony form type).

Regards,  
Tom Levy

### DIFF
--- a/doc/manual/config-advanced.rst
+++ b/doc/manual/config-advanced.rst
@@ -77,11 +77,13 @@ Squid configuration for this might look like::
 
 Self-registration
 `````````````````
-There is also a configuration option to allow teams to self-register with
-the system: ``registration_category_name``. When left empty, no self-registration
-is allowed; when filled with a category name, newly registered teams will
-be placed in this category. During registration, a team can specify their
-affiliation.
+Teams can be allowed to self-register with the system. Each team category can
+be set to allow registration or not. When none of the categories are set to
+allow, self-registration is disabled. When one category is set to allow,
+self-registration is enabled and newly registered teams will be placed in this
+category. When multiple categories are set to allow, teams can choose one of
+them during registration. Teams can also specify their affiliation during
+registration.
 
 Executables
 -----------
--- a/webapp/src/Form/Type/UserRegistrationType.php
+++ b/webapp/src/Form/Type/UserRegistrationType.php
@@ -4,11 +4,13 @@ namespace App\Form\Type;
 
 use App\Entity\Team;
 use App\Entity\TeamAffiliation;
+use App\Entity\TeamCategory;
 use App\Entity\User;
 use App\Service\ConfigurationService;
 use App\Service\DOMJudgeService;
 use App\Utils\Utils;
 use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\EntityRepository;
 use Symfony\Bridge\Doctrine\Form\Type\EntityType;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
@@ -96,6 +98,30 @@ class UserRegistrationType extends AbstractType
                 ],
                 'mapped' => false,
             ]);
+
+        $selfRegistrationCategoriesCount = $this->em->getRepository(TeamCategory::class)->count(['allow_self_registration' => 1]);
+        if ($selfRegistrationCategoriesCount > 1) {
+            $builder
+                ->add('teamCategory', EntityType::class, [
+                    'class' => TeamCategory::class,
+                    'label' => false,
+                    'mapped' => false,
+                    'choice_label' => 'name',
+                    'placeholder' => '-- Select category --',
+                    'query_builder' => function (EntityRepository $er) {
+                        return $er
+                            ->createQueryBuilder('c')
+                            ->where('c.allow_self_registration = 1')
+                            ->orderBy('c.sortorder');
+                    },
+                    'attr' => [
+                        'placeholder' => 'Category',
+                    ],
+                    'constraints' => [
+                        new NotBlank(),
+                    ],
+                ]);
+        }
 
         if ($this->config->get('show_affiliations')) {
             $countries = [];
--- a/etc/db-config.yaml
+++ b/etc/db-config.yaml
@@ -231,11 +231,6 @@
             default_value: ""
             public: true
             description: If not empty, enable teams and jury to send source code to this command. See admin manual for allowed arguments.
-        -   name: registration_category_name
-            type: string
-            default_value: ""
-            public: true
-            description: Team category for users that register themselves with the system. Self-registration is disabled if this field is left empty.
         -   name: data_source
             type: int
             default_value: 0
--- a/webapp/src/Controller/Jury/TeamCategoryController.php
+++ b/webapp/src/Controller/Jury/TeamCategoryController.php
@@ -93,6 +93,7 @@ class TeamCategoryController extends BaseController
             'name' => ['title' => 'name', 'sort' => true],
             'num_teams' => ['title' => '# teams', 'sort' => true],
             'visible' => ['title' => 'visible', 'sort' => true],
+            'allow_self_registration' => ['title' => 'self-registration', 'sort' => true],
         ];
 
         // Insert external ID field when configured to use it
@@ -134,8 +135,9 @@ class TeamCategoryController extends BaseController
                 ];
             }
 
-            $categorydata['num_teams'] = ['value' => $teamCategoryData['num_teams']];
-            $categorydata['visible']   = ['value' => $teamCategory->getVisible() ? 'yes' : 'no'];
+            $categorydata['num_teams']               = ['value' => $teamCategoryData['num_teams']];
+            $categorydata['visible']                 = ['value' => $teamCategory->getVisible() ? 'yes' : 'no'];
+            $categorydata['allow_self_registration'] = ['value' => $teamCategory->getAllowSelfRegistration() ? 'yes' : 'no'];
 
             $team_categories_table[] = [
                 'data' => $categorydata,
--- a/webapp/src/Controller/SecurityController.php
+++ b/webapp/src/Controller/SecurityController.php
@@ -85,13 +85,12 @@ class SecurityController extends AbstractController
         $response = new Response();
         $response->headers->set('X-Login-Page', $this->generateUrl('login'));
 
-        $registrationCategoryName = $this->config->get('registration_category_name');
-        $registrationCategory     = $em->getRepository(TeamCategory::class)->findOneBy(['name' => $registrationCategoryName]);
+        $selfRegistrationCategoriesCount = $em->getRepository(TeamCategory::class)->count(['allow_self_registration' => 1]);
 
         return $this->render('security/login.html.twig', array(
             'last_username' => $lastUsername,
             'error' => $error,
-            'allow_registration' => $registrationCategory !== null,
+            'allow_registration' => $selfRegistrationCategoriesCount !== 0,
             'allowed_authmethods' => $authmethods,
             'auth_xheaders_present' => $request->headers->get('X-DOMjudge-Login'),
             'auth_ipaddress_users' => $auth_ipaddress_users,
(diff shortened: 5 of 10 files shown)


webapp/src/Form/Type/UserRegistrationType.php
```diff
@@ -85,6 +87,39 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+            if ($category) {
+                $registrationCategories[] = $category;
+            }
+        }
+        // if all the category names are invalid, the user won't be able to fill in the form; don't even show it
+        if (empty($registrationCategories)) {
+            throw new HttpException(500, 'Self-registration team categories are invalid');
+        }
+
+        $builder
```
nickygerritsen: I'd say it is indeed better to not add this element to the form if `$registrationCategories` contains only one item and then in the `SecurityController` do a check for if it is only one item and, if so, use that.

thijskh:
Would indeed merge it post-7.2 to not further increase the scope of that

webapp/src/Form/Type/UserRegistrationType.php
```diff
@@ -85,6 +87,39 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'mapped' => false,
             ]);
 
+        $registrationCategoryNames = $this->dj->dbconfig_get('registration_category_names', []);
+        $showCategoryChoice        = count($registrationCategoryNames) !== 1;
+        $registrationCategories    = [];
+        foreach ($registrationCategoryNames as $name) {
+            $category = $this->em->getRepository(TeamCategory::class)->findOneBy(['name' => $name]);
```
eldering: Why not do a single query:
```php
$registrationCategories = $this->em->getRepository(TeamCategory::class)
                             ->findBy(['name' => $registrationCategoryNames]);
```
That's simpler and does less DB queries.

webapp/src/Form/Type/UserRegistrationType.php
tom93: I've changed it to conditionally add the form element, but I think it should still be based on the number of categories in the configuration option (`$registrationCategoryNames`) rather than how many were found in the database (`$registrationCategories`).
Suppose the admin sets the option to two names and one of them is invalid. The intention is clearly to give users a choice. A dropdown with one category seems silly, but it will give a strong hint that the other category is invalid. Removing the dropdown entirely would be very confusing.
There is only a difference when the configuration option is invalid, and I think in that situation debuggablity is more important than presentation. Does that sound reasonable?

[Full diff](https://github.com/tom93/domjudge/compare/self-registration-multiple-categories-v1-hide-with-html-attribute..tom93:self-registration-multiple-categories-v2-conditionally-build-field), prettified diff:

```diff
--- a/webapp/src/Controller/SecurityController.php
+++ b/webapp/src/Controller/SecurityController.php
@@ -125,2 +125,11 @@ class SecurityController extends AbstractController
             $teamName = $registration_form->get('teamName')->getData();
+
+            if (count($registrationCategoryNames) === 1) {
+                $teamCategory = $em->getRepository(TeamCategory::class)->findOneBy(['name' => $registrationCategoryNames[0]]);
+                if (!$teamCategory) {
+                    throw new HttpException(500, sprintf("Team category '%s' not found", $registrationCategoryNames[0]));
+                }
+            } else {
+                // count($registrationCategoryNames) > 1, 'teamCategory' field exists
                 $teamCategory = $registration_form->get('teamCategory')->getData();
+            }
--- a/webapp/src/Form/Type/UserRegistrationType.php
+++ b/webapp/src/Form/Type/UserRegistrationType.php
@@ -88,4 +88,4 @@ class UserRegistrationType extends AbstractType
         $registrationCategoryNames = $this->dj->dbconfig_get('registration_category_names', []);
-        $showCategoryChoice        = count($registrationCategoryNames) !== 1;
+        if (count($registrationCategoryNames) > 1) {
             $registrationCategories = [];
             foreach ($registrationCategoryNames as $name) {
@@ -108,12 +108,11 @@ builder->add('teamCategory', EntityType::class, [
                     'choice_label' => 'name',
-                    'placeholder' => $showCategoryChoice ? '-- Select category --' : false,
+                    'placeholder' => '-- Select category --',
                     'choices' => $registrationCategories,
                     'attr' => [
                         'placeholder' => 'Category',
-                    ] + (
-                        $showCategoryChoice ? [] : ['hidden' => 'hidden']
-                    ),
+                    ],
                     'constraints' => [
                         new NotBlank(),
                     ],
                 ]);
+        }
```

webapp/src/Form/Type/UserRegistrationType.php
```diff
@@ -85,6 +87,39 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'mapped' => false,
             ]);
 
+        $registrationCategoryNames = $this->dj->dbconfig_get('registration_category_names', []);
+        $showCategoryChoice        = count($registrationCategoryNames) !== 1;
```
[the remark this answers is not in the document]
tom93: Changed to `> 1`. See reply to nickygerritsen on checking before loading from the DB.

webapp/src/Form/Type/UserRegistrationType.php
tom93: It doesn't preserve the order of the categories, which I think is important.
Would [this](https://github.com/tom93/domjudge/compare/self-registration-multiple-categories-v2-conditionally-build-field..tom93:self-registration-multiple-categories-v2b-one-db-query) be better? (And should I do the same in `CheckConfigService.php`?)

```diff
--- a/webapp/src/Form/Type/UserRegistrationType.php
+++ b/webapp/src/Form/Type/UserRegistrationType.php
@@ -89,12 +89,16 @@ class UserRegistrationType extends AbstractType
 
         $registrationCategoryNames = $this->dj->dbconfig_get('registration_category_names', []);
         if (count($registrationCategoryNames) > 1) {
+            $unorderedCategories = $this->em->getRepository(TeamCategory::class)->findBy(['name' => $registrationCategoryNames]);
+            $categoryMap         = [];
+            foreach ($unorderedCategories as $category) {
+                $categoryMap[$category->getName()] = $category;
+            }
             $registrationCategories = [];
             foreach ($registrationCategoryNames as $name) {
-                $category = $this->em->getRepository(TeamCategory::class)->findOneBy(['name' => $name]);
                 // silently ignore invalid category names
-                if ($category) {
-                    $registrationCategories[] = $category;
+                if (array_key_exists($name, $categoryMap)) {
+                    $registrationCategories[] = $categoryMap[$name];
                 }
             }
             // if all the category names are invalid, the user won't be able to fill in the form; don't even show it
```

webapp/src/Form/Type/UserRegistrationType.php
eldering: I think that if you want any errors to be debuggable, then it's better to write a notice level log message and/or add a configuration check to see that all categories exist. But I think a team user should not be presented with a weird selection dialog with only one (or even zero) options.

webapp/src/Form/Type/UserRegistrationType.php
eldering: I'd prefer this indeed, although I'd also sacrifice the order to simplify the code.
The code in `CheckConfigService.php` is less important as it gets called rarely; there I'd use whichever is simplest.

webapp/src/Form/Type/UserRegistrationType.php
tom93: Updated to make all the decisions based on the number of categories that exist in the database.
[Diff from previous version](https://github.com/DOMjudge/domjudge/compare/59e33099b6e08a745b743b319a9a98dc5100898e..8e46632f2e7f6a6c70dab688ec3578f6a0e19742)
I already added a configuration check to see that all the categories exist.
The previous version avoided showing a selection dialog with zero options by explicitly throwing an exception.

webapp/src/Form/Type/UserRegistrationType.php
tom93: I changed it to use

    $registrationCategories    = $this->em->getRepository(TeamCategory::class)->findBy(
        ['name' => $registrationCategoryNames],
        ['sortorder' => 'ASC']
    );

which matches the order of the categories in the scoreboard.

doc/manual/config-advanced.rst
```diff
@@ -78,10 +78,11 @@ Squid configuration for this might look like::
 Self-registration
 `````````````````
 There is also a configuration option to allow teams to self-register with
-the system: ``registration_category_name``. When left empty, no self-registration
-is allowed; when filled with a category name, newly registered teams will
-be placed in this category. During registration, a team can specify their
-affiliation.
+the system: ``registration_category_names``. When left empty, no
```
meisterT: I wonder whether this approach is the right one. Here you have a unchecked string field where you enter the category names. Perhaps it's better to add a boolean field to the team_category DB table whether self registration is allowed for this particular category or not.

nickygerritsen: I like this idea. That would also remove a lot of checks whether the list is actually valid

tom93: I've changed it to use a boolean field in the team_category DB table, and rebased (incorporating 464410198477; [steps](https://github.com/DOMjudge/domjudge/compare/40a6cc58723da6c093a369a4c3dc0a68082f53c9...a3eca5b8b1ef51c298a0d5d6000734da5e3b1fd6)).

Some notes:

 - When multiple categories are allowed for self-registration, it is difficult to toggle self-registration on and off (need to go into each category and toggle registration). I thought about adding a master configuration option, "allow_registration", that can be set to false to disable self-registration. (An option with that name existed before commit 9126537e1b. There can be an awkward state where allow_registration is true, but none of the categories are set to allow registration.)

 - I'd like to keep the configuration checker, even though no validation is required any more, so that there is a clear indication of whether self-registration is enabled or disabled. This way, even if the options that control self-registration are moved or changed in the future (e.g. are no longer tied to the team categories), there is a reliable place to check the status. The configuration checker could also display a hint on how to enable self-registration ("go to the Team Categories page"), but I haven't done that.

 - I used `@Serializer\Exclude()` on the new `$allow_self_registration` property in webapp/src/Entity/TeamCategory.php (copied from `$visible`), but I'm not sure if that's correct.

 - I wasn't sure whether to use underscores or camel case for the property name. I used underscores to match recent examples (`$orig_input_filename' in b8e74c1fc53c, `$team_categories` in 4a87d53f34b6), but then in the Twig template it needs to be in camel case for the getter to be inferred. (Also not sure about the doc style for the getter and setter; I tried to match the existing code.)

nickygerritsen: > I've changed it to use a boolean field in the team_category DB table, and rebased (incorporating [4644101](https://github.com/DOMjudge/domjudge/commit/46441019847719337c20a4fa74cd60d10830a474); [steps](https://github.com/DOMjudge/domjudge/compare/40a6cc58723da6c093a369a4c3dc0a68082f53c9...a3eca5b8b1ef51c298a0d5d6000734da5e3b1fd6)).
> 
> Some notes:
> 
> * When multiple categories are allowed for self-registration, it is difficult to toggle self-registration on and off (need to go into each category and toggle registration). I thought about adding a master configuration option, "allow_registration", that can be set to false to disable self-registration. (An option with that name existed before commit [9126537](https://github.com/DOMjudge/domjudge/commit/9126537e1bf325a8310364a293b3a22c94df4ae7). There can be an awkward state where allow_registration is true, but none of the categories are set to allow registration.)

Yeah I'd say if we have the below point it's fine without a master switch?

> * I'd like to keep the configuration checker, even though no validation is required any more, so that there is a clear indication of whether self-registration is enabled or disabled. This way, even if the options that control self-registration are moved or changed in the future (e.g. are no longer tied to the team categories), there is a reliable place to check the status. The configuration checker could also display a hint on how to enable self-registration ("go to the Team Categories page"), but I haven't done that.

Sounds good to keep this checker.

> * I used `@Serializer\Exclude()` on the new `$allow_self_registration` property in webapp/src/Entity/TeamCategory.php (copied from `$visible`), but I'm not sure if that's correct.

This annotation is used so it is not exposed in the API. We could expose it, but then we need to make sure it is only for nonstrict clients. I'm fine with leaving it excluded for now.

> * I wasn't sure whether to use underscores or camel case for the property name. I used underscores to match recent examples (`$orig_input_filename' in b8e74c1fc53c, `$team_categories` in [4a87d53](https://github.com/DOMjudge/domjudge/commit/4a87d53f34b6b6591e4da7943122f335fcb93275)), but then in the Twig template it needs to be in camel case for the getter to be inferred. (Also not sure about the doc style for the getter and setter; I tried to match the existing code.)

We should indeed standardise this at some point...

tom93:
Thanks for merging!

PR #684, opened 2020-01-05
Scoreboard test

eldering:
This now works for me (tm).

- I've added some basic tests, but not yet for FTS.
- I've not used fixtures because I felt they were not worth the effort.

### DIFF
--- a/webapp/tests/Utils/Scoreboard/ScoreboardIntegrationTest.php
+++ b/webapp/tests/Utils/Scoreboard/ScoreboardIntegrationTest.php
@@ -1,0 +1,281 @@
+<?php declare(strict_types=1);
+
+namespace Tests\Utils\Scoreboard;
+
+use App\Entity\Contest;
+use App\Entity\ContestProblem;
+use App\Entity\Judgehost;
+use App\Entity\Judging;
+use App\Entity\Language;
+use App\Entity\Problem;
+use App\Entity\Submission;
+use App\Entity\Team;
+use App\Entity\TeamCategory;
+use App\Service\DOMJudgeService;
+use App\Service\ScoreboardService;
+use App\Utils\Scoreboard\Scoreboard;
+use App\Utils\Scoreboard\TeamScore;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+class ScoreboardTest extends KernelTestCase
+{
+    public const CONTEST_NAME = 'scoretest';
+    public const NUM_PROBLEMS = 3;
+    public const NUM_TEAMS = 3;
+
+    /**
+     * @var \App\Service\DOMJudgeService
+     */
+    private $dj;
+
+    /**
+     * @var \App\Service\ScoreboardService
+     */
+    private $ss;
+
+    /**
+     * @var \Doctrine\ORM\EntityManager
+     */
+    private $em;
+
+    /**
+     * @var App\Entity\Contest
+     */
+    private $contest;
+
+    /**
+     * @var App\Entity\Judgehost
+     */
+    private $judgehost;
+
+    /**
+     * @var App\Entity\Problem[]
+     */
+    private $problems;
+
+    /**
+     * @var App\Entity\Team[]
+     */
+    private $teams;
+
+    protected function setUp()
+    {
+        self::bootKernel();
+
+        $this->dj = self::$container->get(DOMJudgeService::class);
+        $this->ss = self::$container->get(ScoreboardService::class);
+        $this->em = self::$container->get('doctrine')->getManager();
+
+        // Create a contest, problems and teams for which to test the
+        // scoreboard. These get deleted again in tearDown().
+        $this->contest = new Contest();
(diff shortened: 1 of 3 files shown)


webapp/tests/Utils/Scoreboard/ScoreboardIntegrationTest.php
```diff
@@ -0,0 +1,281 @@
...
+        $this->em = self::$container->get('doctrine')->getManager();
+
+        // Create a contest, problems and teams for which to test the
+        // scoreboard. These get deleted again in tearDown().
+        $this->contest = new Contest();
+        $this->contest
+            ->setExternalid(self::CONTEST_NAME)
+            ->setName(self::CONTEST_NAME)
+            ->setShortname(self::CONTEST_NAME)
+            ->setStarttimeString(date('Y').'-01-01 10:00:00 Europe/Amsterdam')
```
nickygerritsen: Does this mean the test will fail between 0:00 and 10:00 on the first of January?

webapp/tests/Utils/Scoreboard/ScoreboardIntegrationTest.php
```diff
@@ -0,0 +1,281 @@
...
+
+        foreach ([ false, true ] as $jury) {
+            $scoreboard = $this->ss->getScoreboard($this->contest, $jury);
+            $this->assertScoresMatch($expected_scores, $scoreboard->getScores());
+        }
+    }
+
+    public function testScoreboardJuryFreeze()
+    {
+        $this->createDefaultSubmissions();
```
nickygerritsen: Why is there no `$this-> recalcScoreCaches()` here but there is one in `testScoreboardsNoFreeze`?

webapp/tests/Utils/Scoreboard/ScoreboardIntegrationTest.php
```diff
@@ -0,0 +1,281 @@
...
+            $this->em->persist($cp);
+        }
+
+        $this->em->flush();
+    }
+
+    protected function tearDown()
+    {
+        // Preserve the data for inspection if a test failed.
+        if ( !$this->hasFailed() ) {
```
meisterT: how useful is this? This test suite has multiple methods and if one fails it will still continue running through the other methods, right?

webapp/tests/Utils/Scoreboard/ScoreboardIntegrationTest.php
```diff
@@ -0,0 +1,281 @@
...
+        $team = $this->teams[1];
+        $this->createSubmission($lang, $this->problems[0], $team, 69*60+00.000, 'correct');
+        $this->createSubmission($lang, $this->problems[0], $team, 72*60+07.824, 'wrong-answer');
+        $this->createSubmission($lang, $this->problems[1], $team, 72*60+39.733, 'wrong-answer');
+        $this->createSubmission($lang, $this->problems[1], $team, 72*60+59.999, 'correct');
+        $this->createSubmission($lang, $this->problems[1], $team, 79*60+00.000, 'wrong-answer');
+        $this->createSubmission($lang, $this->problems[2], $team, 84*60+42, null);
+        $this->createSubmission($lang, $this->problems[2], $team, 85*60+42, 'correct')
+            ->setValid(false);
+    }
```
meisterT: can you please add an explicit comment here that the 3rd team did not submit anything?

webapp/tests/Utils/Scoreboard/ScoreboardIntegrationTest.php
meisterT: That's a great idea to prevent us to submit/merge anything while we are drunk :beers:

eldering: Hmmm, maybe so. The submissions are automatically marked as judged, so I don't really know what would happen in that small time window. But I can change it to Y-1.

webapp/tests/Utils/Scoreboard/ScoreboardIntegrationTest.php
eldering: There is, but it's in the for loop below that changes the freeze time. I'll add a comment to clarify.

webapp/tests/Utils/Scoreboard/ScoreboardIntegrationTest.php
eldering: No, since the data is then not deleted, the next test `setUp()` will fail because of the contest name that already exists and bail out.

webapp/tests/Utils/Scoreboard/ScoreboardIntegrationTest.php
eldering: Done.

eldering:
I've now also added checks for FTS.

webapp/tests/Utils/Scoreboard/ScoreboardIntegrationTest.php
thijskh: This will most likely bite me in the next "happy new year" commit

PR #685, opened 2020-01-11
Fix missing -dev libraries in Development Setup

abhineet255:
After following the documentation, I was required to install these additional libraries in order to get my development environment setup. Including these in the documentation.
```
libcgroup-dev
libcurl4-openssl-dev
libjsoncpp-dev
```

### DIFF
--- a/doc/manual/develop.rst
+++ b/doc/manual/develop.rst
@@ -36,7 +36,8 @@ install the additionally required packages::
   sudo apt install autoconf automake \
     python3-sphinx python3-sphinx-rtd-theme \
     texlive-latex-recommended texlive-latex-extra \
-    texlive-fonts-recommended texlive-lang-european
+    texlive-fonts-recommended texlive-lang-european \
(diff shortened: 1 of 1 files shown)


thijskh:
Thanks for the PR! However, I don’t think this is right. The section you mention lists the *additional* dependencies needed for the development environmeny. What you add are the basic requirements to build DOMjudge and they are already listed in the chapters on building the components.

It could be that we can improve the instructions for a development build, but this just duplicates a subset of the build dependencies.

meisterT:
Perhaps we should mention in
https://www.domjudge.org/docs/manual/develop.html#maintainer-mode-installation
that you have to install both the server and judgehost deps?

abhineet255:
> Perhaps we should mention in
> https://www.domjudge.org/docs/manual/develop.html#maintainer-mode-installation
> that you have to install both the server and judgehost deps?

I think that is accurate. I followed the initial setup instructions then jumped to maintainer mode, which is why I missed the deps listed under judgehost.

PR #686, opened 2020-01-11
Move specification of configuration variables to YAML file and outside of database

nickygerritsen:
So after checking the Symfony docs a bit, I found this thing that allows us to basically 'cache' a YAML file in the Symfony cache directory, which IMHO gives us the best of all worlds:

* The spec of the config is completely outside the database and in a readable YAML file.
* We can 'load the YAML' for every request without performance impact, since we actually load the cached PHP file. In dev-mode Symfony automatically re-caches the file when it changed and in prod mode it only does this if you clear the Symfony cache.

I put the configuration logic in a separate service, since IMHO that makes sense.

The migrations are a bit 'messy': I have build in support for rolling it back and I remove all configuration options from the database that have the default value. I do like the idea that we don't have config options in the DB with the default value and I don't really see how we could do that better without breaking upgrades. The rolling back stuff is then basically 'free'.

Also a lot of the changes is just `s/->dj->dbconfig_get/->config->get` so the PR seems bigger than it actually is I guess.

Input is very much welcome.

### DIFF
--- a/webapp/src/Command/CheckDatabaseConfigurationDefaultValuesCommand.php
+++ b/webapp/src/Command/CheckDatabaseConfigurationDefaultValuesCommand.php
@@ -1,0 +1,103 @@
+<?php declare(strict_types=1);
+
+namespace App\Command;
+
+use App\Service\ConfigurationService;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * Class CheckDatabaseConfigurationDefaultValuesCommand
+ *
+ * @package App\Command
+ */
+class CheckDatabaseConfigurationDefaultValuesCommand extends Command
+{
+    /**
+     * @var ConfigurationService
+     */
+    protected $config;
+
+    /**
+     * CheckDatabaseConfigurationDefaultValuesCommand constructor.
+     *
+     * @param ConfigurationService $config
+     * @param string|null          $name
+     */
+    public function __construct(
+        ConfigurationService $config,
+        string $name = null
+    ) {
+        parent::__construct($name);
+        $this->config = $config;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('domjudge:db-config:check')
+            ->setDescription(
+                'Check if the default values of the database configuration are valid'
+            );
+    }
+
+    /**
+     * @inheritDoc
+     * @throws \Exception
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $style    = new SymfonyStyle($input, $output);
+        $messages = [];
+        foreach ($this->config->getConfigSpecification() as $specification) {
+            $message = sprintf(
+                'Configuration %s (in category %s) is of type %s but has wrong type for default_value (%s)',
+                $specification['name'], $specification['category'],
+                $specification['type'],
+                json_encode($specification['default_value'])
+            );
+            switch ($specification['type']) {
+                case 'bool':
+                    if (!is_bool($specification['default_value'])) {
+                        $messages[] = $message;
+                    }
+                    break;
+                case 'int':
+                    if (!is_int($specification['default_value'])) {
+                        $messages[] = $message;
+                    }
+                    break;
+                case 'string':
+                    if (!is_string($specification['default_value'])) {
+                        $messages[] = $message;
+                    }
+                    break;
+                case 'array_val':
+                    if (!(empty($specification['default_value']) || (
+                            is_array($specification['default_value']) &&
+                            is_int(key($specification['default_value']))))) {
+                        $messages[] = $message;
+                    }
+                    break;
+                case 'array_keyval':
+                    if (!(empty($specification['default_value']) || (
+                            is_array($specification['default_value']) &&
+                            is_string(key($specification['default_value']))))) {
+                        $messages[] = $message;
+                    }
+                    break;
+            }
+        }
+        if (empty($messages)) {
+            $style->success('All default values have the correct type');
+        } else {
+            $style->error('Some default values have the wrong type:');
+            $style->listing($messages);
+        }
+    }
+}
--- a/webapp/src/Migrations/Version20200111104226.php
+++ b/webapp/src/Migrations/Version20200111104226.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use App\Service\ConfigurationService;
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20200111104226 extends AbstractMigration implements ContainerAwareInterface
+{
+    use ContainerAwareTrait;
+
+    public function getDescription(): string
+    {
+        return 'remove unneeded columns from the configuration table';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql',
+            'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('DROP INDEX public ON configuration');
+        $this->addSql('ALTER TABLE configuration DROP type, DROP public, DROP category, DROP description');
+    }
+
+    public function down(Schema $schema): void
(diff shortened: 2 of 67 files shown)


webapp/src/Command/CheckDatabaseConfigurationDefaultValuesCommand.php
```diff
@@ -0,0 +1,100 @@
...
+                    if (!empty($specification['default_value']) && (
+                            !is_array($specification['default_value']) ||
+                            !is_int(key($specification['default_value'])))) {
+                        $messages[] = $message;
+                    }
+                    break;
+                case 'array_keyval':
+                    if (!empty($specification['default_value']) && (
+                            !is_array($specification['default_value']) ||
+                            !is_string(key($specification['default_value'])))) {
```
eldering: The checks for `array_val` and `array_keyval` look wrong to me. How can `$specification['default_value']` be both empty and an array? Shouldn't that be *or*? And the next `||` be `&&`?

webapp/src/Service/DOMJudgeService.php
```diff
@@ -628,6 +637,15 @@ public function getDomjudgeWebappDir(): string
         return $this->params->get('domjudge.webappdir');
     }
 
+    /**
+     * Get the directory used for storing cache files
+     * @return string
+     */
+    public function getCacheDir(): string
```
eldering: For my understanding: where is this used? Also, how does the cacheDir get passed to the config service constructor?

webapp/src/Migrations/Version20200111104226.php
```diff
@@ -0,0 +1,60 @@
...
+        $this->addSql('ALTER TABLE configuration DROP type, DROP public, DROP category, DROP description');
+    }
+
+    public function down(Schema $schema): void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql',
+            'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE configuration ADD type VARCHAR(32) DEFAULT \'NULL\' COLLATE utf8mb4_unicode_ci COMMENT \'Type of the value (metatype for use in the webinterface)\', ADD public TINYINT(1) DEFAULT \'0\' NOT NULL COMMENT \'Is this variable publicly visible?\', ADD category VARCHAR(32) DEFAULT \'Uncategorized\' NOT NULL COLLATE utf8mb4_unicode_ci COMMENT \'Option category of the configuration variable\', ADD description VARCHAR(255) DEFAULT \'NULL\' COLLATE utf8mb4_unicode_ci COMMENT \'Description for in the webinterface\'');
```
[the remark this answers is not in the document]
nickygerritsen: Sure. It was autogenerated so didn't even look at it 🙈

webapp/src/Migrations/Version20200111104415.php
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public function down(Schema $schema): void
+    {
+        $configService = $this->container->get(ConfigurationService::class);
+        $allConfig     = $configService->all();
+
+        foreach ($allConfig as $name => $value) {
+            $this->addSql(
+                'INSERT INTO configuration (name, value) VALUES (:name, :value) ON DUPLICATE KEY UPDATE configid=configid',
```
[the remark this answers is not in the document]
nickygerritsen: This basically means 'update nothing if the value already exists'. Google/SO told me this was the way to do that, since I update configid to itself, i.e. a `noop`. If you have a better way let me know

webapp/src/Service/ConfigurationService.php
```diff
@@ -0,0 +1,211 @@
...
+     * @param string $name         The config name to get the value of
+     * @param mixed  $default      The default value to use if the config
+     *                             option and spec are not found
+     * @param bool   $onlyIfPublic Only return the value if the config is
+     *                             public
+     *
+     * @return mixed The configuration value
+     * @throws Exception If the config can't be found and not default is
+     *                   supplied
+     */
```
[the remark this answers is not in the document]
nickygerritsen: This is what the old behaviour was. I guess it makes sense to not have this anymore, since adding config stuff is easier now. I will remove this functionality.

webapp/src/Command/CheckDatabaseConfigurationDefaultValuesCommand.php
nickygerritsen: Maybe my logic is rusty, but what I want to check: if the value is non-empty but it does contain invalid key types, we print a message. That is what I'm doing right?

webapp/src/Migrations/Version20200111104415.php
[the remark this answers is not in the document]
eldering: Ah, you're right and my comment doesn't make sense: `configid` cannot be taken by a *different* row.

webapp/src/Service/DOMJudgeService.php
nickygerritsen: Ah sorry, it's not used anymore. In the beginning I had ConfigurationService use DOMJudgeService, but that was not possible since DOMJudgeService needs ConfigurationService to read some config, so I injected `$cacheDir` directly (see the `services.yaml` diff)

webapp/src/Command/CheckDatabaseConfigurationDefaultValuesCommand.php
eldering: Ah, sorry, my logic is wrong!
Although I think that the logic as written is a bit confusing: it's checking whether the entry is *not* valid. If the check would be written as:
```php
                    if (!(empty($specification['default_value']) || (
                            is_array($specification['default_value']) &&
                            is_int(key($specification['default_value']))))) { ... }
```
then I'd say it's slightly clearer, since now the whole conditional inside the outer `!( ... )` checks whether the entry *is* valid.

nickygerritsen: Changed it to your suggestion :)

webapp/src/Service/ConfigurationService.php
```diff
@@ -0,0 +1,210 @@
...
+
+        if (isset($spec)) {
+            if ($onlyIfPublic && !$spec['public']) {
+                // If we require public values and the spec is not public,
+                // set the value from the DB to null to not have a value
+                $dbValue = null;
+            }
+        }
+
+        $dbValue = $this->getDbValues()[$name] ?? null;
```
eldering: This should go *before* the code above.

nickygerritsen: Kik

webapp/tests/Service/ScoreboardServiceTest.php
```diff
@@ -0,0 +1,349 @@
...
+            ->method('findAll');
+
+        $this->config->get($itemName, $publicOnly);
+    }
+
+    public function provideInvalidItem()
+    {
+        yield ['does_not_exist', false]; // This item does not exist
+        yield ['does_not_exist', true];
+        yield ['results_priod', true]; // This item exists but is non-public
```
eldering: I assume `results_priod` is a typo.

webapp/src/Service/ConfigurationService.php
```diff
@@ -93,7 +93,7 @@ public function __construct(
      */
     public function get(string $name, bool $onlyIfPublic = false)
     {
-        $spec = $this->getConfigSpecification()[$name];
+        $spec = $this->getConfigSpecification()[$name] ?? null;
```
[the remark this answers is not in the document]
nickygerritsen: Otherwise you get an `undefined index` when passing a `$name` that does not have a spec (one of my tests failed on that until I fixed it).

webapp/tests/Service/ScoreboardServiceTest.php
```diff
@@ -0,0 +1,349 @@
...
+
+class ScoreboardServiceTest extends KernelTestCase
+{
+    /**
+     * @var EntityManagerInterface|MockObject
+     */
+    private $em;
+
+    /**
+     * @var MockObject|ObjectRepository
```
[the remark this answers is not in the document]
nickygerritsen: :( I think PHPStorm puts the default types in alphabetical order and I forgot to fix this one. Fixed now.

webapp/tests/Service/ScoreboardServiceTest.php
nickygerritsen: Whoops.

webapp/src/Service/ConfigurationService.php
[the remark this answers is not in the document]
tom93: I know this is a bit late, but...
It looks like the doc still refers to the old behaviour: "<span>@</span>throws Exception If the config can't be found and not default supplied".
A consequence of the new behaviour is that if an entry is ever removed from the YAML file, migrations can't use ConfigurationService::get to read the value from the DB (but they can use the Configuration repository directly). I ran into this with #678.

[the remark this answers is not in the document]
nickygerritsen: Aha you are right. I think it is fine that you should not be able to read nonexisting configuration values, but we should indeed update the comment and such.

PR #687, opened 2020-01-12
WIP: scoreboard FTS tests and bugfix

eldering:
This adds some test for first to solve and tries to fix a bug. However, I think the bugfix creates a new bug:
Currently, if there is a submission with a valid judging that is part of a rejudging, then that submission is not counted towards previous submissions for FTS, leading to multiple teams having FTS for the same problem.
My last commit fixes that, but I think it opens up another bug, when an earlier incorrect submission has `s.judgehost = NULL`, e.g. when it is part of an ongoing rejudging. The issue is, I think, that `s.judgehost = NULL` does not imply that that submission is unjudged, just that it needs to be picked up for (re)judging, which is a subtly different thing

### DIFF
--- a/webapp/src/Service/ScoreboardService.php
+++ b/webapp/src/Service/ScoreboardService.php
@@ -462,22 +462,18 @@ class ScoreboardService
                     tc.sortorder = :teamSortOrder AND
                     round(s.submittime,4) < :submitTime', $params);
             } else {
-                if ($verificationRequired) {
-                    $verificationRequiredExtra = 'OR j.verified = 0';
-                } else {
-                    $verificationRequiredExtra = '';
-                }
-                $firstToSolve = 0 == $this->em->getConnection()->fetchColumn(sprintf('
+                $verificationRequiredExtra = $verificationRequired ? 'OR j.verified = 0' : '';
+                $firstToSolve = 0 == $this->em->getConnection()->fetchColumn('
                 SELECT count(*) FROM submission s
                     LEFT JOIN judging j USING (submitid)
                     LEFT JOIN team t USING(teamid)
                     LEFT JOIN team_category tc USING (categoryid)
                 WHERE s.valid = 1 AND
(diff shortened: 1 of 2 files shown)


nickygerritsen:
I think you are correct that `s.judgehost IS NULL` is not correct. Shouldn’t be just only check the current valid judging?
This FTS stuff is hard...

webapp/src/Service/ScoreboardService.php
```diff
@@ -462,22 +462,18 @@ public function calculateScoreRow(
...
+                $firstToSolve = 0 == $this->em->getConnection()->fetchColumn('
                 SELECT count(*) FROM submission s
                     LEFT JOIN judging j USING (submitid)
                     LEFT JOIN team t USING(teamid)
                     LEFT JOIN team_category tc USING (categoryid)
                 WHERE s.valid = 1 AND
-                    ((j.valid = 1 AND ( j.rejudgingid IS NULL AND (j.result IS NULL OR j.result = :correctResult %s))) OR
-                      s.judgehost IS NULL) AND
+                    ((j.valid = 1 AND (j.result IS NULL OR j.result = :correctResult '.
+                    $verificationRequiredExtra.')) OR s.judgehost IS NULL) AND
```
eldering: So basically, what we're trying to determine: for each earlier submission, is it potentially correct? Now a submission can at any time have zero or one valid judgings (this is an invariant). That means that the `LEFT JOIN judging j USING (submitid)` always results in one row: either an empty judging or the single valid one. I think this is better expressed by writing the join as:
```sql
LEFT JOIN judging j ON (j.submitid=s.submitid AND j.valid=1)
```
In the line above here, we should therefore check whether the judging is null, or if not, whether it is potentially correct. I think the second part is correct now, so I think that line can be written as:
```sql
(j.judgingid IS NOT NULL AND (j.result IS NULL OR j.result = :correctResult '.$verificationRequiredExtra.')) AND
```

PR #689, opened 2020-01-18
Rename team.externalid to team.icpcid. Fixes #655

### LINKED ISSUE #655: Rename team external ID
Unlike all other external ID's, the team external ID is not used as "API ID". It is shown in the icpc_id API field, so we should rename it, maybe to "ICPC ID" because it is giving a lot of confusion.

nickygerritsen:
For other entries we use the external ID as the 'primary' ID as given by either the primary CCS or another system.
However, for teams we always use team.id for this and team.externalid was a (second) external ID, always representing the ID from the ICPC CMS.
This gave a lot of confusion (as can also be seen in the commit, since I have fixed a few places we used it incorrectly).
So I believe it is better to rename it to something else to not have any confusion.

### DIFF
--- a/webapp/src/Command/ImportEventFeedCommand.php
+++ b/webapp/src/Command/ImportEventFeedCommand.php
@@ -986,7 +986,7 @@ class ImportEventFeedCommand extends Command
                 $this->em->flush();
                 $this->eventLogService->log('teams', $team->getTeamid(),
                                             EventLogService::ACTION_DELETE,
-                                            $this->contestId, null, $team->getExternalid());
+                                            $this->contestId, null, $team->getTeamid());
                 return;
             } else {
                 $this->logger->error('Cannot delete nonexistent team %s', [ $teamId ]);
@@ -1003,7 +1003,7 @@ class ImportEventFeedCommand extends Command
             $team = new Team();
             $team
                 ->setTeamid($teamId)
-                ->setExternalid($icpcId);
+                ->setIcpcid($icpcId);
             $action = EventLogService::ACTION_CREATE;
         }
 
@@ -1063,7 +1063,7 @@ class ImportEventFeedCommand extends Command
         $this->em->flush();
         $this->eventLogService->log('teams', $team->getTeamid(), $action, $this->contestId);
 
-        $this->processPendingEvents('team', $team->getExternalid());
+        $this->processPendingEvents('team', $team->getTeamid());
     }
 
     /**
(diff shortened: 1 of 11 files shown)


webapp/src/Command/ImportEventFeedCommand.php
```diff
@@ -953,7 +953,7 @@ protected function importTeam(array $event)
                 $this->em->flush();
                 $this->eventLogService->log('teams', $team->getTeamid(),
                                             EventLogService::ACTION_DELETE,
-                                            $this->contestId, null, $team->getExternalid());
+                                            $this->contestId, null, $team->getTeamid());
```
meisterT: not `getIcpcId`?

webapp/src/Migrations/Version20200118092658.php
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public function up(Schema $schema) : void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('DROP INDEX externalid ON team');
+        $this->addSql('ALTER TABLE team CHANGE externalid icpcid VARCHAR(255) DEFAULT NULL COLLATE utf8mb4_bin COMMENT \'Team ID in the ICPC system\'');
+        $this->addSql('CREATE UNIQUE INDEX icpcid ON team (icpcid(190))');
```
meisterT: out of curiosity: what does the 190 mean here?

webapp/src/Service/ImportExportService.php
```diff
@@ -625,14 +625,14 @@ protected function importTeamsTsv(array $content, string &$message = null): int
 
             // Set team ID to external ID if it has the literal value 'null' and the external ID is numeric
```
meisterT: nit: fix comment

webapp/src/Command/ImportEventFeedCommand.php
nickygerritsen: Nope, this was a bug: the team ID in the event feed is *always* our `teamid`, never the `icpc_id`.

webapp/src/Migrations/Version20200118092658.php
nickygerritsen: Limit the index to the first 190 characters, which is needed for not-so-new MySQL's, since indices on varchars can not exceed 767 bytes, and utf8mb4 uses 4 bytes, so 190*4 < 767

PR #690, opened 2020-01-18
Add support for showing select boxes in the config screen.

nickygerritsen:
Since we now have a YAML for config options, it was easier to populate 'magic' config option choices from that YAML or the database.

A pictures is better than a 1000 words, so imagine if we have 4:

image
image
image
image

### DIFF
--- a/etc/db-config.yaml
+++ b/etc/db-config.yaml
@@ -138,25 +138,24 @@
                 general: General issue
                 tech: Technical issue
             public: true
-            description: List of additional clarification categories
+            description: List of additional clarification categories.
         -   name: clar_answers
             type: array_val
             default_value:
                 - No comment
                 - Read the problem statement carefully
             public: false
-            description: List of predefined clarification answers
+            description: List of predefined clarification answers.
         -   name: clar_queues
             type: array_keyval
             default_value: []
             public: true
-            description: List of clarification queues
+            description: List of clarification queues.
         -   name: clar_default_problem_queue
             type: string
             default_value: ""
             public: true
-            description: Queue to assign to problem clarifications
-
+            description: Queue to assign to problem clarifications.
 -   category: Display
     items:
         -   name: show_pending
@@ -188,7 +187,11 @@
             type: int
             default_value: 2
             public: true
-            description: "Show compile output in team webinterface? Choices: 0 = never, 1 = only on compilation error(s), 2 = always."
+            description: Show compile output in team webinterface?
+            options:
+                0: never
+                1: only on compilation error(s)
+                2: always
         -   name: show_sample_output
             type: bool
             default_value: false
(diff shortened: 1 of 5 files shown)


etc/db-config.yaml
```diff
@@ -235,12 +239,16 @@
             type: string
             default_value: ""
             public: true
-            description: Team category for users that register themselves with the system. Self-registration is disabled if this field is left empty.
+            description: Team category for users that register themselves with the system
```
eldering: Nitpick: I think these should be full sentences, so should end with a `.` (or `?`).
Also, I think it's best to leave the second sentence to clarify that it can be disabled that way.

etc/db-config.yaml
```diff
@@ -235,12 +239,16 @@
...
         -   name: data_source
             type: int
             default_value: 0
             public: false
-            description: "Source of data. Choices: 0 = all local, 1 = configuration data external, 2 = configuration and live data external"
+            description: Source of data
```
eldering: Although not directly related to this change: I think this needs more clarification. Maybe something like:
`Source of data: used to indicate whether internal or external IDs are exposed in the API. Option 1 is typically used when loading configuration data from the ICPC CMS, and option 2 when running DOMjudge as "shadow system".`

webapp/src/Service/ConfigurationService.php
```diff
@@ -195,4 +199,50 @@ protected function getDbValues(): array
 
         return $this->dbConfigCache;
     }
+
+    /**
+     * Add options to some items
+     *
+     * This method is used to add options that need to be loaded from the
```
eldering: I'd say `to add *predefined* options` to clarify a bit more.

etc/db-config.yaml
nickygerritsen: I will add the dot. The dropdown does show 'Self registration disabled' so I wonder if we should add it to the description as well...?

eldering: Ah, ok. Then the description is fine as is.

webapp/src/Controller/Jury/ConfigController.php
```diff
@@ -103,6 +103,10 @@ private function logUnverifiedJudgings(EventLogService $eventLogService)
     public function indexAction(EventLogService $eventLogService, Request $request)
     {
         $specs = $this->config->getConfigSpecification();
+        foreach ($specs as &$spec) {
```
meisterT: I have to admit, I have no ideas what you do in those four lines.

nickygerritsen: Loading the $spec by reference, then adding options, which basically merges in the options on the fly.
The `unset` is needed because PHP will otherwise use the first time $spec is used after this as a pass-by-reference instead of pass-by-value and shit will hit all kinds of fans.

PR #691, opened 2020-01-18
Add team display name support

nickygerritsen:
I hope I got all places we use the name....
I did it in two commits, adding JSON import support in a separate commit.

### DIFF
--- a/webapp/src/Controller/API/UserController.php
+++ b/webapp/src/Controller/API/UserController.php
@@ -64,24 +64,75 @@ class UserController extends AbstractRestController
      *     name="tsv",
      *     in="formData",
      *     type="file",
-     *     required=true,
+     *     required=false,
      *     description="The groups.tsv files to import."
+     * )
+     * @SWG\Parameter(
+     *     name="json",
+     *     in="formData",
+     *     type="file",
+     *     required=false,
+     *     description="The groups.json files to import."
      * )
      * @SWG\Response(
      *     response="200",
      *     description="Returns a (currently meaningless) status message.",
      * )
-     * @throws BadRequestHttpException
+     * @throws Exception
      */
-    public function addGroupAction(Request $request)
+    public function addGroupsAction(Request $request)
     {
         /** @var UploadedFile $tsvFile */
         $tsvFile = $request->files->get('tsv') ?: [];
-        if ($this->importExportService->importTsv('groups', $tsvFile, $message)) {
+        /** @var UploadedFile $jsonFile */
+        $jsonFile = $request->files->get('json') ?: [];
+        if ((!$tsvFile && !$jsonFile) || ($tsvFile && $jsonFile)) {
+            throw new BadRequestHttpException('Supply exactly one of \'json\' or \'tsv\'');
+        }
+        $message = null;
+        if ($tsvFile && ($result = $this->importExportService->importTsv('groups', $tsvFile, $message))) {
             // TODO: better return all groups here
-            return "New groups successfully added.";
+            return "$result new group(s) successfully added.";
+        } elseif ($jsonFile && ($result = $this->importExportService->importJson('groups', $jsonFile, $message))) {
+            // TODO: better return all groups here
+            return "$result new group(s) successfully added.";
         } else {
             throw new BadRequestHttpException("Error while adding groups: $message");
+        }
+    }
+
+    /**
+     * Add one or more organizations.
+     *
+     * @param Request $request
+     *
+     * @return string
+     * @Rest\Post("/organizations")
+     * @IsGranted("ROLE_ADMIN")
+     * @SWG\Post(consumes={"multipart/form-data"})
+     * @SWG\Parameter(
+     *     name="json",
+     *     in="formData",
+     *     type="file",
+     *     required=true,
+     *     description="The organizations.json files to import."
+     * )
+     * @SWG\Response(
+     *     response="200",
+     *     description="Returns a (currently meaningless) status message.",
+     * )
+     * @throws Exception
+     */
+    public function addOrganizationsAction(Request $request)
+    {
+        /** @var UploadedFile $jsonFile */
+        $jsonFile = $request->files->get('json') ?: [];
+        $message = null;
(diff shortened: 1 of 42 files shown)


webapp/src/Service/ImportExportService.php
```diff
@@ -537,12 +539,46 @@ public function importTsv(string $type, UploadedFile $file, string &$message = n
...
+     * @param string|null  $message
+     * @return int
+     * @throws Exception
+     */
+    public function importJson(string $type, UploadedFile $file, string &$message = null): int
+    {
+        $content = file_get_contents($file->getRealPath());
+        $data = json_decode($content, true);
+        if (json_last_error() !== JSON_ERROR_NONE) {
+            $message = 'File contents is not JSON';
```
eldering: Maybe be a bit more verbose: first, say `is not *valid* JSON` and then append the output of `json_last_error_msg()` so that the user has some chance to understand what's wrong.

webapp/src/Service/ImportExportService.php
```diff
@@ -537,12 +539,46 @@ public function importTsv(string $type, UploadedFile $file, string &$message = n
...
+
+        switch ($type) {
+            case 'groups':
+                return $this->importGroupsJson($data, $message);
+            case 'organizations':
+                return $this->importOrganizationsJson($data, $message);
+            case 'teams':
+                return $this->importTeamsJson($data, $message);
+            default:
+                $message = sprintf('Invalid JSON type %s', $type);
```
eldering: I think it's clearer to say `Invalid import type %s`.

webapp/src/Service/ImportExportService.php
```diff
@@ -577,25 +653,93 @@ protected function importGroupsTsv(array $content, string &$message = null): int
...
+                ->setName($groupItem['name'])
+                ->setVisible($groupItem['visible'] ?? true)
+                ->setSortorder($groupItem['sortorder'] ?? 0);
             $this->em->flush();
             if ($contest = $this->dj->getCurrentContest()) {
-                $this->eventLogService->log('team_category', $teamCategory->getCategoryid(), $action,
-                                            $contest->getCid());
+                $this->eventLogService->log('team_category',
+                    $teamCategory->getCategoryid(), $action,
+                    $contest->getCid());
```
eldering: Nitpick, but *if* you're reindenting, then I think *all* arguments should go onto indented lines and the final `);` on a separate line (following PSR1, although I really don't like putting each single argument on a line of its own).

webapp/src/Service/ImportExportService.php
```diff
@@ -668,21 +853,36 @@ protected function importTeamsTsv(array $content, string &$message = null): int
                     ->from(TeamAffiliation::class, 'a')
                     ->select('a')
                     ->andWhere('a.externalid = :externalid')
-                    ->setParameter(':externalid', $teamItem['team_affiliation']['externalid'])
+                    ->setParameter(':externalid',
+                        $teamItem['team_affiliation']['externalid'])
```
eldering: I think this reformat and below are better left as is: both to not clutter the diff, but also because they're not changing it to PSR either.

nickygerritsen:
Most layout changes happened because I have set a hard limit on line lengths since you mostly change my long lines 🙈. However I will change them (back) as suggested

webapp/src/Controller/API/UserController.php
```diff
@@ -64,27 +64,80 @@ public function __construct(
...
-    public function addGroupAction(Request $request)
+    public function addGroupsAction(Request $request)
     {
         /** @var UploadedFile $tsvFile */
         $tsvFile = $request->files->get('tsv') ?: [];
-        if ($this->importExportService->importTsv('groups', $tsvFile, $message)) {
+        /** @var UploadedFile $jsonFile */
+        $jsonFile = $request->files->get('json') ?: [];
+        if (!$tsvFile && !$jsonFile) {
+            throw new BadRequestHttpException('Neither TSV nor JSON file supplied');
```
meisterT: nit: merge the two exceptions into one which you phrase like this:
`Supply exactly one of 'json' or 'tsv'.`

webapp/src/Controller/API/UserController.php
```diff
@@ -64,27 +64,80 @@ public function __construct(
...
+        } elseif ($tsvFile && $jsonFile) {
+            throw new BadRequestHttpException('Both TSV nor JSON file supplied');
+        }
+        $message = null;
+        if ($tsvFile && $this->importExportService->importTsv('groups', $tsvFile, $message)) {
+            // TODO: better return all groups here
```
meisterT: `importTsv` returns an int, so add this to the message below

webapp/src/Controller/API/UserController.php
```diff
@@ -96,20 +149,38 @@ public function addGroupAction(Request $request)
...
+     * @throws Exception
      */
     public function addTeamsAction(Request $request)
     {
         /** @var UploadedFile $tsvFile */
         $tsvFile = $request->files->get('tsv') ?: [];
-        if ($this->importExportService->importTsv('teams', $tsvFile, $message)) {
+        /** @var UploadedFile $jsonFile */
+        $jsonFile = $request->files->get('json') ?: [];
+        if (!$tsvFile && !$jsonFile) {
```
meisterT: same as above

meisterT: also below

webapp/src/Entity/Team.php
```diff
@@ -58,6 +57,13 @@ class Team extends BaseApiEntity implements ExternalRelationshipEntityInterface
      */
     private $name;
 
+    /**
+     * @var string|null
+     * @ORM\Column(type="string", name="display_name", length=255, options={"comment"="Team display name", "collation"="utf8mb4_bin"},
+     *                            nullable=true)
```
meisterT: does the indentation only look weird in the github UI?

nickygerritsen: What is 'weird'? The nullable is now with name=, but probably should be with type=?

meisterT: Yes, exactly.

PR #693, opened 2020-01-27
group by all selected variables for newer database versions

DerBaer0:
Support for recent MySQL versions. Just like earlier commits:
[674789de9230f1a02bf43fb2a063db87fe6e70c1](https://github.com/DOMjudge/domjudge/pull/606/commits/674789de9230f1a02bf43fb2a063db87fe6e70c1)
[b9b298e50e35aafb7030796ae331b76a37625b79](https://github.com/DOMjudge/domjudge/pull/618/commits/b9b298e50e35aafb7030796ae331b76a37625b79)


nickygerritsen:
Note that specific MySQL 'strict modes' still allow this, but we should indeed make it work without those modes IMHO.

PR #695, opened 2020-02-05
Allow output to be "OUTPUT 0", but not "OUTPUT "

ludopulles:
If you want to add more sample test cases to the interactive example problem, e.g.
```
1
3
1 0 0
```
Then, "OUTPUT 0" is the correct answer but this was not accepted yet.

### DIFF
--- a/sql/files/examples/boolfind_cmp/check_boolfind.c
+++ b/sql/files/examples/boolfind_cmp/check_boolfind.c
@@ -47,8 +47,8 @@ int main(int argc, char **argv)
 		while ( i>=0 && line[i]=='\n' ) line[i--] = 0;
 
 		outputvalid = 1;
-		if ( strncmp(line,"OUTPUT ",7)!=0 ||
-		     strlen(line)<7 || line[7]=='0' ) outputvalid = 0;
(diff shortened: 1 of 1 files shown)


sql/files/examples/boolfind_cmp/check_boolfind.c
```diff
@@ -48,7 +48,7 @@ int main(int argc, char **argv)
 
 		outputvalid = 1;
 		if ( strncmp(line,"OUTPUT ",7)!=0 ||
-		     strlen(line)<7 || line[7]=='0' ) outputvalid = 0;
+		     strlen(line)<7 || line[7]==0 ) outputvalid = 0;
```
eldering: The condition `line[7]=='0'` was meant to check that the number does not have prefix zeros. It does this incorrectly, but the proposed change also isn't effective: if `strlen(line)<7` is not true then `line[7]==0` is also guaranteed to not be true, so the condition is superfluous. I think this line should be changed to:
```c
    (strlen(line)>=8 && line[7]=='0') ) outputvalid = 0;
```

ludopulles:
That makes more sense. I think to allow `OUTPUT 0`, you need `strlen(line) >= 9` but then you also want to exclude the 'empty string', so I added a case checking if `strlen(line) == 7`.

PR #699, opened 2020-02-14
Document how to use REMOTE_USER for authentication

thijskh:
This is a very generic solution for authenticating existing users.
Many (Apache) modules can handle a full authentication flow and
deliver an authenticated user, thereby instantly supporting a range
of methods for DOMjudge.

It does not (yet) support auto creation of users. That might need
more custom work per method, e.g. where to find the full name
or email address? But it's definitely better than what we have now.
Surely we can improve on it further.

Since this does provide a way to authenticate with OpenID Connect,
I think it's time to finally clean up the openid_ configuation
values which have not been working for many releases now.

### DIFF
--- a/doc/manual/config-advanced.rst
+++ b/doc/manual/config-advanced.rst
@@ -35,11 +35,15 @@ Authentication and registration
 -------------------------------
 Out of the box users are able to authenticate using username and password.
 
-Two other authentication methods are available:
+Two other native authentication methods are available:
 
 - IP Address - authenticates users based on the IP address they are accessing
   the system from;
 - X-Headers - authenticates users based on some HTTP header values.
+
+Besides this, DOMjudge can be configured with any provider that can set
+the environment variable ``REMOTE_USER`` to an existing username,
+for example LDAP, SAML, CAS or OpenID connect modules for Apache.
 
 There's an option to let teams register themselves in the system.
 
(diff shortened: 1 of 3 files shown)


doc/manual/config-advanced.rst
```diff
@@ -75,6 +79,32 @@ Squid configuration for this might look like::
...
+Mod Mellon for SAML 2.0, mod Auth CAS, mod OpenID Connect, or mod Kerb for
+Kerberos.
+
+This does not currently allow for auto-provisioning or self-registration,
+the users must already exist in DOMjudge and their DOMjudge username must
+match what is in the ``REMOTE_USER`` variable.
+
+Set up the respective module to authenticate incoming users for the URL
+path of your installation. Then, change ``webapp/config/packages/security.yml``
+change the ``main`` section to look like this::
```
eldering: There is twice `change` in this sentence.

webapp/src/Migrations/Version20200214195715.php
```diff
@@ -0,0 +1,40 @@
+<?php
```
eldering: Do we still need this with the new configuration system? If a setting does not exist in the `etc/db-config.yaml` then it automatically does not show up anymore in the webinterface, right? So we don't need to delete anything, and this also has the (marginal) advantage that if you upgrade and then downgrade again, that your old customized settings are preserved.

thijskh: Yes, come to think of it, the migration is now wholly unnecessary

nickygerritsen: Yep correct

PR #700, opened 2020-02-15
Apache2.2 is not EOL yet

mvr320:
Not 100% if this is still supported/should be supported

### DIFF
--- a/etc/apache.conf.in
+++ b/etc/apache.conf.in
@@ -50,6 +50,8 @@ Alias /domjudge @domserver_webappdir@/public
 
 <Directory @domserver_webappdir@/public>
 Require all granted
+#When Apache <2.4:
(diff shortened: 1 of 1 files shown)


thijskh:
In fact it is EOL: http://httpd.apache.org/docs/current/index.html

Apache 2.4 was released in 2012.

I think we can do without the special exceptions for such legacy software. If you still run it, it should be not so hard to find out how to change it.

mvr320:
> In fact it is EOL: http://httpd.apache.org/docs/2.2/index.html
> 
> Apache 2.4 was released in 2012.
> 
> I think we can do without the special exceptions for such legacy software. If you still run it, it should be not so hard to find out how to change it.

My bad...

Closed the PR

PR #709, opened 2020-02-23
Refactor/scoreboard pod

eldering:
Refactor scoreboard code to use "plain old data" structures instead of classes with private members to remove unnecessary get/set methods.
I would argue that the main reason for using classes is to clarify the data structure here.

### DIFF
--- a/webapp/src/Utils/Scoreboard/ProblemSummary.php
+++ b/webapp/src/Utils/Scoreboard/ProblemSummary.php
@@ -7,82 +7,47 @@ class ProblemSummary
     /**
      * @var int[]
      */
-    protected $numberOfSubmissions = [];
+    public $numSubmissions = [];
 
     /**
      * @var int[]
      */
-    protected $numberOfPendingSubmissions = [];
+    public $numSubmissionsPending = [];
 
     /**
      * @var int[]
      */
-    protected $numberOfCorrectSubmissions = [];
+    public $numSubmissionsCorrect = [];
 
     /**
      * @var float[]
      */
-    protected $bestTimes = [];
+    public $bestTimes = [];
 
     /**
      * @param int $sortorder
-     * @return int
-     */
-    public function getNumberOfSubmissions(int $sortorder): int
-    {
-        return $this->numberOfSubmissions[$sortorder] ?? 0;
-    }
-
-    /**
-     * @param int $sortorder
-     * @return int
-     */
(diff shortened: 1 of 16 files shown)


webapp/src/Utils/Scoreboard/ProblemSummary.php
```diff
@@ -7,82 +7,47 @@ class ProblemSummary
...
      * @var int[]
      */
-    protected $numberOfSubmissions = [];
+    public $numSubmissions = [];
 
     /**
      * @var int[]
      */
-    protected $numberOfPendingSubmissions = [];
+    public $numSubmissionsPending = [];
```
meisterT: numPendingSubmissions?

eldering: I preferred this way since I think it's clear and grammatically correct too, and now things align better:
```php
$numSubmissions
$numSubmissionsPending
$numSubmissionsCorrect
```

webapp/templates/partials/scoreboard_summary.html.twig
```diff
@@ -24,19 +24,19 @@
                     <a {% if link %}href="{{ link }}"{% endif %}>
                         <i class="fas fa-thumbs-up fa-fw"></i>
                         <span style="font-size:90%;" title="number of accepted submissions">
-                                {{ summary.numberOfCorrectSubmissions(sortOrder) }}
+                                {{ summary.numSubmissionsCorrect[sortOrder] }}
```
tom93: This (and below) crashes if sortOrder is -1, which happens when there are no teams because scoreboard_table.html.twig displays the summary of the last sortorder group even if it is empty:
https://github.com/DOMjudge/domjudge/blob/cea1a256301806c5768e7b3b6faffbd13fd55c51/webapp/templates/partials/scoreboard_table.html.twig#L288:L289
Compare this to the summaries of the earlier sortorder groups, which explicitly check for -1:
https://github.com/DOMjudge/domjudge/blob/cea1a256301806c5768e7b3b6faffbd13fd55c51/webapp/templates/partials/scoreboard_table.html.twig#L108:L110

One way to fix this would be to use `... ?? 0` here like the old getters did.
Another way would be to remove the summary of the last sortorder group if the sortorder is -1, which will omit the summary row when there are no teams (the previous behaviour was to display a summary row with all 0's for the stats).

PR #710, opened 2020-02-25
Team manual in RST

thijskh:
This makes the team manual be in RST format from which a PDF will be generated.

Notes:
- The resulting PDF can use more styling to be visually more appealing. Of course this also needs an idea of how we want it too look.
- I skipped the switching based on `submitclient_enabled`. Is that a relevant thing to keep you think?


thijskh:
Here's an example of a built manual: https://djdev.pt-48.utr.surfcloud.nl/domjudge/doc/manual/build/team/domjudge-team-manual.pdf

thijskh:
It seems that the chosen approach (rst2pdf as sphinx plugin) is broken on Debian Buster. This is rather inconvenient. So it may be a good idea to investigate how to implement it slightly differently. Till then on hold.

meisterT:
What's exactly broken?

thijskh:
It is this issue: https://github.com/rst2pdf/rst2pdf/issues/678 fixed in rst2pdf 0.94 which is not in Debian buster. Looking to see if it can be worked around.

thijskh:
It can be worked around.

doc/manual/Makefile
```diff
@@ -15,22 +15,27 @@ docs: config
 # KLUDGE: call make instead of a dependency so that the html target
 # will run after config from inside the doc target.
 docs distdocs:
-	$(MAKE) html
+	$(MAKE) html team
```
eldering, line 18: Having `html` refer to the admin/jury manual and `team` to the (PDF) team manual is a bit weird. Maybe rename `html` to `admin` or so?

doc/manual/overview.rst
```diff
@@ -128,7 +128,7 @@ and Github users mpsijm, sylxjtu.
 Some code has been ported from the ETH Zurich fork by Christoph
 Krautz, Thomas Rast et al.
 
-DOMjudge is Copyright (c) 2004 - 2020 by the DOMjudge developers and contributors.
+DOMjudge is Copyright (c) 2004 - |today| by the DOMjudge developers and contributors.
```
eldering, line 131: Just to confirm: `|today|` only prints a year, not a full date?

thijskh: Yes: https://github.com/DOMjudge/domjudge/pull/710/files#diff-3823844ce2f5e2ce64bf66dd74002ad2R64

meisterT:
What do I need to install to try building this myself? Can you please add this to the admin manual?

thijskh:
In fact, I already added that:
https://github.com/DOMjudge/domjudge/pull/710/files#diff-495bcea1b94d0d3330fc9fa5c7c0f793

meisterT:
Thanks. But then there's something else missing on my machine. I get the following when trying to make the docs:
```
Extension error:
Could not import extension rst2pdf.pdfbuilder (exception: No module named 'rst2pdf')
```

thijskh:
Ugh, Python is such a messy language. Does this help:
```
sudo apt install python-sphinx python-sphinx-rtd-theme
```

nickygerritsen:
Probably need this: https://github.com/DOMjudge/domjudge-packaging/commit/e009096a431ad32204971907052115b1a8591f74

thijskh:
I'd prefer to solve it without pipping around.

nickygerritsen:
> I'd prefer to solve it without pipping around.

If your suggestion works indeed, please tell me so I will update the Gitlab CI docker image

meisterT:
I think I tried this and it didn't help. I can confirm later.

thijskh:
So for al clarity my suggestion is to install those packages in the Python 2 version.

doc/manual/Makefile
thijskh: So the sources of the team manual are now part of the entire manual. So html refers to the HTML version of the built manual. The team target builds a PDF of specifically the team manual.

Issue #701, opened 2020-02-17
Issues with restore_sources2db


thijskh:
There are some issues with restore_sources2db which makes it currently not work:
* submitdir does not store entry point information, making it not possible to reliably reconstruct the db from the source files;
* it uses second precision for the timestamps, this should be microsecond (not natively supported in PHP, need to call `stat`);
* it tries to query a hardcoded "domjudge-test" database (could maybe be addressed by just removing legacy code).

But given these problems that have existed for a while now, it begs the question of whether we should fix all of them or rather drop the facility. The first one seems hardest to fix.

eldering:
I think we can indeed drop this (and also `save_sources2file`) and replace it with import/export functionality to the contest archive format (which is near finalized).

PR #711, opened 2020-02-29
Drop old scripts restore_sources2db and save_sources2file.

eldering:
These are not uptodate anymore because they don't store the
entry point. Probably best to implement similar functionality
using the new ICPC Contest Archive Format.

Closes: #701


thijskh:
Should we then also drop the entire saving of submissions to the $submitdir?

meisterT:
Yes, let's drop that as well

thijskh:
This means we can also should the submitdir concept from `configure`, from the source diff functionality and associated checks in the configchecker.

PR #716, opened 2020-03-16
Add API action to update config.

nickygerritsen:
It mimics the /jury/config code, except that it doesn't require config_ as prefix.
It has some duplicate code now, we might want to refactor that at some point.
Also fix /api/config with no name.

### DIFF
--- a/webapp/src/Controller/API/GeneralInfoController.php
+++ b/webapp/src/Controller/API/GeneralInfoController.php
@@ -2,7 +2,9 @@
 
 namespace App\Controller\API;
 
+use App\Entity\Configuration;
 use App\Entity\Contest;
+use App\Entity\Judging;
 use App\Entity\User;
 use App\Service\CheckConfigService;
 use App\Service\ConfigurationService;
@@ -12,6 +14,7 @@ use Doctrine\ORM\EntityManagerInterface;
 use FOS\RestBundle\Controller\AbstractFOSRestController;
 use FOS\RestBundle\Controller\Annotations as Rest;
 use Nelmio\ApiDocBundle\Annotation\Model;
+use Psr\Log\LoggerInterface;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
 use Swagger\Annotations as SWG;
 use Symfony\Component\HttpFoundation\Request;
@@ -60,6 +63,11 @@ class GeneralInfoController extends AbstractFOSRestController
     protected $router;
 
     /**
+     * @var LoggerInterface
+     */
+    protected $logger;
+
+    /**
      * GeneralInfoController constructor.
      *
      * @param EntityManagerInterface $em
@@ -68,6 +76,7 @@ class GeneralInfoController extends AbstractFOSRestController
      * @param EventLogService        $eventLogService
      * @param CheckConfigService     $checkConfigService
      * @param RouterInterface        $router
+     * @param LoggerInterface        $logger
      */
     public function __construct(
         EntityManagerInterface $em,
@@ -75,7 +84,8 @@ class GeneralInfoController extends AbstractFOSRestController
         ConfigurationService $config,
         EventLogService $eventLogService,
         CheckConfigService $checkConfigService,
-        RouterInterface $router
+        RouterInterface $router,
+        LoggerInterface $logger
     ) {
         $this->em                 = $em;
         $this->dj                 = $dj;
@@ -83,6 +93,7 @@ class GeneralInfoController extends AbstractFOSRestController
         $this->checkConfigService = $checkConfigService;
         $this->router             = $router;
         $this->config             = $config;
+        $this->logger             = $logger;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


nickygerritsen:
Lol huge oversight. Let’s allow only admin?

webapp/src/Controller/API/GeneralInfoController.php
```diff
@@ -256,6 +271,158 @@ public function getDatabaseConfigurationAction(Request $request)
...
+     *     type="object",
+     *     description="The config variables to update",
+     *     required=true
+     * )
+     * @param Request $request
+     * @return \App\Entity\Configuration[]|mixed
+     * @throws \Exception
+     */
+    public function updateConfigurationAction(Request $request)
+    {
```
eldering: It feels like that a lot of this code actually belongs in the `ConfigurationService`. Also, there's a lot of duplication with `Jury/ConfigController.php`.

So I'd prefer seeing what common code can be moved into the `ConfigurationService`.

thijskh: It might indeed be better to do the deduplication before we introduce it instead of in the future

nickygerritsen: I moved the shared code to one spot. It required to inject some services (DOMjudgeService, EventLogService) as arguments, since they could not be autowired in the constructor, since they both need the config service and that results in a circular dependency.

Also added admin security.

webapp/src/Service/ConfigurationService.php
```diff
@@ -183,6 +185,152 @@ function (ConfigCacheInterface $cache) {
...
+     *
+     * @param Request         $request
+     * @param EventLogService $eventLog
+     * @param DOMJudgeService $dj
+     * @param string          $keyPrefix
+     *
+     * @throws NonUniqueResultException
+     */
+    public function saveChanges(
+        Request $request,
```
eldering: Passing the `$request` to the the ConfigurationService feels a bit weird (although it's simple). But that means we can't really use it from a non-web context. Maybe better to pass the relevant information in a different structure?

nickygerritsen: I now convert the request to an array and pass that. That seems to be more usable?

webapp/src/Service/ConfigurationService.php
```diff
@@ -183,6 +185,150 @@ function (ConfigCacheInterface $cache) {
...
+     * Save the changes from the given request
+     *
+     * @param array $data
+     * @param EventLogService $eventLog
+     * @param DOMJudgeService $dj
+     *
+     * @throws NonUniqueResultException
+     */
+    public function saveChanges(
+        array $dataToSet,
```
eldering: Nitpick: the variable name mismatches `$data` above.

nickygerritsen: Whoops had `$data` before but that variable was already used.

webapp/src/Controller/API/GeneralInfoController.php
```diff
@@ -256,6 +272,54 @@ public function getDatabaseConfigurationAction(Request $request)
...
+        $this->config->saveChanges($request->request->all(), $this->eventLogService, $this->dj);
+        $onlypublic = !($this->dj->checkrole('jury') || $this->dj->checkrole('judgehost'));
+        return $this->config->all($onlypublic);
+    }
+
+    private function logUnverifiedJudgings()
```
thijskh: This is now duplicated, probably this copy needs to be removed

thijskh:
Would be great if we had an API test suite. Just dreaming out loud.

nickygerritsen:
> Would be great if we had an API test suite. Just dreaming out loud.

giphy

eldering:
> Would be great if we had an API test suite. Just dreaming out loud.

There is already this: https://github.com/DOMjudge/domjudge-scripts/blob/master/contest-api/check-api.sh

Or did you have something very different in mind?

thijskh:
Yes, that validates the output for schema, I was more thinking of testing the api calls: putting different config updates in this one, check if output of other endpoint then changes accordingly

webapp/src/Controller/API/GeneralInfoController.php
```diff
@@ -256,6 +272,32 @@ public function getDatabaseConfigurationAction(Request $request)
...
+     * @SWG\Response(
+     *     response="200",
+     *     description="The full configuration after change",
+     *     @SWG\Schema(type="object")
+     * )
+     * @SWG\Parameter(
+     *     name="config",
+     *     in="formData",
+     *     type="object",
+     *     description="The config variables to update",
```
meisterT: nit: can you add to the description in what form one has to provide them?

webapp/src/Controller/API/GeneralInfoController.php
```diff
@@ -226,6 +237,7 @@ public function getUserAction()
     /**
      * Get configuration variables
      * @Rest\Get("/config")
+     * @IsGranted("ROLE_ADMIN")
```
meisterT: this shouldn't be here but below at the update code path, right?

webapp/src/Controller/API/GeneralInfoController.php
```diff
@@ -256,6 +272,32 @@ public function getDatabaseConfigurationAction(Request $request)
...
+     *     required=true
+     * )
+     * @param Request $request
+     * @return \App\Entity\Configuration[]|mixed
+     * @throws \Exception
+     */
+    public function updateConfigurationAction(Request $request)
+    {
+        $this->config->saveChanges($request->request->all(), $this->eventLogService, $this->dj);
+        $onlypublic = !($this->dj->checkrole('jury') || $this->dj->checkrole('judgehost'));
```
meisterT: only admin should be able to do this, right?

webapp/src/Controller/API/GeneralInfoController.php
nickygerritsen: Yes

nickygerritsen:
I can update the comment as to how to provide them. I don’t think we can give an exact example.

I will also add a httpie example in the commitb

PR #721, opened 2020-03-22
Fix permissions on webapp/var/* by using group setsid bit.

eldering:


thijskh:
It does not work for me yet.

I get:
```
PHP Fatal error:  Uncaught Symfony\\Component\\Filesystem\\Exception\\IOException: Cannot rename "/tmp/removed-ids.phpRQflL1" to "/opt/domjudge/domserver/webapp/var/cache/prod/ContainerHOlEnwl/removed-ids.php". in /opt/domjudge/domserver/lib/vendor/symfony/filesystem/Filesystem.php:290
```

These are the permissions of the mentioned files:
```
drwxrwsr-x+ 2 thijs www-data 28672 Mar 31 19:28 /opt/domjudge/domserver/webapp/var/cache/prod/ContainerHOlEnwl/
-rw-rwSr--+ 1 thijs www-data 72159 Mar 31 19:28 /opt/domjudge/domserver/webapp/var/cache/prod/ContainerHOlEnwl/removed-ids.php

-rw-r--r-- 1 www-data www-data 72159 Mar 31 19:31 /tmp/systemd-private-1b2d4f84d8fe4c05abcdbe20577105c0-apache2.service-vcDiOf/tmp/removed-ids.phpoieRyp
```

Changing to the user `www-data` I can indeed not move the latter file to the former location (permission denied).

PR #725, opened 2020-03-31
Fix link to documentation from the admin index page.

### LINKED ISSUE #698: 404 error for click 'DOMjudge Manual' in DOMJudge Jury Interface
http://xx.xx.xx.xx/domjudge/doc/manual/build/html/index.html?v=7.2.0

It will report a 404 error after clicking the highlighted hyperlink.

image

thijskh:
The previous link only worked on the maintainer-install of DOMjudge.
Make it so we link to the right location at the regular install,
which also drops the ugly `build/` from the url path. And add a
symlink on the maintainer-install so the docs work there also.

Closes #698


thijskh:
@eldering Can you review my Makefile changes?

PR #730, opened 2020-04-22
Fix Symfony 4.4 / PHP 7.4 deprecations

thijskh:
This fixes another set of deprecation notices in our codebase triggered by either the upgrade of SF 4.4 or PHP 7.4.

One remains that I find hard to find a solution for, any idea @nickygerritsen ?
```
Loading the file "../src/Controller/API/" from the global resource directory "/opt/domjudge/domserver/webapp/src" is deprecated since Symfony 4.4 and will be removed in 5.0.
```
https://github.com/DOMjudge/domjudge/blob/fd5b3ef6ddb090b45b80e4fc69737e20b3107688/webapp/config/routes.yaml#L4

The other remaining notices that I managed to trigger are from libraries that have not yet fixed them in a released version or on a branch we can easily upgrade to.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -2501,16 +2501,16 @@
         },
         {
             "name": "jms/serializer",
-            "version": "3.1.1",
+            "version": "3.6.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/schmittjoh/serializer.git",
-                "reference": "7f8dc86e9168d0112b3cc38ba8cca41b17f409a0"
+                "reference": "e82527ceddedf8dd988a4ce900f4b14fedf4d6f2"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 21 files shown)


nickygerritsen:
I will dive into that error. Don't have a solution ready.

thijskh:
It will only be an error when we upgrade to SF 5.0. So if the other changes are OK for you we can also just ~merge~ rebase and merge these changes now.

nickygerritsen:
It seems that is a thing FosRestBundle does that they have since removed in master (the `DirectoryRouteLoader` that triggers it is gone). So I propose we just wait for an update for that to go away and we merge this PR.

thijskh:
Let's cross that bridge when there's a reason to cross it then

PR #733, opened 2020-04-25
Add info of SLACK

mvr320:
I think the SLACK is easier for experienced people to help (personal opinion)


mvr320:
@thijskh do you agree that the part about the mailinglist should than also removed? To have this info at only 1 location?

meisterT:
I think it's fine to keep the mailing list.

PR #736, opened 2020-04-25
Emit events for all contests an entity belongs to. Fixes #441.

### LINKED ISSUE #441: Removing problems from a contest doesn't emit a `delete` event
When we remove problems from a contest, we do not call `eventlog` for deletion of that problem.
This means that the CDS will think a lot of problems exist in the contest which actually do not.

nickygerritsen:
Wow this was more than I expected. We did not emit a lot of create/update events:
* We only emitted them for the current contest.
* We did not emit any when creating entities, because the ID was not known yet.

Also we didn't emit *any* delete events.

Now we:
* Emit delete events for direct deletion of entities.
* Emit create / update events for all contests an entity belongs to, or all active contests if the entity is 'global'.
* Emit delete events when editing a contest and deleting teams, team categories and/or problems.

We still don't emit the following:
* Cascading delete events. We can try to fix this later, the code has a `TODO` to mark where to do it. We could maybe use the existing `BaseController::getDependentEntities()` method, or we could move it into `EventLogService::log` with a thing similar to `initStaticEvents`, but then for deletions.
* Create / update events when editing a contest and adding or modifying teams, team categories and/or problems. We could, but it's not trivial to check for changes. Furthermore this will happen when a client connects or we have an event that needs it anyway.

### DIFF
--- a/webapp/src/Controller/BaseController.php
+++ b/webapp/src/Controller/BaseController.php
@@ -3,12 +3,15 @@
 namespace App\Controller;
 
 use App\Entity\BaseApiEntity;
+use App\Entity\Contest;
 use App\Entity\Problem;
 use App\Entity\RankCache;
 use App\Entity\ScoreCache;
 use App\Entity\Team;
+use App\Entity\TeamAffiliation;
 use App\Service\DOMJudgeService;
 use App\Service\EventLogService;
+use App\Utils\Utils;
 use Doctrine\Common\Inflector\Inflector;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\MappingException;
@@ -88,19 +91,9 @@ abstract class BaseController extends AbstractController
         $id,
         bool $isNewEntity
     ) {
-        $class        = get_class($entity);
-        $parts        = explode('\\', $class);
-        $entityType   = $parts[count($parts) - 1];
-        $auditLogType = Inflector::tableize($entityType);
+        $auditLogType = Utils::tableForEntity($entity);
 
         $entityManager->flush();
-        if ($endpoint = $eventLogService->endpointForEntity($entity)) {
-            if ($contest = $DOMJudgeService->getCurrentContest()) {
-                $eventLogService->log($endpoint, $id,
-                                      $isNewEntity ? EventLogService::ACTION_CREATE : EventLogService::ACTION_UPDATE,
-                                      $contest->getCid());
-            }
-        }
 
         // If we have no ID but we do have a Doctrine entity, automatically
         // get the primary key if possible
@@ -117,26 +110,38 @@ abstract class BaseController extends AbstractController
             }
         }
 
+        if ($endpoint = $eventLogService->endpointForEntity($entity)) {
+            foreach ($this->contestsForEntity($entity, $DOMJudgeService) as $contest) {
+                $eventLogService->log($endpoint, $id,
+                                      $isNewEntity ? EventLogService::ACTION_CREATE : EventLogService::ACTION_UPDATE,
+                                      $contest->getCid());
+            }
+        }
+
         $DOMJudgeService->auditlog($auditLogType, $id, $isNewEntity ? 'added' : 'updated');
     }
 
     /**
      * Perform the delete for the given entity
+     *
      * @param Request                $request
      * @param EntityManagerInterface $entityManager
      * @param DOMJudgeService        $DOMJudgeService
+     * @param EventLogService        $eventLogService
      * @param KernelInterface        $kernel
      * @param                        $entity
      * @param string                 $description
      * @param string                 $redirectUrl
      * @return \Symfony\Component\HttpFoundation\Response
      * @throws \Doctrine\DBAL\DBALException
+     * @throws \Doctrine\ORM\NoResultException
      * @throws \Doctrine\ORM\NonUniqueResultException
      */
     protected function deleteEntity(
         Request $request,
         EntityManagerInterface $entityManager,
         DOMJudgeService $DOMJudgeService,
+        EventLogService $eventLogService,
         KernelInterface $kernel,
         $entity,
         string $description,
@@ -176,11 +181,8 @@ abstract class BaseController extends AbstractController
         $isError          = false;
         $messages         = [];
         $propertyAccessor = PropertyAccess::createPropertyAccessor();
-        $class            = get_class($entity);
-        $parts            = explode('\\', $class);
-        $entityType       = $parts[count($parts) - 1];
-        $readableType     = str_replace('_', ' ', Inflector::tableize($entityType));
-        $metadata         = $entityManager->getClassMetadata($class);
+        $readableType     = str_replace('_', ' ', Utils::tableForEntity($entity));
+        $metadata         = $entityManager->getClassMetadata(get_class($entity));
         $primaryKeyData   = [];
         foreach ($metadata->getIdentifierColumnNames() as $primaryKeyColumn) {
             $primaryKeyColumnValue = $propertyAccessor->getValue($entity, $primaryKeyColumn);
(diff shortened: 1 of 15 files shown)


webapp/src/Controller/Jury/ContestController.php
```diff
@@ -504,8 +508,61 @@ public function editAction(Request $request, int $contestId)
...
+            // when someone connects to the event feed (or we have a
+            // dependent event) anyway and adding the code here would
+            // overcomplicate this function
+            $deletedEntities = function(Collection $collection, string $idMethod) {
+                /** @var PersistentCollection $collection */
+                $deletedEntities = [];
+                foreach ($collection->getSnapshot() as $oldEntity) {
+                    $found = false;
+                    foreach ($collection->toArray() as $newEntity) {
+                        $newId = call_user_func([$newEntity, $idMethod]);
```
meisterT, line 523: what exactly are you calling in these lines?

nickygerritsen: `getTeamid()` for teams, `getCategoryid()` for team categories and `getProbid()` for contest problems.

webapp/src/Controller/BaseController.php
```diff
@@ -255,6 +262,10 @@ protected function deleteEntity(
                 $entityId = $entity->getTeamid();
             }
 
+            // Get the contests to trigger the event for. We do this before
+            // deleting the entity, since linked data might have vanished
+            $contestsForEvents = $this->getEventContestsForEntity($entity, $DOMJudgeService);
```
eldering, line 267: I think the name of this function is not very clear. Also the variable and function names don't match.
Maybe rename the method to `getContestsForEventsForEntity` although I'm not a big fan of such long names.

webapp/src/Controller/Jury/ContestController.php
```diff
@@ -504,8 +508,61 @@ public function editAction(Request $request, int $contestId)
                 $problem->setContest($contest);
             }
 
+            // Determine the removed teams, team categories and problems.
+            // Note that we do not send out create / update events for
+            // existing / new problems, teams and team categories. This happens
+            // when someone connects to the event feed (or we have a
+            // dependent event) anyway and adding the code here would
+            // overcomplicate this function
```
eldering, line 516: It's not clear to me what this code does exactly and why. Is this already generating a partial cascade of delete events?

nickygerritsen: A no, it checks what teams (/categories/problems) *used* to be set on the contest but are not anymore. It would be wayyyy better to use array_diff here, but that only works on strings and not on objects, because logic.

So the *only* thing I want is that if I remove a team(/category/problem) from a contest, it emits a delete event (since we dont pick these up any other way).

webapp/src/Controller/BaseController.php
```diff
@@ -278,12 +289,23 @@ protected function deleteEntity(
                 $entityManager->remove($entity);
             });
 
+            // Add an audit log entry
             $class        = get_class($entity);
             $parts        = explode('\\', $class);
             $entityType   = $parts[count($parts) - 1];
```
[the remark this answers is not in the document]
nickygerritsen: We can't call `end` on a function return type:
image

webapp/src/Controller/BaseController.php
```diff
@@ -351,4 +373,49 @@ protected function getDependentEntities(string $entityClass, array $relations)
...
+        // Otherwise use the currently active contests
+        $contests = [];
+        if ($entity instanceof Team || $entity instanceof TeamAffiliation) {
+            $possibleContests = $dj->getCurrentContests();
+            foreach ($possibleContests as $contest) {
+                if ($contest->isOpenToAllTeams()) {
+                    $contests[] = $contest;
+                } elseif ($entity->getContests()->contains($contest)) {
+                    $contests[] = $contest;
+                }
```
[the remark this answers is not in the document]
nickygerritsen: Fixed this

webapp/src/Controller/BaseController.php
nickygerritsen: I dont like such a long name either.... But I agree it's not clear.

webapp/src/Controller/BaseController.php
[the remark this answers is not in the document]
eldering: Ok. But I also see that these 4 lines appear at least 3 times in this file, so would make sense to turn it into a small function.

webapp/src/Controller/Jury/ContestController.php
```diff
@@ -504,8 +508,61 @@ public function editAction(Request $request, int $contestId)
...
+            $deletedEntities = function(Collection $collection, string $idMethod) {
+                /** @var PersistentCollection $collection */
+                $deletedEntities = [];
+                foreach ($collection->getSnapshot() as $oldEntity) {
+                    $found = false;
+                    foreach ($collection->toArray() as $newEntity) {
+                        $newId = call_user_func([$newEntity, $idMethod]);
+                        $oldId = call_user_func([$oldEntity, $idMethod]);
```
eldering, line 524: For performance would be better to move this line outside the for-loop over `$newEntity`.

webapp/src/Controller/Jury/ContestController.php
```diff
@@ -504,8 +508,61 @@ public function editAction(Request $request, int $contestId)
...
 
+            // Determine the removed teams, team categories and problems.
+            // Note that we do not send out create / update events for
+            // existing / new problems, teams and team categories. This happens
+            // when someone connects to the event feed (or we have a
+            // dependent event) anyway and adding the code here would
+            // overcomplicate this function
+            $deletedEntities = function(Collection $collection, string $idMethod) {
+                /** @var PersistentCollection $collection */
+                $deletedEntities = [];
```
eldering, line 519: This `$deletedEntities` is different from the variable that the function is assigned to, right? I think for clarity it's better to give them separate names. Also, if I understand correctly, then `getSnaphost()` returns the data that's currenty in the DB, right? Might be helpful to explain that in the comments at the top.

webapp/src/Controller/Jury/ContestController.php
```diff
@@ -504,8 +508,61 @@ public function editAction(Request $request, int $contestId)
...
+
             $this->saveEntity($this->em, $this->eventLogService, $this->dj, $contest,
                               $contest->getCid(), false);
+
+            $teamEndpoint         = $this->eventLogService->endpointForEntity(Team::class);
+            $teamCategoryEndpoint = $this->eventLogService->endpointForEntity(TeamCategory::class);
+            $problemEndpoint      = $this->eventLogService->endpointForEntity(Problem::class);
+
+            // TODO: cascade deletes. Maybe use getDependentEntities()?
+            foreach ($deletedTeams as $deletedTeam) {
```
eldering, line 554: Nitpick: what about shorter `$team` instead of `$deletedTeam`? It's a tight for-loop and it's already clear that it's coming from an array `$deletedTeams`.

webapp/src/Controller/BaseController.php
nickygerritsen: Gave it a shot at a better name

webapp/src/Controller/BaseController.php
```diff
@@ -351,4 +364,61 @@ protected function getDependentEntities(string $entityClass, array $relations)
...
+        return $contests;
+    }
+
+    /**
+     * Return the table name for the given entity
+     * @param $entity
+     *
+     * @return string
+     */
+    protected function tableForEntity($entity)
```
eldering, line 417: Can't this be a static method outside this class? It doesn't depend on anything in this class, right?

webapp/src/Entity/Team.php
```diff
@@ -854,4 +854,18 @@ public function validate(ExecutionContextInterface $context)
...
+     *
+     * @param Contest $contest
+     *
+     * @return bool
+     */
+    public function inContest(Contest $contest): bool
+    {
+        return $contest->isOpenToAllTeams() ||
+            $this->getContests()->contains($contest) ||
+            ($this->getCategory() !== null && $this->getCategory()->getContests()->contains($contest));
```
eldering, line 869: This should use `$this->getCategory()->inContest($contest)`.

webapp/src/Controller/BaseController.php
nickygerritsen: Any suggestion for where to move it?

eldering: `webapp/src/Utils/Utils.php`?

PR #739, opened 2020-04-25
API permissions for public/team

nickygerritsen:
These three endpoints didn't allow team/public access while they should.

I did remember we had some discussion about the awards endpoint that made us decide we shouldn't allow public access, but I can't remember anymore what it was.

Note that the spec also allows public access to judgements and runs but I disagree and have already send a message to the mailinglist to check the opinion of others.

### DIFF
--- a/webapp/src/Controller/API/ClarificationController.php
+++ b/webapp/src/Controller/API/ClarificationController.php
@@ -23,11 +23,12 @@ use Symfony\Component\HttpFoundation\Request;
 class ClarificationController extends AbstractRestController
 {
     /**
-     * Get all the clarifications for this contest
+     * Get all the clarifications for this contest.
+     *
+     * Note that we restrict the returned clarifications in the query builder.
      * @param Request $request
      * @return \Symfony\Component\HttpFoundation\Response
      * @Rest\Get("")
-     * @Security("is_granted('ROLE_JURY') or is_granted('ROLE_JUDGEHOST') or is_granted('ROLE_API_READER')")
      * @SWG\Response(
      *     response="200",
      *     description="Returns all the clarifications for this contest",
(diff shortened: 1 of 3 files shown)


webapp/src/Controller/API/ClarificationController.php
```diff
@@ -27,7 +27,6 @@ class ClarificationController extends AbstractRestController
      * @param Request $request
      * @return \Symfony\Component\HttpFoundation\Response
      * @Rest\Get("")
-     * @Security("is_granted('ROLE_JURY') or is_granted('ROLE_JUDGEHOST') or is_granted('ROLE_API_READER')")
```
eldering, line 30: I was first confused that we now allow anyone to view all clarifications, but then realized this is constrained in the QueryBuilder. Might be good to add a comment here.

webapp/src/Controller/API/SubmissionController.php
```diff
@@ -386,6 +384,22 @@ protected function getQueryBuilder(Request $request): QueryBuilder
...
+            $queryBuilder
+                ->join('s.team', 't')
+                ->join('t.category', 'cat');
+            if ($this->dj->checkrole('team')) {
+                $queryBuilder
+                    ->andWhere('cat.visible = 1 OR s.team = :team')
+                    ->setParameter('team', $this->dj->getUser()->getTeam());
+            } else {
+                // Hide all submissions made by non public teams
+                    $queryBuilder->andWhere('cat.visible = 1');
```
eldering, line 399: Indentation is wrong here.

webapp/src/Controller/API/SubmissionController.php
```diff
@@ -386,6 +384,22 @@ protected function getQueryBuilder(Request $request): QueryBuilder
             $queryBuilder->andWhere('s.submittime < c.endtime');
         }
 
+        if (!$this->dj->checkrole('api_reader') &&
+            !$this->dj->checkrole('judgehost'))
+        {
+            $queryBuilder
+                ->join('s.team', 't')
```
eldering, line 391: Above we should probably already join on team and enforce `t.enabled = 1` in general, because we consider a disabled team as deleted if I'm not mistaken.

nickygerritsen:
We could add something like this to the integration test, if we have two teams that submit submissions and clarifications and a jury member that adds a clarification. We might need an API for that last thing.

Issue #189, opened 2016-01-24
Allow deleting of testcase with associated judgings


eldering:
Currently a testcase that has associated judging_run's cannot be deleted.
Proposal is to allow this, and make the judging_run.testcaseid a dangling reference. Then on the submission details page, we have to filter out links to any missing testcase in/output files.

nickygerritsen:
I did a little digging and this is non-trivial: we use the testcase.probid to load the problem, but we should then use join's on judging and submission

eldering:
Related issue on deleting problems: #243
After thinking more about it, I think that we should not support deleting testcases when these already have associated judging_runs. Besides being non-trivial to handle in the code (and thus bug-prone) by creating dangling references, it creates ambiguity about what it means for the correctness of the submission. If the deleted testcase was the only non-correct judging_run, does that turn the total judging result correct now??

eldering:
Closing this because it technically tricky and also creates ambiguous situations when there are judgings, with some of the judging_runs associated to deleted test cases.

eldering:
Reopening as this is an issue that we ourselves even ran into. We still have to decide how we can properly deal with the dangling references, or possibly delete all associated judgings in this case.

meisterT:
If you want to delete a testcase, you want to rejudge the problem anyway. However, you probably also want to create a proper rejudging to have an overview what has changed. In this case we cannot delete associated judgings.

Another possibility is to mark the testcase as deleted by adding an attribute but not actually delete it. Then there's no dangling reference since the testcase still exist and we can handle it properly.

bumfo:
Sorry for opening duplicated issues. I have searched that but didn't found this one.

An alternative way I think is to mark a test case as disabled, so it won't be run anymore. Should this be another issue called "Allow disabling problem test cases"?

PR #740, opened 2020-04-25
Warn if testcases have been re-ordered (or deleted).

meisterT:
This is in preparation of #189.

### DIFF
--- a/webapp/src/Controller/Jury/SubmissionController.php
+++ b/webapp/src/Controller/Jury/SubmissionController.php
@@ -465,7 +465,6 @@ class SubmissionController extends BaseController
                 $testcase = $runResult[0];
                 if (isset($judgingRunTestcaseIdsInOrder[$cnt])) {
                     if ($testcase->getTestcaseid() != $judgingRunTestcaseIdsInOrder[$cnt]['testcaseid']) {
-                        dump($judgingRunTestcaseIdsInOrder);
                         $sameTestcaseIds = false;
                     }
                 }
--- a/webapp/src/Controller/Jury/SubmissionController.php
+++ b/webapp/src/Controller/Jury/SubmissionController.php
@@ -416,6 +416,7 @@ class SubmissionController extends BaseController
 
         $runs       = [];
         $runsOutput = [];
+        $sameTestcaseIds = true;
         if ($selectedJudging || $externalJudgement) {
             $queryBuilder = $this->em->createQueryBuilder()
                 ->from(Testcase::class, 't')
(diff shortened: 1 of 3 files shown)


webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -449,7 +450,26 @@ public function viewAction(Request $request, int $submitId)
                 ->getQuery()
                 ->getResult();
 
+            $judgingRunTestcaseIdsInOrder = $this->em->createQueryBuilder()
+                ->from(JudgingRun::class, 'jr')
+                ->select('jr.testcaseid')
+                ->andWhere('jr.judging = :judging')
+                ->setParameter(':judging', $selectedJudging)
+                ->orderBy('jr.endtime')
```
nickygerritsen, line 458: If we ever allow running testcases on different machines at the same time (like @ubergeek42 was working on) this can break right?

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -449,7 +450,26 @@ public function viewAction(Request $request, int $submitId)
...
+                ->getQuery()
+                ->getScalarResult();
+
+            $cnt = 0;
             foreach ($runResults as $runResult) {
+                /** @var Testcase $testcase */
+                $testcase = $runResult[0];
+                if (isset($judgingRunTestcaseIdsInOrder[$cnt])) {
+                    if ($testcase->getTestcaseid() != $judgingRunTestcaseIdsInOrder[$cnt]['testcaseid']) {
+                        dump($judgingRunTestcaseIdsInOrder);
```
nickygerritsen, line 468: Leftover dump

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -449,7 +450,26 @@ public function viewAction(Request $request, int $submitId)
...
+                ->orderBy('jr.endtime')
+                ->getQuery()
+                ->getScalarResult();
+
+            $cnt = 0;
             foreach ($runResults as $runResult) {
+                /** @var Testcase $testcase */
+                $testcase = $runResult[0];
+                if (isset($judgingRunTestcaseIdsInOrder[$cnt])) {
+                    if ($testcase->getTestcaseid() != $judgingRunTestcaseIdsInOrder[$cnt]['testcaseid']) {
```
nickygerritsen, line 467: If ``judgingRunTestcaseIdsInOrder[$cnt]` is not set, we should also emit the error right? Maybe remove the `isset` and change the right part of the if to `$judgingRunTestcaseIdsInOrder[$cnt]['testcaseid'] ?? null`?

webapp/src/Controller/Jury/SubmissionController.php
meisterT: fixed by only displaying the message when the judging is done

webapp/src/Controller/Jury/SubmissionController.php
meisterT: done

webapp/src/Controller/Jury/SubmissionController.php
meisterT: Nope. If we have lazy judging the this list will contain less items than the one we are iterating through and that's fine.

meisterT:
Note that https://github.com/DOMjudge/domjudge/pull/742/files has to be merged first.

PR #745, opened 2020-04-25
Feature: Add prometheus metrics

ubergeek42:
This gives us an endpoint that we can use to collect various metrics about domjudge itself. Currently included are the following metrics per contest:

- How many submissions the contest has received
- How many submissions were correct
- How many submissions were ignored
- How many submissions are unverified
- How many submissions are queued

- How many teams the contest has
- How many teams have logged in
- How many teams submitted something
- How many teams submitted something correct

### DIFF
--- a/webapp/config/packages/prometheus.yaml
+++ b/webapp/config/packages/prometheus.yaml
@@ -1,0 +1,11 @@
+services:
+    prometheus_storage:
+      public: false
+      class: \Prometheus\Storage\InMemory
+      # class: \Prometheus\Storage\APC
+    Prometheus\CollectorRegistry: '@prometheus_collector_registry_service'
+    prometheus_collector_registry_service:
+        public: true
+        class: \Prometheus\CollectorRegistry
+        arguments:
+            - '@prometheus_storage'
--- a/webapp/config/packages/prometheus.yaml
+++ b/webapp/config/packages/prometheus.yaml
@@ -2,7 +2,6 @@ services:
     prometheus_storage:
       public: false
       class: \Prometheus\Storage\InMemory
-      # class: \Prometheus\Storage\APC
     Prometheus\CollectorRegistry: '@prometheus_collector_registry_service'
     prometheus_collector_registry_service:
         public: true
--- a/webapp/src/Controller/MetricsController.php
+++ b/webapp/src/Controller/MetricsController.php
@@ -1,0 +1,157 @@
+<?php declare(strict_types=1);
+
+namespace App\Controller;
+
+use App\Service\DOMJudgeService;
+use App\Service\SubmissionService;
+use App\Entity\Team;
+use Doctrine\ORM\EntityManagerInterface;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Annotation\Route;
+
+use Prometheus\CollectorRegistry;
+use Prometheus\RenderTextFormat;
+
+/**
+ * @Route("/prometheus/metrics")
+ * @IsGranted("ROLE_ADMIN")
+ */
+class MetricsController extends AbstractController
+{
+    /**
+     * @var EntityManagerInterface
+     */
+    protected $em;
+
+    /**
+     * @var DOMJudgeService
+     */
+    protected $dj;
+
+    /**
+     * @var SubmissionService
+     */
+    protected $submissionService;
+
+    protected $registry;
+
+    /**
+     * MetricsController constructor.
+     * @param EntityManagerInterface $em
+     * @param DOMJudgeService        $dj
+     * @param SubmissionService      $submissionService
+     * @param CollectorRegistry      $registry
+     */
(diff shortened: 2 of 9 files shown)


meisterT:
As I have no clue about prometheus, does this need to be hooked up in the menu or the index page?

nickygerritsen:
No. You tell Prometheus to look at this url and it generates nice data over time which you can visualize in grafana.
Maybe we should add a grafana dashboard json somewhere that can be used with this?

webapp/src/Controller/MetricsController.php
```diff
@@ -0,0 +1,157 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Annotation\Route;
+
+use Prometheus\CollectorRegistry;
+use Prometheus\RenderTextFormat;
+
+/**
+ * @Route("/prometheus/metrics")
+ * @IsGranted("ROLE_ADMIN")
```
meisterT, line 20: Do we want to use a separate role for this?

meisterT:
> No. You tell Prometheus to look at this url and it generates nice data over time which you can visualize in grafana.

Excuse my ignorance. IIRC, we can hook up Grafana directly to API endpoints. What's the advantage/purpose of Prometheus?

nickygerritsen:
Can’t quote on mobile. But Prometheus has history build in. I think it can also do alerting but Keith knows.
Also grafana auto detects Prometheus variables so you get a fancy dropdown

webapp/src/Controller/MetricsController.php
thijskh: Or if not a separate role, api_reader maybe more appropriate than admin?

eldering: I would say admin (or maybe api_reader) is correct: we're exposing general information about the system and I'd prefer to not have to worry about security implications here. Also, this would only be used by admins anyways.

thijskh: I would not want to use the full access admin account on another machine that only needs read access.

Its an API that only needs to be read from. And I believe it by and large exposes what is in another form available from the existing API. So I think the API readonly user is a good match.

webapp/config/packages/prometheus.yaml
```diff
@@ -0,0 +1,11 @@
+services:
+    prometheus_storage:
+      public: false
+      class: \Prometheus\Storage\InMemory
+      # class: \Prometheus\Storage\APC
```
nickygerritsen, line 5: Since you seem to set and return the values in the same function, does using this commented out APC storage even make sense? Otherwise we might as well remove it.

ubergeek42:
I don't believe you can hook grafana directly up to any api endpoints(I skimmed the data sources it has available and didn't see anything). Grafana is mainly just a visualization layer that interfaces with a bunch of different backends.

In the past at world finals we used Graphite and ran collectd on our systems to collect the metrics. We also ran a shell script on the metrics server which hit our api endpoints(or queried the database directly) and was consumed by collectd and sent to graphite. This worked ok, but is annoying to set up because graphite is finicky.

Prometheus is technically simpler. All the things you monitor simply need to serve a simple text response containing the instantaneous metric values. Prometheus will then store these in it's own database with it's own retention policies. You can indeed do alerting on it as well. I think there's a bit more flexibility in it's query language for getting the metrics values out that graphite had.

So it's mostly down to ease of set up and integration with existing systems.

webapp/config/packages/prometheus.yaml
ubergeek42: This is probably leftover from when I was trying to include some generic http level metrics in the response as well(things like p95/99 response times, request counts, etc). I think inmemory only lives as long as the single request, while apc lives as long as the process is alive(allowing it to aggregate information across requests).

webapp/src/Controller/MetricsController.php
ubergeek42: What is the scope of access for `api_reader`? Does it generally have access to "non-public" data/information during freezetime?

I made this admin just to make sure information doesn't leak if we added some metric like problems solved per team; but I do agree that putting the full admin account credentials on another machine is less than ideal.

I'm not opposed to creating a new metrics role or something like that and using it for just this one endpoint if we think that'd be the best solution.

ubergeek42:
> Maybe we should add a grafana dashboard json somewhere that can be used with this?

I think it'd be neat(and it's something I want to investigate) if we could make a docker image that includes prometheus/grafana and hooks up easily with the other domjudge docker containers to monitor the entire system.

webapp/src/Controller/MetricsController.php
nickygerritsen: Api reader is what kattis/cds uses so it can see everything

ubergeek42: ok, that seems more reasonable to use then. I'll make that change a bit later(and fix up composer.lock)

thijskh:
One final (?) question: the metrics are now under `/metrics` (same level als `/api`, `/jury`, `/team` etc). Do we want this top-level or is `/api/metrics/` a "better" base?

meisterT:
Sounds reasonable to me to put them under the api.

ubergeek42:
@thijskh yeah, if it looks good to you feel free to merge.

PR #746, opened 2020-04-25
Add source map for nvd3 since it is referenced by the JS file and Chr…

meisterT:
…ome complains.


meisterT:
> Make it a symlink instead, just like the .js itself

Thanks, I didn't know that this was a symlink.

Done.

PR #748, opened 2020-04-25
Allow to "delete" testcases even with associated judgings.

### LINKED ISSUE #189: Allow deleting of testcase with associated judgings

meisterT:
This only marks testcases as deleted but doesn't actually delete them.
Since the association with the problem is removed they will not be taken
into account with future (re)judgings of the problem.

All testcases that are ranked after the deleted one are moved up one
rank.

In a follow up change we can actually delete testcases which don't have
any associated judgings.

With commit d873b7c we added a warning to submissions which were judged
either in a different testcase order or with a now deleted testcase.

We may want to offer to rejudge all of those cases together with a
single click but I didn't want to bundle to much in this commit.

Fixes #189.

### DIFF
--- a/webapp/src/Entity/Testcase.php
+++ b/webapp/src/Entity/Testcase.php
@@ -52,7 +52,7 @@ class Testcase
      * @var int
      * @ORM\Column(type="integer", name="probid", length=4,
      *     options={"comment"="Corresponding problem ID", "unsigned"=true},
-     *     nullable=false)
+     *     nullable=true)
      */
     private $probid;
 
@@ -103,6 +103,16 @@ class Testcase
      * @Serializer\Exclude()
      */
     private $sample = false;
+
+    /**
+     * @var boolean
+     * @ORM\Column(type="boolean", name="deleted",
+     *     options={"comment"="Deleted testcases are kept for referential integrity.",
+     *              "default"="0"},
+     *     nullable=false)
+     * @Serializer\Exclude()
+     */
+    private $deleted = false;
 
     /**
      * @ORM\OneToMany(targetEntity="JudgingRun", mappedBy="testcase")
@@ -343,6 +353,20 @@ class Testcase
     }
 
     /**
+     * Set deleted
+     *
+     * @param boolean $deleted
+     *
+     * @return Testcase
+     */
+    public function setDeleted(bool $deleted)
+    {
+        $this->deleted = $deleted;
+
+        return $this;
+    }
+
+    /**
      * Get sample
      *
      * @return boolean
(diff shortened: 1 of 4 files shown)


webapp/src/Entity/Testcase.php
```diff
@@ -52,7 +52,7 @@ class Testcase
      * @var int
      * @ORM\Column(type="integer", name="probid", length=4,
      *     options={"comment"="Corresponding problem ID", "unsigned"=true},
-     *     nullable=false)
+     *     nullable=true)
```
eldering, line 55: Just to check: this doesn't break references when fetching deleted testcases for old judgings?

webapp/src/Migrations/Version20200425120051.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        return 'Adds deleted boolean field to testcase.';
+    }
+
+    public function up(Schema $schema) : void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE testcase ADD deleted TINYINT(1) DEFAULT \'0\' NOT NULL COMMENT \'Deleted testcases are kept for referential integrity.\', CHANGE probid probid INT UNSIGNED DEFAULT NULL COMMENT \'Corresponding problem ID\', CHANGE orig_input_filename orig_input_filename VARCHAR(255) DEFAULT NULL COMMENT \'Original basename of the input file.\'');
```
eldering, line 25: For readability might be good to break over multiple lines, and (nitpick) also change `CHANGE` into `MODIFY` (dropping duplicate column name, to indicate that the column name didn't change).
Also why did `orig_input_filename` change?

webapp/src/Entity/Testcase.php
meisterT: You mean on any of the submissions overview page or the submission page? I have tested both and added the warning (see commit description) to the latter.

webapp/src/Migrations/Version20200425120051.php
meisterT: That's probably a spurious update generated from the script. I'll fix this and the readability tomorrow.

`CHANGE` vs `MODIFY`: I can change this, but since this is autogenerated, do we really care?

eldering: The `MODIFY` is really a nitpick: it doesn't change anything but readability. Feel free to leave as is.

meisterT: Done, let's see if gitlab likes it.

PR #750, opened 2020-04-25
Get rid of submitdir.

meisterT:
This is a follow-up on #711.

Note that this replaces the two old methods of diffing with a new one.
Previously we had:
- diffing with pecl-text-xdiff for which there is no Debian package, so
  it was cumbersome to install, or
- the fallback to actual diff on the filesystem of the files in
  submitdir

Now we use https://github.com/sebastianbergmann/diff which was in our
anyway in our transitive dependencies (now made it explicit).


symfony.lock
```diff
@@ -228,6 +228,9 @@
     "phar-io/version": {
         "version": "2.0.1"
     },
+    "php": {
+        "version": "7.3"
```
nickygerritsen, line 232: I had this as well. Do we know what it does?

eldering, line 232: Is this supposed to be here?

meisterT: :man_shrugging: This was created by composer, I deleted it now.

PR #751, opened 2020-04-25
Drop smart renaming for Java public classes.

### LINKED ISSUE #735: Compile script java_java_detect broken on rename+rejudge
When you submit a Java source that requires a file rename to match a publicly declared class (for example https://github.com/DOMjudge/domjudge/blob/master/tests/test-classname.java), then the rename is done and then the *renamed* entry point is stored with the submission. When the submission is rejudged, this renamed entry point is passed to the compile script. This will not perform a rename anymore, but simply error out.

The simple fix of trying to rename again in the compile script is not the correct one, since if we pass this to any downstream/shadow system, then this will still fail. Effectively, we changed the submission and should store the renamed file. Ugh.

eldering:
This creates problems when rejudging or forwarding to
shadow CCS's because the entry point detected doesn't
match the source file: it was renamed during compilation
but this is not stored in the submission.

Closes #735


sql/files/defaultdata/java_javac_detect/run
```diff
@@ -76,13 +63,8 @@ if [ -z "$MAINCLASS" ]; then
 
 	# Report the entry point, so it can be saved, e.g. for later replay:
 	echo "Info: detected entry_point: $MAINCLASS"
-else
-	cat $TMPFILE
-	rm -f $TMPFILE
 fi
```
meisterT, line 66: Should we add an `else` here so that we only run `DetectMain` once?

So that we run it either as detection or as confirmation of the given entry point.

meisterT: We probably want to align the Kotlin script then.

eldering: Changed in latest version.

PR #754, opened 2020-04-26
Webnotification about the undelivered balloons.

mvr320:


webapp/src/Service/DOMJudgeService.php
```diff
@@ -409,11 +410,24 @@ public function getUpdates(): array
             ->setParameter('status', 'open')
             ->getQuery()->getResult();
 
+        $balloons = $this->em->createQueryBuilder()
+            ->select('b.balloonid', 't.name', 't.room', 'p.name', 'cp.color')
```
thijskh, line 414: Not all this is used in the output now, is it?

mvr320:
I tried to give the text the color of the balloon, this seemed to be not intended in the specs. It would be nice to have the badge the correct color to use the DJ logo with the correct color of which balloon to deliver.

webapp/public/js/domjudge.js
```diff
@@ -599,6 +600,20 @@ function updateMenuInternalErrors(data)
     }
 }
 
+function updateMenuBalloons(data)
```
thijskh, line 603: It sends a notification but does not update the menu. Do we want to add a badge to the menu with the number of undelivered balloons? Like clarifications?

webapp/src/Service/DOMJudgeService.php
```diff
@@ -409,11 +410,24 @@ public function getUpdates(): array
...
+            ->from(Balloon::class, 'b')
+            ->leftJoin('b.submission', 's')
+            ->leftJoin('s.problem', 'p')
+            ->leftJoin('s.contest', 'co')
+            ->leftJoin('p.contest_problems', 'cp', Join::WITH, 'co.cid = cp.contest AND p.probid = cp.problem')
+            ->leftJoin('s.team', 't')
+            ->andWhere('co.cid = :cid')
+            ->setParameter(':cid', $contest->getCid())
```
thijskh, line 422: Just a remark: this is now limited to the current contest which seems fine because it mimicks what Clarifications menu and the balloon page itself do.

webapp/public/js/domjudge.js
```diff
@@ -510,17 +510,18 @@ function toggleRefresh($url, $after, usingAjax) {
 
 function updateMenuAlerts()
 {
-    $.ajax({
-        url: $('#menuDefault').data('update-url')
+	$.ajax({
```
thijskh, line 513: Here and in a few other places the indentation has changed

webapp/public/js/domjudge.js
mvr320: I was indeed planning that in a seperate commit as it was not discussed to do this yet. But my intentions yesterday were to indeed introduce such a method

webapp/public/js/domjudge.js
```diff
@@ -599,6 +602,21 @@ function updateMenuInternalErrors(data)
...
+{
+    var num = data.length;
+    if ( num != 0 ) {
+        for (var i=0; i<num; i++) {
+            var text = (data[i].room !== null) ? data[i].room+': ' : '';
+            text += data[i].pname + ' ' + data[i].name;
+			sendNotification('New balloon:',
+                 {'tag': 'ball_' + data[i].baloonid,
+                  'link': domjudge_base_url + '/jury/balloons',
+                  'body':  text});
```
nickygerritsen, line 615: Nitpick: there is a space too much after the :

mvr320: I'll close this merge request, and fix it in another branch

PR #755, opened 2020-04-26
Limit json updates endpoint

mvr320:


webapp/src/Service/DOMJudgeService.php
```diff
@@ -378,39 +378,48 @@ public function getUpdates(): array
...
-                ->andWhere('clar.sender is not null')
-                ->andWhere('clar.answered = 0')
-                ->setParameter('contest', $contest)
-                ->getQuery()->getResult();
-        }
+        $judgehosts = [];
+        $rejudgings = [];
+        $internal_errors = [];
+
+        if ($this->authorizationChecker->isGranted('ROLE_ADMIN') or $this->authorizationChecker->isGranted('ROLE_JURY')) {
```
nickygerritsen, line 385: admin also has the jury role so you only have to check that. Also you can use `$this->checkRole('jury')`

webapp/public/js/domjudge.js
```diff
@@ -68,15 +68,17 @@ function sendNotification(title, options = {})
...
 
     var link = null;
     if ( typeof options.link !== 'undefined' ) {
         link = options.link;
         delete options.link;
     }
-    options['icon'] = domjudge_base_url + '/apple-touch-icon.png';
 
+    options['icon'] = domjudge_base_url + '/apple-touch-icon.png';
+
```
eldering, line 81: Please be cautious introducing whitespace-only changes.

mvr320: I think I fixed this

mvr320:
> Added one comment. Also your commit from the previous PR is also here, probably we should only merge this after that one.

I think its fixed,

webapp/public/js/domjudge.js
nickygerritsen: The newline seems to have moved, the rest is now ok

PR #756, opened 2020-04-26
Change shadow import to read data from a configuration file

nickygerritsen:
Tested it by shadowing the systestA contest. Tested both the URL and local file flows, and they seem to do what they should do.


etc/shadowing.yaml.dist
```diff
@@ -0,0 +1,36 @@
+# This file is an example file for using DOMjudge as a shadow system.
+# The `import/import-event-feed` command uses the file etc/shadowing.yaml to
+# read configuration options for shadowing another system following the
+# Contest API specification (https://clics.ecs.baylor.edu/index.php?title=Contest_API).
+
+# THe root keys are identifiers used to pass to `import/import-event-feed`
```
meisterT, line 6: nit: s/THe/The/

meisterT: Also "used to pass to `import/import-event-feed`" --> "used to pass to `import/import-event-feed`, e.g. use `import/import-event-feed exampleA` to import the first contest.

"

webapp/src/Command/ImportEventFeedCommand.php
```diff
@@ -263,30 +263,33 @@ protected function execute(InputInterface $input, OutputInterface $output)
         pcntl_signal(SIGTERM, [$this, 'stopCommand']);
         pcntl_signal(SIGINT, [$this, 'stopCommand']);
 
+        if (!$this->readConfig($input->getArgument('contest'))) {
```
meisterT, line 266: I can't attach the comment in unchanged lines, but please add a comment what the return value means.

Also verbosity is spelled wrong above.

webapp/src/Command/ImportEventFeedCommand.php
```diff
@@ -263,30 +263,33 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            return 1;
```
meisterT, line 267: Consider using constants instead of 0/1

webapp/src/Command/ImportEventFeedCommand.php
```diff
@@ -380,17 +367,169 @@ public function stopCommand()
...
     {
-        $this->logger->info('Importing from local file %s', [ $path ]);
+        $this->logger->info('Importing from local file %s', [ $this->feedFile ]);
 
-        $file = fopen($path, 'r');
+        // First, check if the external ID of the contest in DOMjudge matches
+        // the ID of the external contest.
+        $file = fopen($this->feedFile, 'r');
+        $buffer = '';
+        while (!feof($file) || !empty($buffer)) {
```
meisterT, line 495: Consider deduplicating this code with the one below.

etc/shadowing.yaml.dist
```diff
@@ -0,0 +1,37 @@
+# This file is an example file for using DOMjudge as a shadow system.
```
eldering, line 1: I'd propose: `This is an example configuration file for using DOMjudge as a shadow system.`

etc/shadowing.yaml.dist
```diff
@@ -0,0 +1,37 @@
...
+    # By default, DOMjudge checks if the external ID of the contest in DOMjudge
+    # matches with the external system to prevent importing to the wrong
+    # contest. If you do want to import into such a contest, enable this
+    # option. Optional.
+    # allow-external-id-mismatch: true
+    # URL to import from.
+    feed-url: https://example.com/api/contests/exampleA/event-feed
+    # Username to use for authenticating to the feed. Optional.
+    username: user
+    # Passowrd to use for authenticating to the feed. Optional.
```
eldering, line 25: Typo `Passowrd`

etc/shadowing.yaml.dist
```diff
@@ -0,0 +1,37 @@
...
+    # allow-insecure-ssl: true
+exampleB:
+    id: 3
+    # Instead of setting feed-url, you can also read events from a file on disk.
+    # If this is set, `feed-url`, `username` and `password` fields are ignored.
+    feed-file: /home/user/sample-feed/event-feed.ndjson
+    # The base path to use for relative URL's from the feed. Currently only used
+    # for submission files with relative URL's. This field is required if
+    # `feed-file` is set.
+    base-path: /home/user/sample-feed/files/
```
eldering, line 37: Maybe give an example of where the  script would look for submission files then.

nickygerritsen: Well that depends on whats in the feed right? But I can add something

eldering: Shouldn't `x/y` be the other way around? But what I was hoping for is a concrete example with a submission source file.

nickygerritsen: So put in an example json?

PR #760, opened 2020-04-26
Remove duplicate rejudging code.

meisterT:

### DIFF
--- a/webapp/src/Controller/Jury/RejudgingController.php
+++ b/webapp/src/Controller/Jury/RejudgingController.php
@@ -55,7 +55,7 @@ class RejudgingController extends BaseController
     /**
      * @var RejudgingService
      */
-    protected $rejudging;
+    protected $rejudgingService;
 
     /**
      * @var RouterInterface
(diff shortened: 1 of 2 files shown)


webapp/src/Controller/Jury/RejudgingController.php
```diff
@@ -201,6 +187,7 @@ public function indexAction(Request $request)
      */
     public function viewAction(
         Request $request,
+        RejudgingService $rejudgingService,
```
nickygerritsen, line 190: Why not use $this->rejudgingService?

meisterT: I probably did this because of the submissionService below. What is the guiding principle here? When to use the function arg vs adding it in the constructor of the controller?

nickygerritsen: I mostly put it in the constructor if > 1 method needs it.

meisterT: Done.

PR #762, opened 2020-04-26
Move common streaming code to a utils method.

meisterT:


meisterT:
Yes, that what I think as well, but wanted to move slowly and not break (too many) things.

PR #763, opened 2020-04-26
Show external ID and linkt to external CCS on shadow difference page

nickygerritsen:
image

The comment thing we need to think about how to do this and where to store it so I only did the first part of the issue now.

### DIFF
--- a/webapp/templates/jury/partials/submission_list.html.twig
+++ b/webapp/templates/jury/partials/submission_list.html.twig
@@ -9,8 +9,10 @@
 {% endif %}
 
 {% set rowSpan = 1 %}
+{% set tdExtraClass = '' %}
 {% if showExternalResult and showExternalTestcases %}
     {% set rowSpan = 2 %}
+    {% set tdExtraClass = ' thick-border' %}
 {% endif %}
 
 {% if submissions is empty %}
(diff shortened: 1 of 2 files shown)


nickygerritsen:
I'm not sure yet if showing "External" twice is very useful. Better UX is welcome.

webapp/templates/jury/partials/submission_list.html.twig
```diff
@@ -191,13 +202,28 @@
                 {% else %}
                     {% set externalJudgement = submission.externalJudgements.first %}
                 {% endif %}
+                {% if external_ccs_submission_url is not empty %}
+                    {% set externalSubmissionUrl = submission | externalCcsUrl %}
+                {% else %}
+                    {% set sexternalSubmissionUrl = null %}
```
meisterT, line 208: Typo: set **s**external

webapp/templates/jury/partials/submission_list.html.twig
```diff
@@ -191,13 +202,28 @@
...
                 <td>
-                    <a href="{{ link }}">
+                    <a {% if externalSubmissionUrl is not null %}href="{{ externalSubmissionUrl }}"{% endif %}>
                         External
                     </a>
                 </td>
                 <td>
-                    <a href="{{ link }}">
+                    <a {% if externalSubmissionUrl is not null %}href="{{ externalSubmissionUrl }}"{% endif %}>
+                        s{{ submission.externalid }}
```
meisterT, line 217: remove s - it adds more confusion, if external IDs are not numeric but alphanumeric

webapp/templates/jury/partials/submission_list.html.twig
nickygerritsen: I lol'ed

PR #766, opened 2020-04-26
Improve create-cgroups.service file and documentation.

eldering:
At the risk of this getting rejected a description ;-)

Document how the `create-cgroups.service` snippet can get installed as a systemd service. Also mark it as wanted by the multiuser target, so it gets started automaticaly.


eldering:
> I think the installation is already handled by the Makefile, so does it in fact need to be documented separately?

Ah, I wasn't aware of that in `judge/Makefile`. Probably makes sense to move this to `judge/` too then. I'll change the documentation to mention both `domjudge-judgehost` and `create-cgroups` services.

> The service file is referenced by the judgedaemon service file that we ship. So do you need to add the target?

I think it still makes sense: if you run the judgedaemon manually, you still might want to run this service.

thijskh:
Side note, it might not be the best to install services under $libdir. But that was already done in the old code aswell, so this PR does not matter for that. I'm checking what's the best approach.

PR #772, opened 2020-04-26
Deduplicate twig data creation for problems action.

meisterT:

### DIFF
--- a/webapp/src/Controller/PublicController.php
+++ b/webapp/src/Controller/PublicController.php
@@ -181,42 +181,8 @@ class PublicController extends BaseController
      */
     public function problemsAction()
     {
-        $contest            = $this->dj->getCurrentContest(-1);
-        $showLimits         = (bool)$this->config->get('show_limits_on_team_page');
-        $defaultMemoryLimit = (int)$this->config->get('memory_limit');
-        $timeFactorDiffers  = false;
-        if ($showLimits) {
-            $timeFactorDiffers = $this->em->createQueryBuilder()
-                    ->from(Language::class, 'l')
-                    ->select('COUNT(l)')
-                    ->andWhere('l.allowSubmit = true')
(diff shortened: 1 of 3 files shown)


webapp/src/Controller/PublicController.php
```diff
@@ -181,42 +181,8 @@ public function teamAction(Request $request, int $teamId)
...
-        }
-
-        return $this->render('public/problems.html.twig', [
-            'problems' => $problems,
-            'showLimits' => $showLimits,
-            'defaultMemoryLimit' => $defaultMemoryLimit,
-            'timeFactorDiffers' => $timeFactorDiffers,
-        ]);
+        return $this->render('public/problems.html.twig',
+            $this->dj->getTwigDataForProblemsAction(-1));
```
eldering, line 185: This `-1` argument seems a bit hacky. Can't we make `$teamid` an optional argument, or use `null` instead?

webapp/src/Service/DOMJudgeService.php
```diff
@@ -832,4 +832,43 @@ public function getContestStats(Contest $contest): array
             ->getSingleScalarResult();
         return $stats;
     }
+
+    public function getTwigDataForProblemsAction(int $userid): array {
```
eldering, line 836: `$userid` should be understood as `$teamid`, right?

meisterT: thanks, done

webapp/src/Controller/PublicController.php
meisterT: We might want to also clean up getCurrentContest(s), but I think it's separate from this cleanup.

See
https://github.com/DOMjudge/domjudge/blob/master/webapp/src/Service/DOMJudgeService.php#L180-L181 for the `-1` case, but `null` is also allowed - I am not sure if we use that in some places.

PR #777, opened 2020-05-01
Always create a rejudging on rejudge.

### LINKED ISSUE #732: Replace creating non-rejudging rejudge with auto-applying rejudging
Currently there is an option to rejudge a submission (or a set of them) without creating a full blown rejudging.

This complicates the backend code. If we change this to actually always create a full blown rejudging but automatically apply it, this might simplify things. We can either apply each submission after it finished or apply everything at the end, we should decide on one of these.

meisterT:
Fixes #732.

### DIFF
--- a/webapp/src/Controller/API/JudgehostController.php
+++ b/webapp/src/Controller/API/JudgehostController.php
@@ -614,8 +614,10 @@ class JudgehostController extends AbstractFOSRestController
                     $this->dj->auditlog('judging', $judgingId, 'judged',
                                         'compiler-error', $hostname, $contestId);
 
+                    $this->maybeUpdateActiveJudging($judging);
+                    $this->em->flush();
                     if (!$this->config->get('verification_required') &&
-                        $judging->getRejudgingid() === null) {
+                        $judging->getValid()) {
                         $this->eventLogService->log('judging', $judgingId,
                                                     EventLogService::ACTION_UPDATE, $contestId);
                     }
@@ -1044,10 +1046,11 @@ class JudgehostController extends AbstractFOSRestController
                 ->setOutputSystem(base64_decode($outputSystem))
                 ->setMetadata(base64_decode($metadata));
 
+            $this->maybeUpdateActiveJudging($judging);
             $this->em->persist($judgingRun);
             $this->em->flush();
 
-            if ($judging->getRejudgingid() === null) {
+            if ($judging->getValid()) {
                 $this->eventLogService->log('judging_run', $judgingRun->getRunid(),
                                             EventLogService::ACTION_CREATE, $judging->getCid());
             }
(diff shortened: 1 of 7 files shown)


webapp/src/Controller/API/JudgehostController.php
```diff
@@ -614,8 +614,16 @@ public function updateJudgingAction(Request $request, string $hostname, int $jud
                     $this->dj->auditlog('judging', $judgingId, 'judged',
                                         'compiler-error', $hostname, $contestId);
 
+                    if ($judging->getRejudgingid() !== null && $judging->getRejudging()->getAutoApply()) {
+                        $judging->getSubmission()->setRejudging(null);
+                        foreach ($judging->getSubmission()->getJudgings() as $j) {
+                            $j->setValid(false);
+                        }
+                        $judging->setValid(true);
+                        $this->em->flush();
```
eldering, line 623: I think it makes sense to abstract this block of code away in a separate function called `setActiveJudging` or something like that, not sure if as a method on a submission or judging object. That also removes th same duplicate below.

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -1044,10 +1052,18 @@ private function addSingleJudgingRun(
...
+                    $j->setValid(false);
+                }
+                $judging->setValid(true);
+            }
+
             $this->em->persist($judgingRun);
             $this->em->flush();
 
-            if ($judging->getRejudgingid() === null) {
+            if ($judging->getValid()) {
```
eldering, line 1066: Just wondering why we don't have the  `!$this->config->get('verification_required') &&` clause here as above.

webapp/src/Controller/Jury/RejudgingController.php
```diff
@@ -148,6 +149,7 @@ public function indexAction(Request $request)
             $rejudgingdata['endtime']['value']   = Utils::printtime($rejudging->getEndtime(), $timeFormat);
             $rejudgingdata['status']['value']    = $status;
 
+            dump([$rejudging->getRejudgingid(), $rejudging->getEndtime()]);
```
eldering, line 152: Left-over debug.

webapp/src/Service/RejudgingService.php
```diff
@@ -361,11 +332,19 @@ public function calculateTodo(Rejudging $rejudging)
             ->select('COUNT(j)')
             ->andWhere('j.rejudging = :rejudging')
             ->andWhere('j.endtime IS NOT NULL')
+            ->andWhere('j.valid = 0')
```
eldering, line 335: This is only relevant for auto apply rejudgings, right? I don't quite understand it's impact. Might be helpful to add a comment.

meisterT: Jep, added a comment.

webapp/src/Controller/Jury/RejudgingController.php
meisterT: Done.

webapp/src/Controller/API/JudgehostController.php
meisterT: Because one is judging the other is judging_run.

webapp/src/Controller/API/JudgehostController.php
meisterT: Done.

PR #779, opened 2020-05-02
Add option to allow teams to download their own submission. Fixes #714.

### LINKED ISSUE #714: Add a download submission button for teams as an option
Understanding the security reasons why this is not already added. It would be great if there was an option to allow submission downloads

nickygerritsen:
If the option is enabled, this looks like this:
image

Note that the popup is only shown currently for submissions that have a valid judging and are submitted before contest end and thus the button also only appears then. If we want to change this I think it's better to do that in a separate PR, but I also think the current way is fine.

### DIFF
--- a/webapp/src/Controller/Team/SubmissionController.php
+++ b/webapp/src/Controller/Team/SubmissionController.php
@@ -5,18 +5,22 @@ namespace App\Controller\Team;
 use App\Controller\BaseController;
 use App\Entity\Judging;
 use App\Entity\Problem;
+use App\Entity\Submission;
 use App\Entity\Testcase;
 use App\Form\Type\SubmitProblemType;
 use App\Service\ConfigurationService;
 use App\Service\DOMJudgeService;
 use App\Service\SubmissionService;
 use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\NonUniqueResultException;
 use Doctrine\ORM\Query\Expr\Join;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
 use Symfony\Component\Form\FormFactoryInterface;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\Routing\Annotation\Route;
 use Symfony\Component\Validator\Constraints\Language;
 
@@ -72,7 +76,7 @@ class SubmissionController extends BaseController
     /**
      * @Route("/submit", name="team_submit")
      * @param Request $request
-     * @return \Symfony\Component\HttpFoundation\Response
+     * @return Response
      * @throws \Exception
      */
     public function createAction(Request $request)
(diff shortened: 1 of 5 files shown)


thijskh:
@mvr320

webapp/src/Controller/Team/SubmissionController.php
```diff
@@ -135,14 +139,15 @@ public function createAction(Request $request)
...
-     * @throws \Doctrine\ORM\NonUniqueResultException
+     * @return Response
+     * @throws NonUniqueResultException
      */
     public function viewAction(Request $request, int $submitId)
     {
         $verificationRequired = (bool)$this->config->get('verification_required');;
         $showCompile      = $this->config->get('show_compile');
         $showSampleOutput = $this->config->get('show_sample_output');
+        $allowDownload    = (bool)$this->config->get('allow_team_submission_download');
```
meisterT, line 150: what happens if that field does not exist?

webapp/src/Controller/Team/SubmissionController.php
```diff
@@ -201,4 +207,45 @@ public function viewAction(Request $request, int $submitId)
...
+            ->join('s.language', 'l')
+            ->join('s.files', 'f')
+            ->select('s, f')
+            ->andWhere('j.submitid = :submitId')
+            ->andWhere('j.valid = 1')
+            ->andWhere('s.team = :team')
+            ->setParameter(':submitId', $submitId)
+            ->setParameter(':team', $team)
+            ->getQuery()
+            ->getOneOrNullResult();
```
eldering, line 242: Why all these extra joins and conditions? I don't see the need to join on judgings, language or even problem and contest_problem.

webapp/src/Controller/Team/SubmissionController.php
```diff
@@ -201,4 +207,45 @@ public function viewAction(Request $request, int $submitId)
...
+        $submission = $this->em->createQueryBuilder()
+            ->from(Submission::class, 's')
+            ->join('s.judgings', 'j')
+            ->join('s.contest_problem', 'cp')
+            ->join('cp.problem', 'p')
+            ->join('s.language', 'l')
+            ->join('s.files', 'f')
+            ->select('s, f')
+            ->andWhere('j.submitid = :submitId')
+            ->andWhere('j.valid = 1')
```
meisterT, line 237: why do we need a valid judging for this?

webapp/src/Controller/Team/SubmissionController.php
nickygerritsen: Then it returns false since that is in the yaml

webapp/src/Controller/Team/SubmissionController.php
nickygerritsen: We don’t but we show it in the pop-up that does. I can remove this here, then you can access it if you know the url and later possibly update the UI

webapp/src/Controller/Team/SubmissionController.php
nickygerritsen: We don’t need them. Was a copy paste. Will remove it

webapp/src/Controller/Team/SubmissionController.php
eldering: Indeed, so also my comment above: I see no need to restrict this to anything more than the right team and submission.

nickygerritsen: Agreed

webapp/src/Controller/Team/SubmissionController.php
meisterT: The yaml is only imported if people do a db upgrade, IIRC? If they only update the code it doesn't exist in the DB, right?

nickygerritsen: You are not right 🤣. We always read the yaml from the yaml. The db only has values if they differ from the yaml (I.e. if you change them in the UI)

meisterT: Even better, where do we read the yaml?

nickygerritsen: In the configuration service.  ie in the function that gets called here

nickygerritsen:
As discussed we will do the docs in a separate PR that will automate adding a configuration reference to the docs

PR #780, opened 2020-05-02
Update password generate export to reflect accounts.tsv. Fixes #694.

### LINKED ISSUE #694: Review generation of userdata.tsv
The “manage team passwords” feature now generates a file in the format `userdata.tsv`. This seems to be no longer a part of the CLICS CCS spec.

Likely it is better to make the output be `accounts.tsv` and its corresponding format.

nickygerritsen:
https://clics.ecs.baylor.edu/index.php/Contest_Control_System_Requirements#accounts.tsv specifies `For accounts of type team username is on the form "team-nnn" where "nnn" is the zero padded team number as given in teams.tsv.` but I think it would be strange to do that in this export since then you loose the original username.

### DIFF
--- a/webapp/src/Twig/TwigExtension.php
+++ b/webapp/src/Twig/TwigExtension.php
@@ -126,6 +126,7 @@ class TwigExtension extends AbstractExtension implements GlobalsInterface
             new TwigFilter('descriptionExpand', [$this, 'descriptionExpand'], ['is_safe' => ['html']]),
             new TwigFilter('wrapUnquoted', [$this, 'wrapUnquoted']),
             new TwigFilter('hexColorToRGBA', [$this, 'hexColorToRGBA']),
+            new TwigFilter('tsvField', [$this, 'toTsvField']),
         ];
     }
 
@@ -964,5 +965,21 @@ EOF;
         }
 
         return $text;
+    }
+
+    /**
+     * Convert the given string to a field that is safe to use in a TSV file
+     *
+     * @param string $field
+     *
+     * @return string
+     */
+    public function toTsvField(string $field)
+    {
+        return str_replace(
+            ["\\",   "\t",  "\n",  "\r"],
+            ["\\\\", "\\t", "\\n", "\\r"],
+            $field
+        );
     }
 }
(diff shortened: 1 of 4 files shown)


webapp/templates/jury/tsv/accounts.tsv.twig
```diff
@@ -0,0 +1,4 @@
+accounts	1
+{% for row in data %}
+{{ row.type|replace({'\t':' '}) }}	{{ row.fullname|replace({'\t':' '}) }}	{{ row.username|replace({'\t':' '}) }}	{{ row.password|replace({'\t':' '}) }}
```
eldering, line 3: This is modifying data in an unclear way: if indeed a field contained a tab, then this may lead to weird and difficult to trace problems. It seems better to me to replace that tab by an escaped tab like a literal `\t`. That way, it's at least clearer what's going on.

thijskh: (Note that this just mirrors what the previous export did, but the point is valid.)

I propose that we fix it by following this standard "convention" which is wat Jaap says extended to other problematic symbols. Put that into a twig filter and apply it to our tsv exports.
https://en.wikipedia.org/wiki/Tab-separated_values#Conventions_for_lossless_conversion_to_TSV

And I think it'd be great if the CLICS spec would mention explicitly how this is expected to be handled for all tsv's (and obviously preferedly with the convention as specified above)

thijskh:
I think that if someone is interested in WF CLICS compatibility, they need to have their usernames match the format, not covert the usernames in the output.

meisterT:
I agree with Thijs. If you care about WF CLICS compatibility, then you took care of this already when creating the users.

nickygerritsen:
Ok I will update the tab stuff and then we are good to go then

webapp/src/Twig/TwigExtension.php
```diff
@@ -965,4 +966,20 @@ public function hexColorToRGBA(string $text, float $opacity = 1): string
...
+     * Convert the given string to a field that is safe to use in a TSV file
+     *
+     * @param string $field
+     *
+     * @return string
+     */
+    public function toTsvField(string $field)
+    {
+        return str_replace(
+            ["\\", "\t", "\n", "\r"],
```
meisterT, line 980: Do you use double vs single ticks (below) intentionally? Also you might want to align the list elements.

PR #781, opened 2020-05-02
Add simple option to switch between multiple doc versions.

meisterT:
This assumes that the docs for the multiple versions are arranged in a
structure like:
  <SERVER_URL>/docs/manual/<VERSION>/index.html
e.g.
  https://www.domjudge.org/docs/manual/7.2/
  https://www.domjudge.org/docs/manual/7.3/
  https://www.domjudge.org/docs/manual/master/
and that
  <SERVER_URL>/docs/manual/versions.json
contains a version list in JSON format, e.g.
  ["7.2", "7.3", "master"]
.

If the versions.json file is not readable, then the version div remains
unchanged and is not transformed to a select dropdown.

### DIFF
--- a/doc/manual/_static/js/versionswitcher.js
+++ b/doc/manual/_static/js/versionswitcher.js
@@ -1,0 +1,37 @@
+function versionChange(selectList) {
+    url = window.location.href;
+    for (var i = 0; i < 2; i++) {
+        var to = url.lastIndexOf('/');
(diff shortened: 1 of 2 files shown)


meisterT:
Where is the harm of this being in the main repo?

nickygerritsen:
There is no harm but for the main repo there is also no real use right? It’s mainly for the site

Issue #366, opened 2018-04-19
Pin scoreboard header and footer


eldering:
In various cases you might want to view the header and/or footer of the scoreboard while scrolling through a long scoreboard, as suggested by the ICPC WF 2018 jury. We could look into (optionally) pinning the header and/or footer of the scoreboard table, allowing to scroll through the main rows.

mvr320:
I'm currently trying to get a fix for this, atleast half of the problem is fixable

thijskh:
Should somehow take sortorders into account...

nickygerritsen:
The header is fixed in #782 (pun intended). We discussed that pinning the footer is less useful so will close this issue for now. If anyone has a real use case for pinning the footer please let us know.

PR #782, opened 2020-05-02
Fix header scoreboard

mvr320:
If this is acceptable I'll make this 1 commit.

Currently this only makes sense on the static page as top: 0 doesnt really work when the DOMjudge menu is there. But before looking at that I would like to know if this doesn't contain major flaws.

### DIFF
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -659,3 +659,16 @@ function initializeAjaxModals()
 		});
 	});
 }
+
+function pinScoreheader()
+{
+	var static_in_url = new URL(window.location.toString().toLowerCase()).searchParams.get("static")
+	var static_scoreboard = static_in_url==="true" || static_in_url==='1';
+	if(!static_scoreboard){
+		$('.scoreheader th').css('top', $('.fixed-top').css('height'));
+        var resizeObserver = new ResizeObserver(() => {
+            $('.scoreheader th').css('top', $('.fixed-top').css('height'));
+        });
+        resizeObserver.observe($('.fixed-top')[0]);
+	}
+}
--- a/webapp/public/style_domjudge.css
+++ b/webapp/public/style_domjudge.css
@@ -1,6 +1,8 @@
 body {
+    /** Specify the BG color in one place. **/
+    --background-color: white;
     color: black;
-    background-color: white;
+    background-color: var(--background-color);
     font-family: Roboto, sans-serif;
     padding-bottom: 4em;
 }
@@ -203,6 +205,13 @@ img.affiliation-logo {
 }
 .scoreheader th {
     text-align: center;
+    box-shadow: -1px 0px 0px 0px silver inset,
+                0px 2px 0px 0px black;
+    border: none;
+    background: var(--background-color);
+    position: sticky;
+    top: 0px;
+    z-index: 1;
 }
 
 .problempoints {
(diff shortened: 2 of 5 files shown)


mvr320:
If this is acceptable I'll make this 1 commit.
This seems to work for every scoreboard I could find.

webapp/templates/partials/scoreboard.html.twig
```diff
@@ -142,4 +142,23 @@
         Last Update: {{ null | printtime('%a %d %b %Y %T %Z') }}<br/>
         using <a href="https://www.domjudge.org/" target="_top">DOMjudge</a>
     </p>
+
+    <script>
```
thijskh, line 146: Could be in the extrafooter block?

mvr320:
> It seems to unfix again when the scoreboard refreshes.
> 
> I'm not sure about people wanting to fix the summary row for one specific sortorder and I think it's not very intuitve. Maybe it's just me, but is there a need to pin the summary at all? Isn't the header the essential thing to pin?

In https://github.com/DOMjudge/domjudge/issues/366 it was mentioned as an OR, I agree that in this form its not really useful (but the alternative needs more work so I first preferred to have a POC working). Will remove the feature and after suggestion by @meisterT will permanently pin the header which also fixes the refresh.

webapp/templates/partials/scoreboard.html.twig
mvr320: I think I have to copy the code to different files than (so double code) whereas this approach had only 1 place to make changes. But I will move the code.

webapp/public/js/domjudge.js
```diff
@@ -659,3 +659,15 @@ function initializeAjaxModals()
 		});
 	});
 }
+
+function pinScoreheader(){
+	var url_s = window.location;
+	console.log(("I'm here:", url_s));
```
meisterT, line 665: Stale debug log.

mvr320: Fixed it.

webapp/public/style_domjudge.css
```diff
@@ -203,6 +203,13 @@ img.affiliation-logo {
 }
 .scoreheader th {
     text-align: center;
+    box-shadow: -1px 0px 0px 0px silver inset,
+                0px 2px 0px 0px black;
+    border: none;
+    background: white;
```
meisterT, line 209: Does the background have to be white for this to work?

webapp/public/js/domjudge.js
```diff
@@ -659,3 +659,14 @@ function initializeAjaxModals()
 		});
 	});
 }
+
+function pinScoreheader(){
+	var sboard = new URL(window.location).searchParams.get("static");
```
meisterT, line 664: consider using a more descriptive variable name instead of `sboard`.

webapp/public/style_domjudge.css
mvr320: image
I need a background, but there is this: https://caniuse.com/#feat=css-variables, so the background is now equal to the bg of the body but explicit

webapp/public/js/domjudge.js
mvr320: I tried, I also found a mistake that when one uses capitals it fails so thats now also fixed I think

nickygerritsen:
I rebased this onto master so the build should be green.

PR #785, opened 2020-05-03
When the scoreboard is requested to be static, disable Ajax refreshes

thijskh:


nickygerritsen:
Any reason why we drop Ajax support for static scoreboards? I tested it with an externally hosted scoreboard and it worked...

thijskh:
I don’t understand how that can work? Where to send the xhr call to? Since the webserver will only serve static content?

nickygerritsen:
The call gets sent to the page itself. I then look in the html content (since indeed we only get that data) for some specific divs and replace them onto the current page

thijskh:
Alright. Then my fix is not necessary I guess

Issue #425, opened 2018-10-15
Move misc-tools to use API or Symfony


thijskh:
Several of the tools under `misc-tools/` set up a direct database connection to perform queries. It's probably better if these tools use the API instead. This avoids code duplication and breakage due to database changes.

Maybe it's not (yet) possible in every case to perform the desired actions via the API. Perhaps that should then be made possible.

Probably it's a good idea to have some kind of shared code that handles the API setup and calling?

- [x] `create_accounts.in`
- [x] `judging_statistics.in`
- [x] `save_sources2file.in`
- [x] `balloons.in`
- [x] `multi_rejudge.in`
- [x] `simulate_contest.in`
- [x] `eventdaemon.in`
- [x] `combined_scoreboard.in`
- [x] `restore_sources2db.in`

nickygerritsen:
This would also help with #424

thijskh:
Move to API or Symfony code, but in any case should no longer depend on legacy code base

meisterT:
For things like `create_accounts` I propose to just add better documentation how to use the existing API endpoint (https://github.com/DOMjudge/domjudge/blob/master/webapp/src/Controller/API/UserController.php#L208).

WDYT?

PR #787, opened 2020-05-03
Add balloons endpoint.

meisterT:
One step closer to #425 and replacing `balloons.in`.

### DIFF
--- a/webapp/src/Controller/API/BalloonController.php
+++ b/webapp/src/Controller/API/BalloonController.php
@@ -1,0 +1,63 @@
+<?php
+
+
+namespace App\Controller\API;
+
+
+use App\Entity\Contest;
+use App\Service\BalloonService;
+use FOS\RestBundle\Controller\Annotations as Rest;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
+use Swagger\Annotations as SWG;
+use Doctrine\ORM\QueryBuilder;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Intl\Exception\NotImplementedException;
+
+/**
+ * @Rest\Route("/api/v4/contests/{cid}/balloons", defaults={ "_format" = "json" })
+ * @Rest\Prefix("/api/contests/{cid}/balloons")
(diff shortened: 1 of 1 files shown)


webapp/src/Controller/API/BalloonController.php
```diff
@@ -0,0 +1,63 @@
...
+class BalloonController extends AbstractRestController
+{
+    /**
+     * Get all the balloons for this contest.
+     * @param Request $request
+     * @return array
+     * @Rest\Get("")
+     * @SWG\Response(
+     *     response="200",
+     *     description="Returns the ballons for this contest.",
```
eldering, line 36: Typo: `ballons`

meisterT: Done.

webapp/src/Controller/API/BalloonController.php
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * @Rest\Route("/api/v4/contests/{cid}/balloons", defaults={ "_format" = "json" })
+ * @Rest\Prefix("/api/contests/{cid}/balloons")
+ * @Rest\NamePrefix("balloons_")
+ * @SWG\Tag(name="Balloons")
+ * @SWG\Parameter(ref="#/parameters/cid")
+ * @SWG\Response(response="404", ref="#/definitions/NotFound")
+ * @SWG\Response(response="401", ref="#/definitions/Unauthorized")
+ * @Security("is_granted('ROLE_JURY') or is_granted('ROLE_API_READER') or is_granted('ROLE_BALLOON')")
```
nickygerritsen, line 25: I think at some point we should remove jury here and in other places, since jury had api reader automatically

thijskh: It helps to be explicit if you'd ever want to change the definition of roles.

nickygerritsen:
In that case we should add it to some endpoints that don’t have it. Also sometimes in code (“show more if we have role x”). I’ll check the code

webapp/src/Controller/API/BalloonController.php
nickygerritsen: In that case we should add it to some endpoints that don’t have it. Also sometimes in code (“show more if we have role x”). I’ll check the code.

PR #789, opened 2020-05-03
Also terminate user session on a role change of the user.

thijskh:
When a user's password was changed or the Enabled status flipped,
the user is automatically logged out. This was however not the
case for a change in roles. Meaning that e.g. revoking Jury rights
from a user may not be effective as long as they do not log out.
This fixes that so any change in user roles also causes the
session to be invalidated.

### DIFF
--- a/webapp/src/Entity/User.php
+++ b/webapp/src/Entity/User.php
@@ -168,18 +168,20 @@ class User implements UserInterface, EquatableInterface, \Serializable
 
     public function serialize()
     {
-        return serialize(array(
+        return serialize([
             $this->userid,
             $this->username,
             $this->password,
-        ));
+            $this->user_roles,
+        ]);
     }
     public function unserialize($serialized)
     {
         list(
             $this->userid,
             $this->username,
-            $this->password
+            $this->password,
+            $this->user_roles,
         ) = unserialize($serialized);
     }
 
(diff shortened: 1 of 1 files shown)


thijskh:
Think it would be even cleaner to just update the roles for the user when a change has been detected. But that seems not so easy in Symfony.

thijskh:
Maybe this item is a bit more complex after all and might require some more thinking or discussion.

- It's not nice to log out users when changing roles, esp when adding a new role or when the removed role does not affect what they're currently doing. Updating the roles would be a lot nicer (you'd still get an Access Denied if you're e.g. in the jury interface and are no longer jury).
- At the very least we could use array_diff instead so we only log someone out if he has roles removed from his set, not added. But does mean that when adding roles, the user in question still actively has to relogin to make use of it.
- Do we want to log out users when changing their password anyway?

meisterT:
I agree that it's not nice to log users out, note that also the unit tests fail.

Nicky, is there really no way to reload roles in Symfony without log out / back in again?

Logging out users on password change (except if it's the current user) is fair I would say.

nickygerritsen:
I will investigate

nickygerritsen:
See #790.

thijskh:
That solution is definitely better. Have updated the documentation on what happens when you change a password for a logged-in user.

PR #790, opened 2020-05-03
Add event listener that updates the user roles in the token when they change.

nickygerritsen:


webapp/src/EventListener/UpdateUserRolesListener.php
```diff
@@ -0,0 +1,59 @@
...
+        $roles = $token->getRoleNames();
+
+        // If the roles from the token differ from the roles of the user,
+        // update the token
+        if ($user instanceof UserInterface && $roles !== $user->getRoles()) {
+            $token = new UsernamePasswordToken(
+                $user,
+                null,
+                'main',
+                $user->getRoles()
```
thijskh, line 53: Would it be better to just getRoles() once and store it in a variable?

nickygerritsen: It only queries it once since it caches it. We save one function call basically but I can do that if you want

thijskh: Not really

PR #794, opened 2020-05-11
Register: only validate affiliation name uniqueness if "Add new affiliation" is selected

### LINKED ISSUE #793: Register: Affiliation name validated even if not adding new affiliation
During registration, if the user selects "Add new affiliation" and types an affiliation name, but then switches to "Use existing affiliation" or "No affiliation", the affiliation name is still validated for uniqueness.
If the affiliation name is already in use, registration fails and the user is redirected back to the registration page with no visible validation errors.

### Environment
DOMjudge version: master @ fae71ce04, 7.1.3

### Steps to reproduce
1. Enable self-registration.
2. Open the registration page.
3. Enter a username and team name.
4. Select "Add new affiliation".
5. For "Affiliation name", type the name of an existing affiliation.
6. Enter a password.
7. (optional) Register. Should redirect back to registration page with error "This affiliation name is already in use." Enter a password again.
8. Select "Use existing affiliation". The "Affiliation name" field is replaced by a dropdown with existing affiliations.
9. Select an existing affiliation.
10. Register.

### Expected behaviour
Account registered successfully.

### Actual behaviour
Redirected back to registration page, with no visible validation errors. Selecting "Add new affiliation" shows the error "This affiliation name is already in use."

### Fix
The affiliation name field only makes sense when "Add new affiliation" is selected, and should not be validated if a different option is selected.
The validation of uniqueness should be moved into the `$validateAffiliation` function in `configureOptions` and only run when "Add new affiliation" is selected, like the non-blank validation.

### Pull request
#794

tom93:
Also fix property path of violation (team_name => affiliationName).

While we are here, remove redundant property path from teamName validation (that violation error is in the current context).

Fixes #793.

### DIFF
--- a/webapp/src/Form/Type/UserRegistrationType.php
+++ b/webapp/src/Form/Type/UserRegistrationType.php
@@ -91,7 +91,6 @@ class UserRegistrationType extends AbstractType
                     new Callback(function ($teamName, ExecutionContext $context) {
                         if ($this->em->getRepository(Team::class)->findOneBy(['name' => $teamName])) {
                             $context->buildViolation('This team name is already in use.')
-                                ->atPath('teamName')
                                 ->addViolation();
                         }
                     }),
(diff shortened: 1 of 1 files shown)


nickygerritsen:
Looks good! Merged :)

PR #795, opened 2020-05-11
Use relative URL to registration form on the login page instead of absolute

tom93:
Some reverse proxies rewrite the "Host" header, which breaks absolute URLs. Switch to a relative URL like all the other links.

To verify, check the `href` attribute in the HTML source. For example on https://www.domjudge.org/demoweb/login it is
```
<a href="//www.domjudge.org/demoweb/register">Register now</a>.
```
(the other links don't include the domain)


nickygerritsen:
No idea why we used `url`...

PR #796, opened 2020-05-12
Always install python3-yaml, never python-yaml

tom93:
It is required by gen_conf_ref.py, which uses `python3` in the shebang line and so will always run on Python 3, never Python 2.

This is in contrast with the python-sphinx package, which may need to be Python 2 to match the rst2pdf package.

I also moved python3-yaml to the end to emphasise that python-sphinx and rst2pdf are related, is that OK? Should I do the same in the domjudge-packaging repository, in [docker-gitlabci/Dockerfile:10](https://github.com/DOMjudge/domjudge-packaging/blob/8fc6462b6e0f78d4de2112bcf28e33f87a853168/docker-gitlabci/Dockerfile#L10)?

### DIFF
--- a/doc/manual/develop.rst
+++ b/doc/manual/develop.rst
@@ -38,10 +38,10 @@ already listed under
 :ref:`submit client <submit_client_requirements>` requirements)::
 
   sudo apt install autoconf automake bats \
-    python-sphinx python-sphinx-rtd-theme rst2pdf python-yaml
+    python-sphinx python-sphinx-rtd-theme rst2pdf python3-yaml
 
 On Debian 11 and above, install
-``python3-sphinx python3-sphinx-rtd-theme python3-yaml rst2pdf`` instead.
+``python3-sphinx python3-sphinx-rtd-theme rst2pdf python3-yaml`` instead.
 
 When this software is present, bootstrapping can be done by running
 ``make dist``, which creates the ``configure`` script,
--- a/doc/manual/install-workstation.rst
+++ b/doc/manual/install-workstation.rst
@@ -66,11 +66,11 @@ When DOMjudge is configured and site-specific configuration set,
 the team manual can be generated with the command ``make docs``.
 The following should do it on a Debian-like system::
 
-  sudo apt install python-sphinx python-sphinx-rtd-theme python-yaml rst2pdf
+  sudo apt install python-sphinx python-sphinx-rtd-theme rst2pdf python3-yaml
   cd <INSTALL_PATH>/doc/
   make docs
 
 On Debian 11 and above, install
-``python3-sphinx python3-sphinx-rtd-theme python3-yaml rst2pdf`` instead.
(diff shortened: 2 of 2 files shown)


nickygerritsen:
Now I wonder whether it makes sense to change the shebang to just be `python` instead of `python3`. I assume it will then choose the most appropriate Python? Because then that might be better. My knowledge about how Python versions are chosen is...limited.

mvr320:
@nickygerritsen I think your fix would work (most of the time) and if the idea of the manual is to always let people install as little packages as possible that would be the fix with the lowest impact. 

The python versions are chosen based on what is in the PATH, where in (my) ubuntu `/usr/bin/env python` does not point to python3 (and python2 is not installed anymore)

if you would keep it like it is you would run documentation on debian <11 with 2 versions of python and everything would run, if you change it to `env python` on should have a symlink from python->python3 (which I think basicly everyone actually has)

nickygerritsen:
Aka drop the `3` and it should still work right?

@tom93 I do like that you swapped the two packages. So if you can modify your PR to still have that and update the shebang to drop the `3` we can see what the CI does. I can also do that if you want.

tom93:
I think `/usr/bin/env python3` is best.
According to [PEP 394](https://www.python.org/dev/peps/pep-0394/) `python` may refer to python2 or python3, and on Debian 10 it's still python2. I don't see any reason to support Python 2; for example `textwrap.indent` (used [here](https://github.com/DOMjudge/domjudge/blob/a5a24fee21f3f486b3d22e093d147a800171f204/doc/manual/gen_conf_ref.py#L36)) requires Python 3.

Aside: The [Debian Python Policy](https://www.debian.org/doc/packaging-manuals/python-policy/python.html#interpreter_loc) prefers `/usr/bin/python3` and advises against `/usr/bin/env ...` because it bypasses Debian's dependency checking and may run the script in an incomplete local installation. But the gen_conf_ref.py isn't installed system-wide so this is not an issue, and it may be useful to run the script in a local installation such as virtualenv.

nickygerritsen:
> I think leaving it as `/usr/bin/env python3` is best.
> According to [PEP 394](https://www.python.org/dev/peps/pep-0394/) `python` may refer to python2 or python3, and on Debian 10 it's still python2. I don't see any reason to support Python 2; for example `textwrap.indent` (used [here](https://github.com/DOMjudge/domjudge/blob/a5a24fee21f3f486b3d22e093d147a800171f204/doc/manual/gen_conf_ref.py#L36)) requires Python 3.
> 

Aha, I didn't know that. Then let's not try to support something (Python2) that is EOL anyway.

In that case this looks good and I'd say we indeed can change the order in https://github.com/DOMjudge/domjudge-packaging/blob/8fc6462b6e0f78d4de2112bcf28e33f87a853168/docker-gitlabci/Dockerfile#L10 as well.

PR #802, opened 2020-05-16
Add option to verify shadow differences. Fixes #706.

### LINKED ISSUE #706: Improve shadow differences page
We should also display the external submission ID (incl. link it to the primary CCS) on the shadow differences overview pages.

Also add a way to resolve a difference with a comment (I tried to use the normal submission verification functionality for this) and have a way to filter for unverified differences.

nickygerritsen:
I decided to mimic the judging verification logic on the external judgement. This makes it so we can move some logic into shared functions that do the same.

By default we still show all submissions (well the diff actually) but you can switch it to show only verified or unverified ones,

I made it that you can only verify a difference if there is actually a difference.

Note that because the info is stored on the external judgement, if a new one comes in (even with the same result), the verification is gone. Do we want that? We could copy it if the result didn't change for example. Similarly, if we have a new (valid) (internal) judging with a different result, should we remove the verification on the current valid external judgement if it has any? I'm not sure what the best behavior is.

Some screens:

Screenshot 2020-05-16 at 13 16 23
Screenshot 2020-05-16 at 13 16 37
Screenshot 2020-05-16 at 13 17 34
Screenshot 2020-05-16 at 13 17 59

### DIFF
--- a/webapp/src/Migrations/Version20200516101037.php
+++ b/webapp/src/Migrations/Version20200516101037.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20200516101037 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'add verification fields to external judgements';
+    }
+
+    public function up(Schema $schema): void
+    {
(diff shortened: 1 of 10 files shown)


webapp/src/Migrations/Version20200516101037.php
```diff
@@ -0,0 +1,42 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20200516101037 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
```
meisterT, line 17: nit: add a description?

PR #804, opened 2020-05-17
Replace config API calls for every value with a single call to fetch them all

thijskh:
I wanted to add a new config value so it seemed a good time to consolidate this before adding yet another REST call. At least in my install the api calls are not insignificant, although I know some (other) tests in the past did not show big savings here.

The config is fetched initially and at the start of each judge() run.

On the server side the processing should not be heavier to fetch all at once since fetching one specific value already retrieves the full config, after which only the requested value is returned.

Also rename dbconfig to djconfig since the db part is not so relevant in judgehost context.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -13,6 +13,7 @@ if (isset($_SERVER['REMOTE_ADDR'])) {
 require(ETCDIR . '/judgehost-config.php');
 
 $endpoints = [];
+$domjudge_config = [];
 
 function judging_directory(string $workdirpath, array $judging)
 {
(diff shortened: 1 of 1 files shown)


judge/judgedaemon.main.php
```diff
@@ -607,12 +626,15 @@ function fetch_executable(
     // we have gotten a submission for judging
     $endpoints[$endpointID]["waiting"] = false;
 
+    // refresh config at start of judge run
+    djconfig_refresh();
```
meisterT, line 630: nit: I would move this into the judge() function, I think it would be easier to read that way.

PR #807, opened 2020-05-18
Escape newlines and other reserved characters in TSV export

tom93:
Some of the fields can contain newlines, which corrupts the exported TSV files. For example the team name can be made to contain newlines with a bit of effort during self-registration, which breaks teams.tsv.

This PR also uses backslash escapes for tabs, which were previously replaced with spaces.

The [CLICS CCS spec](https://clics.ecs.baylor.edu/index.php?title=Contest_Control_System_Requirements#groups.tsv) doesn't mention handling of special characters; backslash escape sequences seem reasonable.

### DIFF
--- a/webapp/src/Controller/Jury/ImportExportController.php
+++ b/webapp/src/Controller/Jury/ImportExportController.php
@@ -280,7 +280,9 @@ class ImportExportController extends BaseController
             echo sprintf("%s\t%s\n", $type, $version);
             // output the rows, filtering out any tab characters in the data
             foreach ($data as $row) {
-                echo implode("\t", str_replace("\t", " ", $row)) . "\n";
+                echo implode("\t", array_map(function ($field) {
+                    return Utils::toTsvField((string)$field);
+                }, $row)) . "\n";
             }
         });
         $filename = sprintf('%s.tsv', $type);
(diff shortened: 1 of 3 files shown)


nickygerritsen:
Thanks, makes sense.

eldering:
I don't think this is the correct fix: https://tools.ietf.org/html/rfc4180 seems to be the accepted standard for escaping in CSV files, which says that fields containing newlines and commas (or tabs in our case of TSV files) should be enclosed in double quotes. And double quotes in the original text should be replaced by a double double quote.

tom93:
I think that only applies to CSV.
[IANA](https://www.iana.org/assignments/media-types/text/tab-separated-values)'s TSV spec seems the most authoritative, but does not have escaping/quoting of any kind.
[Here](http://dataprotocols.org/linear-tsv/) is a TSV format that uses backslash escapes (and preaches about their advantages).

thijskh:
When adding the Twig filter we followed https://en.m.wikipedia.org/wiki/Tab-separated_values#Conventions_for_lossless_conversion_to_TSV which claims that this is the common convention. In any case, ideally the CLICS spec would make explicit which encoding to use.

thijskh:
But for the fix... why don't we make it a twig template instead, and not move the logic outside of the twig filters?

tom93:
The backslash sequences should probably be processed on import as well (I have a draft at [tsv-import-parse-escapes](https://github.com/tom93/domjudge/commit/tsv-import-parse-escapes)), so I think it makes sense to move the logic out of Twig.

Issue #729, opened 2020-04-22
Improve webtest suite


thijskh:
The current test suite in `webapp/tests/` is a start but not very satisfactory.

Some ways to improve:
- Remove duplication between the tests and clean them up
- Create a good clean template to test e.g. jury interface functionality that can easily be populated for the various data types;
- Add tests using said template;
- Add tests for the API

mathew947:
[ mathew947]()

PR #810, opened 2020-05-22
Tests refactor

nickygerritsen:
This fixes most of the things of #729.

Some notes:
* I added separate commits so it's clearer what I did.
* As stated in one of the commit comments, I moved test classes to a more logical place. However I have no idea where we should put the `GeneralAvailabilityTest` since it is not really tied to a controller or other class. So I left it where it was.
* Same for `ScoreboardIntegrationTest` but I moved it to an `Integration` namespace, so it is at least in a separate folder.

I have not created a base test template yet for for example the jury interface. This can be done later if we have a few jury tests. There are also still some todo's in the comments that I think we can fix now. I will try to fix those first after we have merged this.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -102,7 +102,11 @@
 			"App\\": "webapp/src/"
 		}
 	},
-	"autoload-dev": {},
+	"autoload-dev": {
+		"psr-4": {
+			"App\\Tests\\": "webapp/tests/"
+		}
+	},
 	"config": {
 		"preferred-install": {
 			"*": "dist"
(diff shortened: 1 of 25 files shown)


nickygerritsen:
I have now added tests for the things that had a comment that they were not possible but now are because we have no legacy things anymore. For that I added a nice function that can temporarily change a configuration value, run a callback and then reset it to the original value again which will be useful for more tests I guess.

meisterT:
Can we include in the documentation how to run these tests?

nickygerritsen:
Good idea. I needed to check my this myself as well

doc/manual/tests.rst
```diff
@@ -0,0 +1,47 @@
+Appendix: Running the test suite
```
thijskh, line 1: Does this need to be an appendix or can it be a section in the chapter develop?

PR #813, opened 2020-05-24
Add option to repeat a rejudging N times.

meisterT:
All but the last one of those rejudgings will be automatically
cancelled.

There is no easy comparison across all those rejudgings yet, that will
follow in a separate commit.

### DIFF
--- a/webapp/src/Entity/Rejudging.php
+++ b/webapp/src/Entity/Rejudging.php
@@ -4,6 +4,7 @@ namespace App\Entity;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\Mapping as ORM;
+use JMS\Serializer\Annotation as Serializer;
 
 /**
  * Rejudge group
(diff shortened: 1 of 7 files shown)


webapp/src/Entity/Rejudging.php
```diff
@@ -118,6 +118,26 @@ class Rejudging
...
+
+    /**
+     * @var int
+     *
+     * @ORM\Column(type="integer", name="repeat_rejudgingid",
+     *     options={"comment"="In case repeat is set, this will mark the first rejudgingid.",
+     *              "unsigned"=true,"default"="NULL"},
+     *     nullable=true)
+     */
+    private $repeat_rejudgingid;
```
nickygerritsen, line 139: Shouldn't we change this to an actual entity relation? Then a foreign key will also be created.

meisterT: How? I wanted to copy what we do in other places and don't see any difference to e.g. https://github.com/DOMjudge/domjudge/blob/master/webapp/src/Entity/JudgingRun.php#L47

nickygerritsen: That also has this: https://github.com/DOMjudge/domjudge/blob/master/webapp/src/Entity/JudgingRun.php#L92

We can even leave out the id field itself I believe, it’s not really used and sometimes gives issues if you also have a linked entity.

meisterT: Is that what you had in mind?

nickygerritsen: Beautiful. We could remove the other field but if this works ship it

meisterT: We have this in other places as well so I wonder what's best practice.

meisterT: (we don't need to load any other data than the id itself if it matters)

nickygerritsen: I think best practice is to not have both but we indeed have it a lot. Also you can call something like $em->getreference to not load the whole entity iirc. But as said let’s keep it consistent for now. Maybe we can do a separate PR where we remove all these columns

PR #815, opened 2020-05-27
Allow users with API_READER role to fetch inactive contests

### LINKED ISSUE #814: Can't access inactive contests through API even with role `API_READER`
Trying to access the scoreboard of an inactive contest even with a user with the `API_READER` role results in an `HTTP 404`. While the documentation does not make entirely clear, what the desired behavior would be, line 140 in the API `ScoreboardController.php` clearly indicates that a user with the `API_READER` role should be allowed to access such a contest. The reason why it does not work is the call to `getContestId($request)` in line 139. This call ends up going through `getContestQueryBuilder()` in the `AbstractRestController.php` which excludes deactivated contests regardless of any roles which leads to the `404`.

### Your environment
Version 7.2.0

### Steps to reproduce
- Create contest with deactivation time in the past.
- Create user with `API_READER` role.
- `GET https://domjudge.iti.kit.edu/main/api/v4/contests/<cid>/scoreboard`

### Expected behaviour
The API returns the scoreboard data.

### Actual behaviour
A `404` is thrown.

tim3z:
fix #814

### DIFF
--- a/gitlab/integration.sh
+++ b/gitlab/integration.sh
@@ -176,6 +176,9 @@ fi
 # Start logging again
 set -x
 
+# Delete contest so API check does not fail because of empty results
+echo "DELETE FROM contest WHERE cid =1" | mysql domjudge
+
 # Check the Contest API:
 $CHECK_API -n -C -e -a 'strict=1' http://admin:$ADMINPASS@localhost/domjudge/api
 
--- a/webapp/src/Controller/API/AbstractRestController.php
+++ b/webapp/src/Controller/API/AbstractRestController.php
@@ -138,10 +138,14 @@ abstract class AbstractRestController extends AbstractFOSRestController
             ->from(Contest::class, 'c')
             ->select('c')
             ->andWhere('c.enabled = 1')
-            ->andWhere('c.activatetime <= :now')
-            ->andWhere('c.deactivatetime IS NULL OR c.deactivatetime > :now')
-            ->setParameter(':now', $now)
             ->orderBy('c.activatetime');
+
+        if (!$this->dj->checkrole('api_reader')) {
+            $qb
+                ->andWhere('c.activatetime <= :now')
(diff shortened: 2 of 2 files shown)


nickygerritsen:
Hah now our CI fails, since we return the deactivated contest but it doesn't contain any submissions or clarifications. @eldering I think that is a bug in our validator? It should not always have at least one entry for each endpoint right?

nickygerritsen:
> Hah now our CI fails, since we return the deactivated contest but it doesn't contain any submissions or clarifications. @eldering I think that is a bug in our validator? It should not always have at least one entry for each endpoint right?

Ah that is the `-n` we pass to [check-api.sh](https://github.com/DOMjudge/domjudge/blob/master/gitlab/integration.sh#L180)

eldering:
> Ah that is the `-n` we pass to [check-api.sh](https://github.com/DOMjudge/domjudge/blob/master/gitlab/integration.sh#L180)

Indeed, the point of adding that `-n` option was to make sure that the check doesn't trivially pass by having only empty endpoints. So I wouldn't want to simply remove it. We could delete the `demotest` contest before running this check.

nickygerritsen:
> > Ah that is the `-n` we pass to [check-api.sh](https://github.com/DOMjudge/domjudge/blob/master/gitlab/integration.sh#L180)
> 
> Indeed, the point of adding that `-n` option was to make sure that the check doesn't trivially pass by having only empty endpoints. So I wouldn't want to simply remove it. We could delete the `demotest` contest before running this check.

Adding
```bash
echo "DELETE FROM contest WHERE cid =1" | mysql domjudge
```

just before checking should fix it then. @tim3z could you try adding that?

tim3z:
Seems to have fixed it

nickygerritsen:
Woohooo 😄 I will merge it then.

PR #817, opened 2020-05-27
Process escape sequences in TSV import

tom93:
From the commit message:

> Together with commit e42f594d3 [[PR]](https://github.com/DOMjudge/domjudge/pull/807) (Escape newlines and other reserved characters in TSV export, 2020-05-18), this allows arbitrary data to be exported and imported losslessly via TSV.
> 
> This implementation uses `stripcslashes`, which also supports other escape sequences that are not used in the export format. Those escape sequences can't appear in the exported data, so there's no harm.
> 
> (It's tempting to use `str_replace` instead of `stripcslashes`, and mirror `toTsvField()`, but it's tricky because `str_replace` performs each substitution in order on the entire string. For example `\\t` should be decoded as `\t`, but the naive approaches either process it twice and give `<TAB>`, or disregard precedence and give `\<TAB>`. Here is a correct approach:
> 
>     return str_replace(
>         ["\\\\", "\\t", "\\n", "\\r", "\\\\e"],
>         ["\\\\e", "\t", "\n", "\r", "\\"],
>         explode("\t", rtrim($line, "\r\n"))
>     );
> 
> However, `stripcslashes` is simpler.)

The previous code used `trim()` on each line, which removes the newline from the end but also removes other leading/trailing whitespace. I wasn't sure if it was intentional so I left it as it is, and also explicitly removed trailing newlines in the `parseTsvLine` function.

### DIFF
--- a/webapp/tests/Utils/UtilsTest.php
+++ b/webapp/tests/Utils/UtilsTest.php
@@ -847,4 +847,24 @@ part.";
         $this->assertEquals('team nåme…',   Utils::toTsvField("team nåme…"));
         $this->assertEquals('team🎈name',   Utils::toTsvField("team🎈name"));
     }
+
+    /**
+     * Test Tab Separated Value parsing
+     */
+    public function testParseTsvLine()
+    {
+        $bs  = "\\";
+        $tab = "\t";
+        $this->assertEquals(["team name", "rank"],    Utils::parseTsvLine("team name".$tab."rank"));
+        $this->assertEquals(["team\tname\t", "rank"], Utils::parseTsvLine("team".$bs."t"."name".$bs."t".$tab."rank"));
+        $this->assertEquals(["team\nname\r", "rank"], Utils::parseTsvLine("team".$bs."n"."name".$bs."r".$tab."rank"));
+        $this->assertEquals(["team\\name\\", "rank"], Utils::parseTsvLine("team".$bs.$bs."name".$bs.$bs.$tab."rank"));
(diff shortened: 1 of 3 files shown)


webapp/tests/Utils/UtilsTest.php
```diff
@@ -847,4 +847,24 @@ public function testToTsvField()
         $this->assertEquals('team nåme…',   Utils::toTsvField("team nåme…"));
         $this->assertEquals('team🎈name',   Utils::toTsvField("team🎈name"));
     }
+
```
nickygerritsen, line 850: LIKE! Great that you added a test immediately.

webapp/tests/Utils/UtilsTest.php
```diff
@@ -847,4 +847,24 @@ public function testToTsvField()
...
+    /**
+     * Test Tab Separated Value parsing
+     */
+    public function testParseTsvLine()
+    {
+        $bs  = "\\";
```
nickygerritsen, line 856: What is the reason you uses a variable for these two?

tom93: I thought it was more readable than having to count backslashes, e.g.
```php
Utils::parseTsvLine("team\\\\name\\\\\trank"));
```
vs
```php
Utils::parseTsvLine("team".$bs.$bs."name".$bs.$bs.$tab."rank"));
```

nickygerritsen: Makes sense actually.

PR #821, opened 2020-05-27
Fail gracefully if browser doesn't support ResizeObserver

tom93:
The current Firefox ESR release (68.8.0) doesn't support it yet.

### DIFF
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -611,7 +611,7 @@ function updateMenuBalloons(data)
         for (var i=0; i<num; i++) {
             var text = (data[i].room !== null) ? data[i].room+': ' : '';
             text += data[i].pname + ' ' + data[i].name;
-			sendNotification('New balloon:',
+            sendNotification('New balloon:',
                  {'tag': 'ball_' + data[i].baloonid,
                   'link': domjudge_base_url + '/jury/balloons',
                   'body': text});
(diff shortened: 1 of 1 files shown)


nickygerritsen:
@mvr320 this is one for you I think. Any opinion?

mvr320:
Yes, it does not fix the problem why I introduced the feature, so I would propose fixing by using a different implementation.

thijskh:
@mvr320 SInce it's a more helpful UI but does not actually break anything in older browsers, I would also think it's fine to accept that you get this if you have a more recent browser, and not if you don't.

mvr320:
@thijskh I think in this case when you need the ResizeObserver you have resized the page so the header can end up everywhere, as at that moment its a fixed style object. So I'm not sure if that should be something which is accepted. I think I can get the same behaviour with onresize, but I did not find time yet to test this. If this is currently blocking somewhere I'm fine with making the time this week but otherwise it will take some time.

tom93:
@mvr320 No rush.
For me it appears to work fine without ResizeObserver, even when I resize the page.
It only breaks if I manually change the height of the top navbar using the DOM inspector, and I'm not sure if that can happen in normal circumstances.

thijskh:
After discussion with @mvr320: this PR improves the situation so can be merged; time permitting he will come with a further improvement later.

PR #822, opened 2020-06-02
Support missing id in organization JSON import

tom93:
Merge affiliations based on the shortname instead of the id when the id key is missing (otherwise they are all merged together).

It is common for affiliations to have a NULL externalid, for example those created through the self-registration form.

### DIFF
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -708,12 +708,12 @@ class ImportExportService
     protected function importOrganizationsJson(array $data, string &$message = null): int
     {
         $organizationData = [];
-        foreach ($data as $idx => $group) {
+        foreach ($data as $idx => $organization) {
             $organizationData[] = [
-                'externalid' => @$group['id'],
-                'shortname' => @$group['name'],
-                'name' => @$group['formal_name'],
-                'country' => @$group['country'],
+                'externalid' => @$organization['id'],
+                'shortname' => @$organization['name'],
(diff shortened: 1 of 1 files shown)


nickygerritsen:
`importTeamData` also assumes the affiliation external ID is set. I think we should somehow handle that as well?

tom93:
I had a look at the team import, there are a bunch of things going on there:

 - The `elseif` branch of `importTeamData` (i.e. empty shortname, non-empty externalid) doesn't set the affiliation shortname, but the shortname is required. (That branch is unreachable via `importTeamsTsv` but reachable via `importTeamsJson`.)

 - That `elseif ` branch also doesn't update `$createdAffiliations`, not sure if that's intentional. (It also doesn't call `flush()`, but that doesn't cause problems because there is a flush later on in the loop body. Flushing is required for `find()` to see the new affiliation.)

 - `importTeamsJson` sets externalid for the affiliation but not name/shortname/country. Based on the REST API, I'm guessing the affiliation name and country should be set from the JSON keys "affiliation" and "nationality". Not sure about shortname.

I can have a go at fixing this later.

nickygerritsen:
See https://www.domjudge.org/demoweb/api/contests/nwerc18/organizations for an example JSON output. `shortname` is a DOMjudge-specific thing which will *not* always be there but if it is you can use it. I'd say add it to the array if it exists and use it. Otherwise use something based on the ID.
Adding the flush seems fine btw.

And thanks!

nickygerritsen (approved):

PR #823, opened 2020-06-02
Build executables in chroot.

thijskh:
Build 'DOMjudge executables' inside the chroot.

This PR depends on #798 and vice versa, because when we want to compile teams' submissions inside the chroot, the executables we do the compilation with themselves also need to be built for that system.

Order of merging is: merge this PR into #798, then #798 into master.


thijskh:
Continued in #865

PR #828, opened 2020-06-11
Fix typo (change getIcpcid back to getTeamid in ImportExportService)

tom93:
From commit 0356a0469 (Rename team.externalid to team.icpcid, 2020-01-18).


nickygerritsen:
Makes total sense

PR #830, opened 2020-06-13
Upgrade FosRestBundle to 3.x.

nickygerritsen:
I was hoping we could also already update NelmioApiDocBundle but that one didn't release 4.x yet so that will come at a later stage.

This PR does remove some deprecations warning we had since upgrading Symfony.
I also tried using the Symfony Serializer component instead of JMSSerializer, but that misses the some functionality like excluding fields with an annotation and specifying types. If they ever add that we can switch, which would also mean we can drop the whole `FlattenExceptionHandler` since that is built in there.

I couldn't get it to work that `/api` without a `/` at the end returns the API root instead of a redirect, but I don't think we should really bother with such a small thing.


webapp/config/routes/annotations.yaml
```diff
@@ -1,3 +1,4 @@
 controllers:
-    resource: ../../src/Controller/
+    resource: ../../src/Controller/**/*
```
nickygerritsen, line 2: Note: this change is needed because otherwise the API would also be exposed on `/`

webapp/config/routes/api.yaml
```diff
@@ -0,0 +1,14 @@
+# Add the API controllers twice, both with the /api/v4 as well as the /api prefix
+domjudge_api_v4:
+    type: annotation
+    resource: ../../src/Controller/API/
+    prefix: /api/v4
+    name_prefix: v4_
```
nickygerritsen, line 6: I moved this to a seperate file. The `name_prefix` is needed because it needs to be unique per loaded routing set, otherwise /api would not work.

webapp/src/Serializer/SubmissionVisitor.php
```diff
@@ -85,13 +85,13 @@ public function onPostSerialize(ObjectEvent $event)
             /** @var Submission $submission */
             $submission         = $event->getObject();
             $filesRoute         = $this->router->generate(
-                'submission_files',
+                'v4_submission_files',
```
nickygerritsen, line 88: This name (and the below one) changed because of the name_prefix in the YAML above

PR #833, opened 2020-06-17
Fix output directory of gen_all_secrets when running from a different directory

tom93:
This is a more complete fix than cb1afca77cd4970bdeff0210e80fc041a0f31102.

### DIFF
--- a/etc/gen_all_secrets
+++ b/etc/gen_all_secrets
@@ -13,7 +13,7 @@ create_pw_file()
 	[ -n "$QUIET" ] || echo -n "Running '$SCRIPT'... "
 	touch "$FILE"
 	chmod go= "$FILE"
-	$(dirname $0)/$SCRIPT > "$FILE"
+	"./$SCRIPT" > "$FILE"
 	[ -n "$QUIET" ] || echo "file '$FILE' created."
 }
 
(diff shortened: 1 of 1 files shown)


etc/gen_all_secrets
```diff
@@ -38,7 +38,8 @@ while getopts ':q' OPT ; do
 done
 shift $((OPTIND-1))
 
-create_pw_file gendbpasswords dbpasswords.secret
-create_pw_file genrestapicredentials restapi.secret
-create_pw_file gensymfonysecret symfony_app.secret
-create_pw_file genadminpassword initial_admin_password.secret
+cd "$(dirname "$0")"
```
eldering, line 41: I would propose to only changing this single line. I think that should be enough to make the output go into the `etc` directory (specifically: the directory this script resides in). It also preserves script name in the message on line 13, which becomes more ambiguous with a `./` prefixed.

tom93: I think that would work if line 16 was also changed to `"./$SCRIPT" > "$FILE"`, because `$0` can be a relative path so is unreliable after `cd`.

eldering: Indeed, you're right.

PR #835, opened 2020-06-20
Add command to reset the password of a user.

nickygerritsen:
This is easier than running some PHP code (or use a possibly unsafe online password hash generator) to do the same.

### DIFF
--- a/doc/manual/config-basic.rst
+++ b/doc/manual/config-basic.rst
@@ -40,6 +40,17 @@ A jury or administrative user can also be associated with a team. This
 will enable that user to submit solutions to the system, or resubmit
 edited team solutions.
 
+Resetting the password for a user
+---------------------------------
+
+If you do not have access anymore to any admin user, you can use the following
+command to reset the password of a user to a random value::
+
+  webapp/bin/console domjudge:reset-user-password admin
+
+Replace ``admin`` with the username of the user you want to reset the password for.
+The password will be displayed.
+
 Adding a contest
 ----------------
 You configure a new contest by adding it under the Contests link
--- a/webapp/src/Utils/Utils.php
+++ b/webapp/src/Utils/Utils.php
@@ -1004,18 +1004,36 @@ class Utils
     }
 
     /**
-     * Generate a random password of length 6 with lowercase alphanumeric
-     * characters, except o, 0, l and 1 since these can be confusing.
+     * Generate a random password.
+     *
+     * When $moreEntropy is `true`, generate a password of length 16 with
+     * alphanumeric characters and _ and -. When `false`, generate a password of
+     * length 6 with lowercase alphanumeric, except o, 0, l and 1. `false`
+     * should be used when generating password that will be printed and handed
+     * out. In other cases, use `true`.
+     *
+     * @param bool $moreEntropy
+     *
+     * @return string
      */
-    public static function generatePassword() : string
+    public static function generatePassword(bool $moreEntropy = true) : string
     {
-        $chars = ['a','b','c','d','e','f','g','h','i','j','k','m','n','p','q','r',
-                  's','t','u','v','w','x','y','z','2','3','4','5','6','7','8','9'];
+        if ($moreEntropy) {
+            $chars = array_merge(
+                range('a', 'z'),
+                range('A', 'Z'),
+                range('0', '9'),
+                ['-', '_']
+            );
+        } else {
+            $chars = ['a','b','c','d','e','f','g','h','i','j','k','m','n','p','q','r',
+                      's','t','u','v','w','x','y','z','2','3','4','5','6','7','8','9'];
+        }
 
         $max_chars = count($chars) - 1;
 
         $rand_str = '';
-        for ($i = 0; $i < 6; ++$i) {
+        for ($i = 0; $i < ($moreEntropy ? 16 : 6); ++$i) {
             $rand_str .= $chars[random_int(0, $max_chars)];
         }
 
--- a/webapp/src/Command/ResetUserPasswordCommand.php
+++ b/webapp/src/Command/ResetUserPasswordCommand.php
@@ -1,0 +1,95 @@
+<?php declare(strict_types=1);
+
+namespace App\Command;
+
+use App\Entity\User;
+use App\Utils\Utils;
+use Doctrine\ORM\EntityManagerInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
+
+/**
+ * Class ResetUserPasswordCommand
+ *
+ * @package App\Command
+ */
+class ResetUserPasswordCommand extends Command
+{
+    const STATUS_OK = 0;
+    const STATUS_ERROR = 1;
+
+    /**
+     * @var EntityManagerInterface
+     */
+    protected $em;
+
+    /**
+     * @var UserPasswordEncoderInterface
+     */
+    protected $passwordEncoder;
+
+    /**
+     * ResetUserPasswordCommand constructor.
+     *
+     * @param EntityManagerInterface       $em
+     * @param UserPasswordEncoderInterface $passwordEncoder
+     * @param string|null                  $name
+     */
+    public function __construct(
+        EntityManagerInterface $em,
+        UserPasswordEncoderInterface $passwordEncoder,
+        string $name = null
+    ) {
+        parent::__construct($name);
+        $this->em = $em;
+        $this->passwordEncoder = $passwordEncoder;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('domjudge:reset-user-password')
+            ->setDescription('Reset the password of the given user')
+            ->addArgument(
+                'username',
+                InputArgument::REQUIRED,
+                'The username of the user to reset the password of'
+            );
+    }
+
+    /**
+     * @inheritDoc
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $style = new SymfonyStyle($input, $output);
+
+        $username = $input->getArgument('username');
+        $user = $this->em
+            ->getRepository(User::class)
+            ->findOneBy(['username' => $username]);
+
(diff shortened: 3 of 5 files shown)


webapp/src/Utils/Utils.php
```diff
@@ -1022,6 +1022,32 @@ public static function generatePassword() : string
...
     }
 
+    /**
+     * Generate a strong random password of the given length
+     *
+     * @param int $length
+     *
+     * @return string
+     */
+    public static function generateStrongPassword($length = 16) : string
```
tom93, line 1032: I'm not sure about the name "generateStrongPassword", because it takes a length parameter and e.g. `generateStrongPassword(3)` isn't very strong.

nickygerritsen: In the beginning I wanted to update generatePassword to add options for length and whether to allow more characters. We can also do that?

I will also add unit tests for this function (or the modified generatePassword) btw.

tom93: Oh right, I missed that there is an existing function called "generatePassword". In that case it makes sense to call the new one "generateStrongPassword" (or maybe rename the existing function to "generateShortPassword", or combine them as you suggested).

meisterT: Is there are reason why we make the length configurable even?

webapp/src/Utils/Utils.php
```diff
@@ -1022,6 +1022,32 @@ public static function generatePassword() : string
...
+    {
+        $chars = str_split(
+            'abcdefghijklmnopqrstuvwxyz' .
```
meisterT, line 1035: what about something like this?

```
array_merge(
  range('a', 'z'),
  range('A', 'Z'),
  range('0', '9')
)
```

doc/manual/config-basic.rst
```diff
@@ -40,6 +40,16 @@ A jury or administrative user can also be associated with a team. This
 will enable that user to submit solutions to the system, or resubmit
 edited team solutions.
 
+Resetting the password for a user
+---------------------------------
+
+If you do not have access anymore to any admin user, you can use the following
+command to reset the password of a user to a random value::
```
meisterT, line 47: nit: make it clear that the new password will be displayed and is not a script argument

webapp/src/Utils/Utils.php
nickygerritsen: I like the range. I can make the length non-configurable indeed

nickygerritsen:
I will add a link to the 7.2 documentation for old versions once we have a place where those docs are available. (I could link to the .rst on the 7.2 branch but that's not so nice IMHO)

webapp/src/Utils/Utils.php
mvr320: https://www.ncsc.nl/onderwerpen/authenticatie
I think you should atleast extend with the symbols in asci (if your terminal would not accept unicode symbols) and the length should be variable between 10 and 16 so the amount of possible passwords is still high and bruteforce really becomes undoable?

I think even better if you want to keep the passwords safe is: https://www.zdnet.com/article/fbi-recommends-passphrases-over-password-complexity/ basicly the comic advice we all know so people might actually remember the password so its not downgraded to "test" as this might be remembered (although this really will create different discussions)

webapp/src/Utils/Utils.php
```diff
@@ -1022,6 +1022,30 @@ public static function generatePassword() : string
         return $rand_str;
     }
 
+    /**
+     * Generate a strong random password
+     *
+     * @return string
+     */
+    public static function generateStrongPassword() : string
```
thijskh, line 1030: Or just make generatePassword take a length parameter?

webapp/src/Utils/Utils.php
tom93: Regarding password strength, with 64 options for each character and a length of 16 the number of possible passwords is 64^16 = 7.9 * 10^28. The paper [SHA-1 is a Shambles](https://sha-mbles.github.io/) mentions a cost of 100k USD to compute 2^64 operations of symmetric cryptography, so the cost of cracking such a password is on the order of 10^14 USD and that's plenty.

Using a variable shorter length might seem like it would significantly increase the number of possible passwords, but surprisingly it doesn't: sum(64^n for n in 10..16) = 8.05 * 10^28. It's pretty hard to make the distribution of passwords uniform when the length is variable.

Passphrases are cool but harder to implement because they need a word list, and they are still difficult to memorize.

I think the current scheme (which is already used for other passwords) is fine. Users can easily copy-paste the password or change it.

Blacklisting weak passwords is a good idea. It might make people angry when creating lots of dummy accounts for local testing :)

nickygerritsen: I indeed took the same scheme as used in genadminpasswords, which is basically the characters from base64 + replace `/` with `_` and `+` with `-`.

If we want to change the password generation let's do that later in a separate PR.

doc/manual/config-basic.rst
```diff
@@ -40,6 +40,17 @@ A jury or administrative user can also be associated with a team. This
...
+
+  webapp/bin/console domjudge:reset-user-password admin
+
+Replace `admin` with the username of the user you want to reset the password of.
+The password will be generated and displayed.
```
tom93, line 52: I think \`admin\` is Markdown, and it should be \`\`admin\`\` for RST.

Aside, some alternative wording options:
"Replace \`\`admin\`\` with the username of the user you want to reset the password **for**." (More natural? Same in setArgument.)
"The new password will be displayed." (Hopefully makes it clearer that the password will be updated in the database, and not just generated.)

webapp/src/Utils/Utils.php
nickygerritsen: But generatePassword doesn't support capitals and - and _. Do we want to merge the functions and add an option for that?

tom93: My personal preference is to keep them as separate functions, but rename the old function to generateShortPassword and call the new function generatePassword / generateStrongPassword.
Maybe also update the documentation to explain that generateShortPassword is intended for passwords distributed to users on paper.
The functions can be implemented separately as they are currently, or refactored e.g. using a new private function that takes a length and set of characters. They are short, so it's not a big deal.
This approach prevents misuse (e.g. providing a short length, or calling the old generatePassword function when the intention is to generate a strong password).

thijskh: If you just add length, then you have 16 chars lowercase alphanumeric, should be enough entropy for anyone, no?

nickygerritsen: But then I'd say we also need to change `genadminpassword` to NOT use A-Z, - and _, since then we have two different sets of characters to use for the same thing. But that also feels meh.

thijskh: I'm also fine with adding a $mode = easy/difficult instead of length, and vary charset and length based on it.

In any case, genadminpassword should just use this function (via the console command)?

nickygerritsen: I can just try to use it there and see what happens.

nickygerritsen: Wait but then we need a flag to the command like `--write-to-file` or something that also writes the password to the `etc/initial_admin_password` file

thijskh: Remove the file, say what command to run to set initial value? Or is that too complicated for users?

nickygerritsen: or maybe only run the command for install of the db and not upgrade? The command shows the password so that should help I guess

tom93: I like the password in a file, and it's used in [integration.sh](https://github.com/DOMjudge/domjudge/blob/2e35a3644984f8fd563fff4003e1beddcc27fb6a/gitlab/integration.sh#L44).

[This migration](https://github.com/DOMjudge/domjudge/blob/2e35a3644984f8fd563fff4003e1beddcc27fb6a/webapp/src/Migrations/Version20190803123217.php#L740) currently reads the file, it could be changed to generate the password and write it to the file. That would also work for restapi.secret. (I'm not sure about symfony_app.secret, which might need to exist earlier; by the way, its [recommended length](https://symfony.com/doc/current/reference/configuration/framework.html#secret) is around 32 chars.)

thijskh: Just thinking that maybe we’re creeping the scope here a bit much. So it might also be ok to just merge this with the slightly duplicated code. Then at least we have this useful feature. Not sure on second thought if changing the initial password generation is worth it or just premature optimization

nickygerritsen: I agree, although I think we can still merge the functions. Propsal

* Add `$moreEntropy` to the generate function, update the docs to say that $moreEntropy=false is useful for when you are handing out the passwords on paper.
* If `$moreEntropy = true`, use length 16 and the expanded charset, otherwise not
* Update the tests

And then we can do other stuff later.

nickygerritsen: Should be fixed now.

doc/manual/config-basic.rst
```diff
@@ -40,6 +40,22 @@ A jury or administrative user can also be associated with a team. This
...
+
+If you do not have access anymore to any admin user, you can use the following
+command to reset the password of a user to a random value::
+
+  webapp/bin/console domjudge:reset-user-password admin
+
+Replace ``admin`` with the username of the user you want to reset the password for.
+The password will be displayed.
+
+If you are running DOMjudge ``< 7.2``, see
```
thijskh, line 54: Do we want to add this kind of stuff to the manual? The manual is for the version it ships with. Esp since we now have multiple doc versions also on the website. Just select your version and the docs will match.

webapp/src/Utils/Utils.php
```diff
@@ -1004,18 +1004,36 @@ public static function wrapUnquoted(string $text, int $width = 75, string $quote
...
+     * length 6 with lowercase alphanumeric, except o, 0, l and 1. `false`
+     * should be used when generating password that will be printed and handed
+     * out. In other cases, use `true`.
+     *
+     * @param bool $moreEntropy
+     *
+     * @return string
      */
-    public static function generatePassword() : string
+    public static function generatePassword($moreEntropy = true) : string
```
thijskh, line 1019: Why no type hint?

doc/manual/config-basic.rst
nickygerritsen: Hmm true better not to have it

webapp/src/Utils/Utils.php
nickygerritsen: Since I work too much 🤣 (we still have some projects that don’t allow type hints). I’ll fix it

PR #836, opened 2020-06-20
Add admin password reset steps to the manual

tom93:
A nicer Symfony Console command will be added in #835. These instructions are mainly for older releases that won't have the command.


nickygerritsen:
Perhaps in that case we should do this on the 7.2 (and older) branches? Since #835 makes this obsolete for master.

tom93:
Rebased from master to 7.2.

PR #842, opened 2020-06-26
Make test-memsize.java use more memory faster.

### LINKED ISSUE #839: Make test-memsize.java faster
The file `tests/test-memsize.java` frequently fails the build because the result is _TIMELIMIT_ instead of _RUNERROR_ - it tries to fill the memory but is in some environments too slow to do so.

Change this test program to fill up memory faster. Maybe the change to `test-memsize.cc` can be an inspiration: 7e797237cfbaa60e877e1e993d0aa29bee0dae50.

RagnarGrootKoerkamp:
Fix #839

### DIFF
--- a/tests/test-memsize.java
+++ b/tests/test-memsize.java
@@ -10,12 +10,11 @@ import java.io.*;
 
 class Main {
     public static void main(String[] args) throws Exception {
-		int i;
 		int[][] ar = new int[10240][];
 
-		for(i=0; 1==1; i++) {
-			ar[i] = new int[1024*1024/4];
-			System.out.print("allocated " + (i+1) + " MB of memory\n");
+		for(int i=0; true; i++) {
+			ar[i] = new int[256*1024*1024];
+			System.out.print("allocated " + 256*(i+1) + " MB of memory\n");
 		}
     }
 }


thijskh:
Thanks! I've retried the build a couple of times and it did not fail, at the very least it seems not worse :-)

It did fail once on test-memsize**.cc** unfortunately. So it's still not 100% robust.

tests/test-memsize.java
```diff
@@ -10,12 +10,11 @@
...
     public static void main(String[] args) throws Exception {
-		int i;
 		int[][] ar = new int[10240][];
 
-		for(i=0; 1==1; i++) {
-			ar[i] = new int[1024*1024/4];
-			System.out.print("allocated " + (i+1) + " MB of memory\n");
+		for(int i=0; true; i++) {
+			ar[i] = new int[256*1024*1024];
+			System.out.print("allocated " + 256*(i+1) + " MB of memory\n");
```
eldering, line 17: This is incorrect: `sizeof(int) = 4` so this should be 1GB per iteration. I would argue though that the increment should be a bit smaller (e.g. 128MB) to be able to detect smaller memory limits.

eldering: Already fixed in master.

PR #844, opened 2020-06-27
Fix test-memsize.cc

RagnarGrootKoerkamp:
For some reason c++ (sometimes) takes a long time (~2 seconds) to handle program termination after after bad_alloc is thrown.
With this change, the error is now catched manually and we explicitly abort the program. This seems to be much faster (~0.3 seconds).

#839

### DIFF
--- a/tests/test-memsize.cc
+++ b/tests/test-memsize.cc
@@ -5,6 +5,7 @@
  * @EXPECTED_RESULTS@: RUN-ERROR
  */
 
+#include <cstdlib>
 #include <vector>
 
 using namespace std;
@@ -18,7 +19,8 @@ int main() {
 			v.reserve(1ll * 1024 * 1024);
 			vs.push_back(std::move(v));
 		} catch(const std::bad_alloc&) {
-			// Handle the exception and clean-up manually to prevent slow termination of the program.
+			// Handle the exception and clean-up manually to prevent slow termination of the
+			// program.
 			vs.~vector<vector<char>>();
 			// Make sure to re-raise the exit with SIGABRT.
 			abort();
--- a/tests/test-memsize.cc
+++ b/tests/test-memsize.cc
@@ -1,41 +1,28 @@
 /*
  * This should fail with RUN-ERROR due to running out of memory, which
- * is restricted. The amount allocated may seem to be half of the
- * available, but that is because (GNU implementation) STL vectors by
- * default allocate double the amount of memory requested.
+ * is restricted.
  *
  * @EXPECTED_RESULTS@: RUN-ERROR
  */
 
-#include <iostream>
 #include <vector>
 
 using namespace std;
 
-vector<char> a;
-
-int main()
-{
-	int p, i;
-
-	/*
-	  Watch out: resizing of a vector allocates AT LEAST that much memory!
-	  Testing shows, that (glibc 2.2.5) twice the requested amount is
-	  allocated, so e.g. when you have 64 MB memory available, already when
-	  resizing to more than 32 MB, you run out of memory.
-	*/
-	for(p=4; 1; p*=2) {
-		for(i=p/2; i<p; i+=p/4) {
-			cout << "trying to allocate " << i << " MB... ";
-			a.resize(i*1024*1024,0);
-			if ( a.capacity()<i*1024*1024 ) {
-				cout << "resizing failed." << endl;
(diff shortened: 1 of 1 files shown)


tests/test-memsize.cc
```diff
@@ -1,41 +1,30 @@
...
-			}
-			for(int j=a.capacity()/2; j<a.capacity(); j += 512) a[j] = j%137;
-			cout << "allocated: " << a.capacity()/1024/1024 << " MB." << endl;
+int main() {
+	vector<vector<char>> vs;
+	while(true) {
+		vector<char> v;
+		// Allocate 2GB at a time.
+		try {
+			v.reserve(1ll * 1024 * 1024);
```
tom93, line 19: The comment says 2GB, but it looks like the code allocates 1MB at a time.
If we are rewriting this, why not switch to `malloc`? It should be much simpler.
Does anyone know if there was a particular reason to use `vector` here?

RagnarGrootKoerkamp: Ah right, I'll update the comment.
I've never used C/`malloc`, and I suppose that `vector` is closer to 'real world usage', but other than that it doesn't really matter I suppose.

RagnarGrootKoerkamp: I tried `malloc`, and actually it's just as ugly. Just calling `malloc` and not using the returned pointer gets optimized away, so you have to use the data.
Still, error handling is super slow if I don't catch the `bad_alloc` manually, so the `try..catch` would still be needed.

eldering: > The comment says 2GB, but it looks like the code allocates 1MB at a time.
> If we are rewriting this, why not switch to `malloc`? It should be much simpler.
> Does anyone know if there was a particular reason to use `vector` here?

If I recall mainly because of simulating real world usage. We want to test not only if the memory limits are respected, but also that the handling and judging verdict are as expected. That's why ideally we have each type of test in each language (where it makes sense). This issue shows the value of such test coverage, but in the past we've e.g. also seen that a "memory limit exceeded" verdict is near impossible to implement because of the myriad of different ways this can be handled by different languages/runtime environments.

tom93: > real world usage

Fair enough.

I tested the version in this PR. It turns out `vector.reserve()` doesn't always allocate space, and the `try..catch` doesn't make a difference for me.

How about this, which is a bit faster than the original version:

```c++
const size_t inc_mb = 128;

int main()
{
	vector<vector<char>> vs;
	while(true) {
		vs.emplace_back(inc_mb * 1024 * 1024, 'A');
		cerr << "allocated: " << inc_mb * vs.size() << " MB\n";
	}

	return 0;
}
```

If there are still issues, I think we should reduce the memory limit.

RagnarGrootKoerkamp: In my local testing, I definitely *do* need the try catch, including calling the destructor explicitly. Here's what happens without: (note the delay between `terminate called` and the program actually stopping).

tty

Might very well be that the `reserve()` is optimized away sometimes, but at least your `emplace_back` version seems to work (both `clang++ -O3` and `g++ -O3` don't optimize it away).

tom93: It looks like the delay is caused by a core dump (it's probably creating a massive file called `core`).
Some ways to prevent it:

 - Use `ulimit -c 0` from the shell to disable core dumps. The current setting can be displayed using `ulimit -c`, for me it's 0 by default on Debian.
 - Catch the exception and use `return 1` instead of `abort()`.
 - Call `setrlimit()` to disable core dumps.

I believe the judgehost already disables core dumps, so these are only required if we want the code to behave nicely when executed locally. We should add a comment mentioning `ulimit -c 0` (I'm not sure which systems have core dumps enabled by default). The other two options deviate from real-world usage.

tom93: *_could_ add a comment, not should

RagnarGrootKoerkamp: Never heard of that before, but you're right, a ~4MB coredump was being created. Just stored in a different directory.

With `ulimit -c 0` the runtime goes from 1.5s to 0.5s when running manually, so that's much better. 

I've removed the try/catch from the code and added a comment on ulimit -c 0.

tom93: I updated the comment in commit 1efed521b3121c914a83444ac20c0436c0df3238.

PR #846, opened 2020-06-29
Search for @EXPECTED_RESULTS@: in all submitted files.

RagnarGrootKoerkamp:
`@EXPECTED_RESULTS@: ` is now checked in all submitted files.

Added 2 tests:
- one `NO-OUTPUT` test (doesn't add that much)
- one multiple-file submission where the `@EXPECTED_RESULTS@: ` occurs in an extra `.txt` file.

Let's check in CI whether this deals well with parsing submitted binary files. (There's already tests/test-extra-binary.sh.zip which should pass.)


tests/test-expected-result.c
```diff
@@ -0,0 +1,10 @@
+/*
+ * This should fail with NO-OUTPUT.
+ *
+ * @EXPECTED_RESULTS@: CORRECT, NO-OUTPUT
```
mvr320, line 4: The comment states it should fail but expected result can also be CORRECT, how does this work?

RagnarGrootKoerkamp: Oh right. Artefact from my testing with problemarchive format/BAPCtools. Should just be no-output.

I won't have access to a pc the next few days though.

webapp/src/Service/SubmissionService.php
```diff
@@ -512,11 +512,16 @@ public function submitSolution(
...
         // SQL transaction time below.
         if ($this->dj->checkrole('jury')) {
-            $results = self::getExpectedResults(file_get_contents($files[0]->getRealPath()),
-                $this->config->get('results_remap'));
+            foreach ($files as $rank => $file) {
+                $results = self::getExpectedResults(file_get_contents($file->getRealPath()),
+                    $this->config->get('results_remap'));
+                if ($results !== null){
+                    break;
+                }
```
mvr320, line 523: I assume this is as you expect that only one file has the "EXPECTED_RESULT" which I think should make sense, but as this is for jury solutions (which are not run often and are critical) must this not be really strict and fail if such a mistake is made? Or merge all the expected results?

Although I suspect this will probably (almost) never happen in for real as those are generated via other tools and multifile does not happen often I think? But for this case being strict might be preferred.

RagnarGrootKoerkamp: Yeah this could be more strict. In that case we should also check there's only a single match per file inside `getExpectedResult`.

Also, I don't know how exactly this interacts with the problem archive format import, but in BAPCtools I'm now requiring that a submission in e.g. the wrong answer directory must have the corresponding tag in the expected results list. (It would be weird to only have the time limit exceeded tag in this case.)

nickygerritsen:
@RagnarGrootKoerkamp any chance you can fix the first comment at least? Then we can merge

RagnarGrootKoerkamp:
@nickygerritsen done. forgot about this after the holidays.

I've added some test submissions containing multiple EXPECTED_RESULTS tags that should fail now. Let's see if CI passes with the changes to the `makefile` I did.

webapp/src/Service/SubmissionService.php
```diff
@@ -519,7 +519,7 @@ public function submitSolution(
             foreach ($files as $rank => $file) {
                 $fileResult = self::getExpectedResults(file_get_contents($file->getRealPath()),
                     $this->config->get('results_remap'));
-                if ($fileResult === false) {
+                if ($fileResult === false)
```
thijskh, line 522: Cleaned up one `{` too many here.

PR #847, opened 2020-07-01
Remove standard webserver support for TLS1.0

mvr320:
I think the major reason to have 1.0 was for old win7 clients, those are now really without support so lets try to drop the support.

### DIFF
--- a/etc/apache.conf.in
+++ b/etc/apache.conf.in
@@ -45,6 +45,7 @@
 #ServerName _default_
 #DocumentRoot @domserver_webappdir@/public
 #
+# See https://ssl-config.mozilla.org/ to generate good TLS settings for your server
 #SSLEngine on
 #SSLCertificateFile    /path/to/your/SSL-certificate.crt
 #SSLCertificateKeyFile /path/to/your/SSL-key.key
--- a/etc/nginx-conf.in
+++ b/etc/nginx-conf.in
@@ -41,12 +41,11 @@ server {
 # 	listen   443;
 # 	listen   [::]:443;
 
+    # See https://ssl-config.mozilla.org/ to generate good TLS settings for your server
 # 	ssl on;
 # 	ssl_certificate /path/to/your/SSL-certificate-with-chain.crt;
 # 	ssl_certificate_key /path/to/your/SSL-key.key;
 # 	ssl_session_timeout 5m;
-# 	ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
-#	ssl_ciphers ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA256:ECDHE-RSA-AES256-SHA:ECDHE-RSA-AES128-SHA:DHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-SHA256:DHE-RSA-AES128-SHA256:DHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA:AES256-GCM-SHA384:AES128-GCM-SHA256:AES256-SHA256:AES128-SHA256:AES256-SHA:AES128-SHA:DES-CBC3-SHA;
 # 	ssl_prefer_server_ciphers on;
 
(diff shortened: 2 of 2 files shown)


thijskh:
Is it necessary to explicity have this in our exampe config? Can't we rely on sane defaults in Nginx for these options?

mvr320:
> Is it necessary to explicity have this in our exampe config at all? Can't we rely on sane defaults in Nginx for these options?

Not sure how to interpret that question, with the same reasoning that you dont give these options with apache you also do not have to give them with nginx I think.

Do you prefer a PR which removes the whole setting of the SSL security? On itself the structure does make sense I think as you have the 426 advice, something on the STS so it makes sense to reset the security to the level there (and maybe leave legacy stuff also running on the server with a lower version.... courseware might never be updated??)

thijskh:
So yes my question is why we explicitly in DOMjudge need to add instructions on specific TLS versions and cipher suites, and indeed if we leave them out whether the defaults are reasonable enough for people not to shoot themselves in the foot. I don't use nginx so can't say.

We could add a plug for https://ssl-config.mozilla.org/ instead as a comment in both the Apache and nginx snippets: "# See https://ssl-config.mozilla.org/ to generate good TLS settings for your server"?

nickygerritsen:
> So yes my question is why we explicitly in DOMjudge need to add instructions on specific TLS versions and cipher suites, and indeed if we leave them out whether the defaults are reasonable enough for people not to shoot themselves in the foot. I don't use nginx so can't say.

I believe the nginx defaults are 'allow everything #yolo'

> We could add a plug for https://ssl-config.mozilla.org/ instead as a comment in both the Apache and nginx snippets: "# See https://ssl-config.mozilla.org/ to generate good TLS settings for your server"?

This sounds like a good idea.

mvr320:
> > So yes my question is why we explicitly in DOMjudge need to add instructions on specific TLS versions and cipher suites, and indeed if we leave them out whether the defaults are reasonable enough for people not to shoot themselves in the foot. I don't use nginx so can't say.
> 
> I believe the nginx defaults are 'allow everything #yolo'
> 
> > We could add a plug for https://ssl-config.mozilla.org/ instead as a comment in both the Apache and nginx snippets: "# See https://ssl-config.mozilla.org/ to generate good TLS settings for your server"?
> 
> This sounds like a good idea.

Atleast SSLv3 is disabled in the standard config, but everything as TLSv1* is indeed enabled (atleast for 20.04)
ssl_protocols TLSv1 TLSv1.1 TLSv1.2 TLSv1.3; # Dropping SSLv3, ref: POODLE

But changing this would also downgrade the docker contain setup I think, so I'll also make a PR there with the settings I'll remove here (or the intermediate level recommended by Mozilla).

mvr320:
I'm also not sure if the section about HTTP426 will give more security, I think its asking for an upgrade inside the HTTP/* and not forcing a browser to go to HTTPS
https://httpstatuses.com/426

But I assume people will know this when fiddling with HTTP codes,

thijskh:
The 426 is there to make it fail explicitly when using plain http, instead of silently trying to correct it. Problem with correcting it is that it is not transparant why some things are failing, and when e.g. your API client is mistakingly configured with http://, it will be redirected to HTTPS but you're sending your basic auth credentials cleartext on every request. So we want to make explicit that any http:// request is wrong.

thijskh:
If the defaults of nginx are not sane and we will be using this in Docker anyway, I see the value of having the config inside the file even if it means we need to keep it up to date with advancing insights. So I'm not sure there's much value to moving it to the Docker build step if in the end it turns out that DOMjudge will be shipping default TLS settings one way or the other.

So maybe better to go back to the original plan, perhaps add the mozilla link, and check if the settings are up to date with an Intermediate profile?

mvr320:
> If the defaults of nginx are not sane and we will be using this in Docker anyway, I see the value of having the config inside the file even if it means we need to keep it up to date with advancing insights. So I'm not sure there's much value to moving it to the Docker build step if in the end it turns out that DOMjudge will be shipping default TLS settings one way or the other.
> 
> So maybe better to go back to the original plan, perhaps add the mozilla link, and check if the settings are up to date with an Intermediate profile?

I was wrong, the docker images use the plain HTTP connection, so you put another webserver in front of it.
which makes that this PR could be considered in this form.

PR #848, opened 2020-07-01
make baseurl end with "/"

mvr320:
When installing our new server I forgot to end the url with "/", and ended with broken team docs.

As this would typicly be found only during the contest it might be fine to force it into the format of the example url. Its echoed later in the install but at that point someone does not know that the format has to be respected.

I'm not a pro at autoconf but I think this should fix the problem in the configure file. If there is a better way with autoconf itself I'm fine with changing, I just couldn't find it in the docs howto.

### DIFF
--- a/doc/manual/substitutions.py.in
+++ b/doc/manual/substitutions.py.in
@@ -3,10 +3,10 @@
 rst_prolog = """
 .. |DOMjudge| replace:: @PACKAGE_NAME@
 
-.. |baseurlteam| replace:: @BASEURL@team
+.. |baseurlteam| replace:: @BASEURL@%steam
 
 .. |SOURCESIZE| replace:: 256
 .. |COMPILETIME| replace:: 30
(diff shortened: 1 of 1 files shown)


thijskh:
Probably you followed our manual which says `./configure --enable-static-linking --with-baseurl="https://yourhost.example.edu/domjudge"`

As far as I'm aware, it's only used in the `submit` client and in the team manual. submit is already forgiving: it adds the / at runtme when missing. Maybe we should do the same in `doc/maual/substitutions.py.in`. So that it will just always work.

mvr320:
@thijskh what is the reason to not prefer to force one format to the user?

But I will change the docs generation, and afterwards remove the change to the autoconf.

thijskh:
It might be a matter of preference. But since it's used in two places only, I would say that it's most robust to ensure that those two places can deal with either format. So that regardless of how you (managed to) configure it, it will work?

PR #850, opened 2020-07-02
Speed up inserting static events by checking all of them at once. Fixes #849.

### LINKED ISSUE #849: initStaticEvents currently breaks for a lot of affiliations
As reported on Slack, with 2600+ affiliations `initStaticEvents()` takes a long time.
We should investigate if this is something we can fix.

nickygerritsen:
So when you have a lot of static data, like say 3000 organizations, and you missed a dependency for an event, we checked each of these events individually and then flushed each of them to the database individually. This was very slow (it gave 30s timeouts). Instead, we can check all events at once and also only do one flush.

Now adding a clarification while not having any events yet went from this 30s timeout to 2.5s. I removed the old `checkExistingEvent` function since it's not used anymore. The old `insertEvent` function is still used, but I rewrote it to use the new array version.

### DIFF
--- a/webapp/src/Service/EventLogService.php
+++ b/webapp/src/Service/EventLogService.php
@@ -14,6 +14,7 @@ use Doctrine\Common\Inflector\Inflector;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\ORM\NonUniqueResultException;
+use Doctrine\ORM\Query\Expr\Join;
 use Exception;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\DependencyInjection\ContainerAwareInterface;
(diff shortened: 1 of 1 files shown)


nickygerritsen:
I can try but `addMissingStateEvents` adds at most 5 events (and in normal circumstances only 1) so not sure if we win a lot

thijskh:
Then probably it's indeed not worth it. Unless it makes the code clearer or something like that.

webapp/src/Service/EventLogService.php
```diff
@@ -811,33 +853,35 @@ protected function hasAllDependentObjectEvents(Contest $contest, string $type, a
...
             ->select('e')
             ->andWhere('e.contest = :contest')
             ->andWhere('e.endpointtype = :endpoint')
-            ->andWhere('e.endpointid = :endpointid')
-            ->setParameter(':contest', $event->getContest())
-            ->setParameter(':endpoint', $event->getEndpointtype())
-            ->setParameter(':endpointid', $event->getEndpointid())
-            ->setMaxResults(1)
+            ->andWhere('e.endpointid IN (:endpointids)')
+            ->andWhere('e2.eventid IS NULL')
```
eldering, line 878: Why is this there? And I'm wondering how this code does the same thing as the original single-event one when the check for DELETE is missing.

nickygerritsen: * The e2 is there because we want the last event for every endpointid. If you know a better way then doing a LEFT JOIN with all events after the event and it being NULL, we can do this differently.
* DELETE check: ha you're right, I totally missed the DELETE thing. I think the easiest thing to do is to filter from the array all events with ->getAction() === self::ACTION_DELETE before returning the array.

The DELETE thing never popped up in the CI since we almost never delete stuff.

nickygerritsen: DELETE should be fixed now

PR #851, opened 2020-07-06
Add warning about the judgehost requiring cgroups

nickygerritsen:


nickygerritsen:
> Do we really need this? Is WSL Linux? If not, then I think this is already clear from the fact that we specify Linux as a requirement for the judgehosts, and that cgroups (and libcgroup) are explicitly mentioned.

WSL is full Linux, it has a kernel, it can run X, etc.: https://docs.microsoft.com/en-us/windows/wsl/about . It is 'kinda' like Linux KVM but then on Windows. So I think if we say 'you need Linux', you could see that as that WSL is supported.

eldering:
> WSL is full Linux, it has a kernel, it can run X, etc.: https://docs.microsoft.com/en-us/windows/wsl/about . It is 'kinda' like Linux KVM but then on Windows. So I think if we say 'you need Linux', you could see that as that WSL is supported.

Ok, but at least the virtualization layer seems to miss something, if cgroups can't be used. Anyways, ok, fair enough to add a warning.

nickygerritsen:
> > WSL is full Linux, it has a kernel, it can run X, etc.: https://docs.microsoft.com/en-us/windows/wsl/about . It is 'kinda' like Linux KVM but then on Windows. So I think if we say 'you need Linux', you could see that as that WSL is supported.
> 
> Ok, but at least the virtualization layer seems to miss something, if cgroups can't be used. Anyways, ok, fair enough to add a warning.

Maybe change the warning that in general you need cgroups for the judgehost? Something like:

```
Please note that the judgehost requires support for the memory cgroup
and swap accounting. Platforms that don't support this (for example WSL 1 / 2)
are not supported for the judgehost.
```

Then we make the message a bit more general?

thijskh:
Last approach seems good (focus on the required functionality, not a specific product).

Suggested improvement to the text:
```
The judgehost requires Linux cgroup support for memory and swap accounting.
Platforms that do not provide this support (some virtualization environments, for example WSL 1 / 2)
will not work with the judgehost.
```

nickygerritsen:
Updated

PR #859, opened 2020-08-29
WIP: Delete multiple

mvr320:
This currently only works for the problems and breaks some of the other pages. 

I would like some feedback on if this looks good enough on the frontend side, and if the programming flow makes sense.

I do think before this is merged there should be first some good tests for this.


webapp/public/js/domjudge.js
```diff
@@ -674,3 +674,39 @@ function pinScoreheader()
 		}
 	}
 }
+
+function removeMultipleUpdater(self)
+{
+    /*var newUrl = '/jury/problems/';
```
meisterT, line 680: Is this from debugging?

webapp/public/js/domjudge.js
```diff
@@ -674,3 +674,39 @@ function pinScoreheader()
...
+        }
+    });
+    newUrl = newUrl.slice(0,-1)+'/delete';
+    if(selectedNodes.length>0){
+        window.location = newUrl;
+    }*/
+    var oldUrl = $('#deleteMultiple').attr('href');
+    var postFix = '/delete'
+    var newUrl = ''
+    // We need to either append ,\d or remove ,\d
```
meisterT, line 694: I don't understand this comment

webapp/public/js/domjudge.js
```diff
@@ -674,3 +674,39 @@ function pinScoreheader()
...
+    if(self.checked) {
+        if (oldUrl==''){
+            oldUrl = '/jury/problems//delete';
```
meisterT, line 697: nit: double slash

Also below

webapp/public/js/domjudge.js
```diff
@@ -674,3 +674,39 @@ function pinScoreheader()
...
+    newUrl = newUrl.slice(0,-1)+'/delete';
+    if(selectedNodes.length>0){
+        window.location = newUrl;
+    }*/
+    var oldUrl = $('#deleteMultiple').attr('href');
+    var postFix = '/delete'
+    var newUrl = ''
+    // We need to either append ,\d or remove ,\d
+    if(self.checked) {
+        if (oldUrl==''){
```
meisterT, line 696: not: space before {

(Also below)

webapp/public/js/domjudge.js
```diff
@@ -674,3 +674,39 @@ function pinScoreheader()
 		}
 	}
 }
+
```
meisterT, line 677: Please add a high-level description what's the goal of this method.

webapp/public/js/domjudge.js
mvr320: In this case I make use of that, basicly it becomes:
'/jury/problems/'+problemId+'/delete'

In the final code I'll just make sure the if statement is moved in such a way that it doesn't bother you anymore.

nickygerritsen:
General question: since we do JavaScript anyway, why not turn it into a post / delete with a form or json body? Then we don’t have to do all the stuff with x,y,z

mvr320:
> General question: since we do JavaScript anyway, why not turn it into a post / delete with a form or json body? Then we don’t have to do all the stuff with x,y,z

Basicly I tried to replicate the current method which has such a route.
The POST is the actual applying whereas the GET is the confirmation, 

I'm perfectly fine with altering, but than the other actions should also be rewritten I think. I was planning to later change the functions I now basicly copied to just special cases of a list, namely the list with only 1 item.


Or maybe I do not understand your suggestion, @nickygerritsen

nickygerritsen:
I think you understand what I mean indeed. I wonder if it will make it nicer if we do it that way. Although probably with a bit of cleanup we can get your setup nice as well

webapp/public/js/domjudge.js
```diff
@@ -674,3 +674,32 @@ function pinScoreheader()
...
+function removeMultipleUpdater(self, baseRoute)
```
nickygerritsen, line 678: Instead of all these replaces, why not do the following:

* Save somewhere what the 'general' URL to the delete page is. Something like you already have as argument to this function but them maybe the whole url with a `_REPLACE_ME_` inside it.
* Keep track of an array of items to delete
* `var url = urlTemplate.replace('_REPLACE_ME_', items.join(','))`

Then we don't have all these strange replaces with ,, and ',

nickygerritsen:
I checked out the UI and have an idea. Note that this is a discussion point so I'm not suggesting we change it, just want some opinions: perhaps instead of displaying a trash icon in the table header, we display a red, disabled buton next to the blue '+' button below the table. By default it has the text 'Select rows to delete' and when you have selected one or more rows, it displays 'Delete x problems' and gets enabled. Alternatively we hide the trash icon until you select at least one item, but then probably we also need to set an alt text or a hover with something like 'delete multiple problems' and maybe make it red?

mvr320:
> I checked out the UI and have an idea. Note that this is a discussion point so I'm not suggesting we change it, just want some opinions: perhaps instead of displaying a trash icon in the table header, we display a red, disabled buton next to the blue '+' button below the table. By default it has the text 'Select rows to delete' and when you have selected one or more rows, it displays 'Delete x problems' and gets enabled. Alternatively we hide the trash icon until you select at least one item, but then probably we also need to set an alt text or a hover with something like 'delete multiple problems' and maybe make it red?

I structured the code in such a way that there can be more of these actions over columns, having a button for everything might than be ugly. But its not hard to fix, so I can easily change the layout if more people agree with @nickygerritsen. I thought of having also clarifications to multiple teams for example (everyone in a room for example) where an additional colaction could make sense.

mvr320:
> 
> 
> I checked out the UI and have an idea. Note that this is a discussion point so I'm not suggesting we change it, just want some opinions: perhaps instead of displaying a trash icon in the table header, we display a red, disabled buton next to the blue '+' button below the table. By default it has the text 'Select rows to delete' and when you have selected one or more rows, it displays 'Delete x problems' and gets enabled. Alternatively we hide the trash icon until you select at least one item, but then probably we also need to set an alt text or a hover with something like 'delete multiple problems' and maybe make it red?

@thijskh I want to revive this PR again, which UI do you prefer for this?

PR #861, opened 2020-08-31
Add public / team stats. Fixes #429.

### LINKED ISSUE #429: Add public stats to team/public problems page
It would be great to add a subset of the now only jury available stats also to the team and public pages.

I'm thinking of the submission graph over time (but all rejected verdicts bundled into one, of course nothing after the freeze) and the total number of attempts/corrects.

nickygerritsen:
This moves the stats logic to a service, so we can reuse it on the team/public pages. 

On the team/public problem pages it shows this if enabled:
image

### DIFF
--- a/etc/db-config.yaml
+++ b/etc/db-config.yaml
@@ -240,6 +240,11 @@
             default_value: 0
             public: false
             description: Maximum width of team column on scoreboard. Leave 0 for no maximum.
+        -   name: show_public_stats
+            type: bool
+            default_value: true
+            public: true
+            description: Show submission and problem statistics on the team and public pages.
 -   category: Misc
     description: Miscellaneous configuration options.
     items:
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -674,3 +674,7 @@ function pinScoreheader()
 		}
 	}
 }
+
+$(function () {
+    $('[data-toggle="tooltip"]').tooltip();
+});
--- a/webapp/public/style_domjudge.css
+++ b/webapp/public/style_domjudge.css
@@ -436,3 +436,136 @@ h3.teamoverview {
     padding: 40px 15px;
     text-align: center;
 }
+
+.problem-stats {
+    padding-top: 10px;
+}
+
+.problem-stats .problem-stats-item {
+    display: inline-block;
+    margin-right: 3px;
+    width: 10px;
+    height: 10px;
+    border-radius: 2px;
+    box-shadow: inset 0 0 0 1px rgba(27, 31, 35, .04);
+}
+
+.problem-stats-item.correct-0 {
+    background-color: #eeeeee;
+}
+
+.problem-stats-item.correct-1 {
+    background-color: #dcffe4;
+}
+
+.problem-stats-item.correct-2 {
+    background-color: #bef5cb;
+}
+
+.problem-stats-item.correct-3 {
+    background-color: #85e89d;
+}
+
+.problem-stats-item.correct-4 {
+    background-color: #34d058;
+}
+
+.problem-stats-item.correct-5 {
+    background-color: #28a745;
+}
+
+.problem-stats-item.correct-6 {
+    background-color: #22863a;
+}
+
+.problem-stats-item.correct-7 {
+    background-color: #176f2c;
+}
+
+.problem-stats-item.correct-8 {
+    background-color: #165c26;
+}
+
+.problem-stats-item.correct-9 {
+    background-color: #144620;
+}
+
+.problem-stats-item.incorrect-0 {
+    background-color: #eeeeee;
+}
+
+.problem-stats-item.incorrect-1 {
+    background-color: #ffdce0;
+}
+
+.problem-stats-item.incorrect-2 {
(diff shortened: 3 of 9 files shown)


thijskh:
So the stats are in the team interface but not related to the specific team?

nickygerritsen:
Yes since there is no easy way to go to the public interface if you are logged in

thijskh:
Is it fast enough to expose it to a larger audience? I.e., does it only depend on the scorecache tables?

webapp/templates/partials/analysis/contest_stats.html.twig
```diff
@@ -0,0 +1,26 @@
...
+                Calculated as:<br>
+                average across all teams of [3 people x (number of minutes elapsed - minute last problem solved by team t)]
+              ">Misery
+                    index</abbr>: {{ misc.misery_index|number_format(0) }} minutes
+                <br>
+            {% endif %}
+            # Submissions: {{ misc.total_submissions }}<br>
+            # Accepted Submissions: {{ misc.total_accepted }} <span
+                class="text-muted">(~{{ (misc.total_submissions>0 ? (misc.total_accepted/misc.total_submissions)*100:0)|number_format(0) }}%)</span><br>
+            # of Teams: {{ misc.num_teams }}<br>
```
thijskh, line 23: The third bullet is the only one that has the word "or" in it. Is the "#" even needed or could it just say: "Submissions: 1353"?

nickygerritsen: Note that this didn't change in this PR. But I can agree we can remove the # (and the 'of')

nickygerritsen:
> Is it fast enough to expose it to a larger audience? I.e., does it only depend on the scorecache tables?

It does not use scorecache, since that doesn't contain timing information for all but the last-or-correct submission.

However, for the NWERC data it loads in ~400 ms in debug mode on my laptop in Docker, so it seems to be not that slow.

meisterT:
I am wondering whether we should just fold this into the problemset page instead?

nickygerritsen:
> I am wondering whether we should just fold this into the problemset page instead?

Place it below the problems there? That might make sense

meisterT:
Or the data for each problem in the box for that problem.

nickygerritsen:
So only attempted / correct as numbers? And maybe correct %?
And what about the contest stats and submissions over time? Don't add these?

meisterT:
> So only attempted / correct as numbers? And maybe correct %?

Perhaps a pie chart with absolute numbers? After the results are revealed this could change from incorrect/correct to actual verdicts.


> And what about the contest stats and submissions over time? Don't add these?

For the submissions over time: I could imagine to show a heat map like chart that shows the number of successful submissions for each problem as well.
Each problem box is right now ~300px wide (on my screen at least), so use divide this into X little boxes each ~10px wide (each representing a block of time) and show with a color gradient how many successful submissions there were in that bucket of time, similar to
image

Of course I am just thinking out loud - feel free to ignore me and go ahead with what you already had planned.

nickygerritsen:
I do like the blocks. Do we also want to show somehow which ones (or how many) of them are correct?
For the pie charts: I'm not sure if that is useful? It probably takes more space than 2/3 numbers so what does it add?

Also thinking out loud

meisterT:
What about two rows of blocks then -  one for correct and one for incorrect submissions? (and the raw numbers somewhere)

nickygerritsen:
Let me try to build it, sounds like it might look good

eldering:
> > Is it fast enough to expose it to a larger audience? I.e., does it only depend on the scorecache tables?
> 
> It does not use scorecache, since that doesn't contain timing information for all but the last-or-correct submission.
> 
> However, for the NWERC data it loads in ~400 ms in debug mode on my laptop in Docker, so it seems to be not that slow.

That sounds ok to me as long as it's on a page that doesn't get viewed a lot or has a high refresh rate: 400ms is a lot if this gets hit often in a big contest with 200+ teams, say. So problems overview page sounds fine.

nickygerritsen:
If we do @meisterT his plan it might be that the queries are faster as well. I will check

nickygerritsen:
Updated code, update screenshot in description to reflect new code.

nickygerritsen:
Ah and for timing: the page loads in 250 ms while it loads in 50 ms without the stats, so the speed has increased indeed.

meisterT (approved):

webapp/src/Twig/TwigExtension.php
```diff
@@ -150,6 +150,7 @@ public function getGlobals()
             'refresh_flag' => $refresh_flag,
             'icat_url' => defined('ICAT_URL') ? ICAT_URL : null,
             'external_ccs_submission_url' => $this->config->get('external_ccs_submission_url'),
+            'show_public_stats' => $this->config->get('show_public_stats'),
```
thijskh, line 153: Why is this extra global needed? The template can just see if stats is filled?

nickygerritsen: You are right. This was needed before when it was a separate page. I will change it

thijskh (approved):

PR #865, opened 2020-09-06
Build executables in chroot (rebased after merge of chroot PR)

eldering:
This is an updated version of #823 rebased to current master after #798 was merged.

### DIFF
--- a/judge/build_executable.sh
+++ b/judge/build_executable.sh
@@ -1,0 +1,79 @@
+#!/bin/sh
+
+# Script to build executables.
+#
+# Usage: $0 <workdir>
+
+set -e
+trap error EXIT
+
+cleanexit ()
+{
+	trap - EXIT
+
+	$DJ_LIBJUDGEDIR/chroot-startstop.sh stop
+
+	# Make sure that all files are owned by the current user/group, so
+	# that we can delete the judging output tree without root access.
+	# We also remove group RUNGROUP so that this can safely be shared
+	# across multiple judgedaemons, and remove write permissions.
+	$GAINROOT chown -R "$(id -un):" $WORKDIR
+	chmod -R go-w $WORKDIR
+
+	logmsg $LOG_DEBUG "exiting, code = '$1'"
+	exit $1
+}
+
+# Error and logging functions
+# shellcheck disable=SC1090
+. "$DJ_LIBDIR/lib.error.sh"
+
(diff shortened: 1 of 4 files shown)


judge/judgedaemon.main.php
```diff
@@ -241,12 +241,13 @@ function read_judgehostlog(int $n = 20) : string
...
+    $execpath = "$workdirpath/executable/download/" . $execid;
     $execmd5path = $execpath . "/md5sum";
-    $execdeploypath = $execpath . "/.deployed";
-    $execbuildpath = $execpath . "/build";
-    $execrunpath = $execpath . "/run";
     $execzippath = $execpath . "/executable.zip";
+    $execdeploypath = $execpath . "/.deployed";
+    $execbuilddir = "$workdirpath/executable/build/" . $execid;
+    $execbuildpath = $execbuilddir . "/build";
+    $execrunpath = $execbuilddir . "/run";
```
eldering, line 250: I propose the change the directory structure a bit: make subdirectories `download` and `build` inside a directory `$workdirpath/executable/$execid/`. I think that's a more natural grouping.
Secondly, this makes it easier to make the `build` directory a subdirectory of the chroot, which means that that `build` directory is now clean, and doesn't contain the `etc`, `usr`, `proc`, etc. directories from the chroot bind mounts, which could potentially cause conflicts.

judge/build_executable.sh
```diff
@@ -0,0 +1,88 @@
...
+
+	# Make sure that all files are owned by the current user/group, so
+	# that we can delete the judging output tree without root access.
+	# We also remove group RUNGROUP so that this can safely be shared
+	# across multiple judgedaemons, and remove write permissions.
+	$GAINROOT chown -R "$(id -un):" $WORKDIR
+	chmod -R go-w $WORKDIR
+#	for i in bin etc lib lib64 proc usr; do
+#		rm -rf $i
+#	done
```
eldering, line 22: This commented out cleanup code from me can be removed once we change the directory structure, see below.

eldering (commented):
See comment about changing the directory structure, but otherwise looks ok to me.

I do think that this whole `fetch_executable` function should be cleaned up (e.g. that inline combined run/compare script moved to a separate file), but that can be done afterwards.

thijskh:
Agreed with externalizing the script, I did not take it on yet because of the scope of the PR.

PR #866, opened 2020-09-14
Test to see if exposed URLs are only able to be seen by the correct role

mvr320:
As I was testing with UnitTests I had this as example. The test on itself works but could be more extensive. If people think this is useful I'm fine with extending it to have a higher quality.

Interesting is the /jury/change-contest route which is usable for the team role. On itself its not wrong but I think it doesn't need to be allowed?

### DIFF
--- a/webapp/tests/Controller/ControllerRolesTest.php
+++ b/webapp/tests/Controller/ControllerRolesTest.php
@@ -1,0 +1,160 @@
+<?php declare(strict_types=1);
+
+namespace App\Tests\Controller;
+
+use App\Tests\BaseTest;
+
+class ControllerRolesTest extends BaseTest
+{
+    protected static $roles = [];
+
+    /**
+     * //See: https://www.oreilly.com/library/view/php-cookbook/1565926811/ch04s25.html
+     * Get all combinations of roles with at minimal the starting roles
+     * @return array $results
+     * @var string[] $possible_roles
+     * @var string[] $start_roles
+     */
+    protected function roleCombinations(array $start_roles, array $possible_roles)
+    {
+        // initialize by adding the empty set
+        $results = array($start_roles);
+
+        foreach ($possible_roles as $element) {
+            foreach ($results as $combination) {
+                array_push($results, array_merge(array($element), $combination));
+            }
+        }
+        return $results;
+    }
+
+    /**
+     * Crawl the webpage assume this is allowed and return all other links on the page
+     * @return string[] $urlsToCheck
+     * @var string $url
+     * @var int $statusCode
+     */
+    public function crawlPage(string $url, int $statusCode)
+    {
+        $crawler = $this->client->request('GET', $url);
+        $response = $this->client->getResponse();
+        $message = var_export($response, true);
+        $this->assertEquals($statusCode, $response->getStatusCode(), $message);
+        return array_unique($crawler->filter('a')->extract(['href']));
+    }
+
+    /**
+     * Test that having the role(s) gives access to all visible pages.
+     * This test should detect mistakes where a page is disabled when the user has a
+     * certain role instead of allowing when the correct role is there.
+     * @var string[] $combinations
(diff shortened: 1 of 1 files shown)


nickygerritsen (changes requested):

webapp/tests/Controller/ControllerRolesTest.php
```diff
@@ -0,0 +1,202 @@
+<?php declare(strict_types=1);
+
+namespace App\Tests\Controller;
+
+use App\Tests\BaseTest;
+
+class ClarificationControllerAccessTest extends BaseTest
```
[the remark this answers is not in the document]
mvr320: Good catch, I assumed PHPStorm would tell me about these problems.

webapp/tests/Controller/ControllerRolesTest.php
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    /**
+     * Test that having the jury role is enough to view jury pages.
+     * This test should detect mistakes where a page is disabled when the user has a
+     * certain role instead of allowing when the correct role is there.
+     */
+    public function testJuryAccess()
+    {
+        $roleBaseUrl = '/jury';
```
nickygerritsen, line 78: Would it be better if we combine these three functions (and the same for the last set) and use a `@dataProvider`? They contain quite a bit of duplication now.

mvr320: The reason I chose this was as how I understand UnitTesting you should have a test not do multiple tests. The current code already does more than one thing but having the multiple roles might really create one test which does all things.

I will look into the @dataProvider to see what it actually does. If its indeed this: https://symfony.com/doc/current/testing.html than I think your solution is much better than the current code.

webapp/tests/Controller/ControllerRolesTest.php
```diff
@@ -0,0 +1,202 @@
...
+        $results = array($start_roles);
+
+        foreach ($possible_roles as $element)
+            foreach ($results as $combination)
+                array_push($results, array_merge(array($element), $combination));
+
+        return $results;
+    }
+
+    protected static $roles = [];
```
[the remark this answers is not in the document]
mvr320: I agree with the preference.

webapp/tests/Controller/ControllerRolesTest.php
```diff
@@ -0,0 +1,160 @@
...
+            }
+        }
+        while (array_diff($done,$urlsToCheck));
+
+        // Now check the rights of our user with the current role
+        static::$roles = $role;
+        $this->logIn();
+        $urlsToCheckRole = $this->crawlPage($roleBaseURL, 200);
+        foreach (array_diff($urlsToCheck, $urlsToCheckRole) as $url) {
+            print($url."\n");
```
nickygerritsen, line 135: I assume this is debug?

webapp/tests/Controller/ControllerRolesTest.php
```diff
@@ -0,0 +1,160 @@
...
+    /**
+     * Test that having the team role for example is enough to view pages of that role.
+     * This test should detect mistakes where a page is disabled when the user has a
+     * certain role instead of allowing when the correct role is there.
+     * @var string $roleBaseURL The standard endpoint from where the user traverses the website
+     * @var string[] $baseRoles The standard role of the user
+     * @var string[] $optionalRoles The roles which should not restrict the viewable pages
+     * @dataProvider provideBasePages
+     */
+    public function RoleAccess(string $roleBaseURL, array $baseRoles, array $optionalRoles)
```
nickygerritsen, line 80: You probably removed `test` by accident in the beginning of this method

mvr320:
Locally this now correctly works. For the reviewer, there are some tests which are excluded as I'm not sure about what should be the correct result. I think its better to have the tests now and later check if the breaking parts can be fixed.

PR #867, opened 2020-09-15
Test to see if exposed URLs are only able to be seen by the correct role

mvr320:
Apparantly I cannot reopen after force-push.

This should fix the problems of https://github.com/DOMjudge/domjudge/pull/866

Locally this now correctly works. For the reviewer, there are some tests which are excluded as I'm not sure about what should be the correct result. I think its better to have the tests now and later check if the breaking parts can be fixed.

### DIFF
--- a/webapp/tests/Controller/ControllerRolesTest.php
+++ b/webapp/tests/Controller/ControllerRolesTest.php
@@ -1,0 +1,230 @@
+<?php declare(strict_types=1);
+
+namespace App\Tests\Controller;
+
+use App\Tests\BaseTest;
+
+class ControllerRolesTest extends BaseTest
+{
+    protected static $roles = [];
+    protected static $loginURL = "http://localhost/login";
+
+
+    /**
+     * //See: https://www.oreilly.com/library/view/php-cookbook/1565926811/ch04s25.html
+     * Get all combinations of roles with at minimal the starting roles
+     *
+     * @var string[] $start_roles Base roles for each combination
+     * @var string[] $possible_roles Additional roles to add to the combination
+     * @return array All possible combinations with at least the starting roles
+     */
+    protected function roleCombinations(array $start_roles, array $possible_roles)
+    {
+        // initialize by adding the empty set
+        $results = array($start_roles);
+
+        foreach ($possible_roles as $element) {
+            foreach ($results as $combination) {
+                array_push($results, array_merge(array($element), $combination));
+            }
+        }
+        return $results;
+    }
+
+    /**
+     * Some URLs are not setup in the testing framework or have a function for the
+     * user UX/login process, those are skipped.
+     * @var string $url
+     * @return boolean
+     **/
+    protected function urlExcluded(string $url)
+    {
+        return ($url[0] == '#' ||                                          // Links to local page
+            strpos($url, 'http') !== false ||                        // External links
+            substr($url, 0, 4) == '/doc' ||                     // Documentation is not setup
+            substr($url, 0, 4) == '/api' ||                     // API is not functional in framework
+            strpos($url, '/delete') !== false ||                     // Breaks MockData
+            strpos($url, '/add') !== false ||                        //TODO: Should be fixable
+            strpos($url, '/edit') !== false ||
+            $url == '/logout' ||                                           // Application links
+            $url == '/login' ||
+            strpos($url, 'activate') !== false ||
+            strpos($url, 'deactivate') !== false ||
+            strpos($url, '/jury/change-contest/') !== false ||
+            strpos($url, '/text') !== false ||                       // Pdfs/text/binaries dirty the report
+            strpos($url, '/input') !== false ||                      // TODO: mimetype from the headers
+            strpos($url, '/output') !== false ||
+            strpos($url, '/export') !== false ||
+            strpos($url, '/download') !== false ||
+            strpos($url, '.zip') !== false
+        );
+    }
+
+    /**
+     * Crawl the webpage assume this is allowed and return all other links on the page
+     * @var string $url URL to crawl
+     * @var int $statusCode Expected HTTP status code
+     * @return string[] Found links on crawled URL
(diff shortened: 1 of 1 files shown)


webapp/tests/Controller/ControllerRolesTest.php
```diff
@@ -0,0 +1,221 @@
...
+
+    /**
+     * Crawl the webpage assume this is allowed and return all other links on the page
+     * @return string[] $urlsToCheck
+     * @var string $url
+     * @var int $statusCode
+     */
+    public function crawlPage(string $url, int $statusCode)
+    {
+        // We exclude some possible urls to not break the MockData
```
thijskh, line 66: Why are there two plaes in this file that list URL parts to exclude (also urlExclude()) with somewhat the same content?

webapp/tests/Controller/ControllerRolesTest.php
```diff
@@ -0,0 +1,221 @@
...
+        if (strpos($url, '/delete') !== false) {
+            return [];
+        }
+        // Documentation is not setup in the UnitTesting framework
+        if (substr($url, 0, 4) == '/doc') {
+            return [];
+        }
+        //Remove links to local page, external or application links
+        if ($url[0] == '#' || $url[0] == 'h' || $url == '/logout') {
```
thijskh, line 75: The significance of the letter `h` eludes me. Should really be documented what we're doing here.

webapp/tests/Controller/ControllerRolesTest.php
```diff
@@ -0,0 +1,221 @@
...
+        $this->verifyAccess($combinations, $urlsToCheck);
+    }
+
+    public function provideBasePages()
+    {
+        return [
+            ['/jury', ['admin'],    ['jury','team'], false],
+            ['/jury', ['jury'],     ['admin','team'], false],
+            ['/team', ['team'],     ['admin','jury'], true]
+        ];
```
thijskh, line 171: Really must comment here what this matrix means. Same below.

webapp/tests/Controller/ControllerRolesTest.php
nickygerritsen: Yeah I'd say we should try to merge these two cases? I think it doesn't hurt to exclude the same URL's in both cases?

webapp/tests/Controller/ControllerRolesTest.php
```diff
@@ -0,0 +1,234 @@
...
+        $this->logIn();
+        $urlsToCheck = $this->crawlPage($roleBaseURL, 200);
+        if ($allPages) {
+            $urlsToCheck = $this->getAllPages($urlsToCheck);
+        }
+        $combinations = $this->roleCombinations($baseRoles, $optionalRoles);
+        $this->verifyAccess($combinations, $urlsToCheck);
+    }
+
+    /** The test needs the data of which role to test with its base endpoint and the other optional roles/
```
nickygerritsen, line 165: Can you put this first comment line on the next line, i.e. put `/**` on its own line?

webapp/tests/Controller/ControllerRolesTest.php
```diff
@@ -0,0 +1,234 @@
...
+     * We than check for the role admin which has as base endpoint
```
nickygerritsen, line 166: Nitpick: then

webapp/tests/Controller/ControllerRolesTest.php
```diff
@@ -0,0 +1,234 @@
...
+     * /jury if any combination with the other roles does forbide us from entering a page which the role can see
+     * if its the only role.
+     * For the first row: admin endpoint, [RoleName], [Other Optional roles], no traversing found links
```
nickygerritsen, line 169: Proposal: rename the function to `provideRoleAccessData`, so it is clear it is used in `testRoleAccess`.

Next, instead of returning one big array, we can use yield:
```php
yield ['/jury', ['admin'],    ['jury','team'], false];
yield ['/jury', ['jury'],     ['admin','team'], false];
// etc.
```

Then change the docblock to:
```php
/**
 * Data provider used to test role access. Each item contains:
 * - the page to visit
 * - the base roles the user has
 * - additional roles to add to the user
 * - Whether to also recursively visit linked pages
 */
```

nickygerritsen (changes requested):
Looks good in general. Added some small comments.

webapp/tests/Controller/ControllerRolesTest.php
mvr320: By merging you would remove the failing parts of one test also from the other test where this might not fail right? Its definitely doable and I suspect that our tests are run on the same Mockdata? So one test changing state will have effect on other tests?

webapp/tests/Controller/ControllerRolesTest.php
mvr320: Should be fixed

mvr320: https://github.com/DOMjudge/domjudge/pull/867#discussion_r491326657
Should also be fixed, not 100% sure if we do now not check less but for a general method which crawls everything staying safe might be better.

webapp/tests/Controller/ControllerRolesTest.php
mvr320: Should be fixed by implementing the changes Nicky requested

webapp/tests/Controller/ControllerRolesTest.php
mvr320: external url's start with the char 'h' and I agree it was poorly documented.

webapp/tests/Controller/ControllerRolesTest.php
mvr320: @nickygerritsen is now implemented.

thijskh (approved):

nickygerritsen (approved):

PR #870, opened 2020-09-30
Add full name to self-registration form; close #868

### LINKED ISSUE #868: Self-registration form should allow entering the full name

Currently only the username and email fields can be entered in the self-registration form, but not the name field (for the full name).

In contrast, creating users through the admin interface make it possible to enter both fields.

I suggest to add the field for the "name" to the self-registration form, make it optional, and if not filled-in, use the username value, which should mimic the current behaviour.

### Your environment
- Domjudge 7.3.0

### Steps to reproduce
- Ensure there is a team category with self registration enabled
- Press login
- Click "Register now".

ykazakov:
If full name is not specified,
username will be used for full name

### DIFF
--- a/webapp/src/Controller/SecurityController.php
+++ b/webapp/src/Controller/SecurityController.php
@@ -132,7 +132,9 @@ class SecurityController extends AbstractController
             $plainPass = $registration_form->get('plainPassword')->getData();
             $password  = $passwordEncoder->encodePassword($user, $plainPass);
             $user->setPassword($password);
-            $user->setName($user->getUsername());
+            if ($user->getName() == null) {
+                $user->setName($user->getUsername());
+            }
             $user->addUserRole($team_role);
 
             $teamName = $registration_form->get('teamName')->getData();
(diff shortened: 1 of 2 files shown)


nickygerritsen (approved):

webapp/src/Controller/SecurityController.php
```diff
@@ -132,7 +132,9 @@ public function registerAction(
             $plainPass = $registration_form->get('plainPassword')->getData();
             $password  = $passwordEncoder->encodePassword($user, $plainPass);
             $user->setPassword($password);
-            $user->setName($user->getUsername());
+            if ($user->getName() == null) {
```
thijskh, line 135: Please use strict comparison with `===`

ykazakov: Thanks! Will make a patch in a minute.
(sorry, not an expert in php)

ykazakov: @thijskh just curious: when it makes a different in this situation? I just read that 0 == null returns true and 0 === null returns false, but in our case the name cannot be an integer, can it?

thijskh: So indeed it might not matter much in this specific case. But the policy is to be as type strict as possible in our code. So you're not dependent on some judgement that for this case it "would not matter". Just do it always.

ykazakov: I see. Actually now I see now that there is a function empty(..) to check if the field is empty (e.g., in validation constraints). Should probably use this instead for the sake of consistency.

Issue #869, opened 2020-09-30
Add a possibility to join an existing team for self-registration


ykazakov:
### Description of the problem

In #561 a feature was added for a user to specify a team during the self-registration of users.
However, this form does not let the user to join an existing team. It would be nice to allow self-registered users to join teams.

### Your environment
- DOMjudge 7.3.0

### Steps to reproduce
- Ensure there is a team category with self registration enabled
- Press Login
- Click "Register now".
- Enter new login details with some team
- Press "Register"
- Press "Logout"
- Press "Login"
- Click "Register now".
- Enter new login details with the same team as in the previous step

### Expected behaviour
- A user joins the specified team

### Actual behaviour
- A error message pops up: "ERROR This team name is already in use."

### Any other information that you want to share?
I can see situations which may lead to some security problems, i.e., unwanted users joining teams during contests.
To avoid this situation (and mimic the old behaviour) one could, perhaps, add a flag to team categories to enable joining existing teams during self-registration. In addition, maybe allow the team members to "lock" the team so to prevent other members joining this team.

ykazakov:
I have just created a PR #871 for this issue .
In the patch, it is possible to join teams, which categories allow self registration. So, once self-registration is disabled in these categories, it is no longer possible to join these teams. Or, to prevent joining a particular team, it is sufficient to just change its category to other category which does not allow self-registration. I think it is a sensible solution.

meisterT:
What's the exact user journey for all people involved (original team member, new team member and contest organizer)?

I am worried that any random user could join any team. Without confirmation this sounds suboptimal (and a potential security issue). With confirmation the question is how we would implement confirmation: do the original team members accept the new team member or is that the job of the contest organizers?
We also don't have any email functionality built-in (whether that's a positive or negative is debatable), so currently there's no way to inform old team members that someone wants to join their team.
One could think of a token that the original team member could share with the new team member which they have to enter on registration if they want to join an existing team.

ykazakov:
I could imagine the following workflow:
- Admin creates a team category, which allows self-registration
- Users sign up, join or create teams
- Before the contest starts, admin checks if the teams are OK and disables self-registration

The only problem: currently the team members cannot see who else is in their team. Or am I missing something? This way, if some unintended user joins a team, they cannot report this to the organiser.

[...]

PR #871, opened 2020-10-02
Enable joining teams in self-registration; see #869

ykazakov:
Upon self-registration there will be an additional
option to either select a team to which the user
may join, or to crate a new team

Only teams which categories allow for self-registration
will be offered for selection.

### DIFF
--- a/webapp/src/Controller/Jury/TeamController.php
+++ b/webapp/src/Controller/Jury/TeamController.php
@@ -366,7 +366,10 @@ class TeamController extends BaseController
             throw new NotFoundHttpException(sprintf('Team with ID %s not found', $teamId));
         }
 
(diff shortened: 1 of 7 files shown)


thijskh (changes requested):
As per discussion in issue #869, not ready in current form

ykazakov:
@thijskh What about #872? You could pull out the commit for #869 if you think it is not ready.

mvr320:
@ykazakov what is your plan with this PR? Should it stay open?

ykazakov:
@mvr320 looks like the code has diverge significantly. I do not have time to resolve the conflicts, and there appears to be little interests in #869 and #872 anyway.

PR #873, opened 2020-10-03
Run html/css validation on src files

mvr320:
This scrapes the testinstance to plain html and checks the usage of HTML not in line with W3C recommendations. Currently it would only break when more errors are found than currently in the main branch.

Improvement is possible in a number of ways;
- Testing with a user which has a team role, As I think the admin user does not have this.
- Some code is duplicated from the integration tests, it would help if these are moved to another file which can than be sourced, I'm fine with trying something like that but would think that should be in another PR

The output is now divided into sections both for the file and its problems and another for a problem and which files this occurs as I think fixing /public should be the 1st one but fixing 1 problem at a time might be safer to not get differences in the UI.

All these problems on itself can be labeled as "Good first time" issues in my opinion.

If this is accepted I'll fixup to only 1 commit.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -26,6 +26,34 @@ check_w3c_css:
     - python3 -m json.tool < result
     - exit $RES
 
+check_w3c_html:
+  stage: test
+  before_script:
+    # Takes about 1 minute to pull
+    - echo -e "section_start:`date +%s`:docker\r\e[0Kdocker pull"
+    - time docker pull domjudge/gitlabci:1.0
+    - echo -e "section_end:`date +%s`:docker\r\e[0K"
+  image: docker:stable
+  services:
+    - docker:19-dind
+    - mariadb
+  variables:
+    MYSQL_ROOT_PASSWORD: password
+    DOCKER_HOST: tcp://docker:2375/
+    DOCKER_DRIVER: overlay2
+  script:
+  - |
+    docker run --privileged \
+      -v $CI_PROJECT_DIR:$CI_PROJECT_DIR \
+      --net=host \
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -26,6 +26,35 @@ check_w3c_css:
...
+    MYSQL_ROOT_PASSWORD: password
+    DOCKER_HOST: tcp://docker:2375/
+    DOCKER_DRIVER: overlay2
+  script:
+  - |
+    docker run --privileged \
+      -v $CI_PROJECT_DIR:$CI_PROJECT_DIR \
+      --net=host \
+      -e "TERM=xterm-256color" -e "HOME=$HOME" -e "USER=$USER" -e "MARIADB_PORT_3306_TCP_ADDR=$MARIADB_PORT_3306_TCP_ADDR" -e "MYSQL_ROOT_PASSWORD=$MYSQL_ROOT_PASSWORD" \
+      mvr320/djcihtml:latest /bin/bash -eo pipefail -c "umask 0002; cd $CI_PROJECT_DIR; script --return -qfc \"./gitlab/html.sh\" /dev/null | ts \"[%F %T]\" "
```
nickygerritsen, line 50: Probably better to push this under the domjudge user. Also we can probably keep the source in domjudge-script?

.gitlab-ci.yml
```diff
@@ -26,6 +26,35 @@ check_w3c_css:
...
+  artifacts:
+    when: always
+    paths:
```
nickygerritsen, line 53: Double paths

nickygerritsen (changes requested):

.gitlab-ci.yml
mvr320: Yes when the output was accepted I was going to start making PRs to the gitlabci project assuming thats the source you wanted there

mvr320: This one:
https://github.com/DOMjudge/domjudge-packaging/pull/63

nickygerritsen (approved):

PR #876, opened 2020-10-14
Demo w3c validation on CSS

mvr320:
The pipeline should break to demonstrate the output when something is wrong. So if this is acceptable I'll do a fixup to remove the demostage.

More PRs will be following for other tests independent on this one so this doesn't need to be merged directly. But approval might be nice for my tree...

I will fix the merge conflicts when the code is accepted.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,6 +15,17 @@ check syntax:
   script:
     - ./gitlab/syntax.sh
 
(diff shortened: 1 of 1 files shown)


mvr320:
An example of when there are mistakes can be found in:
https://gitlab.com/DOMjudge/domjudge/-/jobs/791455002

thijskh (approved):

PR #881, opened 2020-10-23
Clean up legacy libs

thijskh:
Remove now unused code from pre-Symfony DOMjudge.

More cleanup can be achieved by refactoring what is left, but that is deliberately left for later; for clarity this pull request is just the first (but in itself valuable) step: removing things that are not in use right now.

Some of this code has been present since the dawn of the project - we recognise its service to the community, so long and we salute you for all the fish.


meisterT (approved):
Deleted code is best code.

meisterT (approved):

nickygerritsen (approved):

eldering (approved):

PR #882, opened 2020-10-27
Update links to CCS specifications

thijskh:


meisterT:
Jaap, are these links already stable?

eldering (approved):

eldering:
> Jaap, are these links already stable?

I'd say pretty stable, but can't guarantee 100%. We can just merge and update if they change.

PR #883, opened 2020-10-28
Web Accessibility Evaluation test added

mvr320:
Alternatives can be found at:
https://www.w3.org/WAI/ER/tools/                                                                                                                                                                                                                The test is useful to make sure that during a contest teams have no problems with the interface. Changing the workspace is more difficult otherwise. 

Currently tests for:
WCAG2AA (See: https://github.com/pa11y/pa11y)


When: https://github.com/DOMjudge/domjudge-packaging/pull/65 is accepted I will increase the speed by removing the `apt update` step which takes quite some time

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -54,6 +54,33 @@ check_w3c_html:
       - public
       - w3cHtmlpublic.json
 
+check_wcag:
+  stage: test
+  before_script:
+    # Takes about 1 minute to pull
+    - echo -e "section_start:`date +%s`:docker\r\e[0Kdocker pull"
+    - time docker pull domjudge/gitlabci:1.0
+    - echo -e "section_end:`date +%s`:docker\r\e[0K"
+  image: docker:stable
+  services:
+    - docker:19-dind
+    - mariadb
+  variables:
+    MYSQL_ROOT_PASSWORD: password
+    DOCKER_HOST: tcp://docker:2375/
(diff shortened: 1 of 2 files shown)


mvr320:
@thijskh as this is mostly affecting Frontend I assume you are the best reviewer?

thijskh:
It seems a significant part of the errors come from Swagger-UI. Or do we want to go and try to fix those upstream?

We surely must get the T value significantly down for these tests to become useful. Why don't we set a lower = more ambitious target, this PR will fail, and we fix things until it passes, then merge? Or even set T=0?

mvr320:
> It seems a significant part of the errors come from Swagger-UI. Or do we want to go and try to fix those upstream?
I think those are the contract errors right? I think these should be part of v2 for testing
> 
> We surely must get the T value significantly down for these tests to become useful. Why don't we set a lower = more ambitious target, this PR will fail, and we fix things until it passes, then merge? Or even set T=0?
My plan was to after this is accepted to indeed pick a page and fix all the W3C(/WCAG) violations, but I think I only want to start with that when the visual regression is also implemented. 

Setting T=0 is not recommended (https://www.linkedin.com/pulse/why-wcag-made-3-level-aa-aaa-different-levels-grade-website-arun-al/) when the test becomes more strict (3A). This current measure prevents us from getting more of these problems, the T is basicly only there to make the script only terminate when we reach to many errors at the end.

thijskh (approved):

PR #884, opened 2020-10-29
Visual regression test

mvr320:
The current master branch and the PR are fully generated and compared
pixel by pixel. If there is to much regression the test fails

A small regression is accepted as the time might differ between the
scraping of the pages and changes of only a pixel will not be noticed.

After https://github.com/DOMjudge/domjudge-packaging/pull/66 is merged I can speedup by removing the apt {update,install} step

In the future this can be extended with different browsers or move to BrowserStack

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -3,6 +3,7 @@ variables:
 
 stages:
   - test
+  - compare
 
 include:
   - template: Dependency-Scanning.gitlab-ci.yml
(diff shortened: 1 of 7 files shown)


thijskh:
How does this work when the PR actually intends to modify the UI?
Can I see what it looks like when it fails?

mvr320:
> 
> 
> How does this work when the PR actually intends to modify the UI?
> Can I see what it looks like when it fails?

I think there we have a "problem" as in that case you should accept that a test fails. We could ofcourse do something with the name of your branch: "UIchange_br_name" where the test does not run for such branches. Or the human solution of actually mentioning that the test fails but this is actually intended (Which I think is better)

But I'm open to every solution, this was more to showcase that it now does work (even with the timing differences) and I think its better than nothing but I'm not experienced enough to tell you what is the perfect solution here so I'm open to whatever you think is preferred of the 2 (or a 3th solution) @thijskh 

I will make something where it actually should fail,

mvr320:
@thijskh this fails:
https://gitlab.com/DOMjudge/domjudge/-/jobs/821434749

Whereas if only the time would change its still accepted,

meisterT:
I am not against it per se, but what are we trying to achieve with a test like this?

mvr320:
> 
> 
> I am not against it per se, but what are we trying to achieve with a test like this?

When you change one thing of the templates this might have an effect in other parts of the website. I think this alerts you when changing the code that it has different effects on UI and you can actually check if this is still good enough. The same way how we have the PHP tests which tell you if functionality implemented does still work after changing the code. So its one of the sorts of tests which was not available yet. 

The other advantage of this test is that its now also the first and maybe only test for some of the pages that certain content is there. If you now change the output of a text block which is reused it will showup in the image as difference (So even when not changing CSS what the main argument for this test is)

But currently I agree with @thijskh that its hard to determine what to do when you can expect the test to fail. So I think it needs some rework before this can be merged,

thijskh:
@meisterT This kind of test will also detect bugs that do not result in an exception. E.g. if a table is now empty instead of having rows, this results in a visual difference in the output.

mvr320:
@thijskh what do you think of this? The current failure is because I changed something as demonstration and when https://github.com/DOMjudge/domjudge/pull/909 is merged the last commits can be removed and handled as custom CSS which can also make the differences less strict.

We now push the result to GitHub to indicate changes as either correct or failing depending on regex after the line:  "Changed URLs:" in either the PR initial message or discussion (from anyone).

mvr320:
Due to 1 *id* not being in master yet we will now trigger the change directly. That will be resolved when this is merged.

thijskh (approved):
Great work, I think this will be useful. Let's merge it and see if there's anything that we need to finetune.

thijskh:
Unfortunately the build failed. Any idea?

thijskh:
It passes on retry. Hope it was a fluke and it does turn out to be robust. If not, we'll fix it.

Issue #887, opened 2020-11-03
Implement new Judgehost API


meisterT:
### Description of the problem
@ubergeek42 proposed a new Judgehost API in https://docs.google.com/document/d/1dMjPOcR2TqS7Z4jNbb5W5vVdHu43BSf79gRngEkfk5w/edit#

This will allow us to parallelize judging across multiple judgehosts, to lazily evaluate test cases after the final verdict for a submission is known or do generic tasks like preheating judgehosts or collecting debug information.

Some related issues are: #258, #215, #162, #639

meisterT:
With some related commits in the past few days I consider this issue closed. The last big issue is https://github.com/DOMjudge/domjudge/issues/215, but that can be and is tracked separately.

PR #888, opened 2020-11-03
Implement new judgehost API

meisterT:
This is just the first step of #887.
I intend to work in the feature branch for a while.


webapp/src/Controller/API/JudgehostController.php
```diff
@@ -1248,4 +1254,91 @@ private function getSubmissionsToJudge(Judgehost $judgehost, $restrictJudgingOnS
...
+            default:
+                throw new BadRequestHttpException('Unknown type requested.');
+        }
+    }
+
+    private function getSourceFiles(string $id) {
+        $queryBuilder = $this->em->createQueryBuilder()
+            ->from(SubmissionFile::class, 'f')
+            ->join('f.submission', 's')
+            ->select('f, s')
```
nickygerritsen, line 1295: I think we can drop selecting s here? Maybe even not load the submission at all but use where f.submission = :submitid?
Same below for exec files

meisterT: Done.

judge/judgedaemon.main.php
```diff
@@ -850,55 +880,50 @@ function judge(array $row)
...
+        logmsg(LOG_ERR, "Fetching executable failed for compile script '" . $judgeTask['compile_script_id'] . "': " . $error);
+        $description = $judgeTask['compile_script_id'] . ': fetch, compile, or deploy of compile script failed.';
+        disable('compile_script', 'compile_script_id', $judgeTask['compile_script_id'], $description, $judgeTask['judgetaskid']);
+        return false;
     }
 
     // Compile the program.
     system(LIBJUDGEDIR . "/compile.sh $cpuset_opt '$execrunpath' '$workdir' " .
-           implode(' ', $files), $retval);
+        implode(' ', $files), $retval);
```
ubergeek42, line 926: :thinking: does this expose us to shell injection vulnerability if you choose a clever filename and upload it? I assume that is probably filtered elsewhere, but it occurred to me while reading this we might want to do some shell escaping magic.

judge/judgedaemon.main.php
```diff
@@ -948,244 +974,238 @@ function judge(array $row)
...
+    );
+
+    if ($result === 'correct') {
+        // Post result back asynchronously. PHP is lacking multi-threading, so
+        // we just call ourselves again.
+        $judgedaemon = BINDIR . '/judgedaemon';
+        $cmd = $judgedaemon
+            . ' -e ' . $endpointID
+            . ' -t ' . $judgeTask['judgetaskid']
+            . ' -j ' . base64_encode(json_encode($new_judging_run))
```
ubergeek42, line 1156: What's the command line length limit? If this includes program output then I could see easily blowing that up if someone increases `output_storage_limit`. I guess the fix would be to store this in a temporary file and pass that to the program instead.

For me locally `getconf ARG_MAX` is `2097152`, and if we're using base64, that works out to ~1.5MiB of data.

I also did a quick google search and see that php does have pthreads, so maybe we can do it async/with threads? No idea if it works / is worth doing though. https://stackoverflow.com/a/15501449/841300

That or we should rewrite in golang :rofl:

judge/judgedaemon.main.php
meisterT: Filenames are restricted in: https://github.com/DOMjudge/domjudge/blob/master/webapp/src/Service/SubmissionService.php#L30

But I agree that we should also add escaping here (in another PR).

judge/judgedaemon.main.php
```diff
@@ -948,244 +974,238 @@ function judge(array $row)
...
+        'output_run'   => rest_encode_file($testcasedir . '/program.out', false),
+        'output_error' => rest_encode_file($testcasedir . '/program.err', $output_storage_limit),
+        'output_system' => rest_encode_file($testcasedir . '/system.out', $output_storage_limit),
+        'metadata' => rest_encode_file($testcasedir . '/program.meta', $output_storage_limit),
+        'output_diff'  => rest_encode_file($testcasedir . '/feedback/judgemessage.txt', $output_storage_limit),
+        'hostname' => $myhost,
+    );
+
+    if ($result === 'correct') {
+        // Post result back asynchronously. PHP is lacking multi-threading, so
```
ubergeek42, line 1150: We do this async so that we can move on to the next testcase without waiting for the roundtrip http request to the server right?

judge/judgedaemon.main.php
ubergeek42: yeah, I definitely don't think we need to do that as part of this PR. Opened https://github.com/DOMjudge/domjudge/issues/960 to track this.

judge/judgedaemon.main.php
meisterT: :eyes: I did some google search back then and didn't find anything. Let's replace it with proper threading then.

re golang: tempting :-)

judge/judgedaemon.main.php
meisterT: Yes, that's the idea. It saves quite some time to do this synchronously.

judge/judgedaemon.main.php
nickygerritsen: I do see this comment linked from the pho page: https://github.com/krakjoe/pthreads/issues/929#issue-410636734

webapp/src/Migrations/Version20201219154651.php
```diff
@@ -0,0 +1,63 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20201219154651 extends AbstractMigration
+{
+    public function getDescription() : string
+    {
+        return '';
```
ubergeek42, line 17: Can we/should we have descriptions? I don't know if we do elsewhere in the migrations...

nickygerritsen: I think most/all have them yes

ubergeek42:
From my rather quick review of this it all looks good to me(aside from the few comments above).

judge/judgedaemon.main.php
meisterT: `pthreads` seems out of question; `parallel` requires PECL IIUC, so deferring this.

To address Keith's initial comment about line length I am using now an intermediate file instead.

webapp/src/Migrations/Version20201219154651.php
meisterT: Done.

webapp/src/Service/ImportProblemService.php
```diff
@@ -524,11 +532,11 @@ public function importZippedProblem(
             $contestProblem->setProblem($problem);
             $contestProblem->setContest($contest);
             $this->em->persist($contestProblem);
+            $this->em->flush();
         }
 
-        $this->em->flush();
```
mvr320, line 529: Can you not move the flush from L530 here adn remove L535?

meisterT: No, in line 532 we rely that the newly persisted object has been flushed.

webapp/src/Controller/Jury/ProblemController.php
```diff
@@ -900,7 +900,7 @@ public function fetchTestcaseAction(int $probId, int $rank, string $type)
             ->join('tc.content', 'tcc')
             ->select('tc', 'tcc')
             ->andWhere('tc.problem = :problem')
-            ->andWhere('tc.ranknumber = :ranknumber')
+            ->andWhere('tc.rank = :ranknumber')
```
nickygerritsen, line 903: Shouldn't this be ranknumber?

webapp/tests/Service/SubmissionServiceTest.php
```diff
@@ -0,0 +1,45 @@
...
+        yield [['wrong-answer'], $defaultPrios, 'wrong-answer'];
+        yield [['correct', 'wrong-answer', NULL], $defaultPrios, 'wrong-answer'];
+        yield [['correct', NULL, 'wrong-answer'], $defaultPrios, NULL];
+        yield [['correct', NULL, 'correct'], $defaultPrios, NULL];
+        yield [['correct', 'wrong-answer', 'timelimit'], $defaultPrios, 'wrong-answer'];
+
+        $modifiedPrios = $defaultPrios;
+        $modifiedPrios['wrong-answer'] = 70;
+        yield [['correct', 'wrong-answer', NULL], $modifiedPrios, NULL];
+        yield [['correct', NULL, 'wrong-answer'], $defaultPrios, NULL];
```
nickygerritsen, line 38: Should be `$modifiedPrios`?

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -1023,49 +917,33 @@ private function addSingleJudgingRun(
...
-            ->getSingleScalarResult();
-
-        $allRuns = array_pad($runs, (int)$numTestCases, null);
+            ->getArrayResult();
+        $runresults = array_column($runs, 'runresult');
 
         $oldResult = $judging->getResult();
 
-        if (($result = $this->submissionService->getFinalResult($allRuns, $resultsPrio)) !== null) {
+        if (($result = $this->submissionService->getFinalResult($runresults, $resultsPrio)) !== null) {
```
nickygerritsen, line 946: nit:

```suggestion
        if (($result = SubmissionService::getFinalResult($runresults, $resultsPrio)) !== null) {
```

nickygerritsen (approved):
Some small things but looks great

webapp/src/Controller/Jury/ProblemController.php
meisterT: I assume this works because there's `getRank`. Let me try with `ranknumber`.

webapp/tests/Service/SubmissionServiceTest.php
meisterT: Good catch!

PR #901, opened 2020-11-07
Allow to pass JSON to API's besides form data. Fixes #895.

### LINKED ISSUE #895: Allow API to use JSON besides form data
The API currently only supports HTTP form data in it's POST body, but it should be easy to also add JSON support.

See https://symfony.com/doc/3.x/bundles/FOSRestBundle/body_listener.html

nickygerritsen:


webapp/config/packages/fos_rest.yaml
```diff
@@ -16,3 +16,4 @@ fos_rest:
             Symfony\Component\HttpKernel\Exception\NotAcceptableHttpException: true
             Symfony\Component\HttpKernel\Exception\NotFoundHttpException: true
             Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException: true
+    body_listener: ~
```
meisterT, line 19: are you kidding me? Is that everything? :-D

meisterT (approved):

PR #902, opened 2020-11-07
Fix Doctrine diff for default values and index lengths

nickygerritsen:


nickygerritsen:
default=null was the... default. We had `default="NULL"` because I generated using MariaDB, which needs a different `server_version` string. BUT Doctrine autodetects it by default, so we don't need it.

eldering (approved):

thijskh (approved):

PR #903, opened 2020-11-07
Update to new Nelmio API doc bundle so we have OpenAPI Spec v3. Fixes #891.

### LINKED ISSUE #891: Upgrade Nelmio API Doc Bundle to 4.0
See https://github.com/nelmio/NelmioApiDocBundle/releases

4.0 uses the newer OpenAPI spec instead of the older Swagger code.

nickygerritsen:
The YAML I have upgraded with the tool mentioned in the link in the issue.
The rest was a lot of manual work, but it seems to work.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -65,7 +65,7 @@
 		"ircmaxell/password-compat": "*",
 		"jms/serializer-bundle": "^3.4",
 		"mbostock/d3": "^3.5",
-		"nelmio/api-doc-bundle": "^3.6",
+		"nelmio/api-doc-bundle": "^4.0",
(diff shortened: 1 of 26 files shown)


eldering (approved):
Very quick glance looks ok. At least no new packages installed ;-)

nickygerritsen:
Found one bug that I thought was fixed upstream but it wasn't so I fixed it again by re-adding `ParameterRefMergeDescriber`

thijskh (approved):
Yes looks good, one doc issue that was quicker to fix instead of to describe :-)

PR #905, opened 2020-11-07
Use pkg-config to determine systemd unit files directory.

### LINKED ISSUE #834: systemd units installed in wrong directory
The systemd units under `judge/` are installed in `/usr/lib/systemd/system` while this should be (on Debian at least) `/lib/systemd/system`. They will not be read from the former.

The systemd manual has some a recipe on how to determine the right installation path with automake:
https://www.freedesktop.org/software/systemd/man/daemon.html#Installing%20systemd%20Service%20Files

Maybe we can follow that. Or just call pkg-config at install time?

eldering:
Follows https://www.freedesktop.org/software/systemd/man/daemon.html#Installing%20systemd%20Service%20Files
Allows overriding the default using --with-systemdsystemunitdir=DIR
If not detected or specified, don't install unit files.

Closes #834

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,7 +124,7 @@ install-domserver-l:
 		$(INSTALL_WEBSITE) -m 0775 -d $(DESTDIR)$(domserver_webappdir)/var/$$d ; \
 	done
 # Make sure that domjudge user and webserver group can write to var/{cache,log}:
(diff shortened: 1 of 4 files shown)


meisterT (approved):

nickygerritsen:
Ah yes thought that was already merged. Then ignore my comment

PR #909, opened 2020-11-08
Add support for loading custom CSS/JS. Fixes #908.

### LINKED ISSUE #908: Custom CSS/JS
Currently we get some questions from people who wish to customize the UI. I propose to have a folder:
webapp/public/custom/{js,css,...}

which will include all files in the folders if these exist. A possible default value there should than tell people to use !important for CSS.

Such a folder can than also be easily be included from the docker image.

It could even be extended for webfonts and favicon which should be out of scope in my opinion (for now).

nickygerritsen:
As stated in #908 my local testing didn't produce any negative performance issues with this.
I tried checking if we could somehow 'precompile' this but I think that requires hooking into the Symfony container builder and I am not sure how we can do that easily.


thijskh (approved):

nickygerritsen:
Update: prechecking was easy after all. See the updated PR.

thijskh:
Should then do the same for banner?

thijskh:
Add to docs that you may need to clear cache when adding such files?

nickygerritsen:
You *have* to clear the cache. We don't have that in the docs yet anywhere it seems. Any suggestion where we should place this?

nickygerritsen:
For banner: we might want to do that, but then we also need it for country flags and team logo's. Let's do that separately if we want it?

mvr320:
> 
> 
> You _have_ to clear the cache. We don't have that in the docs yet anywhere it seems. Any suggestion where we should place this?

I would expect this as tooltip at the config_check_html. And at the common mistakes where java was also discussed. But I cannot find that anymore

nickygerritsen:
Added a doc entry for clearing the cache and referenced it in my note.

thijskh:
I was thinking of css/js/banner because they affect every page (latter the main public page). Team photos much less so.

Flags should not need an if exists at all?

nickygerritsen:
Let's just do them all. Should be easy to make this method I created generic, so I will do that.

thijskh:
Since we ship the flags, maybe we should not fail silently if they are not there but just let it result in a 404

nickygerritsen:
> Since we ship the flags, maybe we should not fail silently if they are not there but just let it result in a 404

I will do this.

nickygerritsen:
Done

doc/manual/config-advanced.rst
```diff
@@ -355,3 +369,15 @@ contest ID to use, and the IDs of the team categories you want to include
...
+
+Clearing the DOMjudge cache
+---------------------------
+
+Some operations require you to clear the DOMjudge cache. To do this, execute
+the `webapp/bin/console` (see the Config checker in the admin interfae for the
+full filesystem path of your installation) binary with the `cache:clear` subcommand::
+
+  webapp/bin/console cache:clear
+
```
thijskh, line 383: Make explicit that this is different from the scoreboard cache that you can also clear?

doc/manual/config-advanced.rst
```diff
@@ -32,6 +32,20 @@ path of your installation) as follows:
   The IDs for affiliations and teams need to be the *external ID*
   if the ``data_source`` setting of DOMjudge is set to external.
 
+It is also possible to load custom CSS and/or JavaScript files. To do so, place
+files ending in `.css` under `public/css/custom/` and/or files ending in `.js`
```
thijskh, line 36: ` must be `` throughout this paragraph

thijskh (changes requested):

doc/manual/config-advanced.rst
nickygerritsen: No? Well at least it is also not the case for all paragraphs above, only for configuration settings.

thijskh: It seems we’re not very consistent in the manual with this. So not an issue for this PR then

thijskh (approved):

PR #910, opened 2020-11-08
W3c fixes

mvr320:
Currently we implement multiple guidelines, I open this PR to bundle these in less tests (or more robust in building) and fix the problems.

### DIFF
--- a/webapp/src/Form/Type/TeamClarificationType.php
+++ b/webapp/src/Form/Type/TeamClarificationType.php
@@ -8,6 +8,7 @@ use App\Service\DOMJudgeService;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\TextareaType;
+use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\FormBuilderInterface;
 
 class TeamClarificationType extends AbstractType
@@ -36,11 +37,9 @@ class TeamClarificationType extends AbstractType
      */
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        $builder->add('recipient', ChoiceType::class, [
-            'choices' => ['Jury' => 'dummy'],
-            'attr' => [
-                'class' => 'select_readonly',
-            ],
+        $builder->add('recipient', TextType::class, [
+	    'data' => 'Jury',
+	    'disabled' => true,
         ]);
 
         $subjects = [];
--- a/webapp/src/Form/Type/TeamClarificationType.php
+++ b/webapp/src/Form/Type/TeamClarificationType.php
@@ -39,7 +39,7 @@ class TeamClarificationType extends AbstractType
         $builder->add('recipient', ChoiceType::class, [
             'choices' => ['Jury' => 'dummy'],
             'attr' => [
-                'readonly' => true,
+                'class' => 'select_readonly',
             ],
         ]);
 
--- a/webapp/templates/jury/clarification.html.twig
+++ b/webapp/templates/jury/clarification.html.twig
@@ -91,7 +91,7 @@ Jury
 
 <div class="col-sm">Queue: <span class="clarification-queue">{{ clar.queue }}
 <button class="btn btn-sm btn-link clarification-queue-change-button"><i class="far fa-edit" title="Change queue"></i></button></span>
-<span class="clarification-queue-form" data-current-selected-queue="{{ clar.queueid }}" data-clarification-id="{{clar.clarid}}">
+<div style="display: inline" class="clarification-queue-form" data-current-selected-queue="{{ clar.queueid }}" data-clarification-id="{{clar.clarid}}">
 <form action="{{ path('jury_clarification_change_queue', {'clarId': clar.clarid }) }}" method="post" class="form-inline">
 <select name="queue" id="queue" class="custom-select custom-select-sm">
 <option value="">Unassigned issues</option>
@@ -103,7 +103,7 @@ Jury
   <span aria-hidden="true">&times;</span>
 </button>
 </form>
-</span>
+</div>
 
 </div>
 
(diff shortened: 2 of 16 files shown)


webapp/templates/jury/clarification.html.twig
```diff
@@ -82,9 +82,9 @@ Jury
 
 <div class="col-sm">Queue: <span class="clarification-queue">{{ clar.queue }}
 <button class="btn btn-sm btn-link clarification-queue-change-button"><i class="far fa-edit" title="Change queue"></i></button></span>
-<span class="clarification-queue-form" data-current-selected-queue="{{ clar.queueid }}" data-clarification-id="{{clar.clarid}}">
+<div style="display: inline" class="clarification-queue-form" data-current-selected-queue="{{ clar.queueid }}" data-clarification-id="{{clar.clarid}}">
```
thijskh, line 85: Is it necessary to add inline styles, or can we do this in the css?

thijskh (commented):
Does this PR intend to make actually visible changes for the user?

webapp/templates/jury/clarification.html.twig
mvr320: We can do this in the CSS, my intention was to first get W3c compliant and when this is done move it to the CSS files. We have this on more locations still.

mvr320:
> 
> 
> Does this PR intend to make actually visible changes for the user?

In most cases as least as possible, the W3C validator complains about certain constructions not being allowed and here I fix most of them (or ignore those) so in the future the tests here bundled together with the role can replace the other html test which is not connected to the role. There are some changes done for WCAG where to me the differences are not that much but the contrast is better, those are things I think you should accept. Ofcourse when I opened this PR I had hoped that I could in the end rebase to the visual change PR so the differences would be a bit easier to spot.

This is ofcourse also a bit ongoing so I could also try to fixup some of the changes so you can cherrypick which changes we actually do accept and which ones should just be ignored by the tests (as I think having to look through 74 commits is a bit much)

Basicly the overall goal of this PR is to define for the tests the roles:
public
team
balloon

where the project is W3C compliant for HTML,CSS,SVG for all roles and also prefer a level of accessibility (WCAG2A,WCAG2AA,Section508) and explicit skip this for jury and admin (if you think DOMjudge should commit to the same standards that's also a choice but I think there should be a clear standard for how far DOMjudge is planning to go)

The W3C compliant so we can reasonably assume that everything will look the same in most browsers (although the current browser landscape is almost only chromium based) and the WCAG because it will remove a lot of issues for on-site contests if someone actually needs the level of accessibility.

mvr320:
More as a test if the UI diff is broken

Changed URLs:
team
public
public/problems

webapp/src/Form/Type/TeamClarificationType.php
```diff
@@ -39,7 +39,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
         $builder->add('recipient', ChoiceType::class, [
             'choices' => ['Jury' => 'dummy'],
             'attr' => [
-                'readonly' => true,
+                'class' => 'select_readonly',
```
thijskh, line 42: The select box is now no longer actually read only. Why is that?

webapp/templates/jury/clarification.html.twig
```diff
@@ -61,7 +61,7 @@ Jury
     </span>
 <span class="clarification-subject-form" data-current-selected-subject="{{ clar.categoryid }}" data-clarification-id="{{clar.clarid}}">
 <form action="{{ path('jury_clarification_change_subject', {'clarId': clar.clarid }) }}" method="post" class="form-inline">
-<select name="subject" id="subject" class="custom-select custom-select-sm">
+<select name="subject" class="subject custom-select custom-select-sm">
```
thijskh, line 64: Do the classes "subject" and "queue" actually exist?

webapp/src/Form/Type/TeamClarificationType.php
mvr320: Because its not according to the w3c standard (as far as I'm aware)

webapp/templates/jury/clarification.html.twig
mvr320: I think the id(subject) was never used in CSS or JS, but was repeated (hence changing of attribute), this just leaves the possibility to use it. I can recheck if this is indeed not used and remove the attribute completely?

thijskh: But your change adds it as a class of the tag?

mvr320: The select tag is used multiple times on the same name (resulting in multiple DOMelements with the same ID) so I kept the identifier by introducing it as a class. I think I checked and the ID was never used anywhere (so could also be removed) or I forgot to check in a change to the JS/CSS where the identifier should be renamed to ".subject" instead of "#subject"

webapp/src/Form/Type/TeamClarificationType.php
thijskh: I'd suggest the following fix which makes it w3c compliant while retaining the intended rendering. Just simplifying it by making it  text field that is disabled. (For some reason I'm not able to push to this branch).

```diff
--- a/webapp/src/Form/Type/TeamClarificationType.php
+++ b/webapp/src/Form/Type/TeamClarificationType.php
@@ -8,6 +8,7 @@ use App\Service\DOMJudgeService;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\TextareaType;
+use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\FormBuilderInterface;
 
 class TeamClarificationType extends AbstractType
@@ -36,11 +37,9 @@ class TeamClarificationType extends AbstractType
      */
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        $builder->add('recipient', ChoiceType::class, [
-            'choices' => ['Jury' => 'dummy'],
-            'attr' => [
-                'class' => 'select_readonly',
-            ],
+        $builder->add('recipient', TextType::class, [
+            'data' => 'Jury',
+            'disabled' => true,
         ]);
 
         $subjects = [];
```

mvr320: You should have the rights, 
image

But I've committed it.

thijskh (approved):
Great, I'll merge it.

PR #912, opened 2020-11-10
Remove explicit foreign key columns since they are not needed.

nickygerritsen:
Oh man, this was a biggie.

Locally my unit tests work and I seem to be able to submit stuff, etc.
But I totally expect that I missed some twig file.
I did check usages for all fields so all PHP should be handled if we have proper type hints, but for Twig I manually checked usages. Most changed usages where from `bla.someid` to `bla.some.someid`.
I do like we can remove a lot of stuff and still have the same result.


thijskh (approved):
The idea is good. It does demonstrate our lack of test coverage. I think the only way to weed out more issues is to merge this to master.

PR #914, opened 2020-11-10
Change default Python interpreter to pypy

pingiun:
We discussed in the slack that the default Python interpreter should be switched to pypy. I believe previously this wasn't done because pypy3 packages weren't readily available. They are now, and it seems many contests are using pypy3 already. 

This PR adds pypy3 to the chroot (and leaves python3 in place for contests that want to supply their own runscript for it), and changes the python interpreter in the run script.

### DIFF
--- a/misc-tools/dj_make_chroot.in
+++ b/misc-tools/dj_make_chroot.in
@@ -133,7 +133,7 @@ if [ "$DISTRO" = 'Debian' ]; then
 INCLUDEDEBS="ca-certificates"
 
 # Packages to install after upgrade (space separated):
-INSTALLDEBS="gcc g++ default-jdk-headless default-jre-headless pypy python3 locales"
+INSTALLDEBS="gcc g++ default-jdk-headless default-jre-headless pypy pypy3 python3 locales"
 # For C# support add: mono-runtime libmono-system2.0-cil
 # However running mono within chroot still gives errors...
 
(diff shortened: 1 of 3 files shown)


nickygerritsen (approved):

nickygerritsen:
Hmmm, I think pypy3 does not exist in Debian stable yet? At least CI is complaining and that uses Debian currently.

pingiun:
It should exist: https://packages.debian.org/search?keywords=pypy3&searchon=names&suite=buster&section=all

thijskh:
The CI uses bionic, not buster, so that must then be bumped to focal (not a bad thing per se)

nickygerritsen:
> The CI uses bionic, not buster, so that must then be bumped to focal (not a bad thing per se)

Yes Michael and me are trying to fix this

mvr320:
@pingiun in the gitlab.ci file you have to replace the gitlabci:1.0 for gitlab:2.0 also for the Docker in Docker jobs,

But maybe you'll be in luck and my PR is fixed before so you can rebase against it.

mvr320:
@pingiun Master now has the gitlabci:2.0 for all tests, so it should be easier to get this one in a working state. Could you rebase and see if it works?

pingiun:
Rebased

thijskh:
This test is supposed to fail but now judged as “correct”: https://github.com/DOMjudge/domjudge/blob/master/tests/test-extra-file.py3.zip

pingiun:
I have no clue why that might be, it seems to have no relation to my change

thijskh:
Yes and no. Your changes expose a bug in the testsuite. So they are triggering the failure, but not the cause. I propose you rebase on current master to see if your PR then passes.

thijskh (approved):

thijskh:
Merging, thanks for your contribution!

PR #915, opened 2020-11-10
Add help text to contest form

### LINKED ISSUE #911: Document contest properties in hover texts
When creating/editing a contest, it's always a bit unclear to me what the _exact_ purpose of all of the following fields is, and how they are different:

- Activate time
- Start time
- Start time enabled
- Contest open to all teams
  Does `no` mean the contest is completely hidden from non-admin/jury users (apart from those explicitly listed)
- Enabled
 
In particular the difference between `enabled` and `active` isn't clear to me. It seems a contest need to be both `enabled` and `active` to judge jury submissions?

I'd suggest to add hover texts to these (or all) of the form fields with a brief explanation of what each field does.

RagnarGrootKoerkamp:
Remaining questions:
- what is the recommended set of settings to play with a contest without it being visible to anybody else (only jury/admins, not teams/public).
- what to put for `start time enabled`
- is `enabled: false` any different from it not being active?

Fix #911

### DIFF
--- a/webapp/src/Form/Type/ContestType.php
+++ b/webapp/src/Form/Type/ContestType.php
@@ -30,7 +30,7 @@ class ContestType extends AbstractExternalIdEntityType
         ]);
         $builder->add('activatetimeString', TextType::class, [
             'label' => 'Activate time',
-            'help' => 'Time when the contest becomes visible for teams to join. Must be in the past to enable judging of jury submissions.',
+            'help' => 'Time when the contest becomes visible for teams to join. Must be in the past to enable submission of jury submissions.',
         ]);
         $builder->add('starttimeString', TextType::class, [
             'label' => 'Start time',
--- a/webapp/src/Form/Type/ContestType.php
+++ b/webapp/src/Form/Type/ContestType.php
@@ -28,9 +28,11 @@ class ContestType extends AbstractExternalIdEntityType
         $builder->add('name', TextType::class);
         $builder->add('activatetimeString', TextType::class, [
             'label' => 'Activate time',
+            'help' => 'Time when the contest becomes visible for teams to join. Must be in the past to enable judging of jury submissions.',
         ]);
         $builder->add('starttimeString', TextType::class, [
             'label' => 'Start time',
+            'help' => 'Absolute time when the contest starts.',
         ]);
         $builder->add('starttimeEnabled', ChoiceType::class, [
             'label' => 'Start time enabled',
@@ -39,21 +41,26 @@ class ContestType extends AbstractExternalIdEntityType
                 'Yes' => true,
                 'No' => false,
             ],
+            'help' => 'Only John knows',
         ]);
         $builder->add('freezetimeString', TextType::class, [
             'label' => 'Scoreboard freeze time',
             'required' => false,
+            'help' => 'Time when the freeze starts: the results of submissions made after this time are not revealed until the scoreboard unfreeze time below has passed.',
         ]);
         $builder->add('endtimeString', TextType::class, [
             'label' => 'End time',
+            'help' => 'Time when the contest ends.',
         ]);
         $builder->add('unfreezetimeString', TextType::class, [
             'label' => 'Scoreboard unfreeze time',
             'required' => false,
+            'help' => 'Time when the final scoreboard is revealed. Usually a few hours after the contest ends, and the award ceremony is over.',
         ]);
         $builder->add('deactivatetimeString', TextType::class, [
             'label' => 'Deactivate time',
             'required' => false,
+            'help' => 'Time when the contest and scoreboard are hidden again. Usually a few hours/days after the contest ends.',
         ]);
         $builder->add('processBalloons', ChoiceType::class, [
             'expanded' => true,
@@ -61,6 +68,7 @@ class ContestType extends AbstractExternalIdEntityType
                 'Yes' => true,
                 'No' => false,
             ],
+            'help' => 'Enables the balloons page. (What is the advantage of disabling it?)',
         ]);
         $builder->add('public', ChoiceType::class, [
             'expanded' => true,
@@ -69,6 +77,7 @@ class ContestType extends AbstractExternalIdEntityType
                 'Yes' => true,
                 'No' => false,
             ],
+            'help' => 'When true, the public scoreboard is enabled and everyone can see it without logging in. When false, only logged in users/teams participating in the contest can see the scoreboard.',
         ]);
         $builder->add('openToAllTeams', ChoiceType::class, [
             'expanded' => true,
@@ -77,6 +86,7 @@ class ContestType extends AbstractExternalIdEntityType
                 'Yes' => true,
                 'No' => false,
             ],
+            'help' => 'When true, any logged in team can join this contest, and the teams/categories listed below are added by default. When false, only the teams/categories listed below will participate.',
         ]);
         $builder->add('teams', EntityType::class, [
             'required' => false,
@@ -85,6 +95,7 @@ class ContestType extends AbstractExternalIdEntityType
             'choice_label' => function (Team $team) {
                 return sprintf('%s (t%d)', $team->getEffectiveName(), $team->getTeamid());
             },
+            'help' => 'List of teams participating in the contest, in case it is not open to all teams.',
         ]);
         $builder->add('teamCategories', EntityType::class, [
             'required' => false,
@@ -93,6 +104,7 @@ class ContestType extends AbstractExternalIdEntityType
             'choice_label' => function (TeamCategory $category) {
                 return $category->getName();
             },
+            'help' => 'List of team categories participating in the contest, in case it is not open to all teams.',
         ]);
         $builder->add('enabled', ChoiceType::class, [
             'expanded' => true,
@@ -100,6 +112,7 @@ class ContestType extends AbstractExternalIdEntityType
                 'Yes' => true,
                 'No' => false,
             ],
+            'help' => 'When false, the contest is hidden from teams (even when active) and judging is disabled. Disabling is a quick way to remove access to it without changing any other settings.',
         ]);
         $builder->add('problems', CollectionType::class, [
             'entry_type' => ContestProblemType::class,
--- a/webapp/src/Form/Type/ContestType.php
+++ b/webapp/src/Form/Type/ContestType.php
@@ -88,7 +88,7 @@ class ContestType extends AbstractExternalIdEntityType
                 'Yes' => true,
                 'No' => false,
             ],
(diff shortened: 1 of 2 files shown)


webapp/src/Form/Type/ContestType.php
```diff
@@ -39,28 +41,34 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'Yes' => true,
                 'No' => false,
             ],
+            'help' => 'Only John knows',
```
meisterT, line 44: Disable to delay the contest start. Enable again after setting a new start time.

webapp/src/Form/Type/ContestType.php
```diff
@@ -39,28 +41,34 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
             'required' => false,
+            'help' => 'Time when the contest and scoreboard are hidden again. Usually a few hours/days after the contest ends.',
         ]);
         $builder->add('processBalloons', ChoiceType::class, [
             'expanded' => true,
             'choices' => [
                 'Yes' => true,
                 'No' => false,
             ],
+            'help' => 'Enables the balloons page. (What is the advantage of disabling it?)',
```
meisterT, line 71: >  (What is the advantage of disabling it?)

This was introduced by @nickygerritsen in 1f3d6d9181ec33127186e52cbba9e12b43ac16e2 but I don't remember why exactly. For this PR, let's just remove this part of the comment.

meisterT (changes requested):
See https://github.com/DOMjudge/domjudge/blob/master/webapp/src/Service/ImportExportService.php#L151 for the settings that we typically enable when playing around with a contest - we nowadays import this typically with the https://github.com/DOMjudge/domjudge/blob/master/misc-tools/import-contest.sh script.

A disabled contest doesn't automatically get active even when activatetime passes.

webapp/src/Form/Type/ContestType.php
nickygerritsen: If you use an external tool to process balloons this gives a minor speed improvement but i I don’t mind to drop it since it’s not that useful

thijskh: Not sure what you mean with external tool but if you disable this, the balloons table will not be filled so you can also not use an external tool I guess. I think it's indeed some kind of optimization that we do not fill another table if you don't do anything with it?

webapp/src/Form/Type/ContestType.php
```diff
@@ -28,9 +28,11 @@ public function buildForm(FormBuilderInterface $builder, array $options)
         $builder->add('name', TextType::class);
         $builder->add('activatetimeString', TextType::class, [
             'label' => 'Activate time',
+            'help' => 'Time when the contest becomes visible for teams to join. Must be in the past to enable judging of jury submissions.',
```
thijskh, line 31: Is the second sentence true? Or should it be "to enable submission of jury submissions"? Honest question, do not know.

RagnarGrootKoerkamp: I don't know either. Figured making a potentially wrong statement is the best way to find out :P

webapp/src/Form/Type/ContestType.php
```diff
@@ -39,28 +41,34 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+            'help' => 'Only John knows',
         ]);
         $builder->add('freezetimeString', TextType::class, [
             'label' => 'Scoreboard freeze time',
             'required' => false,
+            'help' => 'Time when the freeze starts: the results of submissions made after this time are not revealed until the scoreboard unfreeze time below has passed.',
         ]);
         $builder->add('endtimeString', TextType::class, [
             'label' => 'End time',
+            'help' => 'Time when the contest ends.',
```
thijskh, line 53: + "(Submissions will be accepted but shown as 'too-late' and not counted towards the score)"

webapp/src/Form/Type/ContestType.php
meisterT: Indeed, you can't submit as jury before activate time. But if there are submissions (since you moved activatetime after submitting), we'll keep judging.

webapp/src/Form/Type/ContestType.php
RagnarGrootKoerkamp: What effect has `delay the contest start` on the GUI? Will the coundown in the top right just disappear?

webapp/src/Form/Type/ContestType.php
nickygerritsen: I mean the cds. It can also generate balloons

webapp/src/Form/Type/ContestType.php
meisterT: Yes, it will display "countdown paused" or a similar message.

webapp/src/Form/Type/ContestType.php
```diff
@@ -25,12 +25,16 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $this->addExternalIdField($builder, Contest::class);
         $builder->add('shortname', TextType::class);
-        $builder->add('name', TextType::class);
+        $builder->add('name', TextType::class, [
+            'help' => 'Contest name as shown in the top right.'
```
RagnarGrootKoerkamp, line 29: Also added this one here.

webapp/src/Form/Type/ContestType.php
```diff
@@ -77,6 +88,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'Yes' => true,
                 'No' => false,
             ],
+            'help' => 'When true, any logged in team can join this contest, and the teams/categories listed below are added by default. When false, only the teams/categories listed below will participate.',
```
RagnarGrootKoerkamp, line 91: Can someone confirm this assumption is indeed correct?

nickygerritsen: Nope when true the other fields are hidden (or should be) and everyone automatically is part of the contest. When false only the teams / cats you select are part. We don’t have a way to join as a team

RagnarGrootKoerkamp: Ah right. They are hidden indeed.

webapp/src/Form/Type/ContestType.php
thijskh: I think this helptext applies to 'shortname', and 'name' is 'Contest name in full as shown on the scoreboard'

webapp/src/Form/Type/ContestType.php
```diff
@@ -25,12 +25,16 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $this->addExternalIdField($builder, Contest::class);
         $builder->add('shortname', TextType::class);
-        $builder->add('name', TextType::class);
+        $builder->add('name', TextType::class, [
+            'help' => 'Contest name as shown in the top right.'
+        ]);
         $builder->add('activatetimeString', TextType::class, [
             'label' => 'Activate time',
+            'help' => 'Time when the contest becomes visible for teams to join. Must be in the past to enable submission of jury submissions.',
```
nickygerritsen, line 33: Instead of:

> Time when the contest becomes visible for teams to join

isn't

> Time when the contest becomes visible for teams

Since there is no actual joining.

thijskh (approved):

RagnarGrootKoerkamp:
Still need to test that this text actually shows up nicely on the webpage. What's the easiest way to do that?

nickygerritsen:
> Still need to test that this text actually shows up nicely on the webpage. What's the easiest way to do that?

Checkout code, docker-compose up (maybe twice if it errors), http://localhost:12345 and log in with admin and the password from etc/initial_admin_password in the docker container

thijskh:
image

RagnarGrootKoerkamp:
Hmm, is the `Yes`/`No` button text weird in combination with the `true`/`false` description text?

thijskh:
Agreed, I would use "enabled"/"disabled"

RagnarGrootKoerkamp:
Rewrote `yes/no` to `enabled/disabled`, and changed the form titles such that `enabled/disabled` is a sensible answer (instead of `yes/no`). This may be a bit too much though. Personally I like the look of the `yes/no` buttons slightly better because they look more distinct.

Also removed the manual line wrapping from the help text in the top right, and prevented text wrapping of the time spec.

img-2020-11-11-143342
img-2020-11-11-143358

thijskh:
Sorry, I already fixed the wrapping in master. So you can rebase on that.

Enabled/disabled as labels is indeed too much. Also it is redundant in places like: "Enable public scoreboard? Enable/Disable."

RagnarGrootKoerkamp:
Ok, back to `Yes`/`No` button labels, but `enabled`/`disabled`  (instead of `true`/`false`) in the title and descriptions.

img-2020-11-11-145312

thijskh (approved):
I am good with it but would appreciate another review

webapp/src/Form/Type/ContestType.php
```diff
@@ -24,59 +24,75 @@ class ContestType extends AbstractExternalIdEntityType
...
         $builder->add('starttimeEnabled', ChoiceType::class, [
-            'label' => 'Start time enabled',
+            'label' => 'Start time countdown enabled',
             'expanded' => true,
             'choices' => [
                 'Yes' => true,
                 'No' => false,
             ],
+            'help' => 'Disable to delay the contest start and stop the countdown. Enable again after setting a new start time.',
+            # What happens when the countdown is disabled, but the start time is in the past/time 'crosses' the start time?
```
meisterT, line 49: In that case, we still do the same and say "start delayed" and disallow looking at problems/submissions:
image

We may want to do just forbid this (in a separate commit), but for now just remove this line.

webapp/src/Form/Type/ContestType.php
```diff
@@ -24,59 +24,75 @@ class ContestType extends AbstractExternalIdEntityType
...
         ]);
         $builder->add('freezetimeString', TextType::class, [
             'label' => 'Scoreboard freeze time',
             'required' => false,
+            'help' => 'Time when the freeze starts: the results of submissions made after this time are not revealed until the scoreboard unfreeze time below has passed.',
         ]);
         $builder->add('endtimeString', TextType::class, [
             'label' => 'End time',
+            'help' => 'Time when the contest ends. Submissions made after this time will be accepted but shown as \'too-late\' and not counted towards the score.',
```
meisterT, line 58: - "accepted **and judged**"
- "but shown **(to teams and public)**"

meisterT (commented):
Two nits

meisterT:
Thanks!

meisterT (approved):

PR #916, opened 2020-11-10
The gitlabci has pa11y on different PATH

mvr320:
Due to a problem with puppeteer in docker the binary is installed in another location

This should only be merged when https://github.com/DOMjudge/domjudge-packaging/pull/68 is handled

The chroot error can be removed by unmounting earlier but is not problematic for actually terminating. As a bonus this updates everything to 2.0

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -11,7 +11,7 @@ include:
 
 check syntax:
   stage: test
-  image: domjudge/gitlabci:1.0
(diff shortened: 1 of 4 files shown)


nickygerritsen:
I am going to bump `domjudge/gitlabci` to `2.0` for the change we made there since it is a new Ubuntu release. Can you update the .gitlab-ci.yml to reflect this?

nickygerritsen (approved):

PR #919, opened 2020-11-12
Add JSON files to import docs and add info about how to do it in the UI.

nickygerritsen:
As discussed on Slack, my attempt at improving the import docs.

### DIFF
--- a/doc/manual/import.rst
+++ b/doc/manual/import.rst
@@ -124,10 +124,10 @@ Prepare a file called ``teams.json`` which contains the teams.
 It should be a JSON array with objects, each object should contain the following
 fields:
 
-- ``id``: the (integer) team ID. Must be unoque
+- ``id``: the (integer) team ID. Must be unique
 - ``icpc_id`` (optional): an external ID, e.g. from the ICPC CMS, may be empty
-- ``group_ids``: an array with one element: the category ID this team belomgs to
-- ``name``: the team name as used on the web inteface
+- ``group_ids``: an array with one element: the category ID this team belongs to
+- ``name``: the team name as used in the web inteface
 - ``display_name`` (optional): the team display name. If provided, will display
   this instead of the team name in certain places, like the scoreboard
 - ``organization_id``: the external ID of the team affiliation this team belongs to
@@ -164,8 +164,8 @@ Each of the following lines must contain the following elements separated by tab
 
 - the (integer) team ID. Must be unique
 - an external ID, e.g. from the ICPC CMS, may be empty
-- the category ID this team belomgs to
-- the team name as used on the web inteface
+- the category ID this team belongs to
+- the team name as used in the web inteface
 - the institution name as used on the scoreboard
 - the institution short name as used in the jury interface and certain exports
 - a country code in form of ISO 3166-1 alpha-3
--- a/doc/manual/import.rst
+++ b/doc/manual/import.rst
@@ -22,10 +22,12 @@ Prepare a file called ``groups.json`` which contains the team categories.
 It should be a JSON array with objects, each object should contain the following
 fields:
 
-- ``id``: the category ID
-- ``name``: the name of the team category
-- ``hidden`` (defaults to ``false``): whether the team category is hidden
-- ``sortorder`` (defaults to ``0``): the sort order of the team category
+- ``id``: the (integer) category ID to use. Must be unique
+- ``name``: the name of the team category as shown on the scoreboard
+- ``hidden`` (defaults to ``false``): if ``true``, teams in this category will
+  not be shown on the scoreboard
+- ``sortorder`` (defaults to ``0``): the sort order of the team category to use
+  on the scoreboard. Categories with the same sortorder will be grouped together.
 
 Example ``groups.json``::
 
@@ -35,10 +37,10 @@ Example ``groups.json``::
     "hidden": true
   }, {
     "id": "47",
-    "name": "Netherlands"
+    "name": "Participants"
   }, {
     "id": "23",
-    "name": "United Kingdom"
+    "name": "Spectators"
   }]
 
 To import the file using the jury interface, go to `Import / export`, select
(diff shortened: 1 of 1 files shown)


doc/manual/import.rst
```diff
@@ -1,17 +1,56 @@
...
+There are two formats to import team categories: a JSON format and a legacy TSV format.
+
+Using JSON
+^^^^^^^^^^
+
+Prepare a file called ``groups.json`` which contains the team categories.
+It should be a JSON array with objects, each object should contain the following
+fields:
+
+- ``id``: the category ID
```
thijskh, line 25: Something you can just make up?

doc/manual/import.rst
```diff
@@ -1,17 +1,56 @@
...
+- ``name``: the name of the team category
+- ``hidden`` (defaults to ``false``): whether the team category is hidden
```
thijskh, line 27: The explanations of the fields do not really add information that is not already in the name

doc/manual/import.rst
```diff
@@ -1,17 +1,56 @@
...
+
+Example ``groups.json``::
+
+  [{
+    "id": "13337",
+    "name": "Companies",
+    "hidden": true
+  }, {
+    "id": "47",
+    "name": "Netherlands"
```
thijskh, line 38: Maybe not too clear an example because country information is also in affiliations...

doc/manual/import.rst
```diff
@@ -26,13 +65,95 @@ Example ``groups.tsv``::
...
     http --check-status -b -f POST "<API_URL>/users/groups" tsv@groups.tsv
 
+Importing team affiliations
+---------------------------
+
+.. note::
+
+    The team TSV import automatically imports team affiliations as well.
+
+Prepare a file called ``organizations.json`` which contains the teams.
```
thijskh, line 82: “which contains the _affiliations_”

doc/manual/import.rst
```diff
@@ -53,13 +174,20 @@ Example ``teams2.tsv``::
    2	447837	25	Pleading not FAUlty	Friedrich-Alexander-University Erlangen-Nuremberg	FAU	DEU	INST-43
 
 
-To import the file run the following command::
+To import the file using the jury interface, go to `Import / export`, select
+`groups` under `Tab-separated import`, select your file and click `Import`.
```
thijskh, line 178: Select _teams_

thijskh (changes requested):
Alright, it’s very explicit not but maybe a tad redundant? But definitely better than before

doc/manual/import.rst
nickygerritsen: So remove all explanations for both formats? That is also not ideal since then we don’t know formal name vs name?

doc/manual/import.rst
nickygerritsen: Good point. I think they should at least be integer so I will add something

doc/manual/import.rst
nickygerritsen: A better idea? WF uses them for super regions but otherwise it is mostly not used, although NWERC indeed uses countries I think (to have a “best from NL”)

nickygerritsen:
> Alright, it’s very explicit not but maybe a tad redundant? But definitely better than before

We can also remove the tsv parts that are legacy? But then the last chapter is odd

doc/manual/import.rst
thijskh: Better is to provide actual information?. E.g.: name as shown on the scoreboard. Hidden means the teams can participate but will not be shown on the public scoreboard.

doc/manual/import.rst
thijskh: At least in the past we used categories like "Participants", "Spectators", "Companies" and "Contest staff" (hidden true)

thijskh:
Let's keep the redundancy as is and refactor the text later separately (if at all)

doc/manual/import.rst
```diff
@@ -121,12 +124,13 @@ Prepare a file called ``teams.json`` which contains the teams.
 It should be a JSON array with objects, each object should contain the following
 fields:
 
-- ``id``: the team ID
+- ``id``: the (integer) team ID. Must be unoque
```
thijskh, line 127: unique

doc/manual/import.rst
```diff
@@ -121,12 +124,13 @@ Prepare a file called ``teams.json`` which contains the teams.
...
 - ``icpc_id`` (optional): an external ID, e.g. from the ICPC CMS, may be empty
-- ``group_ids``: an array with one element: the category ID
-- ``name``: the team name
-- ``display_name`` (optional): the team display name
-- ``organization_id``: the external ID of the team affiliation
+- ``group_ids``: an array with one element: the category ID this team belomgs to
```
thijskh, line 129: belongs

doc/manual/import.rst
```diff
@@ -121,12 +124,13 @@ Prepare a file called ``teams.json`` which contains the teams.
...
+- ``name``: the team name as used on the web inteface
```
thijskh, line 130: in the web interface

doc/manual/import.rst
```diff
@@ -158,12 +162,12 @@ Prepare a file called ``teams2.tsv`` which contains the teams.
...
 Each of the following lines must contain the following elements separated by tabs:
 
-- the team ID
+- the (integer) team ID. Must be unique
 - an external ID, e.g. from the ICPC CMS, may be empty
-- the category ID
-- the team name
-- the institution name
-- the institution short name
+- the category ID this team belomgs to
```
thijskh, line 167: belongs

doc/manual/import.rst
```diff
@@ -158,12 +162,12 @@ Prepare a file called ``teams2.tsv`` which contains the teams.
...
+- the team name as used on the web inteface
```
thijskh, line 168: used in the web interface

thijskh (approved):
Besides from some typos, fine to merge

PR #920, opened 2020-11-13
Problem attachments

nickygerritsen:
The [problem package format](https://icpc.io/problem-package-format/spec/problem_package_format#attachments) has the notion of attachments.

During the BAPC preliminaries we got the question whether we supported this, which we didn't yet.
So I came up with a simple implementation.

Some screenshots:

Jury interface:
Screenshot 2020-11-13 at 12 13 39

Team / public problemset page:
Screenshot 2020-11-13 at 12 13 45


RagnarGrootKoerkamp:
Nice!
Some thoughts:
- There can be multiple/many attachments. Download as a zip for contestants may be nicer than individual downloads. Although download as a file is nice if it's only one.
- For interactive problems (the typical case where attachments are used), download of samples should be disabled. The sample input typically contains instructions on how to run the interactor, which are not meant for contestants to see. In our case, we provide replacement samples as an attachment.

nickygerritsen:
> Nice!
> Some thoughts:
> 
> * There can be multiple/many attachments. Download as a zip for contestants may be nicer than individual downloads. Although download as a file is nice if it's only one.
> * For interactive problems (the typical case where attachments are used), download of samples should be disabled. The sample input typically contains instructions on how to run the interactor, which are not meant for contestants to see. In our case, we provide replacement samples as an attachment.

I think the latter requires something in the spec then, since that is not really a nice case. But if the samples contain stuff the team should not see then it is not a sample imho. Since you ship real samples in the zip anyway, wouldn’t marking the samples as non sample solve this?
For the former I can cook something up that if you have more than file it offers a zip. But if you do that then why not put a zip in attachments/?

RagnarGrootKoerkamp:
* Yeah, agree that we should talk about what it means for something to be in `data/samples/`.
   I'd still like them to be in the `samples` directory, since the PDF is rendered using the sources there, but I see your point as well.

* I'd prefer to not put a zip file in my git repository, but I see your point. I could always convert multiple files to a zip when exporting it for DOMjudge import.

nickygerritsen:
I wonder what the others think about the zip? Adding a “all attachments” button if w he’s e > 1 that creates a zip is trivial so we could do it

thijskh:
How many attachments do we normally expect a problem to have? I can see why you'd want a zip file if they are many but it's already such a button fest on that page. If it's just one or two or three normally I don't see much value for the zip button.

nickygerritsen:
@RagnarGrootKoerkamp what do you think? Do you often have more than say 3?

RagnarGrootKoerkamp:
Since Kattis provides individual download links, let's go with that.

As said before, we can always provide zip files in the attachments directory if needed.

thijskh:
So can this be merged in current form?

nickygerritsen:
If others agree yes

webapp/src/Controller/Team/ProblemController.php
```diff
@@ -134,6 +135,64 @@ public function problemTextAction(int $probId)
...
+     * @Route(
+     *     "/{probId<\d+>}/attachment/{attachmentId<\d+>}",
+     *     name="team_problem_attachment"
+     *     )
+     * @param int    $probId
+     * @param int    $attachmentId
+     * @return \Symfony\Component\HttpFoundation\Response
+     * @throws \Doctrine\ORM\NonUniqueResultException
+     */
+    public function attachmentAction(int $probId, int $attachmentId)
```
meisterT, line 148: Is this just duplicated code with above? Can you dedup it somewhere?

nickygerritsen: I will try this

nickygerritsen: I gave it a try and removed even more duplicated code. Can you check if you agree with the approach?

webapp/src/Migrations/Version20201113094653.php
```diff
@@ -0,0 +1,40 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20201113094653 extends AbstractMigration
+{
+    public function getDescription() : string
+    {
+        return '';
```
meisterT, line 17: nit: add a description

meisterT (approved):

PR #923, opened 2020-11-14
Fix num teams query for contests.

nickygerritsen:
We can't load it in one query, since we will get data double if you add both a team directly and through a category.

Feels a bit ugly to do 3 queries, but the only other option I see would be to do a DISTINCT instead of a count and I think that is even slower.

### DIFF
--- a/webapp/src/Controller/Jury/ContestController.php
+++ b/webapp/src/Controller/Jury/ContestController.php
@@ -305,9 +305,9 @@ class ContestController extends BaseController
             if ($contest->isOpenToAllTeams()) {
                 $contestdata['num_teams'] = ['value' => '<i>all</i>'];
             } else {
-                $teamCount = $em
+                $teamIds = $em
                     ->createQueryBuilder()
-                    ->select('COUNT(t.teamid) AS cnt')
+                    ->select('DISTINCT t.teamid')
                     ->from(Team::class, 't')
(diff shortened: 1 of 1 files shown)


eldering:
I think it would be better to just fetch all team ids and then do a unique count in code. That's only a single query and prevents a lot of duplicate and bulky query code.

eldering (changes requested):
See my comment: let's not do 3 bulky SQL queries where one suffices.

nickygerritsen:
> See my comment: let's not do 3 bulky SQL queries where one suffices.

If you believe that is faster I can update to it. Will to this afternoon.

eldering:
> > See my comment: let's not do 3 bulky SQL queries where one suffices.
> 
> If you believe that is faster I can update to it. Will to this afternoon.

I'm not sure it's faster (although could be: the overhead of each query maybe be bigger than data transfer), but mostly it leads to cleaner code.

nickygerritsen:
Probably speed is not a real issue anyway, this page probably doesn’t contain a million contests and isn’t visited too often

nickygerritsen:
This seems to work just fine :)

eldering (approved):
Also looks fine to me.

thijskh:
Why is this not `COUNT(DISTINCT teamid)`?

PR #927, opened 2020-11-18
This shorter syntax seems to work

mvr320:
Updating the gitlabci:(1.0->2.0) would be slightly easier when we can use this so we do not have to pass a lot of the options.

As I'm not 100% why the other syntax is chosen we might need to check if something fails here.


nickygerritsen (approved):
If it works ship it. I also don’t know why we did it 🤷‍♂️

PR #928, opened 2020-11-18
Use anchors to build 2 database jobs

mvr320:
Test the integration test also against the traditional MySQL server. In
the future we can extend this with different judgedaemons or different
base images.

This does yield a speed penalty currently.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -23,7 +23,8 @@ include:
   timeout: 10m
 
 .retry_time: &long_job
-  timeout: 20m
+  timeout: 45m
+# Due to the retry this will be worst case 3*timeout before the job fails
 
 check syntax:
   <<: *retry_job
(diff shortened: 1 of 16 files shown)


nickygerritsen (approved):
Sweet

mvr320:
Removed the non-generated integration test and added the comments back.

mvr320:
@nickygerritsen can you check if the migration makes sense (Do I indeed have to change this much for the Entity files?)

Yesterday evening this terminated correctly, so I dont expect this build to fail and should be in mergeable state.

nickygerritsen (commented):
There is a `webapp/src/Controller/Jury/.ProblemController.php.swp` file still left.

Did you generate the migration using `webapp/bin/console doctrine:migrations:diff`? I'm not sure if that works with renaming columns though.

PR #931, opened 2020-11-22
Enable caching of integration job

mvr320:
As I'm not 100% sure if it will indeed speedup currently

As discussed with @nickygerritsen this was disabled as Gitlab had issues but this should store the caches on S3 to speedup the downloading steps in the integration job

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -186,12 +186,15 @@ visual_compare:
     when: always
     paths:
       - gitlabartifacts
+  cache:
+    key: integration
+    paths:
+      - lib/vendor/
 
 # TODO: Re-enable when gitlab is in better shape...
 #  cache:
 #    key: integration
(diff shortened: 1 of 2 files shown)


nickygerritsen (approved):

mvr320:
So we actually hit a timeout when uploading the chroot, so maybe actually making a separate image for the integration jobs with: deb caching (https://manpages.debian.org/unstable/debootstrap/debootstrap.8.en.html) might be an idea? So the cache is already put into the dockerimage and would only get larger when we need new debs?

For now the lib/vendor works but the chroot would need additional work to actually give a real advantage.

nickygerritsen:
So then let’s do lib/vendor for now and the other thing later

nickygerritsen (approved):

nickygerritsen (approved):

PR #933, opened 2020-11-23
W3m

meisterT:


meisterT:
Ok, this works as expected and you can uncollapse the section `Short error description` and see something like this:

image

PR #935, opened 2020-11-29
Ignore failures to run fix_permissions script if you aren't root

ubergeek42:
Because everything else has it and it should be possible to run install as non-root user.


eldering (approved):
LGTM.
I actually have this same change in my local, uncommitted changes. ;-)

PR #938, opened 2020-12-08
The carriage return should not be in the regex

mvr320:
I'm not of the generation who has used a typewriter, so I forgot to take care of those.

In the future the whole file should be rewritten to be better understandable, I want to wait till this has proven itself.

Also implemented @RagnarGrootKoerkamp his idea, as its better than nothing

### DIFF
--- a/gitlab/visualcompare.sh
+++ b/gitlab/visualcompare.sh
@@ -51,9 +51,9 @@ do
         STATE="success"
     fi
     MANY=`ls $dir|wc -l`
-    FILE="$dir"
+    FILE="browse/$dir"
     if [ $MANY -eq 1 ]; then
-        FILE=$dir/`ls $dir`
+        FILE="file/$dir/"`ls $dir`
         CHANGE=1
     elif [ $MANY -gt 1 ]; then
         CHANGE=1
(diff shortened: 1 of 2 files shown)


gitlab/visualcompare.sh
```diff
@@ -64,8 +64,16 @@ do
         -X POST \
         -H "Authorization: token $GH_BOT_TOKEN_OBSCURED" \
         -H "Accept: application/vnd.github.v3+json" \
-        -d "{\"state\": \"$STATE\", \"target_url\": \"$CI_JOB_URL/artifacts/file/$FILE\", \"description\":\"UI changes\", \"context\": \"UI diffs ($dir)\"}"
-      exit 1
+        -d "{\"state\": \"$STATE\", \"target_url\": \"$CI_JOB_URL/artifacts/$FILE\", \"description\":\"UI changes\", \"context\": \"UI diffs ($dir)\"}"
+      if [ $STATE=="failure" ]; then
+        cat <<EOF > $failingchanges/README
+List you changed endpoints in a message on the PR. Regex is allowed.
```
nickygerritsen, line 70: This sentence feels odd?

mvr320: I think because it is, changed it a bit and made the syntax also slightly better

nickygerritsen (approved):

gitlab/visualgithubprdiscussion.py
```diff
@@ -18,7 +18,7 @@ def toMessage(url):
     index=message.find(BEGIN_RGX)
     if index>-1:
         for line in message[index:].split('\n')[1:]:
```
thijskh, line 20: Probably clearer to use `splitlines()` here and do away with the strip below?

mvr320: Indeed better, is implemented now.

Locally you can see the difference now with:
```python3 gitlab/visualgithubprdiscussion.py public 910```
Where public is in the regex and is therefor a wanted change
```python3 gitlab/visualgithubprdiscussion.py anotherendpoint 910```
which is not in the regex and therefor would be a failing change

thijskh (approved):

PR #939, opened 2020-12-12
Add filter on result on jury submissions page.

nickygerritsen:
As requested by @RagnarGrootKoerkamp .

Screenshot:
image

### DIFF
--- a/webapp/src/Twig/TwigExtension.php
+++ b/webapp/src/Twig/TwigExtension.php
@@ -156,6 +156,7 @@ class TwigExtension extends AbstractExtension implements GlobalsInterface
             new TwigFilter('printtimeHover', [$this, 'printtimeHover'], ['is_safe' => ['html']]),
             new TwigFilter('printResult', [$this, 'printResult'], ['is_safe' => ['html']]),
             new TwigFilter('printValidJuryResult', [$this, 'printValidJuryResult'], ['is_safe' => ['html']]),
+            new TwigFilter('printValidJurySubmissionResult', [$this, 'printValidJurySubmissionResult'], ['is_safe' => ['html']]),
             new TwigFilter('printHost', [$this, 'printHost'], ['is_safe' => ['html']]),
             new TwigFilter('printYesNo', [$this, 'printYesNo']),
             new TwigFilter('printSize', [Utils::class, 'printSize'], ['is_safe' => ['html']]),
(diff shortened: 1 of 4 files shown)


webapp/templates/jury/partials/submission_list.html.twig
```diff
@@ -72,11 +72,27 @@
...
+                {% endif %}
+            {% elseif submission.judgings.first is empty or submission.judgings.first.result is empty %}
+                {% if submission.judgehost %}
+                    {% set result = 'judging' %}
+                {% else %}
+                    {% set result = 'queued' %}
+                {% endif %}
+            {% else %}
+                {% set result = submission.judgings.first.result %}
+            {% endif %}
```
ubergeek42, line 88: this is ugly, but I don't see an obvious way to make it clearer

ubergeek42 (approved):
nice! :tada:

webapp/templates/jury/partials/submission_list.html.twig
nickygerritsen: Yes I had the same issue.... We could put this in a twig function and reuse it? But then too-late is hard

thijskh:
Instead of keep adding text boxes, would it be difficult to make it one box?

Filter on: [ [Language C] [Result correct] ]

nickygerritsen:
> Instead of keep adding text boxes, would it be difficult to make it one box?
> 
> Filter on: [ [Language C] [Result correct] ]

I wonder if that is not less user friendly? Probably technically it's possible but non-trivial (since we use 3 different AJAX requests for the top 3 boxes and none for the bottom one).

RagnarGrootKoerkamp:
Gitlab does this, and I don't like it. I much prefer separate boxes.

webapp/templates/jury/partials/submission_list.html.twig
thijskh: Maybe we should not be doing this in twig in the first place?
Something for the controller to determine what the 'verdict' is,including juding and queued. And also add a too-late flag as well that twig can act on?

nickygerritsen: Yes this might be better. I will try it

nickygerritsen: Gave it a try

webapp/src/Twig/TwigExtension.php
```diff
@@ -549,6 +550,53 @@ public function printValidJuryResult($result): string
...
 
+    /**
+     * Print the given submission for the jury, assuming it is valid
+     *
+     * @param Submission $submission
+     * @param bool       $forDisplay
+     *
+     * @return string
+     */
+    public function printValidJurySubmissionResult(Submission $submission, bool $forDisplay = true)
```
meisterT, line 561: nit: add `: string`

meisterT (approved):

PR #942, opened 2020-12-12
Add API actions to add a single team, organization and group

nickygerritsen:
This uses the existing ImportExportService to be able to add single entities. You can `POST` both JSON as well as form data.

### DIFF
--- a/webapp/src/Controller/API/AbstractRestController.php
+++ b/webapp/src/Controller/API/AbstractRestController.php
@@ -107,18 +107,14 @@ abstract class AbstractRestController extends AbstractFOSRestController
     /**
      * Render the given data using the correct groups
      *
-     * @param Request  $request
      * @param mixed    $data
-     * @param int      $statusCode
-     * @param string[] $headers
-     *
-     * @return Response
+     * @param string[] $extraheaders
      */
     protected function renderData(
         Request $request,
         $data,
         int $statusCode = Response::HTTP_OK,
-        array $headers = []
+        array $extraheaders = []
     ): Response {
         $view = $this->view($data);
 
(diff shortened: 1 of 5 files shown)


nickygerritsen:
This might need additions to the documentation, although I'd say we have the API docs and these map directly to them.

If documentation is needed, I'd like to hear a proposal on what to add. And I didn't create it yet since I first wanted opinions on whether this is a good idea anyway.

thijskh:
I don’t think we need additional documentation besides the generated API docs for straightforward CRUD endpoints.

webapp/src/Controller/API/GroupController.php
```diff
@@ -66,6 +70,44 @@ public function singleAction(Request $request, string $id)
...
+     *         mediaType="multipart/form-data",
+     *         @OA\Schema(ref="#/components/schemas/TeamCategory")
+     *     ),
+     *     @OA\MediaType(
+     *         mediaType="application/json",
+     *         @OA\Schema(ref="#/components/schemas/TeamCategory")
+     *     )
+     * )
+     * @OA\Response(
+     *     response="200",
```
mvr320, line 95: Why should this not be 201? https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201

nickygerritsen: 201 might make sense. We should then also set the Location header

webapp/src/Controller/API/AbstractRestController.php
```diff
@@ -105,12 +106,20 @@ protected function performSingleAction(Request $request, string $id)
...
+     *
      * @return Response
      */
-    protected function renderData(Request $request, $data): Response
-    {
+    protected function renderData(
+        Request $request,
+        $data,
+        int $statusCode = Response::HTTP_OK,
+        array $headers = []
```
meisterT, line 121: nit: let's call this `extraHeaders` or similar?

webapp/src/Controller/API/AbstractRestController.php
```diff
@@ -105,12 +106,20 @@ protected function performSingleAction(Request $request, string $id)
 
     /**
      * Render the given data using the correct groups
-     * @param Request $request
-     * @param mixed $data
+     *
+     * @param Request  $request
```
meisterT, line 110: Do we need the `@param` annotation here with the type hints below?

meisterT (approved):

webapp/src/Controller/API/AbstractRestController.php
nickygerritsen: Nope will remove them

PR #944, opened 2020-12-19
Add contest name to title elements.

thijskh:
So if you have multiple tabs open you can see which one is which.

I've done this for pages that are specific to a contest: scoreboard, problemset, submissions, balloons, analysis. It does not make sense to me to do it globally. e.g. "Configuration settings NWERC20" is not a good title since the configuration settings are not contest specific.

### DIFF
--- a/webapp/templates/jury/analysis/contest_overview.html.twig
+++ b/webapp/templates/jury/analysis/contest_overview.html.twig
@@ -1,6 +1,6 @@
 {% extends "jury/base.html.twig" %}
 
-{% block title %}Analysis - Contest{% endblock %}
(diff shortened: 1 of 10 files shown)


nickygerritsen (commented):
I believe some of these pages can be accessed without an active contest and just display a message. If so these are now broken. Might be good to add a null check?

Otherwise looks good!

ubergeek42 (approved):
Nice!

thijskh:
@nickygerritsen How is a "null check" different from the `| default('')` that I'm using?

nickygerritsen:
> @nickygerritsen How is a "null check" different from the `| default('')` that I'm using?

lol was looking on phone. But does that work with `current_contest.something` if `current_contest` is null? If so go ahead and merge

PR #945, opened 2020-12-20
Validate team and role in submit client.

meisterT:

PR #946, opened 2020-12-21
Fix API exception renderer to display messages correctly for configured errors.

nickygerritsen:
Also display usable message when a team is not linked when submitting.

This is the better fix for #945.

In prod mode when you don't have the team permission you now get:

```bash
$ http --auth admin:admin POST http://localhost:12345/api/contests/2/submissions
HTTP/1.1 403 Forbidden
Access-Control-Allow-Origin: *
Cache-Control: no-cache, private
Connection: keep-alive
Content-Type: application/json
Date: Mon, 21 Dec 2020 16:41:17 GMT
Server: nginx/1.18.0 (Ubuntu)
Transfer-Encoding: chunked
Vary: Accept

{
    "code": 403,
    "message": "You need to be a team to add a submission"
}
```

And when you do have the permission but no team linked:

```bash
$ http --auth admin:admin POST http://localhost:12345/api/contests/2/submissions  problem=dummy language=dummy
HTTP/1.1 400 Bad Request
Access-Control-Allow-Origin: *
Cache-Control: no-cache, private
Connection: keep-alive
Content-Type: application/json
Date: Mon, 21 Dec 2020 16:42:08 GMT
Server: nginx/1.18.0 (Ubuntu)
Transfer-Encoding: chunked
Vary: Accept

{
    "code": 400,
    "message": "User does not belong to a team"
}
```

In dev mode you get the backtrace as well, but you also get this message.

### DIFF
--- a/webapp/src/Controller/API/SubmissionController.php
+++ b/webapp/src/Controller/API/SubmissionController.php
@@ -113,7 +113,7 @@ class SubmissionController extends AbstractRestController
      * @return int
      * @Rest\Post("")
      * @OA\Post()
-     * @IsGranted("ROLE_TEAM")
+     * @IsGranted("ROLE_TEAM", message="You need to have the Team Member role to add a submission")
      * Uploading an array of files in swagger is not supported, see
      * @OA\RequestBody(
      *     required=true,
--- a/webapp/config/services.yaml
+++ b/webapp/config/services.yaml
@@ -59,7 +59,7 @@ services:
         public: true
         class: App\FosRestBundle\FlattenExceptionHandler
         arguments:
-            - '@fos_rest.exception.codes_map'
+            - '@fos_rest.exception.messages_map'
             - '%kernel.debug%'
 
     Metadata\MetadataFactoryInterface: '@jms_serializer.metadata_factory'


webapp/src/Controller/API/SubmissionController.php
```diff
@@ -113,7 +113,7 @@ public function singleAction(Request $request, string $id)
      * @return int
      * @Rest\Post("")
      * @OA\Post()
-     * @IsGranted("ROLE_TEAM")
+     * @IsGranted("ROLE_TEAM", message="You need to be a team to add a submission")
```
thijskh, line 116: I would reword this as "You need to have the Team Member role to..." so it's more clear what exact action you need to do on the user to fix it.

thijskh (approved):
It works, great. One comment on the contents of the feedback string.

The submit client handles this because it will output the given JSON in case of an error. It might be overkill for the submit client to attempt to parse it and if there's a "code" and "message" in it, display it differently? I think so. The output is now:

```
Do you want to continue? (y/n) 
{"code":400,"message":"User does not belong to a team"}
[Dec 22 20:35:06.698] ./submit[2036200]: error: Submission failed (code 400)
```

```
Do you want to continue? (y/n) 
{"code":403,"message":"You need to be a team to add a submission"}
[Dec 22 20:35:43.321] ./submit[2036204]: error: Submission failed (code 403)

```

nickygerritsen:
Both ideas sound good. I can try to give my non-php skills a try

nickygerritsen:
Oh you already did the submit client part 🤣

thijskh:
I didn't in the end. It seemed too much hassle for what it's worth since the actual display of the JSON is not that bad. The output of submit is very technical already.

PR #947, opened 2020-12-23
Don't swallow the first leading newline for program output.

meisterT:
While the default validator typically ignores whitespace, it may be
configured differently. Also it can become confusing when looking at
debugging output.

Thanks to Kevin Thiotanry for reporting!


nickygerritsen (approved):

thijskh:
I don't follow this yet. So we swallow the newline somewhere else, and add it back in the twig template?

meisterT:
No, we don't swallow it. The browser does.

See https://codepen.io/meisterT/pen/gOwGoaQ for a minimal example.

PR #950, opened 2020-12-23
Improve web tests by moving more to shared functions.

nickygerritsen:
We can always do more, but I think this is a good start. The ProblemControllerTest still has a lot of code but I don't see any fast way to optimize it.
The other two test classes have less code now.


thijskh (approved):

thijskh:
Great improvement. To test it out a little I've added some basic tests of the Balloon and Config controllers.

Did not get submitting the Config settings form to work yet, but that is probably me.

PR #952, opened 2020-12-30
Phpunit8

mvr320:
This is just to test if the outcome is as expected.

Before this can be merged there are some things which should be tested/changed:
- Test with symfony bridge (suggestion by composer, see https://github.com/DOMjudge/domjudge/pull/954)
- changes for the gitlabci image (now currently added to the file itself)
- Coverage might be able in gitlab (https://aymdev.io/en/blog/post/php-tests-reports-with-gitlab-ci or https://www.cypress.io/blog/2019/10/22/show-code-coverage-on-gitlab-ci/) but the current uploading as artifact is good enough in my opinion

When the coverage is better the CRAP can be used to limit the amount of "low quality" code which is accepted per PR but looking at the values now I think that is not worth it yet.

This current version seems to also alert on deprecated code, so the advice from Symfony to use the bridge might not be needed, both PR's are fine to merge (but only 1 out of 2)

### DIFF
--- a/gitlab/unit-tests.sh
+++ b/gitlab/unit-tests.sh
@@ -17,4 +17,20 @@ cd /opt/domjudge/domserver
 export APP_ENV="test"
 
 # Run phpunit tests.
-lib/vendor/bin/phpunit -c webapp/phpunit.xml.dist --log-junit ${CI_PROJECT_DIR}/unit-tests.xml --colors=never
+php -dpcov.enabled=1 -dpcov.directory=webapp/src lib/vendor/bin/phpunit -c webapp/phpunit.xml.dist --log-junit ${CI_PROJECT_DIR}/unit-tests.xml --colors=never --coverage-html=${CI_PROJECT_DIR}/coverage-html > phpunit.out
+CNT=$(sed -n '/Generating code coverage report/,$p' phpunit.out | wc -l)
+FILE=deprecation.txt
+sed -n '/Generating code coverage report/,$p' phpunit.out > ${CI_PROJECT_DIR}/$FILE
+if [ $CNT -eq 1 ]; then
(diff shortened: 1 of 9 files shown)


gitlab/unit-tests.sh
```diff
@@ -17,4 +18,9 @@ cd /opt/domjudge/domserver
 export APP_ENV="test"
 
 # Run phpunit tests.
-lib/vendor/bin/phpunit -c webapp/phpunit.xml.dist --log-junit ${CI_PROJECT_DIR}/unit-tests.xml --colors=never
+apt update
```
nickygerritsen, line 21: Although we build this in the docker image later I assume, why do we update twice?

webapp/phpunit.xml.dist
```diff
@@ -11,10 +11,6 @@
         <ini name="error_reporting" value="-1" />
         <server name="APP_ENV" value="test" force="true" />
         <server name="SHELL_VERBOSITY" value="-1" />
-        <server name="SYMFONY_PHPUNIT_REMOVE" value="" />
-        <server name="SYMFONY_PHPUNIT_VERSION" value="6.5" />
-        <ini name="date.timezone" value="UTC" />
```
nickygerritsen, line 16: Don't we need the timezone anymore? I remember we needed it for some tests if we ran it on calca

mvr320: This was what `auto` happened, I see no problem in setting this explicit so I'll add it back if this will be the final PR

gitlab/unit-tests.sh
mvr320: Probably copy/paste error.

nickygerritsen (approved):

PR #953, opened 2020-12-30
Repeat firefox request to warm up cache

mvr320:
Other alternatives would be to use:
- https://wkhtmltopdf.org/
- Different CI image where chromium does not use the snap version

Example of failure can be found at: https://gitlab.com/DOMjudge/domjudge/-/jobs/938481152/artifacts/file/screenshotsmaster/public.html-ff.png

### DIFF
--- a/gitlab/visualpr.sh
+++ b/gitlab/visualpr.sh
@@ -124,6 +124,9 @@ do
         urlpath=$(sed "s/$prefix//g"<<<$file)
         # Small risk of collision
         storepath=$(sed "s/\//_s_/g"<<<$urlpath)
+	# Sometimes not all images load, (See https://bugzilla.mozilla.org/show_bug.cgi?id=1412061)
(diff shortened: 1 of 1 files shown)


gitlab/visualpr.sh
```diff
@@ -124,6 +124,7 @@ do
         urlpath=$(sed "s/$prefix//g"<<<$file)
         # Small risk of collision
         storepath=$(sed "s/\//_s_/g"<<<$urlpath)
+        firefox -screenshot /dev/zero http://localhost/$url/$urlpath
```
meisterT, line 127: This should be `/dev/null` right? And perhaps add a comment why you are doing this.

mvr320: Although both will do what I need, I think the `/dev/null` is indeed better together with the comment. Will fix.

meisterT (approved):

PR #957, opened 2021-01-05
Tests if the ContestController works

mvr320:
Currently the card header does not work

Add different contest names

Tried a possible SQL problem, Unicode problem, space problem and
something with really a lot of length

### DIFF
--- a/webapp/tests/Controller/Jury/ContestControllerTest.php
+++ b/webapp/tests/Controller/Jury/ContestControllerTest.php
@@ -1,0 +1,210 @@
+<?php declare(strict_types=1);
+
+namespace App\Tests\Controller\Jury;
+
+use App\Entity\Contest;
+
+class ContestControllerTest extends JuryControllerTest
+{
+    protected static $baseUrl         = '/jury/contests';
+    protected static $shortTag        = 'contest';
+    protected static $deleteEntities  = array('name' => ['Demo practice session']);
+    protected static $getIDFunc       = 'getCid';
+    protected static $exampleEntries  = ['Demo contest','Demo practice session'];
+
+    protected static $DOM_elements    = array('h1' => ['Contests'],
+                                              'h3' => array('admin' => ['Current contests', 'All available contests'],
+                                                            'jury' => []));
+    protected static $className       = Contest::class;
+
+    /*public function pRrovideAddEntity () : \Generator
+    {
+        $timeStrings = $this->getTimes();
+
+        $contests  = [
+            'NewContest',
+            '⛀⛁⛂⛃⛄⛅⛆⛇⛈⛉⛊⛋⛌⛍⛎⛏',
+            'name with spaces',
+            str_repeat("A", 255)
+        ];
+
+        foreach ($contests as $contest) {
+            yield [$contest,'11111', $timeStrings[0]];
+        }
+        // Try all possible values for the radio buttons
+        // TODO: In the future test if the config option also works
+        for($i=0; $i<2**5; $i++)
+        {
+            $bin = sprintf( "%05d", decbin($i));
+            yield ['RadioTest', $bin, $timeStrings[0]];
+        }
+        // Try all timings
+        foreach($timeStrings as $timeString)
+        {
+            yield ['TimeInterval', '11111', $timeString];
+        }
+    }*/
+
+    /*public function getTimes () : array
+    {
+        $res = [[
+            "2020-01-01 09:00:00 Europe/Amsterdam",
+            "2021-01-01 09:00:00 Europe/Amsterdam",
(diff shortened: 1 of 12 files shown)


webapp/tests/Controller/Jury/ContestControllerTest.php
```diff
@@ -0,0 +1,204 @@
...
+     * @var string baseContest The basecontest we use throughout these tests
+     * @dataProvider provideBasePage
+     */
+    public function testContestsOverview (
+	string $role,
+	int $statuscode,
+	string $baseContest) : void
+    {
+    	static::$roles = [$role];
+        $this->logIn();
```
nickygerritsen, line 27: The new tests log in automatically, so i think if you want something like this best is to first log out

webapp/tests/Controller/Jury/ContestControllerTest.php
```diff
@@ -0,0 +1,204 @@
...
+
+    /**
+     * Test that jury role can NOT add a new contest
+     */
+    public function testCheckAddContestJury () : void
+    {
+        static::$roles = ['jury'];
+        $this->logIn();
+        $this->verifyPageResponse('GET', static::$baseUrl, 200);
+        $this->assertSelectorNotExists('a:contains('.static::$addButton.')');
```
nickygerritsen, line 69: Also check that the /add url gives a 403?

webapp/tests/Controller/Jury/ContestControllerTest.php
```diff
@@ -0,0 +1,204 @@
...
+     * @var string radioValues The Radio settings of the contest
+     * @var array timeString The time for a certain field as string
+     * @dataProvider provideAddContest
+     */
+     public function testCheckAddContest (
+        string $contest,
+        string $radioValues,
+        array $timeString) : void
+    {
+        $this->logIn();
```
nickygerritsen, line 84: This login call should not be needed at all i think

webapp/tests/Controller/Jury/ContestControllerTest.php
```diff
@@ -0,0 +1,204 @@
...
+
+    /**
+     * Data provider used to test if we can add contests
+     * - the to be created contest
+     * - radio button boolean values (Bit string)
+     */
+    public function provideAddContest () : \Generator
+    {
+        //* - The different times of the contest
+        //TODO: Try things like: "admin'--"
```
nickygerritsen, line 157: I wonder if we should really test these sql injection things. This is one of the reasons we use a framework and not direct queries; the framework takes care of it.

webapp/tests/Controller/Jury/ContestControllerTest.php
```diff
@@ -0,0 +1,204 @@
...
+
+    /**
+     * Remove the contest
+     */
+    /* TODO: fix this
+    public function testCheckDeleteContest () : void
+    {
+        $contestShort = 'demo';
+        $contestLong = 'Demo contest';
+        $cid = '2';
```
nickygerritsen, line 192: Proposal: use the EntityManagerInterface to get the contest with the given short name?

nickygerritsen (commented):
Formatting is....strange but i guess you’ll fix that later.

mvr320:
> 
> 
> Formatting is....strange but i guess you’ll fix that later.

WSL... so yes will fix it later.

webapp/tests/Controller/Jury/ContestControllerTest.php
mvr320: Sounds good, do we already do this somewhere? Otherwise it will take me some extra time

nickygerritsen: Maybe not in the tests. I will check when on a bigger device

webapp/tests/Controller/Jury/ContestControllerTest.php
mvr320: I agree with you that SQL injection should not happen (and if it does... upstream has to fix it) but they are interesting edge cases which are also teamnames.... but as in code it already fails its a TODO which might actually be removed in the end.

Atleast in my contest I always had people who tried DROP table... and especially in the beginning I was slightly afraid of having to enter these.. (never worked ofcourse)

I'll just remove the TODO for now.

webapp/tests/Controller/Jury/ContestControllerTest.php
mvr320: they login as admin I assume?

nickygerritsen: As what you define as the static $roles variable

PR #959, opened 2021-01-06
Add Content-Security-Policy in PHP so we can have unsafe-eval when th…

nickygerritsen:
…e profiler is enabled

### DIFF
--- a/webapp/src/EventListener/AddContentSecurityPolicyListener.php
+++ b/webapp/src/EventListener/AddContentSecurityPolicyListener.php
@@ -1,0 +1,38 @@
+<?php declare(strict_types=1);
+
+namespace App\EventListener;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\ResponseEvent;
+use Symfony\Component\HttpKernel\Profiler\Profiler;
+
+class AddContentSecurityPolicyListener implements EventSubscriberInterface
+{
+    /**
+     * @var Profiler|null
+     */
+    protected $profiler;
+
+    public function __construct(?Profiler $profiler)
+    {
+        $this->profiler = $profiler;
+    }
+
+    /**
+     * @inheritDoc
+     */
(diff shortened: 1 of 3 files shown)


thijskh (approved):

mvr320:
Why can you not do this in the configure step?

You only need this for maintainer-install right?

nickygerritsen:
> Why can you not do this in the configure step?
> 
> You only need this for maintainer-install right?

You can switch to dev/prod mode every easily in any install by changing one line in `.env.local` (and i sometimes do this to debug stuff)

thijskh:
I'm not sure maintainer-install and running in dev mode are necessarily the same thing?

mvr320:
> 
> 
> > Why can you not do this in the configure step?
> > You only need this for maintainer-install right?
> 
> You can switch to dev/prod mode every easily in any install by changing one line in `.env.local` (and i sometimes do this to debug stuff)

Everyday I learn something new. But than I do indeed agree your solution is better/actually good

webapp/src/EventListener/AddContentSecurityPolicyListener.php
```diff
@@ -0,0 +1,38 @@
...
+        return [ResponseEvent::class => 'onKernelResponse'];
+    }
+
+    public function onKernelResponse(ResponseEvent $event)
+    {
+        // Set the correct CSP based on whether the profiler is enabled, since
+        // the profiler requires 'unsafe-eval' for script-src 'self'
+        $response = $event->getResponse();
+        $cspExtra = $this->profiler ? " 'unsafe-eval'" : "";
+        $csp = "default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'$cspExtra; img-src 'self' data:";
```
meisterT, line 35: Can we add a space between `'` and `$` for readability?

nickygerritsen: Done

PR #961, opened 2021-01-09
Clean up all entities.

meisterT:
Mostly remove useless comments, e.g. for a method called `setFoo` we
don't need a comment "Sets foo".

Set proper type hints whereever possible.


eldering (approved):
LGTM

PR #964, opened 2021-01-10
Only admin role should be able to upload executables

mvr320:
Move executable upload to /add page

Minor:
Reformatted PHPcode
Added return types
Removed redundant type annotations
Removed not used imports

PHPstorm also reports on:
https://github.com/mvr320/domjudge/blob/858b03f0427c49f9fd03f7db292a022aa27c6cc5/webapp/src/Controller/Jury/ExecutableController.php#L199

Where DOMJudge has 2 functions, which only has 1 argument and 1 which has 2, I'm not sure if its suggestion is correct.
Followed PHPstorm suggestions on imports

### DIFF
--- a/webapp/src/Controller/Jury/ExecutableController.php
+++ b/webapp/src/Controller/Jury/ExecutableController.php
@@ -13,15 +13,17 @@ use App\Service\DOMJudgeService;
 use App\Service\EventLogService;
 use App\Utils\Utils;
 use Doctrine\ORM\EntityManagerInterface;
+use Exception;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
 use Symfony\Component\Form\Exception\InvalidArgumentException;
 use Symfony\Component\Form\Extension\Core\Type\FileType;
 use Symfony\Component\Form\Extension\Core\Type\SubmitType;
 use Symfony\Component\Form\Extension\Core\Type\TextareaType;
+use Symfony\Component\Form\FormInterface;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\StreamedResponse;
-use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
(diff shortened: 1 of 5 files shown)


webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -86,65 +88,13 @@ public function __construct(
     /**
      * @Route("", name="jury_executables")
      * @param Request $request
```
nickygerritsen, line 90: We can also get rid of this

webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -206,14 +156,81 @@ public function indexAction(Request $request)
         ]);
     }
 
+    /**
+     * @Route("/add", name="jury_executable_add")
+     * @IsGranted("ROLE_ADMIN")
+     * @param Request $request
```
nickygerritsen, line 162: We don’t need the @param and @return

webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -206,14 +156,81 @@ public function indexAction(Request $request)
...
+        }
+
+        return $this->render('jury/executable_add.html.twig', [
+            'form' => $form->createView(),
+        ]);
+    }
+
     /**
      * @Route("/{execId}", name="jury_executable")
      * @param Request $request
```
nickygerritsen, line 229: Here as well. More below which you can find

webapp/src/Controller/Jury/ExecutableController.php
mvr320: Why is the param not needed?

nickygerritsen: It’s already in the type of the function arg. Only when that’s not possible we should add it

mvr320: https://softwareengineering.stackexchange.com/questions/365176/are-docblock-typehints-redundant-when-using-strict-typing

Ok... I understand, will remove them

webapp/src/Controller/Jury/ExecutableController.php
mvr320: Should now be removed together with PHPDoc parts which are just trivial

nickygerritsen (approved):

PR #965, opened 2021-01-10
Restrict post problem upload

mvr320:


webapp/src/Controller/Jury/ProblemController.php
```diff
@@ -1102,6 +1080,10 @@ public function deleteAttachmentAction(Request $request, int $attachmentId): Res
     /**
      * @Route("/{testcaseId<\d+>}/delete_testcase", name="jury_testcase_delete")
      * @IsGranted("ROLE_ADMIN")
+<<<<<<< HEAD
+=======
```
nickygerritsen, line 1084: Woops

nickygerritsen (commented):
If we keep the form at the index action why is handling it in a different action? As in what is the benefit of it?

webapp/src/Controller/Jury/ProblemController.php
mvr320: > 
> 
> If we keep the form at the index action why is handling it in a different action? As in what is the benefit of it?

If I understand Symfony correctly we will always allow al types of HTTP actions, so when you allow GET (IsGranted(Jury)) you also allow POST, by moving such actions to another route we can set an alternative IsGranted. You can also do this in code, but now all pages which accept such post handle this on an alternative route and the design can be:
Main page of an Entity (problemS, ExecutableS, is allowed for both Jury/Admin) all mutations to such entities will be done on another endpoint, so a Jury member can post to the page, but it will just not do anything.

Symfony seems to have Security as IsGranted++ which might do the same thing, but I think this is easier to understand on whats happening, and in this case the UI has the space for the upload problems form and I would expect it there, although moving it to another page is also fine by me.

I just wanted to remove the problem that when you check if something is POST, that you also need to check if someone is admin, as in this case I find the functions which do this long and hard to understand. This was the easiest thing I could come up with, but if you think this is actually making it harder to comprehend I'm fine with either moving to another page, or closing this PR (or removing the 2nd commit from it).

nickygerritsen: Hmmmm separate functions make sense... I think we have the following options:
* Do as you suggested. We do create the form in 2 places in that case, but it's not much code. However it feels like a non-good practice to split creating and handling the form?
* Move the whole form to a separate page. Not really that useful if it has like 1 file input and a button...
* Extract the form handling logic in a separate function but call it from the index action. We still need to check for admin but at least it's in a separate function?

What do you and others think?

mvr320: > Hmmmm separate functions make sense... I think we have the following options:
> 
>     * Do as you suggested. We do create the form in 2 places in that case, but it's not much code. However it feels like a non-good practice to split creating and handling the form?
We could let the creation also be done by the endpoint, so when you create it it returns the form, IMO this would solve your issue, and is also ugly IMO
> 
>     * Move the whole form to a separate page. Not really that useful if it has like 1 file input and a button...
This is consistent, the advantage of this is that its really an action which you don't overlook, I've in the past had that I forgot to change the contest when wanting to upload problems as the form somehow did not feel as 1 form but as 2 different.. I would call this a PEBKAC but it would be solved in that way. The 1 extra click is not problematic I think (as this is only before the contest)
> 
>     * Extract the form handling logic in a separate function but call it from the index action. We still need to check for admin but at least it's in a separate function?
I think this is not *clean* and creates the problem that you need to check the code if something is allowed as that role (as Jury can still post, but we just make sure that it does nothing, but does work when you have different rights... for the agent the system will react different depending on the role although the HTTP method is allowed), I'm not saying that this is worse than my solution its just different.
> 
> 
> What do you and others think?

Another alternative is to be explicit which methods are allowed and discourage the other ones, so when we only implement code for GET, throw an 405 when you POST etc.

nickygerritsen: For the last point: the `@Route` annotation should have a `methods={}` option to limit allowed HTTP methods.

Personally I'm fine with moving the form.

mvr320: @nickygerritsen what do you think of this?

I'll have to check if I can do more cleanup, but this should have most of it.

nickygerritsen: Good in general, some comments:

* The import/export has a link to 'problem archive' which can go away
* The rest of the page assumes the currently selected contest always, should we drop the contest choice? Although you might want to load problems before the contest becomes active? Not sure.

mvr320: > 
> 
> Good in general, some comments:
> 
>     * The import/export has a link to 'problem archive' which can go away
Fixed
> 
>     * The rest of the page assumes the currently selected contest always, should we drop the contest choice? Although you might want to load problems before the contest becomes active? Not sure.

I think it would be good to move the selector to the top of the page, so you can import/export for also non active contests. But I prefer to do that in another PR (as the goal here was to get the problem/executable pages in a generic format).

nickygerritsen: Separate PR for that sounds good

nickygerritsen (approved):

PR #970, opened 2021-01-18
The contestId is lost when the entity is removed,

mvr320:
The return type of getCid does not allow this

I'm not 100% if the eventlogservice would need the contestId, but I think keeping it there is better than having contests with an ID which can be null.


nickygerritsen (approved):
The `{` can go on the same line as the `if` but ✅ otherwise.
maybe add a comment why we do this?

PR #972, opened 2021-01-21
Implement CCS specs submission upload (icpc/ccs-specs#14).

nickygerritsen:
This implements the changes described in icpc/ccs-specs#14.

Since the spec is not officially merged yet I will leave this as a draft for now.

Some open questions:
* Should we internally switch to using `problem_id` and `language_id` in the submit client as well? Since we are doing a new major release this could be a breaking change. Other option would be to still allow both options for now but switch in the submit client, then later drop the old option. In that case I propose to not *document* the old fields. Leaving the `code` argument makes sense, since I think using the base64-encoded ZIP format the CCS specs use internally is not a good idea.
* For now I made a distinction between form and JSON body types. The primary reason is that in a form in SwaggerUI you can not have nested objects, which means you can not do the `files[0].data` thingy there. Browsers and HTTPie do support it, so maybe we should merge the two? If we still have both the fields with and without `_id` this does make the spec a little messy, since we can not set `required=` to anything in that case.
* There is no check whether files are in the root of the ZIP and filenames are valid but the submission service checks this already, so I don't think we need anything for this.
* It has no tests (yet). Probably best to do this together with the tests for the old/`code` behavior.

### DIFF
--- a/webapp/src/Controller/API/SubmissionController.php
+++ b/webapp/src/Controller/API/SubmissionController.php
@@ -118,7 +118,7 @@ class SubmissionController extends AbstractRestController
      * Add a submission to this contest
      * @Rest\Post("")
      * @OA\Post()
-     * @Security("is_granted('ROLE_TEAM') or is_granted('ROLE_API_WRITER')", message="You need to have the Team Member role or be an admin to add a submission")
+     * @Security("is_granted('ROLE_TEAM') or is_granted('ROLE_API_WRITER')", message="You need to have the Team Member role to add a submission")
      * @OA\RequestBody(
      *     required=true,
      *     @OA\MediaType(
@@ -357,15 +357,15 @@ class SubmissionController extends AbstractRestController
             // CCS spec format, files are a ZIP, get them and transform them into a file object
             $filesList = $request->request->get('files');
             if (!is_array($filesList) || count($filesList) !== 1 || !isset($filesList[0]['data'])) {
-                throw new BadRequestHttpException("The 'files' attribute should be an array with a single item, containing an object with a base64 encoded data field");
+                throw new BadRequestHttpException("The 'files' attribute must be an array with a single item, containing an object with a base64 encoded data field");
             }
 
             if (isset($filesList[0]['mime']) && $filesList[0]['mime'] !== 'application/zip') {
-                throw new BadRequestHttpException("The 'files[0].mime' attribute should be application/zip if provided");
+                throw new BadRequestHttpException("The 'files[0].mime' attribute must be application/zip if provided");
             }
 
             $data        = $filesList[0]['data'];
-            $decodedData = base64_decode($data);
+            $decodedData = base64_decode($data, true);
             if ($decodedData === false) {
                 throw new BadRequestHttpException("The 'files[0].data' attribute is not base64 encoded");
             }
--- a/webapp/src/Controller/API/SubmissionController.php
+++ b/webapp/src/Controller/API/SubmissionController.php
@@ -2,28 +2,34 @@
 
 namespace App\Controller\API;
 
+use App\Entity\Contest;
 use App\Entity\ContestProblem;
 use App\Entity\Language;
 use App\Entity\Problem;
 use App\Entity\Submission;
 use App\Entity\SubmissionFile;
+use App\Entity\Team;
 use App\Service\ConfigurationService;
 use App\Service\DOMJudgeService;
 use App\Service\EventLogService;
 use App\Service\SubmissionService;
+use App\Utils\Utils;
 use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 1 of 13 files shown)


nickygerritsen:
Note that I did do tests locally obviously and they seemed to work.

nickygerritsen:
Updated this PR to reflect the updates to the spec PR: you can now specify a submission ID (as admin) while uploading, which is needed when this is used by a proxy that can handle the CCS being offline.

This required some custom cases related to where to get a submission ID from. The new assumption is that the submission ID is `externalid`, unless null. In that case it is `submitid`.
We could solve this in a cleaner way (codewise) by adding an extra field in the DB that contains that value (i.e. `submission.realid` or similar), then we can get rid of most special cases (but still need some). I'm not a fan of that and think these small special cases are not too bad.

Also did 2 small things:
* We should the external ID of the submission on the submission lists in the jury interface. We already showed it on the details page but now it is (better) visible in the overview.
* I made `externalid` in the API not be emitted in strict mode and renamed it to `external_id` to reflect other entities.

nickygerritsen:
The spec is now final, so this is now reviewable.

webapp/src/Controller/API/SubmissionController.php
```diff
@@ -219,23 +318,120 @@ public function addSubmissionAction(Request $request)
...
+
+        if ($request->request->has('files')) {
+            // CCS spec format, files are a ZIP, get them and transform them into a file object
+            $filesList = $request->request->get('files');
+            if (!is_array($filesList) || count($filesList) !== 1 || !isset($filesList[0]['data'])) {
+                throw new BadRequestHttpException("The 'files' attribute should be an array with a single item, containing an object with a base64 encoded data field");
+            }
+
+            if (isset($filesList[0]['mime']) && $filesList[0]['mime'] !== 'application/zip') {
+                throw new BadRequestHttpException("The 'files[0].mime' attribute should be application/zip if provided");
```
thijskh, line 364: should be or must be?

webapp/src/Controller/API/SubmissionController.php
```diff
@@ -219,23 +318,120 @@ public function addSubmissionAction(Request $request)
...
+            }
+        }
+
+        $tempFiles = [];
+
+        if ($request->request->has('files')) {
+            // CCS spec format, files are a ZIP, get them and transform them into a file object
+            $filesList = $request->request->get('files');
+            if (!is_array($filesList) || count($filesList) !== 1 || !isset($filesList[0]['data'])) {
+                throw new BadRequestHttpException("The 'files' attribute should be an array with a single item, containing an object with a base64 encoded data field");
```
thijskh, line 360: should be or must be?

webapp/src/Controller/API/SubmissionController.php
```diff
@@ -219,23 +318,120 @@ public function addSubmissionAction(Request $request)
...
+            }
+
+            if (isset($filesList[0]['mime']) && $filesList[0]['mime'] !== 'application/zip') {
+                throw new BadRequestHttpException("The 'files[0].mime' attribute should be application/zip if provided");
+            }
+
+            $data        = $filesList[0]['data'];
+            $decodedData = base64_decode($data);
```
thijskh, line 368: Do we want to set the strict flag? Seems good to be strict I think.

webapp/src/Controller/API/SubmissionController.php
```diff
@@ -219,23 +318,120 @@ public function addSubmissionAction(Request $request)
...
+                $files = [$files];
+            }
         }
 
         // Now submit the solution
-        $team       = $this->dj->getUser()->getTeam();
         $submission = $this->submissionService->submitSolution(
             $team, $problem, $problem->getContest(), $language,
-            $files, null, null, $entryPoint, null, null, $message
+            $files, null, null, $entryPoint, $submissionId, $time, $message
```
thijskh, line 422: So the supplied ID we store as an external ID. Is this compliant with the spec? Especially the over-designed "proxy accepts submissions for later forwarding" use case?

thijskh (changes requested):
I think I really expected to find some kind of test to be added. With an actual json being posted.

nickygerritsen:
> I think I really expected to find some kind of test to be added. With an actual json being posted.

I can try to add that.

webapp/src/Controller/API/SubmissionController.php
nickygerritsen: Yep sounds good.

webapp/src/Controller/API/SubmissionController.php
nickygerritsen: That is exactly the reason we need this. What we do is when this is merged, we return the external ID if it is set, which the proxy can send to us.

thijskh (approved):

PR #973, opened 2021-01-21
Add POST /api/users to add a new user using the API. Fixes #877.

### LINKED ISSUE #877: "Create individual user" Route for internal API
Our instance will not have team based competitions but will be individual members.  

I want to be able to add a user programmatically through the internal API that mimics the "Add Users" section of the UI.

API would allow creation of new user with username/password and optionally assign to a team, if no team matches the requested team than auto-create the team

### Additional context

Currently it looks like the only way to accomplish this is to script together the user/teams API and the user/accounts API to create a team with the user/teams API, then add a user with the user/accounts API.

nickygerritsen:


webapp/src/Controller/API/UserController.php
```diff
@@ -292,6 +294,90 @@ public function singleAction(Request $request, string $id)
         return parent::performSingleAction($request, $id);
     }
 
+    /**
+     * Add a new user
+     *
+     * @Rest\Post()
+     * @Security("is_granted('ROLE_ADMIN') or is_granted('ROLE_API_WRITER')")
```
mvr320, line 301: Is admin not also a API WRITER?

nickygerritsen: HEY you are right. So we need only api writer here.

tuupke (approved):

thijskh (approved):

PR #977, opened 2021-01-23
Php inspections

mvr320:
Most of the trivial errors PhpStorm reports. Will squash everything if this is accepted as most should be trivial changes.


nickygerritsen (approved):
Didn’t look in detail but in general this is good, so go ahead

PR #979, opened 2021-01-25
Language fix

mvr320:
@nickygerritsen @meisterT this fixes the problem that http://localhost:12345/jury/languages/add works again, but I'm not sure if these variables should indeed be empty.

As the problem is already there on the Add page, where Language is an empty object the problem will not be here, but I'm not sure if somewhere else there is an assumption that a Language always has correct values.


nickygerritsen (approved):
This is indeed the intended fix

PR #980, opened 2021-01-25
Contest page does not break on Language without compile script.

mvr320:
Currently its possible to upload a language without a compile script. On
itself this might be a problem, but as the config page is the place to
report these problems it should not assume a Language to be correct.

### DIFF
--- a/webapp/src/Service/CheckConfigService.php
+++ b/webapp/src/Service/CheckConfigService.php
@@ -493,7 +493,11 @@ class CheckConfigService
             $languageerrors[$langid] = $errors;
 
             $morelanguageerrors[$langid] = '';
-            if ($compile = $language->getCompileExecutable()->getExecid()) {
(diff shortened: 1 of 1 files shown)


mvr320:
Another question is ofcourse if it should ever be possible to have a language without a compile script.

webapp/src/Service/CheckConfigService.php
```diff
@@ -485,6 +485,7 @@ public function checkLanguagesValidate()
         $languageerrors = $scripterrors = [];
         $result = 'O';
         foreach ($languages as $language) {
+            dump($language);
```
nickygerritsen, line 488: Whoops

nickygerritsen (approved):

webapp/src/Service/CheckConfigService.php
mvr320: Fixed

nickygerritsen (approved):

PR #981, opened 2021-01-27
[Test] Give PSR-12 suggestions

mvr320:
To make reviewing PRs slightly easier I tried to implement a test with php-cs-fixer (suggestion by @nickygerritsen) this should pick the PSR12 guidelines by default (and otherwise I can tweak it as far as I understand). As this has a lot of suggestions, should this be further tweaked to skip certain things?

Otherwise I propose to make 1 PR with only fixes.

### DIFF
--- a/.github/workflows/phpcodesniffer.yml
+++ b/.github/workflows/phpcodesniffer.yml
@@ -1,0 +1,22 @@
+name: PHP CodeSniffer
+
+# Controls when the action will run. 
+on: ['pull_request']
+jobs:
+  phpcs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # important!
+      - name: Install PHP_CodeSniffer
+        run: |
+          curl -OL https://raw.githubusercontent.com/DOMjudge/action-php-codesniffer/master/staticphar/phpcs.phar
+          php phpcs.phar --version
+      - uses: domjudge/action-php-codesniffer@v1
+        with:
+          files: "**.php" # you may customize glob as needed
+          phpcs_path: php phpcs.phar
+          standard: .gitlab/workflows/phpruleset.xml
+          fail_on_warnings: false
+          fail_on_errors: false
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -232,3 +232,18 @@ php linter:
       webapp/public
       webapp/config
 
+php-cs-fixer:
+  <<: *tiny_job
+  image:
+    name: cytopia/php-cs-fixer
+    entrypoint: [""]
+  artifacts:
+    when: always
+    reports:
+      codequality: code-quality.json
+    paths:
+      - code_quality_debug.json
+  script:
+    - /usr/bin/php-cs-fixer fix --format=gitlab --dry-run --diff . > code-quality.json || true
+    - cp code-quality.json code_quality_debug.json
+
--- a/.github/workflows/phpruleset.xml
+++ b/.github/workflows/phpruleset.xml
@@ -1,0 +1,346 @@
+<?xml version="1.0"?>
+<ruleset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="PSR12" xsi:noNamespaceSchemaLocation="../../../phpcs.xsd">
+    <description>The PSR-12 coding standard.</description>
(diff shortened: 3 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -241,3 +241,15 @@ integration_mariadb:
     - name: mariadb
       alias: sqlserver
 
+php-cs-fixer:
+  image: 
+    name: cytopia/php-cs-fixer
+    entrypoint: [""]
+  script:
+    - /usr/bin/php-cs-fixer fix --dry-run --diff .
```
nickygerritsen, line 249: If we run this with `--format=gitlab` and write it to a JSON + add an artifact like this:
```yaml
    artifacts:
        when: on_failure
        reports:
            codequality: code-quality.json
```

I think it at least shows up in a tab in the GitLab pipeline.

nickygerritsen (commented):
I like the idea, but we should indeed check if we agree or want to tweak.

.gitlab-ci.yml
mvr320: I wonder if there is an github actions which can also do this, so the extra config to get it to GitHub is not needed. (I searched but didn't find the same feature yet)

mvr320: @nickygerritsen this is now implemented and can be found at: https://gitlab.com/DOMjudge/domjudge/-/pipelines/248383076/codequality_report

I think keeping this there and just always accepting the test would be fine? So have the report but dont fail and work more with the github actions which are now introduced.

.gitlab-ci.yml
```diff
@@ -241,3 +241,39 @@ integration_mariadb:
...
+  script:
+    - parallel-lint --colors .
+
+phpcs PSR12:
+  stage: test
+  image: pipelinecomponents/php-codesniffer:latest
+  script:
+    - phpcs -s -p --colors --extensions=php --standard=PSR12 .
+
+php-compatibility 7.3:
```
nickygerritsen, line 267: Shouldn't this be 7.2? I'm also fine just calling it `php-compatibility`

nickygerritsen (approved):

.gitlab-ci.yml
mvr320: The copied stuff had 7.3, will indeed change it.

.github/workflows/phpcodesniffer.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # important!
+      - name: Install PHP_CodeSniffer
+        run: |
+          curl -OL https://squizlabs.github.io/PHP_CodeSniffer/phpcs.phar
+          php phpcs.phar --version
+
+      - uses: mvr320/action-php-codesniffer@v1
```
eldering, line 17: I'm not sure what this refers to, but if it's an external repo/site, can we also move it inside this repo to make things self-contained and also clear what we're using?

eldering (commented):
I'm assuming this will only suggest issues with PSR-12, not actually force fixing them, right? That's fine with me.
One comment though on the external dependency.

.github/workflows/phpcodesniffer.yml
mvr320: I assume you refer to the mvr320/ part? its a github action which I had to fix. I dont have the rights to move it to the DOMjudge organisation but if someone does this we can directly archive it also (as I see no reason in maintaining it as DOMjudge)

if its about the downloading of the phpcs.phar, you would lose using the latest version but that could also be moved to the action or directly commited to a directory in the domjudge repo (to pin to a specific version of the sniffer)

I'm fine with all options, just not 100% sure what you want me to change.

mvr320:
> 
> 
> I'm assuming this will only suggest issues with PSR-12, not actually force fixing them, right? That's fine with me.
> One comment though on the external dependency.

image

Yes it will only comment on the `Files changed` tab and upload a report to gitlab but none of these tests should be failing even if new violations are introduced. So the idea is to only have gentle nudges to follow the PSR-12 standard.

eldering (approved):
LGTM

thijskh:
Why is the build failing?

thijskh (commented):
I am OK with this change but my experience is that it can also lead to a lot of (in my opinion) needless/nitpicky churn. So we should guard whether the signalled issues are actually useful/important enough. We can tweak that as we go.

PR #997, opened 2021-03-21
When giving back a judging, only give back runs from the current judg…

nickygerritsen:
…ehost that are not finished yet.

### DIFF
--- a/webapp/src/Controller/API/JudgehostController.php
+++ b/webapp/src/Controller/API/JudgehostController.php
@@ -15,9 +15,7 @@ use App\Entity\JudgingRunOutput;
 use App\Entity\Rejudging;
 use App\Entity\Submission;
 use App\Entity\SubmissionFile;
-use App\Entity\Testcase;
 use App\Entity\TestcaseContent;
-use App\Entity\User;
 use App\Service\BalloonService;
 use App\Service\ConfigurationService;
 use App\Service\DOMJudgeService;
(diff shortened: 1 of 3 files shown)


webapp/src/Controller/API/JudgehostController.php
```diff
@@ -812,37 +810,47 @@ public function internalErrorAction(Request $request)
 
         if (in_array($disabled['kind'], ['problem', 'language', 'judgehost']) && $judgingId) {
             // Give back judging if we have to.
-            $this->giveBackJudging((int)$judgingId);
+            $hostname = $request->request->get('hostname');
+            /** @var Judgehost $judgehost */
+            $judgehost = $this->em->createQueryBuilder()
```
meisterT, line 815: why not use `getRepository->find` here?

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -812,37 +810,47 @@ public function internalErrorAction(Request $request)
...
+                    // For the other runs, we need to reset the judge task if it belongs to the current judgehost
+                    if ($run->getJudgetask()->getHostname() === $judgehost->getHostname()) {
+                        $run->getJudgetask()->setHostname(null);
+                    }
+                }
+
+                $this->em->flush();
             });
 
             $this->dj->auditlog('judging', $judgingId, 'given back', null,
```
meisterT, line 856: should we also update this?

webapp/src/Controller/API/JudgehostController.php
nickygerritsen: Copy paste from somewhere else, that sounds better

webapp/src/Controller/API/JudgehostController.php
nickygerritsen: What do you propose?

meisterT: Perhaps mention the judge host in the message?

meisterT (approved):

Issue #936, opened 2020-12-07
Error: Maximum execution time of 30 seconds exceeded for generate password


zargarzadehm:
### Description of the problem
 We had 90 teams in a contest and every team had a user, to generate a password for these teams I got this error:
```
Error: Maximum execution time of 30 seconds exceeded
```

### Your environment
 - DOMjudge server version 7.3.1
 - docker
 - Webserver nginx

### Steps to reproduce
  - create a relatively large number of teams with users for those teams
  - generate a password for them

### Expected behaviour
  *  Create passwords for all teams

### Actual behaviour
  *  500 Internal Server Error

### Any other information that you want to share?
![2020-12-07 00 07 36](https://user-images.githubusercontent.com/35424774/101398635-a3192d80-38e3-11eb-9809-9460eb35f9db.jpg)

  * i resolved this issue with add line `set_time_limit(300);` to function [generatePasswordsAction](https://github.com/DOMjudge/domjudge/blob/75b2dded8f3b12bf0bddbfa3d2778e436867b69a/webapp/src/Controller/Jury/UserController.php#L320) in  `webapp/src/Controller/Jury/UserController.php`

  * Another solution is set to limit teams that can generate password or option for select teams

  * another solution that @thijskh suggested in the slack channel was to lower the hash cost.

nickygerritsen:
I wonder what the clean solution(s) would be:
* We set `set_time_limit(-1)` on this page and add a warning on the page: `Please be aware that this will take some time`.
* We move the code to a command.
* We do both, by moving the code to a service that both can use.

mvr320:
Why is it a bad idea to do this in batches? So reset only **n** accounts per pass? As I think you probably also want to paperprint these batches and its very hard to find out if something takes "some" time or is just stuck.

nickygerritsen:
We could do something similar to the scoreboard cache refresh page, which outputs something after each processed item.
Doing it in batches would mean someone needs to press the button a couple of times, that doesn't sound very user friendly.

thijskh:
Maybe we should explore lowering the hash cost. Generating 90 passwords does not seem like an extreme usecase.

nickygerritsen:
I agree that 90 is not that much. Looking at some ICPC contests we have around ~128 teams and thus users.
We currently have a cost of 10. Lowering it basically means a higher attack chance right? It seems that lowering the cost by `1` halves the time to generate a password, but also the time to verify one and thus to bruteforce.

[...]

PR #1001, opened 2021-03-23
hash cost for users decreased

mvr320:
Symfony does advice 'Sodium' which could be used as alternative, or we
could set all values lower and advice a reasonable amount of hashes per
time.

Also added a config check to verify if hashing time is reasonable.

Locally 7 is to simple, but I think thats something someone should propose ranges for..

This is the start for https://github.com/DOMjudge/domjudge/issues/936

### DIFF
--- a/webapp/src/Service/CheckConfigService.php
+++ b/webapp/src/Service/CheckConfigService.php
@@ -17,6 +17,7 @@ use Doctrine\Common\Inflector\Inflector;
 use Doctrine\Inflector\InflectorFactory;
 use Doctrine\ORM\EntityManagerInterface;
 use Symfony\Component\Routing\RouterInterface;
+use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 
 /**
@@ -60,6 +61,11 @@ class CheckConfigService
      */
     protected $debug;
 
+    /**
+     * @var UserPasswordEncoderInterface
+     */
+    protected $passwordEncoder;
+
     public function __construct(
         bool $debug,
         EntityManagerInterface $em,
@@ -67,7 +73,8 @@ class CheckConfigService
         DOMJudgeService $dj,
         EventLogService $eventLogService,
         RouterInterface $router,
-        ValidatorInterface $validator
+        ValidatorInterface $validator,
+        UserPasswordEncoderInterface $passwordEncoder
     ) {
         $this->debug           = $debug;
         $this->em              = $em;
@@ -76,9 +83,10 @@ class CheckConfigService
         $this->eventLogService = $eventLogService;
         $this->router          = $router;
         $this->validator       = $validator;
+        $this->passwordEncoder = $passwordEncoder;
     }
 
-    public function runAll()
+    public function runAll() : array
     {
         $results = [];
 
@@ -97,6 +105,7 @@ class CheckConfigService
             'filesizememlimit' => $this->checkScriptFilesizevsMemoryLimit(),
             'debugdisabled' => $this->checkDebugDisabled(),
             'tmpdirwritable' => $this->checkTmpdirWritable(),
+            'hashtime' => $this->checkHashTime(),
         ];
 
         $results['Configuration'] = $config;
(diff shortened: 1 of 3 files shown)


meisterT:
If you generate passwords for 200 users, how long does it take before/after your change?

webapp/src/Controller/Jury/ConfigController.php
```diff
@@ -16,6 +16,7 @@
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
+use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
```
nickygerritsen, line 19: Seems like this is not needed (anymore)?

webapp/src/Service/CheckConfigService.php
```diff
@@ -337,6 +346,36 @@ public function checkTmpdirWritable()
...
 
+    public function checkHashTime()
+    {
+        $tmpuser = new User();
+        $counter = 0;
+        $time_start = microtime(true);
+        do {
+            $plainPassword = 'ryanpass';
+            $encoded = $this->passwordEncoder->encodePassword($tmpuser, $plainPassword);
+            // $user->setPassword($encoded);
```
nickygerritsen, line 357: This comment can be removed

webapp/src/Service/CheckConfigService.php
```diff
@@ -337,6 +346,36 @@ public function checkTmpdirWritable()
...
+            $time_end = microtime(true);
+            $counter++;
+        } while ( ($time_end - $time_start) < 4);
```
nickygerritsen, line 360: This 4 is the number of seconds right? Maybe we should lower it a bit if that's the case? I guess that we can also determine a sane value in 2 seconds.

mvr320:
> 
> 
> If you generate passwords for 200 users, how long does it take before/after your change?

I can do 429 user hashes in 4s if the logic of the config page is indeed correct.

meisterT:
And with the old settings?

webapp/src/Service/CheckConfigService.php
```diff
@@ -337,6 +346,36 @@ public function checkTmpdirWritable()
...
+
+        $length_suggestion = 'When you have a contest duration more than a week increase the password length.';
```
nickygerritsen, line 362: This is not related to the hashing algo right? But more a general thing.

webapp/src/Service/CheckConfigService.php
```diff
@@ -337,6 +346,36 @@ public function checkTmpdirWritable()
...
+        if ($counter>300) {
+            return ['caption' => 'User password hashing',
+                'result' => 'W',
+                'desc' => sprintf('Hashing is to simple for small sized contests (Did %d hashes). %s',
```
nickygerritsen, line 366: `s/to/too`
Also maybe we need to add somewhere in the docs how to increase / lower the hashing cost?

webapp/src/Service/CheckConfigService.php
```diff
@@ -337,6 +346,36 @@ public function checkTmpdirWritable()
...
+                                  $counter, $length_suggestion)];
+        } elseif ($counter<100) {
+            return ['caption' => 'User password hashing',
+                'result' => 'W',
+                'desc' => sprintf('Hashing is to expensive for medium sized contests  (Did %d hashes). %s',
```
nickygerritsen, line 371: s/to/too`

nickygerritsen (changes requested):

webapp/src/Service/CheckConfigService.php
mvr320: Yes and no, if you want more secure passwords the first thing which should be increased is the length (citing thijs) and not the hashing cost. I think i'll remove it and make something for that specific usecase?

webapp/src/Service/CheckConfigService.php
mvr320: I now delay the page for 4s, will change it to 2s

webapp/src/Service/CheckConfigService.php
mvr320: Thats something I wanted to ask you, I assume the encodePassword will already do the cost (so its not a promise to the DB?)

webapp/src/Service/CheckConfigService.php
nickygerritsen: True. We can't really check the length of passwords, so maybe just something we should put somewhere indeed as a note.

webapp/src/Service/CheckConfigService.php
nickygerritsen: Sounds good.

webapp/src/Service/CheckConfigService.php
nickygerritsen: Correct.

webapp/src/Controller/Jury/ConfigController.php
mvr320: When I remove it things break, its used at L77

nickygerritsen: Line 77 is a comment. You are looking at the wrong file I think

mvr320: You think correctly.

webapp/src/Service/CheckConfigService.php
mvr320: Done

webapp/src/Service/CheckConfigService.php
```diff
@@ -60,14 +61,20 @@ class CheckConfigService
      */
     protected $debug;
 
+    /**
+     * @var $passwordEncoder
```
nickygerritsen, line 65: this @var is quite useless.

webapp/src/Service/CheckConfigService.php
```diff
@@ -60,14 +61,20 @@ class CheckConfigService
...
+     */
+    protected $passwordEncoder;
```
nickygerritsen, line 67:
```suggestion
    protected UserPasswordEncoderInterface $passwordEncoder;
```

webapp/src/Service/CheckConfigService.php
mvr320: All of these "@Var" should be removed right? (in that section)

nickygerritsen (approved):

PR #1002, opened 2021-03-23
Advice on hashing cost verses lifetime of user

mvr320:
This is only a suggestion to the user as we do not know the password length.

### DIFF
--- a/webapp/src/Service/CheckConfigService.php
+++ b/webapp/src/Service/CheckConfigService.php
@@ -14,6 +14,7 @@ use App\Entity\Testcase;
 use App\Entity\User;
 use App\Utils\Utils;
 use BadMethodCallException;
+use DateInterval;
 use Doctrine\Inflector\InflectorFactory;
 use Doctrine\ORM\EntityManagerInterface;
 use Symfony\Component\Routing\RouterInterface;
@@ -66,6 +67,11 @@ class CheckConfigService
      */
     protected $passwordEncoder;
 
+    /**
+     * @var int
+     */
+    protected $hashCost;
+
     public function __construct(
         bool $debug,
         EntityManagerInterface $em,
@@ -74,7 +80,8 @@ class CheckConfigService
         EventLogService $eventLogService,
         RouterInterface $router,
         ValidatorInterface $validator,
-        UserPasswordEncoderInterface $passwordEncoder
+        UserPasswordEncoderInterface $passwordEncoder,
+        int $hashCost
     ) {
         $this->debug           = $debug;
         $this->em              = $em;
@@ -84,6 +91,7 @@ class CheckConfigService
         $this->router          = $router;
         $this->validator       = $validator;
         $this->passwordEncoder = $passwordEncoder;
+        $this->hashCost        = $hashCost;
     }
 
     public function runAll() : array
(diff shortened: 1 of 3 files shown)


webapp/src/Service/CheckConfigService.php
```diff
@@ -391,6 +392,33 @@ public function checkContestsValidate()
...
+        // Fetch all active and future contests
+        $contests = $this->dj->getCurrentContests(null, true);
+
+        $contestWarning = [];
+        $result = 'O';
+        foreach ($contests as $contest) {
+            $cid = $contest->getCid();
+            $duration = $contest->getEndtime()-$contest->getStarttime();
+            $multiplier = 60*60*24*7;
+            if ($duration/$multiplier>1) {
```
nickygerritsen, line 406: Maybe `$this->getStartTimeObject()->add(new \DateInterval('P1W')) < $this->getEndTimeObject()` is more readable?

mvr320: I had to google this... but yes its more readable.

mvr320:
> 
> 
> Passwords are for users, not contests, so the security depends on how long a user account is active, not how long the contest is active?

Valid point, but I assume that a user is valid from start to end? I would say that its still up to the admin to verify if the warning is applicable for that contest. Are you against this merge or should it be implemented differently?

thijskh:
I'm not enthousiastic because I think it conflates two things: lifetime of a password and contest length, which are only loosly coupled. I think this just confuses and muddies the concepts for users.

What about: read out the cost factor, show a statically configured advice for that factor?
If <= 7 then : Your cost factor is 7, this is appropriate when your user's passwords are in use for a week or less. If they live longer, please raise the cost factor.

mvr320:
> 
> 
> I'm not enthousiastic because I think it conflates two things: lifetime of a password and contest length, which are only loosly coupled. I think this just confuses and muddies the concepts for users.
> 
> What about: read out the cost factor, show a statically configured advice for that factor?
> If <= 7 then : Your cost factor is 7, this is appropriate when your user's passwords are in use for a week or less. If they live longer, please raise the cost factor.

Ok, makes sense and is doable, I will also advice on the password length but I think your suggestion is indeed better.

webapp/src/Service/CheckConfigService.php
```diff
@@ -385,6 +394,21 @@ public function checkHashTime() : array
             'desc' => sprintf('Hashing cost is reasonable (Did %d hashes).', $counter)];
     }
 
+    public function checkHashStrength() : array
+    {
+        if ($this->cryptCost>7) {
```
nickygerritsen, line 399: Add spaces around `>`

nickygerritsen (approved):

thijskh:
As it appears the current advice is not complete. As the chosen hash cost may also significantly affect API performance and thus judging througput. I think we need to be explicit about that. It's not only the length of the user account lifetime but also how good you want the judging to perform.

mvr320:
> 
> 
> As it appears the current advice is not complete. As the chosen hash cost may also significantly affect API performance and thus judging througput. I think we need to be explicit about that. It's not only the length of the user account lifetime but also how good you want the judging to perform.

I think we could do something like:
https://symfony.com/doc/current/security/named_encoders.html
assuming that someone makes the password longer, the hash can be less costly for the judgedaemon.

So I would propose to have the password for restapi a bit longer, and the hash maybe less costly?

thijskh:
I've continued the discussion in #936 since that seems more appropriate than a closed PR

PR #1007, opened 2021-03-24
[BugFix] HTTP500 fixed

mvr320:


eldering:
For a next time, it would be good to add a somewhat more explicit commit messsage:
Was this fixing how we return a 500 error, or just a specific 500 error caused by a bug in the code? If the latter, what was the bug?

PR #1008, opened 2021-03-24
Properly handle internal errors / giving back judgings again.

meisterT:


webapp/src/Controller/API/JudgehostController.php
```diff
@@ -828,32 +828,56 @@ public function internalErrorAction(Request $request): ?int
...
-    protected function giveBackJudging(int $judgingId, Judgehost $judgehost)
+    protected function giveBackJudging(int $judgingId, ?Judgehost $judgehost)
     {
         /** @var Judging $judging */
         $judging = $this->em->getRepository(Judging::class)->find($judgingId);
         if ($judging) {
             $this->em->transactional(function () use ($judging, $judgehost) {
                 /** @var JudgingRun $run */
                 foreach ($judging->getRuns() as $run) {
+                    if ($judgehost === null) {
```
nickygerritsen, line 841: I don’t follow how this is null for an internal error? Don’t we set it on line 822?

meisterT: Only if we disable the judgehost, e.g. because disk was full.

meisterT: Fixed now.

nickygerritsen (approved):

PR #1009, opened 2021-03-24
Add DOMjudge team to admin user if in dev mode.

meisterT:


mvr320:
Currently checking it out in the maintainer container

I think the codecoverage should also be changed (I'll do that) and looking at this PR the project should allow a -1%? and target should be 0%?

mvr320:
@meisterT so if you would have a non maintainer install, and turn Symfony in debugging + reinstall the DB this would also apply right? But not when you turn Symfony in debug but dont reinstall the DB?

webapp/src/DataFixtures/DefaultData/UserFixture.php
```diff
@@ -52,13 +53,19 @@ public function load(ObjectManager $manager)
...
-            $adminuser
+            $adminUser = new User();
+            $adminUser
                 ->setUsername('admin')
                 ->setName('Administrator')
-                ->setPassword($this->passwordEncoder->encodePassword($adminuser, trim($adminpasswordContents)))
+                ->setPassword($this->passwordEncoder->encodePassword($adminUser, trim($adminpasswordContents)))
                 ->addUserRole($this->getReference(RoleFixture::ADMIN_REFERENCE));
-            $manager->persist($adminuser);
+            if (isset($_ENV['APP_ENV']) && $_ENV['APP_ENV'] == 'dev') {
```
mvr320, line 62: So when a "normal" install is put in debug mode, and the DB is reset this would apply but not when the DB is not recreated right?

mvr320 (approved):

meisterT:
Yes, my reasoning was that you would rarely put your normal installation in debug mode and then reset the DB, but there's also no real harm if that happens.

nickygerritsen (approved):
This might break my test that assumes an admin has no team. If so I’ll fix it

meisterT:
No, CI is green - we do a regular install there.

nickygerritsen:
> No, CI is green - we do a regular install there.

OK cool. Modified my tests anyway since locally I have a maintainer one

PR #1010, opened 2021-03-24
Make it more obvious how to use DEBUG.

meisterT:
I always do it wrong when I want to enable this, so let's be a bit more
explicit about it.


nickygerritsen (approved):

etc/common-config.php
```diff
@@ -7,13 +7,19 @@
 // Log to syslog facility; do not define to disable.
 define('SYSLOG', LOG_LOCAL0);
 
-// Set DEBUG as a bitmask of the following settings.
-// Of course never to be used on live systems!
 
-define('DEBUG_PHP_NOTICE', 1); // Display PHP notice level warnings
-define('DEBUG_JUDGE', 8); // Display judging scripts debug info
+// Display PHP notice level warnings.
+define('DEBUG_PHP_NOTICE', (1 << 0));
```
thijskh, line 12: So this is not actually config. So leave it out of the file entirely?

meisterT: I don't understand, can you elaborate what your suggestion is? Removing that line doesn't help with my goal to make it easier to understand

thijskh: Well, I think it does because it is not something you can configure, while the rather similarly looking define('DEBUG') a few lines is something you _can_ configure. So at least to me it would be clearer if we'd only put things in common-config.php that are actually config items you can change.

Now this file contains defines of constants, defines that are actually config options, and code that interprets these options, all in one file of 24 lines.

Looking a bit further how to improve understandability, I suggest:
* Remove the DEBUG_PHP_NOTICE constant altogether. If DEBUG is enabled you always get notices. Notices should be few and only occur when there's actually a problem with the code.
* Move the setting of error_reporting (both the basic setting and the DEBUG variant) out of config files and into code files. I think that webapp/ does not need this: if running in DEBUG mode, you always get all error reporting. So the code can just be moved to the start of judgedaemon.
* The setting DEBUG_JUDGE seems overloaded as it both influences the judgehost and some webapp API calls at the same time. I think it's clearer if we move the ability to set debugging for the judgehost either in etc/judgehost-config.php or as a commandline parameter to the daemon. That profiling is disabled for some API calls is in my view only tangentially related to debugging the judgehost (more to debugging the webapp). So I'd create a setting for that that is in domserver-config.php.

eldering: I guess it makes sense to split this `DEBUG` bitmask constant up into separate parts: this was useful before we moved to symfony and we had more shared code between backend and frontend. Since Symfony essentially enforces its own way of having these settings, it doesn't really make sense anymore to have these here globally.

meisterT: Thanks, that makes sense - I'll rework my PR and split this up.

thijskh: I've taken the liberty to add a suggested solution to your PR.

I think this is just something that we did not yet clean up properly when removing old code before. I think the only 'contestible' choice is that I've removed the option to reenable the profiler for the judgehost user via config or a parameter. That seems to me a bit overengineering for the rare cases that you need that. In those cases, it's easy to just change the file or the permissions of your judgehost user. KISS.

webapp/src/EventListener/ProfilerDisableListener.php
```diff
@@ -45,20 +45,16 @@ public function __construct(KernelInterface $kernel, DOMJudgeService $dj, ?Profi
...
-    public static function getSubscribedEvents()
+    public static function getSubscribedEvents(): array
     {
         return [RequestEvent::class => 'onKernelRequest',];
     }
 
-    public function onKernelRequest()
+    public function onKernelRequest(): void
     {
         // Disable the profiler for users with the judgehost permission but not the admin one,
```
nickygerritsen, line 55: Very nit:
```suggestion
        // Disable the profiler for users with the judgehost permission but not the admin one
```

etc/common-config.php
nickygerritsen: I'm personally fine with this suggestion.

eldering (approved):

etc/common-config.php
meisterT: How do you feel about adding info about how to profile judgedaemon calls to the maintainer docs?

doc/manual/develop.rst
```diff
@@ -94,6 +94,24 @@ run the ``*clean`` targets: otherwise e.g. ``paths.mk`` will
 be deleted before subdirectory ``*clean`` targets are called that
 depend on information in it.
 
+Debugging
+---------
+To enable debugging in the web interface, make sure you have set
```
nickygerritsen, line 99: Line 67 also contains this information, although it doesn't hurt to also put it here.

thijskh: Thanks, I've moved the two paragraphs there to sections where I think they are more appropriate.

PR #1012, opened 2021-03-25
Idea controller fixes

mvr320:

### DIFF
--- a/webapp/src/Controller/Jury/ClarificationController.php
+++ b/webapp/src/Controller/Jury/ClarificationController.php
@@ -375,7 +375,7 @@ class ClarificationController extends AbstractController
         }
 
         $subject = $request->request->get('subject');
-        list($cid, $probid) = explode('-', $subject);
+        [$cid, $probid] = explode('-', $subject);
 
         $contest = $this->em->getReference(Contest::class, $cid);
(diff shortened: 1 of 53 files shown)


webapp/src/Controller/Jury/ClarificationController.php
```diff
@@ -260,12 +253,13 @@ protected function getProblemShortName(int $probid, int $cid) : string
     protected function getClarificationFormData(Team $team = null) : array
     {
         $em = $this->getDoctrine()->getManager();
+        $sp = function($lteam) {sprintf("%s (t%s)", $lteam->getEffectiveName(), $lteam->getTeamid());};
```
mvr320, line 256: This probably should be done differently

nickygerritsen: Maybe put it in a separate function? Otherwise you need the `return` at least.

PHP 7.4+ could do:

```php
$sp = fn($lteam) => sprintf("%s (t%s)", $lteam->getEffectiveName(), $lteam->getTeamid());
```

mvr320: Ok... I'll take it just out of this PR

nickygerritsen (approved):

PR #1014, opened 2021-03-25
Remove temporary ApiDescriber fix

mvr320:
The linked GitHub issue has been closed and this seems to work.

@nickygerritsen not sure if I understood the problem, but for me the ApiDoc works and the example endpoint you posted in that issue has the correct amount of parameters.


nickygerritsen:
I will test tomorrow. What I remember is you could not submit in the api doc since it thought vars needed to be supplied twice

nickygerritsen:
Nope, try to use the UI to get a clarification. It will make the fields red since it exposes `id` twice. So seems we still need the fix.

nickygerritsen:
@mvr320 I see Nelmio API Doc Bundle 4.2.0 has the fix. If I update to it locally on your branch it seems to work. IMHO we can (and should) update to it.

mvr320:
> 
> 
> @mvr320 I see Nelmio API Doc Bundle 4.2.0 has the fix. If I update to it locally on your branch it seems to work. IMHO we can (and should) update to it.

Ok, I'll update (and also checkout if those are the only files we should alter/change)

nickygerritsen (approved):

PR #1016, opened 2021-03-26
Dependency doctrine

mvr320:
After doing this a `grep -r "DoctrineCache" --exclude-dir cache webapp` results nothing, so I think its safe to remove this?


nickygerritsen (approved):

nickygerritsen:
If CI is green gogogo

mvr320:
> 
> 
> If CI is green gogogo

CI was green before (this is only a squashed commit now)

PR #1023, opened 2021-03-28
Allow .env.local to override variables so that we can set DATABASE_URL with the correct MySQL version.

nickygerritsen:
So as noted running `webapp/bin/console doctrine:migrations:diff` gave a huge diff when you had a database that was not MySQL 5.7.0, because we specify that version in `load_db_secrets.php`.

Since it seems this server version string is only useful for creating migrations, which is only useful for people developing on DOMjudge, it seemed logical to make the `.env.local` (which we already use to run in `APP_ENV=dev` mode) to override the database string. With this change you can add something like this to your `.env.local`:

```bash
DATABASE_URL=mysql://domjudge:domjudge@mariadb:3306/domjudge?serverVersion=mariadb-10.5.9
```

Assuming you are running MariaDB 10.5.7. This makes the migration diff work again.

Also made a small fix in an index specification, since it would result in a diff every time.

### DIFF
--- a/doc/manual/develop.rst
+++ b/doc/manual/develop.rst
@@ -71,6 +71,24 @@ maintainer-install`` when the file did not exist before.
 For more details see
 `https://symfony.com/doc/current/configuration/dot-env-changes.html`.
 
+The ``webapp/.env.local`` file can also be used to overwrite the database
+version. This is needed to automatically generate migrations based on the
+current database compared to the models. To set the correct version, add a line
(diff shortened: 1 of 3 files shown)


meisterT:
Can you document this also somewhere?

nickygerritsen:
> Can you document this also somewhere?

Done!

meisterT (approved):

PR #1024, opened 2021-03-28
Sort rejudgings on relevant status

mvr320:
We now prefer:
x% done > ready > cancelled+applied

image

I assume that sorting on the ids is not relevant here (and would only delay the page) as you can do this in the UI.

### DIFF
--- a/webapp/src/Controller/Jury/RejudgingController.php
+++ b/webapp/src/Controller/Jury/RejudgingController.php
@@ -102,18 +102,18 @@ class RejudgingController extends BaseController
             ->getQuery()->getResult();
 
         $table_fields = [
-            'rejudgingid' => [
-                'title' => 'ID',
-                'sort' => true,
-                'default_sort' => true,
-                'default_sort_order' => 'desc'
-            ],
+            'rejudgingid' => ['title' => 'ID', 'sort' => true],
             'reason' => ['title' => 'reason', 'sort' => true],
             'startuser' => ['title' => 'startuser', 'sort' => true],
             'finishuser' => ['title' => 'finishuser', 'sort' => true],
             'starttime' => ['title' => 'starttime', 'sort' => true],
             'endtime' => ['title' => 'finishtime', 'sort' => true],
-            'status' => ['title' => 'status', 'sort' => true],
+            'status' => [
+                'title' => 'status',
+                'sort' => true,
+                'default_sort' => true,
+                'default_sort_order' => 'asc'
+            ],
         ];
 
         $timeFormat       = (string)$this->config->get('time_format');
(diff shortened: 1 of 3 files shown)


webapp/src/Controller/Jury/RejudgingController.php
```diff
@@ -103,15 +103,20 @@ public function indexAction(Request $request): Response
             'rejudgingid' => [
                 'title' => 'ID',
                 'sort' => true,
-                'default_sort' => true,
-                'default_sort_order' => 'desc'
+                //'default_sort' => true,
```
meisterT, line 106: delete?

webapp/src/Controller/Jury/RejudgingController.php
```diff
@@ -164,9 +173,20 @@ public function indexAction(Request $request): Response
...
         }
 
+        dump($rejudgings_table);
+        # Sort the rejudging table to have the more relevant things on top.
+        usort($rejudgings_table,
+            function ($l, $r): bool {
+                return $l['sort'] > $r['sort'];
+            }
+        );
+        dump($rejudgings_table);
```
meisterT, line 188: remove

webapp/src/Controller/Jury/RejudgingController.php
```diff
@@ -164,9 +173,20 @@ public function indexAction(Request $request): Response
                 'actions' => [],
                 'link' => $this->generateUrl('jury_rejudging', ['rejudgingId' => $rejudging->getRejudgingid()]),
                 'cssclass' => $class,
+                'sort' => $status_sort,
             ];
+
         }
 
+        dump($rejudgings_table);
```
meisterT, line 181: remove

webapp/src/Controller/Jury/RejudgingController.php
```diff
@@ -141,16 +146,20 @@ public function indexAction(Request $request): Response
 
             if ($rejudging->getEndtime() !== null) {
                 $status = $rejudging->getValid() ? 'applied' : 'canceled';
+                $status_sort = 2;
```
meisterT, line 149: s/status_sort/sort_order/ ?

mvr320:
@nickygerritsen what would be the preferred way to make a Test for this? So how can I mock the data that there is one rejudging at x% done, or even better. What is the preferred way to enter judging results back in the unit testing?

nickygerritsen:
> @nickygerritsen what would be the preferred way to make a Test for this? So how can I mock the data that there is one rejudging at x% done, or even better. What is the preferred way to enter judging results back in the unit testing?

I think fixtures: https://symfonycasts.com/screencast/phpunit/fixtures-fast-db

webapp/tests/Controller/Jury/RejudgingControllerTest.php
```diff
@@ -0,0 +1,60 @@
...
+        $this->logOut();
+        $this->logIn();
+        $this->verifyPageResponse('GET', '/jury/rejudgings', $http);
+        if($http===200) {
+            foreach(['No rejudgings defined',' Add new rejudging','Rejudgings'] as $element) {
+                self::assertSelectorExists('body:contains("'.$element.'")');
+            }
+        }
+        // TODO: static::$roles is not always reset between tests
+        static::$roles = $oldRoles;
```
meisterT, line 27: Move this to the usual `tearDown` function and then delete here and below.

mvr320: I agree that is a solution, but I think this should be done in the BaseTest instead of the individual tests. I tried it in the BaseTest but some other tests actually seem to rely on this behavior. 

So I prefer to handle this in a separate PR and fix all of them in one go, or did you mean another `tearDown()`?

meisterT: Oh well, if we have tests that rely on this, it's not great, but you're right it's not the problem of this change, so go ahead.

meisterT (approved):

PR #1025, opened 2021-03-28
Speed up rejudgings and show progress while creating, canceling and finishing them.

nickygerritsen:
The speedup is handled by doing mass SQL inserts.

Since it still takes a while to create big rejudgings, we now show a progress bar to indicate how far along the progress is. We also set a timeout of 0 when creating rejudgings, allowing it to finish.
I tried to minimize the duplicated code and fixes some typo's along the way.

Gif of it in action:
Screen Recording 2021-03-28 at 15 34 58

### DIFF
--- a/webapp/src/Service/RejudgingService.php
+++ b/webapp/src/Service/RejudgingService.php
@@ -11,6 +11,7 @@ use App\Entity\Submission;
 use App\Entity\Team;
 use App\Entity\User;
 use App\Utils\Utils;
+use BadMethodCallException;
 use Doctrine\DBAL\DBALException;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\NonUniqueResultException;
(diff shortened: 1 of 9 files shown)


nickygerritsen:
Note that since this gif canceling/applying also has a progress bar. It looks similar to the create one.

webapp/templates/jury/rejudging_finish.html.twig
```diff
@@ -13,38 +13,60 @@
...
+                return responseReader.read().then(function(result) {
+                    if (result.done) { return; }
 
-                lastResponseLength = response.length;
+                    var chunk = new TextDecoder().decode(result.value);
+                    // Make sure we start at the last {
+                    var data = JSON.parse(chunk.substring(chunk.lastIndexOf('{')));
 
-                $status.append(output);
+                    $pogress
```
mvr320, line 48: is this intentional?

nickygerritsen: The typo? No. Fixed.

webapp/src/Service/RejudgingService.php
```diff
@@ -183,9 +203,9 @@ public function finishRejudging(Rejudging $rejudging, string $action, callable $
...
 
         if ($rejudging->getEndtime()) {
-            $error = sprintf('Rejudging already %s.', $rejudging->getValid() ? 'applied' : 'canceled');
+            $error = sprintf('<div class="alert alert-danger">Rejudging already %s.</div>', $rejudging->getValid() ? 'applied' : 'canceled');
             if ($progressReporter) {
-                $progressReporter($error, true);
+                $progressReporter(0, '', $error);
                 return false;
             } else {
                 throw new \BadMethodCallException($error);
```
mvr320, line 211: @nickygerritsen

meisterT (approved):

PR #1026, opened 2021-03-29
When creating judgetasks, compute per problem data once instead of for each testcase.

meisterT:


meisterT:
This is also done as part of https://github.com/DOMjudge/domjudge/pull/1025 so I am closing this one.

PR #1027, opened 2021-03-29
Restructure import/export page

thijskh:
Attempt at improving the import/export page by reordering the items and putting them into groups related to the type of content they import/export instead of the file type.

Add buttons on the contests, problems and teams page to refer to the import page as requested by @meisterT.

Includes tests for the page itself and for all exports. No tests yet for importing things.

# Was

Screenshot_2021-03-29 Import and export - DOMjudge(1)

# Is
Screenshot_2021-03-29 Import and export - DOMjudge


webapp/templates/jury/import_export.html.twig
```diff
@@ -6,110 +6,120 @@
...
                 {{ form_end(icpccms_form) }}
             </div>
         </div>
     </div>
 
+    <h2 id="results">Results</h2>
+    <div class="row">
+        <div class="col-md-6">
+        </div>
+        <div class="col-md-6">
```
nickygerritsen, line 83: `col-md-push-6` on this one instead of the empty `col-md-6` above?

webapp/templates/jury/import_export.html.twig
```diff
@@ -6,110 +6,120 @@
...
+            <div class="tile">
+                <h3>Export &lt;html&gt;</h3>
+                    <ul>
+                    {% for sort_order in sort_orders %}
+                        <li>
+                            <a href="{{ path('jury_html_export', {'type': 'results'}) }}?sort_order={{ sort_order }}"><code>results.html</code> for sort order {{ sort_order }}</a>
```
nickygerritsen, line 89: If we're changing this anyway:

```suggestion
                            <a href="{{ path('jury_html_export', {'type': 'results', 'sort_order': sort_order}) }}"><code>results.html</code> for sort order {{ sort_order }}</a>
```

also works and looks nicer IMHO.

webapp/tests/Controller/Jury/ImportExportControllerTest.php
```diff
@@ -0,0 +1,218 @@
...
+    - { name: Scala }
+    - { name: 'POSIX shell' }
+    - { name: Swift }
+problems:
+    - { label: boolfind, name: 'Boolean switch search', color: limegreen, rgb: '#32CD32' }
+    - { label: fltcmp, name: 'Float special compare test', color: yellow, rgb: '#FFFF00' }
+    - { label: hello, name: 'Hello World', color: magenta, rgb: '#FF00FF' }
+
+HEREDOC;
+        yield["2", $yaml];
```
nickygerritsen, line 137: Nit: space before [. Also a few more below.

meisterT:
Nice improvement. I am wondering whether we should remove some dropdown and form and replace it with auto-detection based on base name and extension of the file, e.g. `groups.tsv` and `accounts.json` are obvious what you want to upload.

nickygerritsen (approved):
Very nice! I like the tests :smile:

Some very small comments.

thijskh:
@meisterT It's not very explicit currently that the files are actually mandatorily named that way. But if this is guaranteed, we can indeed do away with the dropdowns and the TSV/JSON difference.

Begs the question whether we can get away with just one "drop your files here" upload which processes anything you throw at it (contest.yaml, accounts.tsv, hello.zip). But I'm guessing that this is not worth it currently and we can find better ways to improve DOMjudge.

meisterT:
After my changes
image

meisterT:
> @meisterT It's not very explicit currently that the files are actually mandatorily named that way. But if this is guaranteed, we can indeed do away with the dropdowns and the TSV/JSON difference.

That's something for Nicky/Jaap.

nickygerritsen:
> > @meisterT It's not very explicit currently that the files are actually mandatorily named that way. But if this is guaranteed, we can indeed do away with the dropdowns and the TSV/JSON difference.
> 
> That's something for Nicky/Jaap.

I think *we* always name them like this since we use ccs config repo's, but I think there are also users who use the import/export to do this importing and they will not follow the naming we use.
Even more so: we mostly use the API to do the importing/exporting.

I do like the idea of auto detecting but not sure how we could do it easily.

meisterT:
Can we at least auto detect the extension / filetype?

webapp/templates/jury/import_export.html.twig
thijskh: Doesn't work for me...

nickygerritsen:
We can do that yes. Then the only problem is that TSV has accounts but JSON doesn't and JSON has orgs and TSV doesn't. So if we have one dropdown with all 4 types, we return an error if you select an unsupported type?

webapp/templates/jury/import_export.html.twig
nickygerritsen: 😢 apparently it's `col-md-offset-6`: https://getbootstrap.com/docs/3.4/css/#grid-offsetting

thijskh:
Indeed, my original thought of keeping them separate is that it doesn't improve clarity much for me if some types work with some format and other do not.

Dare to dream: can we just do away with two formats for the same kind of data? Just drop support for one or the other?

nickygerritsen:
> Indeed, my original thought of keeping them separate is that it doesn't improve clarity much for me if some types work with some format and other do not.
> 
> Dare to dream: can we just do away with two formats for the same kind of data? Just drop support for one or the other?

If only we could. The json is new for this year so I expect the old format to stick around for a bit. At some point it could go away

PR #1032, opened 2021-03-29
Fix judge shellescape

eldering:
Use `escapeshellarg`, or actually our own wrapper, and a few other small fixes.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -257,10 +257,8 @@ function fetch_executable(
     $execbuildpath   = $execbuilddir . '/build';
     $execrunpath     = $execbuilddir . '/run';
     if (!is_dir($execdir) || !file_exists($execdeploypath)) {
-        $rm_exec_dir_builddir = escapeshellcmd("rm -rf '$execdir' '$execbuilddir'");
-        system($rm_exec_dir_builddir);
-        $mk_builddir = escapeshellcmd("mkdir -p '$execbuilddir'");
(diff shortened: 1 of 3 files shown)


nickygerritsen (approved):

judge/judgedaemon.main.php
```diff
@@ -947,8 +941,13 @@ function compile(array $judgeTask, string $workdir, string $workdirpath, array $
...
     // Compile the program.
-    $compile_cmd = escapeshellcmd(LIBJUDGEDIR . "/compile.sh $cpuset_opt '$execrunpath' '$workdir' " .
-                                  implode(' ', $files));
+    $compile_cmd = LIBJUDGEDIR . "/compile.sh $cpuset_opt " .
+        implode(' ', array_map('dj_escapeshellarg', array_merge([
+            $execrunpath,
+            $workdir,
+        ], $files)));
+    logmsg(LOG_NOTICE, "Compile command: ".$compile_cmd);
+    logmsg(LOG_NOTICE, "Compile files: ".implode(' ', $files));
```
meisterT, line 950: this is part of the previous command, isn't it?

meisterT (approved):

judge/judgedaemon.main.php
eldering: > this is part of the previous command, isn't it?

Indeed, and that's leftover debug stuff. Changed it into a single `LOG_DEBUG` message.

PR #1036, opened 2021-03-31
Fix quotes that crashed the judgehost and chmod 'run' after compiled

bkorecic:
The quotes that I removed caused the judgehost to crash. The line I added makes the 'run' file executable, which was necessary when the run file was automatically written instead of, for example, a C++ executable which has the permission by default.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -331,11 +331,12 @@ function fetch_executable(
                     $buildscript .= "javac -cp ./ -d ./ '$source'.java\n";
                     $buildscript .= "echo '#!/bin/sh' > run\n";
                     // no main class detection here
-                    $buildscript .= "echo 'java -cp ./ '$source' >> run\n";
+                    $buildscript .= "echo 'java -cp ./ '$source >> run\n";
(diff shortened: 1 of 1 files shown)


judge/judgedaemon.main.php
```diff
@@ -403,6 +404,7 @@ function fetch_executable(
             if ($retval!==0) {
                 return [null, "Failed to build executable in $execdir."];
             }
+            chmod($execrunpath, 0755);
```
meisterT, line 407: instead of doing this here let's do `chmod +x run` in the corresponding java and python scripts above

meisterT (commented):
Thanks for your PR!

judge/judgedaemon.main.php
bkorecic: Thank you for your review! Why though? I thought it made more sense to do it when the executable is created, the same way it is done with the build file once it is created (being consistent with that).

meisterT: Sounds reasonable

meisterT (approved):

PR #1037, opened 2021-04-02
Add button to request remaining runs.

### LINKED ISSUE #162: Judge all test files on request even if lazy eval is enabled
With lazy evaluation, judging stops at the first error with highest priority.

Sometimes, it's useful to have the possibility to request the judgement for all test files for a single submission even if lazy evaluation is enabled for that problem.

meisterT:
Use-case: The contest administrator enables lazy judging because of
limited resources, but the jury wants for particular
submissions/judgings the results of all testcases.

The judgetasks got downgraded in priority already when the final verdict
was determined, so we don't have to do that explicitly here.

Fixes #162.


nickygerritsen (approved):
Beautiful

PR #1038, opened 2021-04-02
Country flags in api

nickygerritsen:


nickygerritsen:
For the country flags I regenerated them using the mentioned repository. Mauritania didn't have the correct flag, so I used my Inkscape and SVG edit skills to update the flag with the new red borders. Copied the color from the Wikipedia flag SVG.
Also updated codes.txt to reflect changes in country names where applicable.

Note that the CCS spec suggests:

> A server is recommended to provide flags of size around 56x56 and 160x160.

But I think it's not very useful to add 160x160 (ish) flags only to expoose them through the API. What do others think?
Our 80x60 flags are close enough to 56x56 IMHO.

meisterT:
Are there actually tools that use flags from the API instead of bringing their own?

nickygerritsen:
The ICPC tools as of https://github.com/icpctools/icpctools/commit/e55beb3cb9ed521473ce377686cf1d01053aa262
And the idea for the change in the CCS spec is that the API supports it and thus there is one place where you configure whether flags are available instead of in every tool.

webapp/src/Utils/Utils.php
```diff
@@ -243,8 +243,10 @@ class Utils
         "FJI" => "Fiji",
         "FIN" => "Finland",
         "FRA" => "France",
-        "GUF" => "French Guiana",
+        // Removed since not an "official" country
+        // "GUF" => "French Guiana",
         "PYF" => "French Polynesia",
+        // Removed since not an "official" country
         "ATF" => "French Southern Territories",
```
meisterT, line 250: Did you mean to comment this line?

meisterT (approved):

webapp/src/Utils/Utils.php
nickygerritsen: I did

thijskh:
Maybe thinking a bit too much outside of the box, but what about something like https://github.com/lipis/flag-icon-css?
So we have nice SVGs that will work well in any format instead of PNGs in hardcoded dimensions?

nickygerritsen:
I didn’t take something like this for two reasons:
* we now have “glossy” flags. Do we want to keep that? Can we with this solution?
* how do we expose them as png in the api? Can we?

meisterT:
Although I find the flags that we currently have a little bit more visually appealing, this will also look fine, SVGs are great and the repository is kept up to date. So overall I like Thijs' idea.

meisterT:
Ah, for the API: can we change the accepted formats there to include SVGs? It's likely better for downstream clients to have a high quality data source that they can convert to their needs instead of "arbitrary" resolution PNGs.

nickygerritsen:
I will put this on the ccs spec agenda

nickygerritsen:
I have updated this PR to use the SVG library suggested, which existed as a composer dependency so we can even easily keep it up to date.

Note that I am not using a `<span>` tag like they suggested, since that requires more CSS changes on our side to make it line up. [Any browser released in the last 10 years](https://caniuse.com/svg-img) supports rendering an SVG image inside an `` tag, so that seems like the easier solution.

The API now exposes both the square as well as the 4x3 flags.

I also got rid of our big countries array, since Symfony has a built in class that can do this for us.

nickygerritsen:
Changed URLs: 
/public

doc/manual/config-advanced.rst
```diff
@@ -10,10 +10,6 @@ You can place the images under the path `public/images/` (see
 the Config checker in the admin interfae for the full filesystem
 path of your installation) as follows:
 
-- *Country flags* are shown when the ``show_flags`` configuration option
```
meisterT, line 13: why did you remove this?

webapp/src/Controller/API/GeneralInfoController.php
```diff
@@ -286,6 +288,48 @@ public function getConfigCheckAction(): JsonResponse
...
+     */
+    public function countryFlagAction(string $countryCode, string $size): Response
+    {
+        // This API action exists for two reasons
+        // - Relative URL's are relative to the API root according to the CCS spec. This
+        //   means that we need to have an API endpoint for files.
+        // - This makes it that we can not return a flag if flags are disabled.
+
+        if (!$this->config->get('show_flags')) {
+            throw new NotFoundHttpException('country flag not found');
```
meisterT, line 314: Well, we might have found it, but it's not enabled :-)

webapp/src/Controller/API/GeneralInfoController.php
```diff
@@ -286,6 +288,48 @@ public function getConfigCheckAction(): JsonResponse
...
+        }
+
+        $alpha3code = strtoupper($countryCode);
+        if (!Countries::alpha3CodeExists($alpha3code)) {
+            throw new NotFoundHttpException('country flag not found');
```
meisterT, line 319: mention the missing country code in the message?

webapp/src/Controller/API/GeneralInfoController.php
```diff
@@ -286,6 +288,48 @@ public function getConfigCheckAction(): JsonResponse
...
+        }
+        $alpha2code = strtolower(Countries::getAlpha2Code($alpha3code));
+        $flagFile = sprintf('%s/public/flags/%s/%s.svg', $this->dj->getDomjudgeWebappDir(), $size, $alpha2code);
+
+        if (!file_exists($flagFile)) {
+            throw new NotFoundHttpException('country flag not found');
```
meisterT, line 325: How is this different from the one above?

webapp/src/DataFixtures/Test/SampleAffilicationsFixture.php
```diff
@@ -0,0 +1,30 @@
...
+use Doctrine\Bundle\FixturesBundle\Fixture;
+use Doctrine\Persistence\ObjectManager;
+
+class SampleAffilicationsFixture extends Fixture
+{
+    public function load(ObjectManager $manager)
+    {
+        $affiliationData = [
+            // short name, name,                                                country
+            ['FAU',        'Friedrich-Alexander-Universität Erlangen-Nürnberg', 'DEU'],
```
meisterT, line 15: :heart_eyes:

doc/manual/config-advanced.rst
nickygerritsen: Whoops

webapp/src/Controller/API/GeneralInfoController.php
nickygerritsen: Well I didn’t want to expose this difference but if you think we should I can change it

webapp/src/Controller/API/GeneralInfoController.php
nickygerritsen: If we somehow not have an SVG for a valid alpha3/2 code

webapp/src/Controller/API/GeneralInfoController.php
meisterT: Yep, I think we should make this explicit to ease debugging.

nickygerritsen:
Changed URLs:
/api/doc
/jury/affiliations
/jury/affiliations/1
/jury/affiliations/1/edit
/jury/affiliations/add
/jury
/jury/scoreboard
/style/.*
/team/scoreboard
/team/team/2

thijskh (approved):

PR #1039, opened 2021-04-02
Allow dots (.) in external identifiers. Fixes #1030.

### LINKED ISSUE #1030: Add support for dots in identifiers
The CCS spec now allows a dot in an identifierL icpc/ccs-specs#20. We should also allow this.

nickygerritsen:
As the comment states we do not check if the identifiers do not start or end with invalid characters.
If someone has an idea to easily do this and still allow the check in `ImportExportService::importContestYaml` to be easy, I am all ears. This is the regex I had to check for beginning/end, very ugly:

```
/^(([a-zA-Z0-9_][a-zA-Z0-9_.-]*[a-zA-Z0-9-])|[a-zA-Z0-9])$/
```

### DIFF
--- a/webapp/src/Controller/API/SubmissionController.php
+++ b/webapp/src/Controller/API/SubmissionController.php
@@ -333,6 +333,10 @@ class SubmissionController extends AbstractRestController
 
         if ($submissionId = $request->request->get('id')) {
             if ($this->isGranted('ROLE_API_WRITER')) {
+                if (preg_match(DOMJudgeService::EXTERNAL_IDENTIFIER_REGEX, $submissionId) !== 1) {
+                    throw new BadRequestHttpException(sprintf("ID %s is not valid", $submissionId));
+                }
+
                 // Check if we already have a submission with this ID
                 $existingSubmission = $this->em->createQueryBuilder()
                     ->from(Submission::class, 's')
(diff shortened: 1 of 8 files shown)


webapp/src/Controller/API/SubmissionController.php
```diff
@@ -333,6 +333,10 @@ public function addSubmissionAction(Request $request): string
 
         if ($submissionId = $request->request->get('id')) {
             if ($this->isGranted('ROLE_API_WRITER')) {
+                if (preg_match(DOMJudgeService::EXTERNAL_IDENTIFIER_REGEX, $submissionId) !== 1) {
```
meisterT, line 336: Do we actually care? As in: can we allow more than in the spec?

nickygerritsen: Well we expose them through the API as their ID. Personally I don’t care and we barely use them, but I can imagine people using the api assuming they are safe to store as file names for example. Which is exactly what the CDS does

meisterT: I wouldn't hard fail on something that is irrelevant to us.

nickygerritsen: I can remove this check if you want. But what about setting external ids of other entities? If you used the UI we already didn’t allow it before. Do you want to drop that?

meisterT: That's a fair point. Feel free to submit as is.

PR #1040, opened 2021-04-02
Add option to specify priority for rejudgings.

### LINKED ISSUE #639: Set rejudgings as lower priority for pickup by a judgehost
Currently, when you start a rejudging with a big set of submissions, this will DoS the judging queue, since those submissions have the same priority as any normal, new submissions.

In most cases, you'd probably like to make them lower priority.

meisterT:
Fixes #639.

### DIFF
--- a/webapp/src/Form/Type/RejudgingType.php
+++ b/webapp/src/Form/Type/RejudgingType.php
@@ -44,6 +44,16 @@ class RejudgingType extends AbstractType
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder->add('reason', TextType::class);
+        $builder->add('priority', ChoiceType::class,
+            [
+                'choices' => [
+                    'low' => 'low',
+                    'default' => 'default',
+                    'high' => 'high',
+                ],
+                'data' => 'default',
+            ]
+        );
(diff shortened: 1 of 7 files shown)


webapp/src/Form/Type/RejudgingType.php
```diff
@@ -44,6 +44,16 @@ public function __construct(EntityManagerInterface $em)
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder->add('reason', TextType::class);
+        $builder->add('priority', ChoiceType::class,
+            [
+                'choices' => [
+                    'low' => 'low',
```
nickygerritsen, line 50: Why not set the keys (or values, I always forget) to the actual prio numbers? Symfony will validate you actually pass a valid value so you can’t “hack” the form and specify something else (well technically because we have the Ajax in between you could but 🤷‍♂️)

meisterT: You could also use `curl` or `httpie`, and I don't want to accept arbitrary priorities.

nickygerritsen: Fair point

Issue #990, opened 2021-02-08
Deprecated composer dependencies


nickygerritsen:
Our current `master` has the following deprecated composer packages:

```
Package doctrine/doctrine-cache-bundle is abandoned, you should avoid using it. No replacement was suggested.
Package doctrine/reflection is abandoned, you should avoid using it. Use roave/better-reflection instead.
Package phpunit/php-token-stream is abandoned, you should avoid using it. No replacement was suggested.
Package twig/extensions is abandoned, you should avoid using it. No replacement was suggested.
Package zendframework/zend-code is abandoned, you should avoid using it. Use laminas/laminas-code instead.
```

Do we want to get rid of them? Below a list of where they come from.

# doctrine/doctrine-cache-bundle

- [x] This is a root dependency which is deprecated and not used anymore for Symfony 5. I think we can just remove it (and maybe add `doctrine/cache` if we still use it

# doctrine/reflection

- [x] Used by `doctrine/common`. We could update that if we update doctrine/doctrine-bundle.

# phpunit/php-token-stream

- [ ] Used by `phpunit/php-code-coverage` 7.x and 8.x. 9.x requires PHP 7.3 but since it is only for the tests we could do that.

# twig/extensions

- [x] See https://github.com/twigphp/Twig-extensions. Not sure which extensions we actually use but we can probably figure that out.

# zendframework/zend-code

- [x] Used by `ocramius/proxy-manager` which in turn is used by `jms/serializer`, `doctrine/doctrine-bundle` and `doctrine/doctrine-bundle`. `jms/serializer` we can just upgrade. For the latter two we could upgrade again.

thijskh:
`twig/extensions` is gone now

thijskh:
I disagree with `phpunit/php-token-stream`. If we support PHP 7.2 we must also be able to run the unit tests against it. If we do not currently do so you could even see that as a bug/missing feature. But in any case I don't think we should actively break that. Let's just accept that one.

nickygerritsen:
> I disagree with `phpunit/php-token-stream`. If we support PHP 7.2 we must also be able to run the unit tests against it. If we do not currently do so you could even see that as a bug/missing feature. But in any case I don't think we should actively break that. Let's just accept that one.

Agreed. We can update when we move to a newer php

PR #1041, opened 2021-04-02
Cleanup: upgrade deprecations

thijskh:
This attempts to resolve most of #990.

### DIFF
--- a/webapp/src/Doctrine/HashPasswordSubscriber.php
+++ b/webapp/src/Doctrine/HashPasswordSubscriber.php
@@ -4,7 +4,8 @@ namespace App\Doctrine;
 
 use App\Entity\User;
 use Doctrine\Common\EventSubscriber;
-use Doctrine\Common\Persistence\Event\LifecycleEventArgs;
+use Doctrine\ORM\Event\LifecycleEventArgs;
+use Doctrine\ORM\Event\PreUpdateEventArgs;
 use Doctrine\ORM\Events;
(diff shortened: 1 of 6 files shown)


webapp/src/Doctrine/HashPasswordSubscriber.php
```diff
@@ -4,7 +4,8 @@
 
 use App\Entity\User;
 use Doctrine\Common\EventSubscriber;
-use Doctrine\Common\Persistence\Event\LifecycleEventArgs;
+use Doctrine\ORM\Event\LifecycleEventArgs;
```
nickygerritsen, line 7: I think we’re not using this import anymore?

nickygerritsen (approved):

webapp/src/Doctrine/HashPasswordSubscriber.php
thijskh: We are using it in `public function preUpdate(PreUpdateEventArgs $args)`

nickygerritsen: That is PreUpdateEventArgs, not LifecycleEventArgs?

thijskh: *We are using it in `public function prePersist(LifecycleEventArgs $args)`

nickygerritsen: Now I see it 🤓

nickygerritsen (approved):

PR #1043, opened 2021-04-03
Clean up judgehost diskspace by default.

### LINKED ISSUE #943: Judgehosts should (optionally) clean up old judging data
For long running contests(or even short ones where you don't have a ton of disk space) it'd be nice for judgehosts to have an option to clean up old judging data, instead of just throwing an internal error and stopping judging.

meisterT:
This behavior can be disabled by starting the judgedaemon with
`--diskspace_error`.
If that flag is passed (or not enough diskspace is freed up), an
internal error is posted.

Judging directories are cleaned up in order of their last modification.

Fixes #943.

image

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -230,11 +230,12 @@ function usage()
 {
     echo "Usage: " . SCRIPT_ID . " [OPTION]...\n" .
         "Start the judgedaemon.\n\n" .
-        "  -d       daemonize after startup\n" .
-        "  -n <id>  daemon number\n" .
-        "  -v       set verbosity to LEVEL (syslog levels)\n" .
-        "  -h       display this help and exit\n" .
-        "  -V       output version information and exit\n\n";
+        "  -d                daemonize after startup\n" .
+        "  -n <id>           daemon number\n" .
+        "  --diskspace_error send internal error on low diskspace\n" .
+        "  -v                set verbosity to LEVEL (syslog levels)\n" .
+        "  -h                display this help and exit\n" .
+        "  -V                output version information and exit\n\n";
     exit;
 }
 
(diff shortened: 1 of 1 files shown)


judge/judgedaemon.main.php
```diff
@@ -438,6 +439,7 @@ function fetch_executable(
 if (isset($options['n'])) {
     $options['daemonid'] = $options['n'];
 }
+if (isset($options))
```
nickygerritsen, line 442: You miss something here

meisterT: Done.

nickygerritsen (approved):

judge/judgedaemon.main.php
```diff
@@ -612,29 +613,63 @@ function fetch_executable(
...
+                    return filemtime($a) <=> filemtime($b);
+                });
+                $after = $before = disk_free_space(JUDGEDIR);
+                logmsg(LOG_INFO,
+                    "🗑 Low on diskspace, cleaning up (" . count($candidateDirs) . " potential candidates).");
+                $cnt = 0;
+                foreach ($candidateDirs as $d) {
+                    $cnt++;
+                    logmsg(LOG_INFO, "  - deleting $d");
+                    system('rm -rf ' . dj_escapeshellarg("$d"), $retval);
```
eldering, line 640: No need to write `$d` inside an interpolated string.

eldering (approved):
LGTM except for a nitpick

thijskh:
Is it just me who thinks the _ should be a - in the commandline parameter name?

meisterT:
Thijs, no opinion here, so feel free to change.

eldering:
> Is it just me who thinks the _ should be a - in the commandline parameter name?

I'd also have a slight preference for `-`.

PR #1044, opened 2021-04-04
Allow to hide judgehosts that haven't checked in in a while.

### LINKED ISSUE #857: Cannot delete a judgehost from the web interface

I cannot seem to delete a (disconnected) judgehost from the web interface. If I click on the bin icon and press "Delete" nothing happens.

Screenshot 2020-07-28 at 21 47 30

Clicking on the status icon and then on "Delete judgehost" does not work either:

Screenshot 2020-07-28 at 21 53 40

I could however delete the `example-judgehost` that appears with the new domserver installation and that never got connected. 

The judge hosts that I try to delete have pending [aborted] or completed judgings. Could it be the problem? If that is a problem there should probably be some message or the delete button should be disabled.

### Your environment
- DOMjudge 7.3.0 (built from the sources)
- Ubuntu 20.04 server running inside an LXD container (not sure if that matters)
- LXC version 4.3
- $ uname -a
Linux judgehost2 5.4.0-42-generic #46-Ubuntu SMP Fri Jul 10 00:24:02 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

meisterT:
You can do this manually on individual judgehosts, but the main purpose
is to do this once after switching to a different set of judgehosts.

Fixes #857.

### DIFF
--- a/webapp/src/Controller/Jury/JudgehostController.php
+++ b/webapp/src/Controller/Jury/JudgehostController.php
@@ -83,6 +83,7 @@ class JudgehostController extends BaseController
             ->from(Judgehost::class, 'j')
             ->leftJoin('j.restriction', 'r')
             ->select('j', 'r')
+            ->andWhere('j.hidden = 0')
             ->orderBy('j.hostname')
             ->getQuery()->getResult();
 
@@ -127,6 +128,7 @@ class JudgehostController extends BaseController
         $time_warn = $this->config->get('judgehost_warning');
         $time_crit = $this->config->get('judgehost_critical');
         $judgehosts_table = [];
+        $all_checked_in_recently = true;
         foreach ($judgehosts as $judgehost) {
             $judgehostdata    = [];
             $judgehostactions = [];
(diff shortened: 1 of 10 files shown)


webapp/src/Controller/Jury/JudgehostController.php
```diff
@@ -83,6 +83,7 @@ public function indexAction(Request $request): Response
             ->from(Judgehost::class, 'j')
             ->leftJoin('j.restriction', 'r')
             ->select('j', 'r')
+            ->andWhere('j.hidden = 0')
```
nickygerritsen, line 86: Not really important but I think you could write false instead of 0 here

nickygerritsen (approved):

webapp/src/Controller/Jury/JudgehostController.php
meisterT: We use in most places 0/1, don't we?

nickygerritsen: Not sure but if we do keep it indeed

PR #1045, opened 2021-04-04
Use integer auto incrementing primary key for judgehosts.

nickygerritsen:
This makes indices faster on other tables (judging, judgetask) since they are no longer using string indices.


webapp/src/Controller/Jury/JudgehostController.php
```diff
@@ -98,10 +99,10 @@ public function indexAction(Request $request): Response
 
         $now           = Utils::now();
         $contest       = $this->dj->getCurrentContest();
-        $query         = 'SELECT judgehost, SUM(IF(endtime, endtime, :now) - GREATEST(:from, starttime)) AS `load`
-                          FROM judging
+        $query         = 'SELECT hostname, SUM(IF(endtime, endtime, :now) - GREATEST(:from, starttime)) AS `load`
+                          FROM judging INNER JOIN judgehost on judging.judgehostid = judgehost.judgehostid
```
thijskh, line 103: This could be avoided by just keying the workloads in the `$result` array and below by the judgehostid.

webapp/src/Controller/Jury/JudgehostController.php
```diff
@@ -357,7 +358,7 @@ public function deleteAction(Request $request, string $hostname): Response
      */
     public function activateAction(RouterInterface $router, Request $request, string $hostname): RedirectResponse
     {
-        $judgehost = $this->em->getRepository(Judgehost::class)->find($hostname);
+        $judgehost = $this->em->getRepository(Judgehost::class)->findOneBy(['hostname' => $hostname]);
```
thijskh, line 361: This UI action and the one below could just use the ID as the parameter as would other UI parts. Internally, use judgehostid as the reference where we can.

webapp/src/Controller/Jury/JudgehostController.php
```diff
@@ -370,7 +371,7 @@ public function activateAction(RouterInterface $router, Request $request, string
      */
     public function deactivateAction(RouterInterface $router, Request $request, string $hostname): RedirectResponse
     {
-        $judgehost = $this->em->getRepository(Judgehost::class)->find($hostname);
+        $judgehost = $this->em->getRepository(Judgehost::class)->findOneBy(['hostname' => $hostname]);
```
thijskh, line 374: As above.

webapp/src/Migrations/Version20210404121354.php
```diff
@@ -0,0 +1,92 @@
...
+
+        // First, add the new field on the judgehost table. For this to work, we first need to add a unique index on hostname, since otherwise foreign keys will break
+        $this->addSql('CREATE UNIQUE INDEX hostname ON judgehost (hostname)');
+        $this->addSql('ALTER TABLE judgehost DROP PRIMARY KEY');
+        $this->addSql('ALTER TABLE judgehost ADD judgehostid INT UNSIGNED AUTO_INCREMENT NOT NULL COMMENT \'Judgehost ID\' FIRST, ADD PRIMARY KEY (judgehostid)');
+
+        // Now add the new judgehostid field on judgetasks and judgings, including indices
+        $this->addSql('DROP INDEX hostname_valid_priority ON judgetask');
+        $this->addSql('DROP INDEX hostname_jobid ON judgetask');
+        $this->addSql('DROP INDEX specific_type ON judgetask');
```
thijskh, line 33: Now we have two sequential migrations in master that create big indices only to immediately drop them. We squash commits for more clarity - what are the thoughts about squashing migrations?

thijskh (changes requested):

webapp/src/Migrations/Version20210404121354.php
nickygerritsen: I am fine with this if others are ok with manually fixing their db (probably just recreating it)

nickygerritsen: Should be done

thijskh (approved):

PR #1046, opened 2021-04-10
Fix for SQL error on unverified submissions view

mvr320:
@nickygerritsen I'm not 100% if the Fixture is loaded in the correct way here?

### DIFF
--- a/webapp/src/Service/SubmissionService.php
+++ b/webapp/src/Service/SubmissionService.php
@@ -173,7 +173,7 @@ class SubmissionService
             if ($restrictions['verified']) {
                 $queryBuilder->andWhere('j.verified = 1');
             } else {
-                $queryBuilder->andWhere('j.verified = 0 OR (j.verified IS NULL AND s.judgehost IS NULL)');
+                $queryBuilder->andWhere('j.verified = 0 OR j.verified IS NULL');
             }
         }
 
--- a/webapp/tests/Controller/Jury/SubmissionControllerTest.php
+++ b/webapp/tests/Controller/Jury/SubmissionControllerTest.php
@@ -1,0 +1,41 @@
+<?php declare(strict_types=1);
+
+namespace App\Tests\Controller\Jury;
+
+use App\DataFixtures\Test\SampleSubmissionsFixture;
+use App\Tests\BaseTest;
+use Generator;
+
+class SubmissionControllerTest extends BaseTest
+{
+    protected static $roles = ['jury'];
+    protected static $baseURL = '/jury/submissions';
+
+    /**
+     * Test that the basic building blocks of the index page are there.
+     */
+    public function testIndexBasic(): void
+    {
+        $this->verifyPageResponse('GET', static::$baseURL, 200);
+    }
+
+    /**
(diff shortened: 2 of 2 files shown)


reset_db.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+echo 'drop database domjudge' | mysql -pdomjudge
```
nickygerritsen, line 2: Don’t add this magic file 🙈

webapp/src/Service/SubmissionService.php
```diff
@@ -173,7 +173,7 @@ public function getSubmissionList(array $contests, array $restrictions, int $lim
             if ($restrictions['verified']) {
                 $queryBuilder->andWhere('j.verified = 1');
             } else {
-                $queryBuilder->andWhere('j.verified = 0 OR (j.verified IS NULL AND s.judgehost IS NULL)');
+                $queryBuilder->andWhere('j.verified = 0 OR (j.verified IS NULL AND j.judgehost IS NULL)');
```
nickygerritsen, line 176: @meisterT are we even setting the judgeost on a judging?

webapp/tests/Controller/Jury/SubmissionControllerTest.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Test the filtered views do throw server errors
+     *
+     * @dataProvider provideViews
+     */
+    public function testIndexViewFilter(string $filter, array $fixtures): void
+    {
+        if (!empty($fixtures)) {
+            static::$fixtures = $fixtures;
```
nickygerritsen, line 30: This will not work. Call $this->loadFixture

reset_db.sh
mvr320: I think we should... but maybe in a different PR.

webapp/tests/Controller/Jury/SubmissionControllerTest.php
mvr320: Should now be fixed with a foreach loop.

webapp/tests/Controller/Jury/SubmissionControllerTest.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Test the filtered views do throw server errors
+     *
+     * @dataProvider provideViews
+     */
+    public function testIndexViewFilter(string $filter, array $fixtures): void
+    {
+        foreach($fixtures as $fixture) {
+            $this->loadFixture($fixture);
```
nickygerritsen, line 30: Ah lol there is also loadFixtures 🤣

mvr320: Thats probably better.

webapp/src/Service/SubmissionService.php
meisterT: We right now set the judgehost only when we pop back the compilation result on compile error. We should probably change that so that we remove that field from the table, but it's independent from this PR.

For this PR I would just remove the `AND ...` part.

meisterT (approved):

PR #1055, opened 2021-04-18
Add command to load development fixtures

mvr320:
@nickygerritsen is there an easier way of doing this? I needed this to check why an Unit test was failing.

### DIFF
--- a/doc/manual/develop.rst
+++ b/doc/manual/develop.rst
@@ -166,5 +166,10 @@ have to type these filters manually. If you use such an IDE, just make sure to
 specify the `webapp/phpunit.xml.dist` file as a PHPUnit configuration file and
 it should work.
 
+Loading development fixture data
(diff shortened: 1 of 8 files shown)


nickygerritsen:
I’m not sure what `bin/console doctrine:fixtures:load` has as options? Maybe you van provide it a path or something

mvr320:
> 
> 
> I’m not sure what `bin/console doctrine:fixtures:load` has as options? Maybe you van provide it a path or something

I check that before and I couldn't find it, so I would like this to be merged.

nickygerritsen (approved):

meisterT:
Looks good but please also mention in the developer docs.

mvr320:
> 
> 
> Looks good but please also mention in the developer docs.

I added a small section for it,

PR #1061, opened 2021-04-21
Dont inherit $roles of other Unit tests

mvr320:
The static roles leaked different values to other tests. Due to this
there was a difference when using the standard value or explicit setting
this in the test. This difference was shown when you either only run 1
test or run everything where this can either succeed or fail based on
the values of other tests.

@nickygerritsen I'm not sure if this is the correct explanation (or even the correct fix) but this works on my side.


nickygerritsen (approved):
Explanation makes total sense, so LGTM

PR #1062, opened 2021-04-23
Rejudging contest

mvr320:
This should provide only the rejudgings of the relevant contest.

I'll squash this before merging.

@meisterT is this indeed the feature which you wanted?

### DIFF
--- a/webapp/src/Controller/Jury/RejudgingController.php
+++ b/webapp/src/Controller/Jury/RejudgingController.php
@@ -4,7 +4,6 @@ namespace App\Controller\Jury;
 
 use App\Controller\BaseController;
 use App\Entity\Contest;
-use App\Entity\ContestProblem;
 use App\Entity\Judgehost;
 use App\Entity\JudgeTask;
 use App\Entity\Judging;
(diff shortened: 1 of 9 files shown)


webapp/src/Controller/Jury/RejudgingController.php
```diff
@@ -94,12 +95,28 @@ public function __construct(
      */
     public function indexAction(Request $request): Response
     {
+        $curContest = $this->dj->getCurrentContest();
         /** @var Rejudging[] $rejudgings */
         $rejudgings = $this->em->createQueryBuilder()
+            ->select('r')
+            ->from(Rejudging::class, 'r');
+        if ($curContest !== NULL) {
+            $rejudgings = $rejudgings->join(Submission::class, 's', Join::WITH, "s.rejudging = r.rejudgingid")
```
nickygerritsen, line 104: Why not ->join(r.submission, s)?

mvr320: Because when the `Join::WITH` is removed the JOIN is done without a "ON" in the query. I'm not sure how to fix that as I agree this is kinda ugly.

nickygerritsen: Also if you use r.submission instead of submisssion::class? Shouldn’t be the case

mvr320: I think this is now better.

nickygerritsen (approved):

meisterT (approved):
Thanks!

PR #1066, opened 2021-04-25
Repush add shortdesk

mvr320:
There is some optimization still to be done, the new forloops will often find the same variables for the DB and can be optimized away.

This PR is first to see if the current code does still work.

The commits are done in such a way that the indentation can be followed, I'll rewrite the commits when this is approved to make more sense for the history.


webapp/src/Entity/Contest.php
```diff
@@ -381,6 +381,11 @@ public function getShortname(): string
         return $this->shortname;
     }
 
+    public function getShortDesc(): string
```
nickygerritsen, line 384: I’d be more of a fan to use getShortDescription

webapp/tests/Unit/Controller/Jury/tmpTest.php
```diff
@@ -0,0 +1,52 @@
+webapp/tests/Unit/Controller/Jury/ContestControllerTest.php
```
nickygerritsen, line 1: What’s this file?

webapp/src/Controller/BaseController.php
```diff
@@ -191,70 +182,72 @@ protected function deleteEntity(
...
-                                    $dependentEntities = $this->getDependentEntities($table, $relations);
-                                    if (!empty($dependentEntities)) {
-                                        $dependentEntitiesReadable = [];
-                                        foreach ($dependentEntities as $dependentEntity) {
-                                            $parts                       = explode('\\', $dependentEntity);
-                                            $dependentEntityType         = $parts[count($parts) - 1];
-                                            $dependentEntitiesReadable[] = str_replace(
-                                                '_', ' ',
-                                                $inflector->tableize($inflector->pluralize($dependentEntityType))
+        foreach($entities as $entity) {
```
nickygerritsen, line 185: Nit: missing space

webapp/src/Controller/BaseController.php
```diff
@@ -137,28 +137,19 @@ protected function saveEntity(
...
+    protected function deleteEntities(
         Request $request,
         EntityManagerInterface $entityManager,
         DOMJudgeService $DOMJudgeService,
         EventLogService $eventLogService,
         KernelInterface $kernel,
-        $entity,
-        string $description,
+        array $entities,
         string $redirectUrl
```
nickygerritsen, line 153: Is this to prepare for multidelete?

mvr320: yes, as getting this right is hard so I rather do the backend logic first and thann have the frontend and the needed route for it.

webapp/tests/Unit/Controller/Jury/tmpTest.php
mvr320: This branch originally also had tests... in refactoring I moved these to this tmp file, should not be needed in this form

webapp/src/Entity/Contest.php
mvr320: I agree, will fix it.

nickygerritsen (approved):

mvr320:
@nickygerritsen I changed some things in the commits, but I think the tests if everything works for lists with not only 1 element can be done later when the multidelete is actually added.

PR #1067, opened 2021-04-25
Semisecure lower hashingcost

mvr320:


nickygerritsen (approved):

webapp/src/Command/ResetUserPasswordCommand.php
```diff
@@ -32,27 +32,17 @@ class ResetUserPasswordCommand extends Command
...
-     *
-     * @param EntityManagerInterface       $em
-     * @param UserPasswordEncoderInterface $passwordEncoder
-     * @param string|null                  $name
-     */
     public function __construct(
         EntityManagerInterface $em,
         UserPasswordEncoderInterface $passwordEncoder,
-        string $name = null
+        ?string $name = null
```
nickygerritsen, line 38: we can just drop the name arg (and pass nothing to the parent)

nickygerritsen (changes requested):

nickygerritsen (approved):

mvr320:
I found a bug so will first make tests before merging this.

mvr320:
Some things have changed so if someone can check again that would be nice.

PR #1068, opened 2021-04-26
[DEBUG] see if the coverage.xml is correct

mvr320:
@nickygerritsen it works, so I think this should just be merged.

### DIFF
--- a/gitlab/unit-tests.sh
+++ b/gitlab/unit-tests.sh
@@ -2,6 +2,8 @@
 
 set -euxo pipefail
 
+. gitlab/dind.profile
+
 # Set up
 "$( dirname "${BASH_SOURCE[0]}" )"/base.sh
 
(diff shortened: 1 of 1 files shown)


gitlab/unit-tests.sh
```diff
@@ -35,6 +37,7 @@ curl https://api.github.com/repos/domjudge/domjudge/statuses/$CI_COMMIT_SHA \
   -H "Authorization: token $GH_BOT_TOKEN_OBSCURED" \
   -H "Accept: application/vnd.github.v3+json" \
   -d "{\"state\": \"$STATE\", \"target_url\": \"${CI_JOB_URL/$ORIGINAL/$REPLACETO}/artifacts/$FILE\", \"description\":\"Symfony deprecations\", \"context\": \"Symfony deprecation\"}"
-cd $DIR
+section_start_collap uploadcoverage "Upload code coverage"
 set +u # Uses some variables which are not set
-. gitlab/uploadcodecov.sh
+. $DIR/gitlab/uploadcodecov.sh
+section_start uploadcoverage
```
nickygerritsen, line 43: this should be section_end i think?

nickygerritsen (changes requested):

gitlab/unit-tests.sh
mvr320: Wow, yes it should be, will fix.

nickygerritsen (approved):

PR #1069, opened 2021-04-26
Organization, team and contest images in API

nickygerritsen:
The API spec allows us to expose team photos, organization logos and the contest banner in the API. This PR adds this.

This is useful for example if you want to use the ICPC Resolver directly with DOMjudge without a CDS in between, since these images should then appear in the Resolver.

The `apiRelativeUrl` method in `DOMJudgeService` can later also be used for #1038 to have a bit less duplicate code.

Note that the tests, controllers and visitors all have a bit of duplicated code, but they are a bit too different for me to think of an easy way to unify them. Differences:
* Team always uses teamid for the filename, while organization uses the ID as exposed in the API. Contest always uses banner.png, no per contest file.
* Team and organization both need the contest AND entity ID in the API URL, while contest only needs the contest ID (but named `id` and not `cid`).
* Since contest uses a single file, we can not use the current `assetPath` method, although we could easily extend that one.

I think that if we try to unify these three we get so many if statements, it will not be that readable anymore. But thoughts are of course welcome.

### DIFF
--- a/webapp/src/Controller/API/ContestController.php
+++ b/webapp/src/Controller/API/ContestController.php
@@ -5,6 +5,7 @@ namespace App\Controller\API;
 use App\Entity\Contest;
 use App\Entity\ContestProblem;
 use App\Entity\Event;
+use App\Entity\TeamAffiliation;
 use App\Service\ConfigurationService;
 use App\Service\DOMJudgeService;
 use App\Service\EventLogService;
@@ -23,6 +24,7 @@ use Nelmio\ApiDocBundle\Annotation\Model;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
 use OpenApi\Annotations as OA;
+use Symfony\Component\HttpFoundation\BinaryFileResponse;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
@@ -104,7 +106,12 @@ class ContestController extends AbstractRestController
      *     description="Returns all contests visible to the user (all contests for privileged users, active contests otherwise)",
      *     @OA\JsonContent(
      *         type="array",
-     *         @OA\Items(ref=@Model(type=Contest::class))
+     *         @OA\Items(
+     *             allOf={
+     *                 @OA\Schema(ref=@Model(type=Contest::class)),
+     *                 @OA\Schema(ref="#/components/schemas/Banner")
+     *             }
+     *         )
      *     )
      * )
      * @OA\Parameter(ref="#/components/parameters/idlist")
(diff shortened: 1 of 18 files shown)


webapp/src/Controller/API/ContestController.php
```diff
@@ -139,6 +151,40 @@ public function singleAction(Request $request, string $cid) : Response
...
+            ->getOneOrNullResult();
+
+        if ($contest === null) {
+            throw new NotFoundHttpException(sprintf('Object with ID \'%s\' not found', $id));
+        }
+
+        $banner = sprintf('%s/public/images/banner.png', $this->dj->getDomjudgeWebappDir());
+
+        if (!file_exists($banner)) {
+            throw new NotFoundHttpException('contest banner not found');
```
mvr320, line 180: Capital?

webapp/tests/Unit/Controller/API/ContestControllerTest.php
```diff
@@ -0,0 +1,55 @@
...
+            'penalty_time'               => 20,
+            'start_time'                 => '2020-01-01T11:00:00+00:00',
+            'end_time'                   => '2023-01-01T16:00:00+00:00',
+            'duration'                   => '26309:00:00.000',
+            'scoreboard_freeze_duration' => '1:00:00.000',
+            'id'                         => '2',
+            'external_id'                => 'demo',
+            'name'                       => 'Demo contest',
+            'shortname'                  => 'demo',
+            'banner'           => [
```
mvr320, line 21: indentation,

thijskh:
> The apiRelativeUrl method in DOMJudgeService can later also be used for #1038 to have a bit less duplicate code.

Will you fix that?

> Note that the tests, controllers and visitors all have a bit of duplicated code, but they are a bit too different for me to think of an easy way to unify them. Differences:

One way is to unify them more on our side. Maybe it actually makes sense to not use `banner.png` but `banner_<cid>.png` so you can host multiple contests on one install? Or is this too much overhead? Maybe it's also not worth it.

About the returned files. I think it would be excellent it by default these come with e.g. an ETag and last-modified header and ideally respect if-modified-since. Are any of these easy to achieve by leveraring Symfony features without too much custom code?

nickygerritsen:
> Will you fix that?
> 

I will do this.

> One way is to unify them more on our side. Maybe it actually makes sense to not use `banner.png` but `banner_<cid>.png` so you can host multiple contests on one install? Or is this too much overhead? Maybe it's also not worth it.
> 

Also not sure if this is worth it...

> About the returned files. I think it would be excellent it by default these come with e.g. an ETag and last-modified header and ideally respect if-modified-since. Are any of these easy to achieve by leveraring Symfony features without too much custom code?

Last-Modified was already sent. ETag is now also being sent and we verify the tags. Was indeed very simple.

nickygerritsen:
> > Will you fix that?
> 
> I will do this.
>
Fixed this while rebasing

nickygerritsen:
Changed URLs:
api_s_doc

meisterT:
What's the state of this?

nickygerritsen:
> What's the state of this?

awaiting approval. Should be done unless you see something

nickygerritsen:
Oh there is a conflict again I see. Will fix that

webapp/tests/Unit/Controller/API/BaseTest.php
```diff
@@ -185,6 +203,11 @@ public function testListContestNotFound()
         if (($apiEndpoint = $this->apiEndpoint) === null) {
             static::markTestSkipped('No endpoint defined');
         }
+        if ($apiEndpoint === 'contests') {
+            // We can not test this, since /contests always exists. Assert that true is true to not make the test risky
```
meisterT, line 207: :-D

meisterT (approved):

PR #1075, opened 2021-05-07
Remove shell escape chars as they are passed different to tail

mvr320:
image

The running man is without these and looks fine, also included a copy of the line when the escape chars are removed.

I'm not sure if the running man should be escaped (and that the escape itself should also be fixed) or if this is in general a good solution. It works in my terminal but I'm not sure if this is correct.

### DIFF
--- a/lib/lib.error.php
+++ b/lib/lib.error.php
@@ -64,7 +64,11 @@ function logmsg(int $msglevel, string $string)
             fflush(STDLOG);
         }
         if (defined('SYSLOG')) {
-            syslog($msglevel, $string . "\n");
+            // Remove the ANSI escape characters,
+            // When logging direct to the terminal the checkmark should be green, this
+            // does not work in log so remove this specific control char.
+            // This removes both the start and the reset of the color.
+            syslog($msglevel, preg_replace('# ?\\x1b\[[0-9];?[0-9]*m#', '', $string) . "\n");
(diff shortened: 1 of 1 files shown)


meisterT:
Is your screenshot from a terminal or the log file?

mvr320:
> Is your screenshot from a terminal or the log file?

its me ssh'ng from tilix into my judge and doing a `tail -f` on the /var/log/domjudge.log

meisterT:
Ok, I expected something like this.

So we should only write these to the terminal, but not to the file.

I don't have access to my laptop right now, but I can look into this next weekend

mvr320:
@meisterT For me this is fixed with my last commits, but I don't need it for Wednesday so next weekend is fine by me.

lib/lib.error.php
```diff
@@ -64,7 +64,7 @@ function logmsg(int $msglevel, string $string)
             fflush(STDLOG);
         }
         if (defined('SYSLOG')) {
-            syslog($msglevel, $string . "\n");
+            syslog($msglevel, preg_replace('# ?\\x1b\[[0-9];?[0-9]*m#', '', $string) . "\n");
```
meisterT, line 67: Please leave a comment here that explains the idea of the regex.

meisterT (approved):

PR #1089, opened 2021-05-14
Exclude Migrations/Command from coverage

mvr320:
Migrations is write once code so doesn't benefit from coverage.
For commands the database ends up in a lock with the current DB
structure for what I tested.

### DIFF
--- a/webapp/phpunit.xml.dist
+++ b/webapp/phpunit.xml.dist
@@ -29,6 +29,9 @@
     <filter>
         <whitelist processUncoveredFilesFromWhitelist="true">
             <directory suffix=".php">src</directory>
+            <exclude>
+                <directory suffix=".php">src/Migrations</directory>
+            </exclude>
         </whitelist>
     </filter>
 


nickygerritsen:
> Migrations is write once code so doesn't benefit from coverage.
> For commands the database ends up in a lock with the current DB
> structure for what I tested.

I don’t follow how command locks the db. Unit tests do work, don’t they? Also at some point I do like to cover commands

mvr320:
> 
> 
> > Migrations is write once code so doesn't benefit from coverage.
> > For commands the database ends up in a lock with the current DB
> > structure for what I tested.
> 
> I don’t follow how command locks the db. Unit tests do work, don’t they? Also at some point I do like to cover commands

I tried this for the reset password command, and it always ends up with a database lock as the transaction is reverted in the unit tests. I might have done this wrongly ofcourse.

Shall I take the command part out of the commit or do you also disagree with the migrations?

nickygerritsen:
Migration is fine

Issue #1096, opened 2021-05-20
Rejudge remaining broken


mvr320:
Currently when running a competition in master:

When doing a partial judging, when you press the judge remaining (so when there is already a non correct verdict) there is a database error.

thijskh:
This seems to be fixed in main?

nickygerritsen:
I believe you fixed this in 1f38d4beb465b754c2549a35d35ea7d68414dd05

mvr320:
I agree, will see if I finish the tests for this.

thijskh:
> I believe you fixed this in 1f38d4b

No, that is rejudging, this is about judging remaining runs of a submission if it has stopped judging after finding e.g. a wrong-answer. But in either case, it's fixed.

PR #1097, opened 2021-05-21
[BugFix] alter judgetask to do a full judging

mvr320:
closes https://github.com/DOMjudge/domjudge/issues/1096

### DIFF
--- a/webapp/tests/Unit/Controller/Jury/SubmissionControllerTest.php
+++ b/webapp/tests/Unit/Controller/Jury/SubmissionControllerTest.php
@@ -20,7 +20,7 @@ class SubmissionControllerTest extends BaseTest
     }
 
     /**
-     * Test the filtered views do throw server errors
+     * Test the filtered views have correct queries
      *
      * @dataProvider provideViews
      */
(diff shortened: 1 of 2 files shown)


meisterT:
I am not sure the fix is what we want.

The column hostname is gone since 0036cdaf8454b164ea9f677dd217a1000de2d072 (cc @nickygerritsen) but it didn't get deleted but was replaced with judgehost. I /assume/ but didn't double check that we probably want to change the condition to check for hostname. I'll look more closely later today.

mvr320:
> 
> 
> I am not sure the fix is what we want.
> 
> The column hostname is gone since [0036cda](https://github.com/DOMjudge/domjudge/commit/0036cdaf8454b164ea9f677dd217a1000de2d072) (cc @nickygerritsen) but it didn't get deleted but was replaced with judgehost. I /assume/ but didn't double check that we probably want to change the condition to check for hostname. I'll look more closely later today.

Ok, I wrongly assumed that this was part of the API changes, I'll first make some tests for this and see if I can alter the code to set the hostname.

webapp/tests/Unit/Controller/Jury/SubmissionControllerTest.php
```diff
@@ -38,4 +38,20 @@ public function provideViews(): Generator
             }
         }
     }
+
+    /*
+     * When lazy evaluation is enabled testcases are evaluated until a non-correct
```
meisterT, line 43: Remove this? Or fix?

mvr320:
As this is now purely about the tests and I still have a lot of those small branches with tests I'll do this in a different PR.

PR #1100, opened 2021-05-25
Show submit button on problem set page. Fixes #1098.

### LINKED ISSUE #1098: Feature request: reordering the submission steps
I am proposing to reorder the steps tied to the green Submit button. From my point of view, the steps should be

1. Problem.
2. Source files.
3. Language.

Currently, it is Source files, Problem, Language.

Steps 2 and 3 can be swapped if needed but it would be helpful if Step 1 is to select the problem. 
Feel free to comment on this.

nickygerritsen:
Screenshots:
image
image

Note that I changed the order of the problem / language autodetect javascript logic, making it that we still select a language if a problem is already populated, but not the other wat around anymore. Maybe we should just have 2 ifs there, that we update both fields separately?

### DIFF
--- a/webapp/templates/team/problems.html.twig
+++ b/webapp/templates/team/problems.html.twig
@@ -7,6 +7,7 @@
         problem_text_path: 'team_problem_text',
         problem_sample_testcase_path: 'team_problem_sample_testcase',
         problem_attachment_path: 'team_problem_attachment',
-        problem_sample_zip_path: 'team_problem_sample_zip'
+        problem_sample_zip_path: 'team_problem_sample_zip',
+        show_submit_button: true
     } %}
 {% endblock %}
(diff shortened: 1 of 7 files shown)


webapp/templates/team/problems.html.twig
```diff
@@ -7,6 +7,7 @@
         problem_text_path: 'team_problem_text',
         problem_sample_testcase_path: 'team_problem_sample_testcase',
         problem_attachment_path: 'team_problem_attachment',
-        problem_sample_zip_path: 'team_problem_sample_zip'
+        problem_sample_zip_path: 'team_problem_sample_zip',
+        show_submit_button: true
```
thijskh, line 11: Why is this always true? Should it only be true when the contest is started and not ended?

nickygerritsen: It is only true for teams, not for public (which reuses the same Twig snippet that we include).
The whole problemset page is not shown if the contest is not running (and if you go there you get an empty page)

thijskh: Fair enough

thijskh (approved):

PR #1101, opened 2021-05-29
Double check to prevent $properties being null in problem importing

yang-er:
Double check to prevent $properties being null.

If we upload an archive with `domjudge-problem.ini` like

```
name = True Friends
timelimit = 1.0
```

without the quotation marks, the token `True` will be parsed as boolean. Then the `Friends` will be treated as an invalid token, after which the parse_ini_string function gives a warning and returns null. So that `$problemProperties` will be null.

image

When checking external id, `array_key_exists` would say parameter 2 "array expected, null given".

### DIFF
--- a/webapp/src/Service/ImportProblemService.php
+++ b/webapp/src/Service/ImportProblemService.php
@@ -127,7 +127,7 @@ class ImportProblemService
 
         // Double check to prevent $properties being null, which may throw at
         // array_key_exists saying parameter 2 "array expected, null given".
-        if ($propertiesString !== false && $properties === null) {
+        if ($propertiesString !== false && $properties === false) {
             $properties   = [];
             $messages[]   = "The given <code>domjudge-problem.ini</code> is invalid, ignoring.";
         }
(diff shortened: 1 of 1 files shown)


thijskh (commented):
Thanks! Indeed a useful improvement. I'd like to suggest to decomplicate the code a bit though. Now line 126 checks for propertiesString being false, then sets properties to empty array, and your newly added if statement also does this.

Maybe rewrite  from line 126 along these lines:
* $properties = []:
* if ($propertiesString !== false) {
**  call the parse function
** return message if result is false, otherwise assign to $properties
}

What do you think?

yang-er:
@thijskh Have changed the code. However, the CI failed with `visual_compare` item, which is not related to this PR, I think.

mvr320:
> 
> 
> @thijskh Have changed the code. However, the CI failed with `visual_compare` item, which is not related to this PR, I think.

You should probably rebase against the main branch. As you don't have a feature which was implemented in main.

thijskh (approved):

PR #1107, opened 2021-06-10
remove redundant access check in scoreboard API

### LINKED ISSUE #1106: Team Member role cannot query scoreboard from api for private contest she has access to
A user with only the Team Member role has access to public contests and contests her team or team category has access to.
`/api/v4/contests/{cid}/scoreboard` should be able to query the scoreboard for a contest the user has access to even when the contest is not public.

In `webapp/src/Controller/API/ScoreboardController.php` access is checked as follows
```php
$inactiveAllowed = $this->isGranted('ROLE_API_READER');
$accessAllowed   = ($inactiveAllowed && $contest->getEnabled()) || (!$inactiveAllowed && $contest->isActive());
if (!$accessAllowed) {
    throw new AccessDeniedHttpException();
}
```
Meaning, when the user does not have the api reader role, the contest has to be active which requires the contest to be public.
```php
public function isActive(): bool
{
    return $this->getEnabled() &&
        $this->getPublic() &&
        ($this->deactivatetime == null || $this->deactivatetime > time());
}
```
I think in this context `isActive` should not require the contest to be public.

chistopher:
fixes #1106.

The removed checks are done implicitly by calling `getContestId()`.

I don't know enough about domjudge to add a specific test for this, though :(

### DIFF
--- a/webapp/src/Controller/API/ScoreboardController.php
+++ b/webapp/src/Controller/API/ScoreboardController.php
@@ -125,12 +125,8 @@ class ScoreboardController extends AbstractRestController
         }
 
         /** @var Contest $contest */
-        $contest         = $this->em->getRepository(Contest::class)->find($this->getContestId($request));
-        $inactiveAllowed = $this->isGranted('ROLE_API_READER');
-        $accessAllowed   = ($inactiveAllowed && $contest->getEnabled()) || (!$inactiveAllowed && $contest->isActive());
-        if (!$accessAllowed) {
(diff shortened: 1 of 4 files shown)


nickygerritsen:
Thanks! I will see if I can add the tests before merging.

nickygerritsen:
I added the test and verified locally. Let's wait for CI and if it is green I will merge.

nickygerritsen:
Thanks @chistopher ! It seems to all work

PR #1109, opened 2021-06-11
Support contest award configuration and rendering on scoreboard

### LINKED ISSUE #1102: Consider displaying award information on the scoreboard?
In the competition held for our school, it is required that the top 13 be the gold medal, the top 20 be the silver medal, and the top 40 be the bronze medal. In order to increase the look and feel of the rankings, I modified the `scoreboard.twig` file to realize the three colors of gold, silver and bronze for the template winning teams in the first column of the ranking, as shown in the figure below. 

As we all know, there are also gold awards, silver awards and bronze awards in the algorithm competition (especially in ICPC, there are 10% teams for gold, 20% teams for silver, 30% teams for bronze). Why not consider adding this feature to domjudge?

shuibinlong:
Feature for fixes #1102.

Changes:
- Add award attributes for contest: these attributes could be configured in jury's contest add/edit page.
    -  `process_awards`: Is to process awards for this contest
    -  `contest_team_category_for_awards`: Only teams belonging to these categories can be awarded
    - `gold_awards`: number of gold medals for awards
    - `silver_awards`: number of silver medals for awards
    -  `bronze_awards`: number of bronze medals for awards
- Render awards' color on scoreboard if `process_awards=True`

Here are some Pictures for these feature:
- Awards in scoreboard

- Configuration in jury pages

### DIFF
--- a/webapp/src/Entity/Contest.php
+++ b/webapp/src/Entity/Contest.php
@@ -173,7 +173,7 @@ class Contest extends BaseApiEntity
     /**
      * @var boolean|null
      * @ORM\Column(type="boolean", name="process_awards",
-     *     options={"comment"="Are there awards for this contest?","default"=0},
+     *     options={"comment"="Whether to process awards for this contest","default"=0},
      *     nullable=false)
      * @Serializer\Exclude()
      */
@@ -187,12 +187,12 @@ class Contest extends BaseApiEntity
      *               )
      * @Serializer\Exclude()
      */
-    private $awardsCategories;
+    private $awards_categories;
 
     /**
      * @var int|null
      * @ORM\Column(type="smallint", length=3, name="gold_awards",
-     *     options={"comment"="Number of gold medals","unsigned"="true","default"=0},
+     *     options={"comment"="Number of gold awards","unsigned"="true","default"=0},
      *     nullable=false)
      * @Serializer\Exclude()
      */
@@ -201,7 +201,7 @@ class Contest extends BaseApiEntity
     /**
      * @var int|null
      * @ORM\Column(type="smallint", length=3, name="silver_awards",
-     *     options={"comment"="Number of silver medals","unsigned"="true","default"=0},
+     *     options={"comment"="Number of silver awards","unsigned"="true","default"=0},
      *     nullable=false)
      * @Serializer\Exclude()
      */
@@ -210,7 +210,7 @@ class Contest extends BaseApiEntity
     /**
      * @var int|null
      * @ORM\Column(type="smallint", length=3, name="bronze_awards",
-     *     options={"comment"="Number of bronze medals","unsigned"="true","default"=0},
+     *     options={"comment"="Number of bronze awards","unsigned"="true","default"=0},
      *     nullable=false)
      * @Serializer\Exclude()
      */
@@ -387,7 +387,7 @@ class Contest extends BaseApiEntity
         $this->submissions      = new ArrayCollection();
         $this->internal_errors  = new ArrayCollection();
         $this->team_categories  = new ArrayCollection();
-        $this->awardsCategories = new ArrayCollection();
+        $this->awards_categories = new ArrayCollection();
     }
 
     public function getCid(): int
@@ -688,10 +688,9 @@ class Contest extends BaseApiEntity
      *
      * @return Contest
      */
-    public function setProcessAwards($processAwards)
+    public function setProcessAwards(bool $processAwards): Contest
     {
         $this->processAwards = $processAwards;
-
         return $this;
     }
 
@@ -700,7 +699,7 @@ class Contest extends BaseApiEntity
      *
      * @return boolean
      */
-    public function getProcessAwards()
+    public function getProcessAwards(): bool
     {
         return $this->processAwards;
     }
@@ -710,22 +709,22 @@ class Contest extends BaseApiEntity
      */
     public function getAwardsCategories(): Collection
     {
-        return $this->awardsCategories;
+        return $this->awards_categories;
     }
 
-    public function addAwardsCategories(TeamCategory $awardsCategory): self
+    public function addAwardsCategories(TeamCategory $awardsCategory): Contest
     {
-        if (!$this->awardsCategories->contains($awardsCategory)) {
-            $this->awardsCategories[] = $awardsCategory;
+        if (!$this->awards_categories->contains($awardsCategory)) {
+            $this->awards_categories[] = $awardsCategory;
         }
 
         return $this;
     }
 
-    public function removeAwardsCategories(TeamCategory $awardsCategory): self
+    public function removeAwardsCategories(TeamCategory $awardsCategory): Contest
     {
-        if ($this->awardsCategories->contains($awardsCategory)) {
-            $this->awardsCategories->removeElement($awardsCategory);
+        if ($this->awards_categories->contains($awardsCategory)) {
+            $this->awards_categories->removeElement($awardsCategory);
         }
 
         return $this;
@@ -738,10 +737,9 @@ class Contest extends BaseApiEntity
      *
      * @return Contest
      */
-    public function setGoldAwards($goldAwards)
+    public function setGoldAwards(int $goldAwards): Contest
     {
         $this->goldAwards = $goldAwards;
-
         return $this;
     }
 
@@ -750,7 +748,7 @@ class Contest extends BaseApiEntity
      *
      * @return integer
      */
-    public function getGoldAwards()
+    public function getGoldAwards(): int
     {
         return $this->goldAwards;
     }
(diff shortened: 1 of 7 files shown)


webapp/src/Entity/Contest.php
```diff
@@ -170,6 +170,52 @@ class Contest extends BaseApiEntity
...
+     *                inverseJoinColumns={@ORM\JoinColumn(name="categoryid", referencedColumnName="categoryid", onDelete="CASCADE")}
+     *               )
+     * @Serializer\Exclude()
+     */
+    private $awardsCategories;
+
+    /**
+     * @var int|null
+     * @ORM\Column(type="smallint", length=3, name="gold_awards",
+     *     options={"comment"="Number of gold medals","unsigned"="true","default"=0},
```
nickygerritsen, line 195: Maybe we should call it also `awards` here and not `medals`?

webapp/src/Entity/Contest.php
```diff
@@ -170,6 +170,52 @@ class Contest extends BaseApiEntity
      */
     private $b = 0;
 
+    /**
+     * @var boolean|null
+     * @ORM\Column(type="boolean", name="process_awards",
+     *     options={"comment"="Are there awards for this contest?","default"=0},
```
nickygerritsen, line 176: Nitpick: `Whether to process awards for this contest` (or maybe `generate` instead of `process`)

webapp/src/Entity/Contest.php
```diff
@@ -341,6 +387,7 @@ public function __construct()
         $this->submissions      = new ArrayCollection();
         $this->internal_errors  = new ArrayCollection();
         $this->team_categories  = new ArrayCollection();
+        $this->awardsCategories = new ArrayCollection();
```
nickygerritsen, line 390: Not for you but we should really make our variables consistent in casing...

webapp/src/Entity/Contest.php
```diff
@@ -634,6 +681,128 @@ public function getProcessBalloons(): bool
...
     }
 
+    /**
+     * Set processAwards
+     *
+     * @param boolean $processAwards
+     *
+     * @return Contest
+     */
+    public function setProcessAwards($processAwards)
```
nickygerritsen, line 691: We could add `bool` as typehint here (and in the return of the method below)

webapp/src/Entity/Contest.php
```diff
@@ -634,6 +681,128 @@ public function getProcessBalloons(): bool
...
+    }
+
+    /**
+     * Set goldAwards
+     *
+     * @param integer $goldAwards
+     *
+     * @return Contest
+     */
+    public function setGoldAwards($goldAwards)
```
nickygerritsen, line 741: Same here, we could add `int` as typehint

webapp/src/Form/Type/ContestType.php
```diff
@@ -75,6 +76,35 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+            'help' => 'Whether to award medals for this contest.',
+        ]);
+        $builder->add('awardsCategories', EntityType::class, [
+            'required' => false,
+            'class' => TeamCategory::class,
+            'multiple' => true,
+            'choice_label' => function (TeamCategory $category) {
+                return $category->getName();
+            },
+            'help' => 'The categories could be awarded for this contest if process awards.',
```
nickygerritsen, line 94:
```suggestion
            'help' => 'The categories that will receive awards for this contest.',
```

webapp/templates/partials/scoreboard_table.html.twig
```diff
@@ -122,7 +142,7 @@
             {% set color = score.team.category.color %}
         {% endif %}
         <tr class="{{ classes | join(' ') }}" id="team:{{ score.team.teamid }}">
-            <td class="scorepl">
+            <td class="scorepl{{awardColor}}">
```
nickygerritsen, line 145: Maybe always add a space to the class here so we don't have to add it to the variable? Does not really hurt.

webapp/templates/partials/scoreboard_table.html.twig
```diff
@@ -114,6 +119,21 @@
             {% set previousTeam = null %}
         {% endif %}
 
+        {# process awards color #}
+        {% set awardColor = '' %}
+        {% if showLegends and processAwards and score.team.category.name in awardsCategories %}
```
nickygerritsen, line 124: Later we can add an option whether the award numbers are global (i.e. we give 4 total gold awards) or per category (i.e. we give 4 gold awards per category). For now this is the most useful behavior so I like it.

Also is the `name` needed here? Not sure

webapp/templates/partials/scoreboard_table.html.twig
```diff
@@ -358,6 +399,16 @@
                 {{ cMax }} 96%);
         }
     {% endfor %}
+
```
nickygerritsen, line 402: Why not this to the [general CSS file](https://github.com/DOMjudge/domjudge/blob/master/webapp/public/style_domjudge.css)?

nickygerritsen (changes requested):
Looks very good in general! I added some pretty small comments.

One question and two remarks:
* On your screenshot the 12th and 13th team do not have a gold medal. Why is that?
* Tests are missing but I will add them later.
* We should also update the awards API endpoint with this data. I will do this after we have merged this; we might want to use the data from that endpoint to have the logic only in one place. I will have a look at that.

shuibinlong:
@nickygerritsen because the 12th and 13th team is not in the categories which is set by `awardCategories`, these feature is designed for those teams that are not included in the list of winners and they just join these contest for other reasons, for example tourist's teams in every WF.

In addition, I will carefully read your comments on my code, and try to fix it, thanks for the review.

webapp/src/Entity/Contest.php
shuibinlong: Sure.

webapp/src/Entity/Contest.php
shuibinlong: Sure.

webapp/src/Entity/Contest.php
shuibinlong: Fine, maybe `awards_categories`?

webapp/src/Entity/Contest.php
shuibinlong: Sure.

webapp/src/Entity/Contest.php
shuibinlong: Sure

webapp/templates/partials/scoreboard_table.html.twig
shuibinlong: It seems that `awardsCategories` is an array of categories' name

webapp/templates/partials/scoreboard_table.html.twig
shuibinlong: Sound's great.

webapp/src/Entity/Contest.php
nickygerritsen: For now sure. We will check later to make it clean, we have some (like `removedIntervals`) that are ugly still.

webapp/templates/partials/scoreboard_table.html.twig
nickygerritsen: I wonder how. Anyway if it works it's fine. I will investigate further after merging.

nickygerritsen:
> @nickygerritsen because the 12th and 13th team is not in the categories which is set by `awardCategories`, these feature is designed for those teams that are not included in the list of winners and they just join these contest for other reasons, for example tourist's teams in every WF.

Makes total sense!

> In addition, I will carefully read your comments on my code, and try to fix it, thanks for the review.

Great. As said I think it looks good already!

webapp/templates/partials/scoreboard_table.html.twig
shuibinlong: The default setting is `awardColor=''`, so its length is 0, actually if a team does not win any award, there will be no spaces in this class.

webapp/templates/partials/scoreboard_table.html.twig
shuibinlong: Actually you are right. I just tested for it, team categories' name is not needed here, thank you.

webapp/templates/jury/contest.html.twig
```diff
@@ -106,6 +106,32 @@
...
+                    <td>
+                        {% if contest.processAwards %}
+                            <a>{{ contest.goldAwards|default('-') }} Gold Medal</a>
+                            <br>
+                            <a>{{ contest.silverAwards|default('-') }} Silver Medal</a>
+                            <br>
+                            <a>{{ contest.bronzeAwards|default('-') }} Bronze Medal</a>
+                            <br>
+                            {% for category in contest.awardsCategories %}
+                                For all teams from
```
nickygerritsen, line 124: I will probably move this outside the loop and only show the category names + links

nickygerritsen (approved):
I see some small things that could improve, but I will do those myself tomorrow.

Thanks for the great feature!

webapp/templates/jury/contest.html.twig
shuibinlong: Great.

PR #1111, opened 2021-06-25
Warn role user

mvr320:
As reported by Azeeto in Slack,

When someone has no roles in DOMjudge but does a login (s)he is redirected to public. This only adds a warning explaining this behaviour.

### DIFF
--- a/webapp/src/Controller/RootController.php
+++ b/webapp/src/Controller/RootController.php
@@ -45,6 +45,10 @@ class RootController extends BaseController
                 return $this->redirectToRoute('jury_balloons');
             }
         }
+        $this->addFlash(
+            'danger',
+            'Your user has no DOMjudge roles, you are deauthenticated.'
+        );
         return $this->redirectToRoute('public_index');
     }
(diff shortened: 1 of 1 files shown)


webapp/src/Controller/RootController.php
```diff
@@ -45,6 +45,10 @@ public function redirectAction(AuthorizationCheckerInterface $authorizationCheck
                 return $this->redirectToRoute('jury_balloons');
             }
         }
+        $this->addFlash(
+            'danger',
+            sprintf("User has no DOMjudge roles, user deauthenticated.)
```
nickygerritsen, line 50: Why a sprintf? Also I think you are missing a “

nickygerritsen (changes requested):

webapp/src/Controller/RootController.php
mvr320: I should indeed remove that.

nickygerritsen (approved):

PR #1112, opened 2021-07-08
Move Unit tests to dedicated folder

mvr320:
I think this was introduced from a PR which was opened before the split of E2E and Unit.


nickygerritsen (approved):

mvr320:
@nickygerritsen I had to change the tests, so I would like your re-approval on the changes.

nickygerritsen (approved):

PR #1117, opened 2021-07-16
Coverity fixes

mvr320:
I think some of these should be cherry picked and some can just be ignored,

The Migration one is in my opinion a FP and should just be labeled as such.
The one about the close I'm not sure of if this is indeed the intended behaviour but it would need better documentation in my opinion.

### DIFF
--- a/lib/lib.misc.c
+++ b/lib/lib.misc.c
@@ -231,7 +231,8 @@ void daemonize(const char *_pidfile)
 
 	/* Close all other file descriptors. */
 	maxfd = sysconf(_SC_OPEN_MAX);
-	for(fd=3; fd<maxfd; fd++) close(fd);
+	/* Failure here is expected for some fds */
+	for(fd=3; fd<maxfd; fd++) if ( close(fd)!=0 ) continue;
 
 	/* Start own process group, detached from any tty */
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

lib/lib.misc.c
```diff
@@ -231,7 +231,8 @@ void daemonize(const char *_pidfile)
 
 	/* Close all other file descriptors. */
 	maxfd = sysconf(_SC_OPEN_MAX);
-	for(fd=3; fd<maxfd; fd++) close(fd);
+	/* Failure here is expected for some fds */
+	for(fd=3; fd<maxfd; fd++) if ( close(fd)!=0 ) continue;
```
eldering, line 235: I'm not really a fan of this syntax, as it introduces additional control logic that effectively does nothing, but leaves you wondering why it's there. I'd just mark this as a "won't fix / false positive" in coverity as in cases like this there's really nothing to do to handle an error.

mvr320: I get your point, can I assume that the comment is also not needed? If so I'll close the PR.

eldering: Indeed, that would be my suggestion.

PR #1120, opened 2021-07-16
Fix CID: 1405295

mvr320:
The filedescriptor is left open when the program exits. If closing fails
the code will now only warn as trying to fix this at this point (in a
failure) makes little sense.

### DIFF
--- a/judge/runpipe.cc
+++ b/judge/runpipe.cc
@@ -226,6 +226,7 @@ void resize_pipe(int fd)
 		if ( fscanf(f, "%d", &max_pipe_size)!=1 ) {
 			max_pipe_size = -2;
 			warning(errno, "could not read from '%s'", PROC_MAX_PIPE_SIZE);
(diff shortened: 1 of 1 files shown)


judge/runpipe.cc
```diff
@@ -226,6 +226,7 @@ void resize_pipe(int fd)
 		if ( fscanf(f, "%d", &max_pipe_size)!=1 ) {
 			max_pipe_size = -2;
 			warning(errno, "could not read from '%s'", PROC_MAX_PIPE_SIZE);
+			if ( fclose(f)!=0 ) warning(errno, "failed to close pipe");
```
eldering, line 229: This should be the same warning as below: `warning(errno, "could not close '%s'", PROC_MAX_PIPE_SIZE);` as this is not a pipe, but the file `/proc/sys/fs/pipe-max-size`.

eldering (changes requested):

eldering (approved):

PR #1124, opened 2021-07-21
Fix adding team with a user after recent validation update. Fixes #1122

### LINKED ISSUE #1122: Adding user via frontend fails
```
<html><body>
<!--StartFragment-->

This collection should contain 1 element or more. | 0 | Caused by:                                                             Symfony\Component\Validator\ConstraintViolation {#2984 ▼   root: Symfony\Component\Form\Form {#610 …}   path: "data.users[0].user_roles"   value: Doctrine\Common\Collections\ArrayCollection {#1518 …} }
-- | -- | --


<!--EndFragment-->
</body>
</html>
```

nickygerritsen:
Also add tests for adding teams.

### DIFF
--- a/webapp/src/Controller/Jury/TeamController.php
+++ b/webapp/src/Controller/Jury/TeamController.php
@@ -429,20 +429,8 @@ class TeamController extends BaseController
                 // If we do not want to add a user, remove it again
                 $team->removeUser($user);
             } else {
-                // Otherwise, add the team role to it
-                /** @var Role $role */
-                $role = $this->em->createQueryBuilder()
-                    ->from(Role::class, 'r')
-                    ->select('r')
-                    ->andWhere('r.dj_role = :team')
-                    ->setParameter(':team', 'team')
-                    ->getQuery()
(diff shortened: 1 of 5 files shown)


webapp/src/Controller/Jury/TeamController.php
```diff
@@ -429,20 +429,8 @@ public function addAction(Request $request): Response
...
-                    ->from(Role::class, 'r')
-                    ->select('r')
-                    ->andWhere('r.dj_role = :team')
-                    ->setParameter(':team', 'team')
-                    ->getQuery()
-                    ->getOneOrNullResult();
-                $user->addUserRole($role);
-                $user->setTeam($team);
-                // Also set the user's name to the team name
+                // Otherwise, set the user's name to the team name
```
mvr320, line 432: Why is the user created at all? I don't think its needed for the normal add page. But thats probably not part of your PR.

mvr320 (approved):
As far as I can see it fixes the issue and improves on the tests.

PR #1129, opened 2021-07-28
Remount build directory for shared gitlab-runners

mvr320:
Some of the current gitlab runners have the /build with nodev,noexec
which breaks the chroot installation. Remounting is currently still
allowed. For debugging the mount table is added.


mvr320:
Difference can be seen here:
https://gitlab.com/DOMjudge/domjudge/-/jobs/1458541370

We start with noexec,nodev and do still get a working chroot.

nickygerritsen (approved):

PR #1132, opened 2021-07-31
Add support for import multiple accounts per team and IP addresses for users.

nickygerritsen:


misc-tools/import-accounts-multiuser.py
```diff
@@ -0,0 +1,81 @@
...
+
+import_non_team_accounts()
+
+team_regex = r"^[^0-9]*0*([0-9]+)([^0-9]*)$"
+accounts = get_team_accounts()
+for account in accounts:
+    team_data = re.search(team_regex, account["username"])
+    team_id = team_data.group(1)
+    user_suffix = team_data.group(2)
+    ip_mapping = {
```
meisterT, line 64: Where did we take the IP addresses from before this script?

misc-tools/import-accounts-multiuser.py
```diff
@@ -0,0 +1,81 @@
...
+httpie_options = sys.argv[2]
+
+# Note: this script uses httpie so we can use the same commandline flags as the base import-contest.sh script
+
+# Import all non-team accounts. We use the existing CSV import functionality for this,
+# because for jury members it creates teams and a category, which we can only do in the
+# API if we know a contest ID. We do not know one at this point
+def import_non_team_accounts():
+    with open('accounts.tsv') as csv_file:
+        with open('accounts_non_team.tsv', 'w') as csv_output:
```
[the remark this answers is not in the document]
nickygerritsen: See tyhe comment above this method for that reason. If you have a better idea it is very welcome.

misc-tools/import-accounts-multiuser.py
nickygerritsen:
```sql
UPDATE user set IP = concat('10.1.1.', team_id) WHERE team_id IS NOT NULL
```

I think we did something ugly like this.

thijskh:
Should this be in `import/`, or in `misc-tools/`?

Should it perhaps not end in `.py`?

nickygerritsen:
> Should this be in `import/`, or in `misc-tools/`?
> 

I put it in the same place as import-contest.sh 
> Should it perhaps not end in `.py`?

No opinion 🤷‍♂️

I wonder if we even keep this after wf2020. I PR’d it to that branch at least

misc-tools/import-accounts-multiuser.py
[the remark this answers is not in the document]
meisterT: Just double checking: do we actually need the contest ID in the API to create what we do create or can we get it of that restriction instead?

misc-tools/import-accounts-multiuser.py
meisterT: Can we add an optional column to accounts.tsv instead with the IP address mapping?

nickygerritsen: I can add it. What if it is not present, set no IP?

misc-tools/import-accounts-multiuser.py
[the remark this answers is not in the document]
nickygerritsen: The API requires it. There is no underlying reason. But I think it is hard to get rid of it, since we use the api/contests/xxx prefix everywhere. I can investigate if we can get rid of it but I highly doubt it. Especially for teams which can differ per contest

webapp/src/Service/ImportExportService.php
```diff
@@ -1094,7 +1094,7 @@ protected function importAccountsTsv(array $content, string &$message = null): i
                     // Note that https://ccs-specs.icpc.io/ccs_system_requirements#accountstsv
                     // assumes team accounts of the form "team-nnn" where
                     // nnn is a zero-padded team number.
-                    $teamId = preg_replace('/^[^0-9]*0*([0-9]+)$/', '\1', $line[2]);
+                    $teamId = preg_replace('/^[^0-9]*0*([0-9]+)[^0-9]*$/', '\1', $line[2]);
```
meisterT, line 1097: You probably want to update the comment above

nickygerritsen:
> Do we already have a test for this code?

Let me check but I don’t think so. I will try to add one

nickygerritsen:
> > Do we already have a test for this code?
> 
> Let me check but I don’t think so. I will try to add one

Now we do, even find a bug (we didn't set the members field for jury teams)

webapp/src/Service/ImportExportService.php
nickygerritsen: Done

meisterT (approved):
Let's merge it into main?

nickygerritsen:
> Let's merge it into main?

My plan: merge into wf2020, cherry pick into main. Agreed? Otherwise I need to rebase first this change onto main.

meisterT:
Whatever works best for you.

PR #1133, opened 2021-08-06
Store user on submission.

nickygerritsen:
Also show submissions on user page and allow to rejudge based on user.

### DIFF
--- a/webapp/src/Command/ImportEventFeedCommand.php
+++ b/webapp/src/Command/ImportEventFeedCommand.php
@@ -1720,7 +1720,7 @@ class ImportEventFeedCommand extends Command
                 // Submit the solution
                 $contest    = $this->em->getRepository(Contest::class)->find($this->contestId);
                 $submission = $this->submissionService->submitSolution(
-                    $team, $contestProblem, $contest, $language, $filesToSubmit, null,
+                    $team, null, $contestProblem, $contest, $language, $filesToSubmit, null,
                     null, $entryPoint, $submissionId, $submitTime,
                     $message
                 );
--- a/webapp/src/Controller/API/SubmissionController.php
+++ b/webapp/src/Controller/API/SubmissionController.php
@@ -234,8 +234,9 @@ class SubmissionController extends AbstractRestController
             }
         }
 
-        // By default, use the team of the user
-        $team = $this->dj->getUser()->getTeam();
+        // By default, use the user and team of the user
+        $user = $this->dj->getUser();
+        $team = $user->getTeam();
         if ($teamId = $request->request->get('team_id')) {
             $idField = $this->eventLogService->externalIdFieldForEntity(Team::class) ?? 'teamid';
             $method  = sprintf('get%s', ucfirst($idField));
@@ -269,6 +270,8 @@ class SubmissionController extends AbstractRestController
                     throw new BadRequestHttpException(
                         sprintf("Team %s not found or not enabled", $teamId));
                 }
+
+                $user = $team->getUsers()->first() ?: null;
             } elseif (!$team) {
                 throw new BadRequestHttpException(sprintf('User does not belong to a team'));
             } elseif ((string)call_user_func([$team, $method]) !== (string)$teamId) {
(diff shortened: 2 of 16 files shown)


nickygerritsen:
Changed URLs:
jury_s_rejudgings_s_add
jury_s_users_s_1
jury_s_users_s_2
jury_s_users_s_3

webapp/src/Command/ImportEventFeedCommand.php
```diff
@@ -1720,7 +1720,7 @@ protected function importSubmission(array $event)
                 // Submit the solution
                 $contest    = $this->em->getRepository(Contest::class)->find($this->contestId);
                 $submission = $this->submissionService->submitSolution(
-                    $team, $contestProblem, $contest, $language, $filesToSubmit, null,
+                    $team, null, $contestProblem, $contest, $language, $filesToSubmit, null,
```
meisterT, line 1723: That poses the question whether we need to expose the user to the secondary CCS. It's extremely unlikely that we need to act on this info anyway, but right now the secondary CCS doesn't have enough info, right?

webapp/src/Controller/API/SubmissionController.php
```diff
@@ -236,6 +236,7 @@ public function addSubmissionAction(Request $request, ?string $id): Response
 
         // By default, use the team of the user
         $team = $this->dj->getUser()->getTeam();
+        $user = $this->dj->getUser();
```
meisterT, line 239: nit: switch this with the previous line and then write it as `$team = $user->getTeam();`

webapp/src/Controller/API/SubmissionController.php
```diff
@@ -269,6 +270,8 @@ public function addSubmissionAction(Request $request, ?string $id): Response
                     throw new BadRequestHttpException(
                         sprintf("Team %s not found or not enabled", $teamId));
                 }
+
+                $user = $team->getUsers()->first() ?: null;
```
meisterT, line 274: Do we want to allow to specify a user_id as well?

meisterT (approved):

webapp/src/Command/ImportEventFeedCommand.php
nickygerritsen: Since user is not in the spec at all i wonder how we do this…

webapp/src/Controller/API/SubmissionController.php
nickygerritsen: Do we? Might make sense actually. Do we want to make sure the user is part of the team? I can add it and a test for it

meisterT: We can also do this separately from this PR

nickygerritsen: I do like the idea. I think my proposal makes sense and is easy to implement

webapp/src/Controller/API/SubmissionController.php
```diff
@@ -278,6 +282,21 @@ public function addSubmissionAction(Request $request, ?string $id): Response
...
         }
 
+        if ($userId = $request->request->get('user_id')) {
+            // If the current user is an admin or API writer, allow it to specify the user
+            if ($this->isGranted('ROLE_API_WRITER')) {
+                // Load the user
+                /** @var User|null $user */
+                $user = $this->em->getRepository(User::class)->find($userId);
+
+                if (!$user || !$user->getTeam() || $user->getTeam()->getTeamid() !== $team->getTeamid()) {
```
meisterT, line 292: I know it's more code, but it would be more user friendly to expose which of these conditions failed

meisterT (approved):

webapp/src/Controller/API/SubmissionController.php
nickygerritsen: I thought it more a security leak. Or do you think it does not matter? It's easy to add

meisterT: I don't think there's any info worth protecting in here.

PR #1137, opened 2021-08-13
Remove JudgehostRestriction feature

mvr320:
This is to trigger the pipeline, I think a quick grep for restrictions will only show other restrictions which are another type of restrictions.


mvr320:
Changed URLs:
jury
jury_s_change-contest_s_-1
jury_s_judgehosts
jury_s_judgehosts_s_1
jury_s_judgehosts_s_1_s_activate
jury_s_config_s_check
jury_s_judgehosts_s_edit_s_multiple
jury_s_judgehosts_s_edit_s_multiple@include_hidden=0

webapp/src/Migrations/Version20210813195818.php
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public function up(Schema $schema) : void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE judgehost DROP FOREIGN KEY judgehost_ibfk_1');
+        $this->addSql('DROP TABLE judgehost_restriction');
+        $this->addSql('DROP INDEX restrictionid ON judgehost');
```
thijskh, line 27: What is the “change polltime”?

thijskh (changes requested):
I think I’m missing a docs update.

mvr320:
> 
> 
> I think I’m missing a docs update.

You mean the mention that the feature was removed? I'll add that.

webapp/src/Migrations/Version20210813195818.php
mvr320: That should not be there, its part of the doctrine:migrations:diff wanting the change the whole DB.

thijskh:
I meant just removing the chapter in the manual. I think the manual is not the place for upgrade notes.

thijskh (approved):
Good work. Ready to merge when build passes and commits are squashed

webapp/src/Migrations/Version20210813195818.php
nickygerritsen: Our dev docs state how to fix this (set a car in .env.local)

PR #1142, opened 2021-08-16
Replace C++ submit client with fully standalone submit client.

nickygerritsen:


doc/manual/install-workstation.rst
```diff
@@ -19,22 +19,25 @@ Command line submit client
 DOMjudge comes with a command line submit client which makes it really
 convenient for teams to submit their solutions to DOMjudge.
 
-In order to build the submit client, you need libcURL, libJSONcpp and
-optionally libmagic. To install this on Debian-like distributions::
+In order to use the submit client, you need python 3, the python requests 
+library and optionally the python magic library. To install this on
```
thijskh, line 23: Add here at end of sentence: "installed on the team's workstation"

doc/manual/install-workstation.rst
```diff
@@ -19,22 +19,25 @@ Command line submit client
...
-  sudo yum install libcurl-devel jsoncpp-devel file-devel
+  sudo yum install python3 python3-requests python3-magic
 
-Then run (adapt the URL to your environment; look at the Config check
-in the admin web interface if unsure)::
+You can now copy this client from ``submit/submit`` to the workstations.
 
-  ./configure --enable-static-linking --with-baseurl="https://yourhost.example.edu/domjudge"
-  make submitclient
+To configure the baseurl used for the submit client you have four options:
```
thijskh, line 34: Reword as: "The submit client needs to know the base URL of the domserver where it should submit to. You have four options to configure this:"

doc/manual/install-workstation.rst
```diff
@@ -19,22 +19,25 @@ Command line submit client
...
 
-You can now copy this client from ``submit/submit`` to the workstations.
+* Let teams pass it using the ``--url`` argument.
```
thijskh, line 36: Make it the lowest opton

doc/manual/install-workstation.rst
```diff
@@ -19,22 +19,25 @@ Command line submit client
...
+* Set it as an environment variable called ``SUBMITBASEURL``.
+* Use the ``submit_wrapper.sh`` script and set ``SUBMITBASEURL`` there.
```
thijskh, line 38: This is no longer a useful thing to have since if you can edit the wrapper you can edit the client itself. git rm submit_wrapper and change "four" in suggested text above to "three"

doc/manual/install-workstation.rst
```diff
@@ -19,22 +19,25 @@ Command line submit client
...
-in the admin web interface if unsure)::
+You can now copy this client from ``submit/submit`` to the workstations.
 
-  ./configure --enable-static-linking --with-baseurl="https://yourhost.example.edu/domjudge"
-  make submitclient
+To configure the baseurl used for the submit client you have four options:
 
-You can now copy this client from ``submit/submit`` to the workstations.
+* Let teams pass it using the ``--url`` argument.
+* Set it as an environment variable called ``SUBMITBASEURL``.
```
thijskh, line 37: .. "e.g. in ``/etc/profile.d/``"?

submit/submit_online.bats
```diff
@@ -41,7 +42,7 @@ setup() {
 }
 
 @test "binary file emits warning" {
-    cp ./submit $BATS_TMPDIR/binary.c
+    cp $(which bash) $BATS_TMPDIR/binary.c
```
thijskh, line 45: Too bad this nice self referencing hack now disappears :-)

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+import argparse, sys, logging, os, stat, requests, json, time
+try:
+    import magic
+except ModuleNotFoundError:
+    # Ignore, magic is optional
+    magic = None
+
+# By default no base URL is set and it is expected to use either the
+# SUBMITBASEURL environment variable or the -u/--url argument.
+# If you want to hardcode a default base URL, set it here
```
thijskh, line 19: Reword as:

# Set the default base URL to submit to (optional). It can be overridden by the
# SUBMITBASEURL environment variable or the -u/--url argument.

thijskh: Love the rendering of this comment, but intended was:
```python
# Set the default base URL to submit to (optional). It can be overridden by the
# SUBMITBASEURL environment variable or the -u/--url argument.
```

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+baseurl = ''
+
+# Use a specific API version, set to empty string for default
+api_version = 'v4/'
```
thijskh, line 23: So why is this string not empty by default?

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+
+# Last modified time in minutes after which to warn for submitting
+# an old file.
+warn_mtime = 5
+
+num_warnings = 0
```
thijskh, line 29: Add "# End of configurable settings" text above to make it clear that num_warnings is not also something you can tweak

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+    verbosity = 'ERROR'
+
+logging.basicConfig(format='[%(asctime)s] {}[{}]: %(message)s'.format(sys.argv[0], os.getpid()), datefmt='%b %d %H:%M:%S', level=loglevels[verbosity])
+
+contest_id = ''
+
+if 'SUBMITBASEURL' in os.environ:
+    baseurl = os.environ['SUBMITBASEURL']
+if 'SUBMITCONTEST' in os.environ:
+    contest_id = os.environ['SUBMITCONTEST']
```
thijskh, line 360: Undocmented feature?

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+    try:
+        st = os.stat(filename)
+    except FileNotFoundError:
+        usage("Can not find file `{}'".format(filename))
+    
+    logging.debug("submission file {}: `{}'".format(index + 1, filename))
+
+    # Do some checks on submission file and warn user
+    if not stat.S_ISREG(st.st_mode):
+        warn_user("`{}' is not a regular".format(filename))
```
thijskh, line 443: is not a regular _file_

thijskh (approved):

thijskh:
I think this already brings us quite a bit closer to resolving the current installability problems btw.

doc/manual/install-workstation.rst
```diff
@@ -19,22 +19,25 @@ Command line submit client
 DOMjudge comes with a command line submit client which makes it really
 convenient for teams to submit their solutions to DOMjudge.
 
-In order to build the submit client, you need libcURL, libJSONcpp and
-optionally libmagic. To install this on Debian-like distributions::
+In order to use the submit client, you need python 3, the python requests
```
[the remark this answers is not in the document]
nickygerritsen: Sure. Although the packages in the lines below are still called python3

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+
+def read_problems() -> list:
+    '''Read all problems for the current contest from the API
+    
+    Returns:
+        The problems or None if an error occurred
+    '''
+
+    try:
+        endpoint = 'contests/' + my_contest['id'] + '/problems'
```
mvr320, line 134: I think you can also use `format` here,

submit/submit_online.bats
nickygerritsen: Yeah. I first tried using bats but that is not a binary 😅

submit/submit
nickygerritsen: Copied it from the c++ version. I think we can surely leave it empty. Let me test

submit/submit
nickygerritsen: Bats uses it. Should we document it somewhere?

thijskh: I guess? I'm a fan of either having a feature and documenting it or not having it at all.

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+        return None
+
+    if not isinstance(data, list):
+        logging.warning('REST API returned unexpected JSON data for endpoint contests')
+        return None
+
+    contests = []
+
+    for contest in data:
+        if ('id' not in contest or 'shortname' not in contest or contest['id'] == ''
```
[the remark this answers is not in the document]
meisterT: Consider using `not` for checking whether strings are empty.

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+            or contest['shortname'] == ''):
+            logging.warning('REST API returned unexpected JSON data for contests')
+            return None
```
meisterT, line 79: So one contest didn't contain all fields and we're not returning any data?

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+        return None
+
+    if not isinstance(data, list):
+        logging.warning('REST API returned unexpected JSON data for endpoint languages')
+        return None
+
+    languages = []
+
+    for item in data:
+        if ('id' not in item or 'extensions' not in item
```
meisterT, line 107: Same as above with () and line breaks

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+            or item['id'] == '' or not isinstance(item['extensions'], list)
```
meisterT, line 108: and None

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+            or len(item['extensions']) == 0):
+            logging.warning('REST API returned unexpected JSON data for languages')
+            return None
+        language = {
+            'id': item['id'],
+            'name': item['name'],
+            'entry_point_required': item['entry_point_required'] or False,
+            'extensions': {item['id']}
```
meisterT, line 116: Just on the phone, so can't check, but does it make sense to have a dict here instead of a list.?

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+        return None
+
+    languages = []
+
+    for item in data:
+        if ('id' not in item or 'extensions' not in item
+            or item['id'] == '' or not isinstance(item['extensions'], list)
+            or len(item['extensions']) == 0):
+            logging.warning('REST API returned unexpected JSON data for languages')
+            return None
```
meisterT, line 111: Same question as above

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+        return None
+
+    if not isinstance(data, list):
+        logging.warning('REST API returned unexpected JSON data for endpoint problems')
+        return None
+
+    problems = []
+
+    for problem in data:
+        if ('id' not in problem or 'label' not in problem
```
meisterT, line 147: Same things as above

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+
+    Returns:
+        The endpoint contents
+
+    Raises:
+        RunTimeError when the response is not JSON or the HTTP status code is non 2xx
+    '''
+
+    if baseurl == '':
+        raise RuntimeError('No baseurl set')
```
meisterT, line 171: Suggest passing the option or permanently setting it in the environment

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+
+    try:
+        response = requests.get(url)
+    except requests.exceptions.ConnectionError as e:
+        raise RuntimeError(e)
+
+    if response.status_code >= 300:
+        print(response.text)
+        if response.status_code == 401:
+            raise RuntimeError('authentication failed, please check your DOMjudge credentials.')
```
meisterT, line 185: Where/how

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+        contests_part_one = '''For CONTEST use the ID or short name as shown in the top-right contest
+selection box in the webinterface.'''
+        if contests is not None and len(contests) == 1:
+            contests_part_two = "Currently this defaults to the only active contest '{}'".format(contests[0]['shortname'])
+    else:
+        contests_part_one = 'For CONTEST use one of the following:'
+        for contest in contests:
+            contests_part_one += '\n    {:<10} - {}'.format(contest['shortname'], contest['name'])
+
+    if problems is None or len(problems) == 0:
```
meisterT, line 225: `if not problems`

Same below

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+    data = {
+        'problem': my_problem['id'],
+        'language': my_language['id'],
+    }
+    if entry_point is not None:
+        data['entry_point'] = entry_point
+
+    files = []
+    for filename in filenames:
+        files.append(('code[]', open(filename, 'rb')))
```
meisterT, line 281: A list comprehension would be more pythonic here

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+        error('parsing REST API output: {}'.format(e))
+
+    if (not isinstance(submission, dict)
+        or not 'id' in submission
+        or not isinstance(submission['id'], str)):
+        error('REST API returned unexpected JSON data')
+
+    print('Submission received, id = s{}'.format(submission['id']))
+
+version_text = '''
```
meisterT, line 312: I don't see any version info here. Should it contain any?

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+problem_id = args.problem or ''
+language_id = args.language or ''
+entry_point = args.entry_point
+if args.url:
+    baseurl = args.url
+
+logging.info('set verbosity to {}'.format(verbosity))
+
+# Make sure that baseurl terminates with a '/' for later concatenation.
+if baseurl != '' and baseurl[-1:] != '/':
```
meisterT, line 373: Does it make sense to continue at all if baseurl is not set after this point?

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+contests = read_contests()
+if contests is None:
+    logging.warning('could not obtain active contests')
+
+my_contest = None
+my_language = None
+my_problem = None
+
+if contest_id == '':
+    if contests is None or len(contests) == 0:
```
meisterT, line 385: `if not contests`

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+    languages = read_languages()
+    problems = read_problems()
+
+parser.epilog = get_epilog()
+
+if args.help:
+    parser.print_help()
+    exit(0)
+
+if baseurl == '':
```
meisterT, line 410: See my comment above. Move this up as much as possible

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+    usage('no url specified')
+
+if my_contest is None:
```
meisterT, line 413: Move up to the opposite check

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+    usage('no (valid) contest specified')
+
+if languages is None:
+    logging.warning('could not obtain language data')
+
+if problems is None:
+    logging.warning('could not obtain language data')
```
meisterT, line 420: Problem data

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+problem_id = problem_id.lower()
+for problem in problems:
+    if problem['id'].lower() == problem_id or problem['label'].lower() == problem_id:
+        my_problem = problem
+        break
+
+if my_problem is None:
+     usage('no known problem specified or detected')
+
+if my_language is None:
```
meisterT, line 494: Move up before problem checks

doc/manual/install-workstation.rst
nickygerritsen: I will add this (here?). Indeed command line option overrides the env var.

submit/submit
```diff
@@ -0,0 +1,534 @@
+#!/usr/bin/env python3
+
+#
+# submit -- command-line submit program for solutions.
```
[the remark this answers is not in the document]
nickygerritsen: Copy-paste from C++ version but agreed

submit/submit
```diff
@@ -0,0 +1,534 @@
...
+def confirm(message: str) -> bool:
+    answer = ''
+    while answer not in ['y', 'n']:
+        answer = input(' (y/n) '.format(message)).lower()
+    return answer == 'y'
+
+def warn_user(msg: str):
+    global num_warnings
+    if args.quiet:
+        logging.debug('user warning #{}: {}'.format(num_warnings, msg))
```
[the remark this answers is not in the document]
nickygerritsen: Did not know this was a thing. So much easier

submit/submit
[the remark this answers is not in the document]
nickygerritsen: Will add more breaks. It is already two lines.

What do you mean with `not`? Does `not contest['shortname']` work for empty strings (and None)?

submit/submit
nickygerritsen: This is how we did it in the C++ version. Do you think we should skip it and use the others?

doc/manual/install-workstation.rst
mvr320: That may be checked in the bats test.

submit/submit
nickygerritsen: It's a set right, not a dict? At least that is what I found about `{item}`. I do this to make sure the items are unique (since we insert both the ID and extension). Can change it to a list but then we need to make it unique.

submit/submit
nickygerritsen: Or f-strings, TIL

submit/submit
nickygerritsen: It was there before, but we dropped it to make the submit client standalone. However, we still pass it to the `--version` flag, so that is why this variable is named like this

doc/manual/install-workstation.rst
thijskh: I think this just "does what you'd expect" of any commandline tool

submit/submit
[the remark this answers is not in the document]
meisterT: Yep, it does. See https://docs.python.org/3/library/stdtypes.html#truth-value-testing

submit/submit
nickygerritsen: The main problem with this, and a lot of your comments below: the 'epilog' for the argument list is calculated using variables we only determine below. If we would stop here, you can't render the epilog so `--help` would show way less. Is that something that is acceptable? Note: this is how the C++ version did it, but I am totally fine with changing this to make the flow better.

submit/submit
nickygerritsen: Same as above for the epilog. We need get_epilog before this call, but currently need get_epilog AFTER reading problems and languages. Any suggestion?

submit/submit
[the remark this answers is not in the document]
nickygerritsen: You saw my comment too quickly 😅 . I tested it and found this out as well

submit/submit
meisterT: Yep

submit/submit
meisterT: Ah, ok, yeah that works

submit/submit
meisterT: Do we want to give this a separate version number  then?

cc @thijskh

submit/submit
meisterT: Ok, let's keep it like this and I'll stare hard at it when it's committed

submit/submit
thijskh: If the contest does not contain the required data in the JSON structure something is terribly broken, right? Better to just bail than continue siliently?

nickygerritsen: I am fine with both, but indeed it would be strange if one contest does contain the correct fields and the other does not

submit/submit
thijskh: We could hardcode a version number.
Or we can insert a version number/git commit at tarball build time instead of at runtime.
Or just leave it...

nickygerritsen: I will leave it for now and let someone else decide and implement this

thijskh:
Just to document. The previous submit client made some effort to reuse the curl connections for the series of API queries it does. Requests has a built in pooling system that should do this automatically, so the new client should be fine in this regard.

meisterT (approved):

mvr320:
> This pull request **introduces 1 alert** when merging [e052137](https://github.com/DOMjudge/domjudge/commit/e052137757ef1d26b2bffef0058fee3604afa0ac) into [467a06f](https://github.com/DOMjudge/domjudge/commit/467a06fec865083d4cdf9000413e1ea22509a98c) - [view on LGTM.com](https://lgtm.com/projects/g/DOMjudge/domjudge/rev/pr-e8736718f96f2cb6baca6b232f6fcb053ffbb3f8)
> 
> **new alerts:**
> 
>     * 1 for Unreachable code

Thats my mistake,

submit/submit
```diff
@@ -0,0 +1,546 @@
...
+    error('Entry point required but not specified nor detected.')
+
+logging.debug(F"contest is `{my_contest['shortname']}'")
+logging.debug(F"problem is `{my_problem['label']}'")
+logging.debug(F"language is `{my_language['name']}'")
+logging.debug(F"entry_point is `{entry_point or '<None>'}'")
+logging.debug(F"url is `{baseurl}'")
+
+if not args.assume_yes:
+    print('Submission information:')
```
mvr320, line 528:
```suggestion
    global num_warnings
    print('Submission information:')
```

nickygerritsen: This is not in a method but already in a global scope right? Does this even work?

mvr320: You're right, but than the LGTM message makes little sense, I assumed function due to indentation.

nickygerritsen: Yeah I don't get LGTM either

thijskh: DNLGTM

PR #1144, opened 2021-08-16
api/endpoint Unit tests

mvr320:
Also some small refactoring to reduce code duplication. These are not the most interesting features test for /api/judgehosts. The idea is to do the same for the other endpoints to get some simple coverage.

### DIFF
--- a/webapp/tests/Unit/Controller/API/JudgehostControllerTest.php
+++ b/webapp/tests/Unit/Controller/API/JudgehostControllerTest.php
@@ -1,0 +1,60 @@
+<?php declare(strict_types=1);
+
+namespace App\Tests\Unit\Controller\API;
+use Generator;
+
+class JudgehostControllerTest extends BaseTest
+{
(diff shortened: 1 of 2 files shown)


webapp/tests/Unit/Controller/API/JudgehostControllerTest.php
```diff
@@ -0,0 +1,62 @@
...
+        static::markTestSkipped('No API relation between contest and judgehost.');
+    }
+
+    /**
+     * Test that the endpoint does not return anything for objects that don't exist
+     *
+     * @dataProvider provideSingleNotFound
+     */
+    public function testSingleNotFound(string $id): void {
+        //TODO: This should fails
```
mvr320, line 59: This PR is mostly for what should be the outcome here.

mvr320: I made a decision, in a new iteration we can do something smart here.

nickygerritsen (approved):

PR #1145, opened 2021-08-17
Improve test coverage api

thijskh:


webapp/tests/Unit/Controller/API/GeneralInfoControllerTest.php
```diff
@@ -2,14 +2,121 @@
 
 namespace App\Tests\Unit\Controller\API;
 
+use App\DataFixtures\Test\SampleSubmissionsFixture;
 use App\Service\DOMJudgeService;
-use App\Tests\Unit\BaseTest;
+use App\Tests\Unit\BaseTest as BaseBaseTest;
```
nickygerritsen, line 7: Can't this line go completely then?

webapp/tests/Unit/Controller/API/GeneralInfoControllerTest.php
```diff
@@ -2,14 +2,121 @@
...
 use Generator;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
 use Symfony\Component\Intl\Countries;
 
 class GeneralInfoControllerTest extends BaseTest
 {
+    private $API_VERSION = 4;
```
nickygerritsen, line 14: Why a private var and not a `const`? Because they are public? (note: I think if we later use PHP 7.4+ only consts can also be private)

nickygerritsen (commented):
Although I wasn't the designated reviewer, some very small comments

webapp/tests/Unit/Controller/API/ConfigControllerTest.php
```diff
@@ -0,0 +1,164 @@
...
+        $this->verifyApiJsonResponse('GET', $this->endpoint . '/check', 401);
+        $this->verifyApiJsonResponse('GET', $this->endpoint . '/check', 403, 'demo');
+    }
+
+    /**
+     * Test the config checker endpoint returns expected content
+     */
+    public function testConfigCheckerWorksForAdmin()
+    {
+        // In the test setup, the config check retuns some errors so expected result is 500
```
mvr320, line 138:
```suggestion
        // In the test setup, the config check returns some errors so expected result is 500
```

webapp/tests/Unit/Controller/API/BaseTest.php
```diff
@@ -84,7 +84,7 @@ protected function removeTestContainer(): void
     }
 
     /**
-     * Verify the given API URL produces the given status code and return the body as JSON
+     * Verify the given API URL produces the given status code and return the body
      *
      * @param string      $method   The HTTP method to use
```
mvr320, line 89: I think some of these params can be removed, as they are already in the function signature.

webapp/tests/Unit/Controller/API/ConfigControllerTest.php
```diff
@@ -0,0 +1,164 @@
...
+class ConfigControllerTest extends BaseTest
+{
+    private $endpoint = '/config';
+
+    /**
+     * Test that same public config variables are returned for all role types
+     */
+    public function testConfigReturnsPublicVariables()
+    {
+        foreach([null, 'demo', 'admin'] as $user) {
```
mvr320, line 14: I personally prefer the @dataProvider here so its easy to see for which role the test fails,

nickygerritsen: Agreed (or add a message with the role to the asserts)

webapp/tests/Unit/Controller/API/BaseTest.php
thijskh: But not documented?

webapp/tests/Unit/Controller/API/MetricsControllerTest.php
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public function testPrometheusNoAccess()
+    {
+        $this->verifyApiResponse('GET', "/metrics/prometheus", 401);
+    }
+
+    /**
+     * Test that the Prometheus metrics give the basic expected output format
+     */
+    public function testPrometheuMetricss()
```
mvr320, line 24:
```suggestion
    public function testPrometheusMetrics()
```

webapp/tests/Unit/Controller/API/ConfigControllerTest.php
```diff
@@ -0,0 +1,164 @@
...
+            static::assertIsArray($response);
+            $secretvars = ['script_memory_limit', 'clar_answers', 'external_ccs_submission_url', 'data_source'];
+            foreach($secretvars as $secretvar) {
+                static::assertArrayNotHasKey($secretvar, $response);
+            }
+        }
+    }
+
+    /**
+     * Test that secret config variables are returned for admin
```
mvr320, line 50: Really nitpick, in this case the comment is very similar to the function name.

webapp/tests/Unit/Controller/API/BaseTest.php
mvr320: Ok, I agree.

nickygerritsen (approved):

mvr320 (approved):
I think it looks good apart from the nitpicks which can also be fixed later.

PR #1148, opened 2021-08-19
Improve test coverage api

thijskh:
The tests for balloons and awards are limited because we do not yet have completed judgings in the fixtures.

### DIFF
--- a/webapp/src/Controller/API/ProblemController.php
+++ b/webapp/src/Controller/API/ProblemController.php
@@ -95,9 +95,11 @@ class ProblemController extends AbstractRestController implements QueryObjectTra
 
             $objects = [];
             foreach ($ordinalArray->getItems() as $item) {
-                /** @var ContestProblemWrapper $contestProblemWrapper */
-                $contestProblemWrapper = $item->getItem();
-                $contestProblem        = $contestProblemWrapper->getContestProblem();
+                /** @var ContestProblemWrapper|ContestProblem $contestProblem */
+                $contestProblem = $item->getItem();
+                if ($contestProblem instanceof ContestProblemWrapper) {
+                    $contestProblem = $contestProblem->getContestProblem();
+                }
                 $probid                = $this->getIdField() === 'p.probid' ? $contestProblem->getProbid() : $contestProblem->getExternalId();
                 if (in_array($probid, $ids)) {
                     $objects[] = $item;
--- a/webapp/tests/Unit/Controller/API/BaseTest.php
+++ b/webapp/tests/Unit/Controller/API/BaseTest.php
@@ -10,7 +10,7 @@ use Symfony\Bundle\FrameworkBundle\KernelBrowser;
 
 abstract class BaseTest extends BaseBaseTest
 {
-    protected static $rootEndpoints = ['contests','judgehosts'];
+    protected static $rootEndpoints = ['contests','judgehosts','users'];
     
     /** @var KernelBrowser */
     protected $client;
@@ -341,6 +341,9 @@ abstract class BaseTest extends BaseBaseTest
     {
         if (($apiEndpoint = $this->apiEndpoint) === null) {
             static::markTestSkipped('No endpoint defined');
+        }
+        if (in_array($apiEndpoint, static::$rootEndpoints)) {
+            static::markTestSkipped('Endpoint does not have contests prefix');
         }
         // Note that the 42 and 123 here is a contest that doesn't exist
         if ($apiEndpoint === 'contests') {
--- a/webapp/tests/Unit/Controller/API/BaseTest.php
+++ b/webapp/tests/Unit/Controller/API/BaseTest.php
@@ -366,7 +366,6 @@ abstract class BaseTest extends BaseBaseTest
             static::markTestSkipped('No endpoint defined');
         }
         $url = $this->helperGetEndpointURL($apiEndpoint,$id);
-        var_dump($url);
         $this->verifyApiJsonResponse('GET', $url, 404, $this->apiUser);
     }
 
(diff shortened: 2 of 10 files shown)


webapp/src/Controller/API/ProblemController.php
```diff
@@ -95,9 +95,11 @@ public function listAction(Request $request): Response
 
             $objects = [];
             foreach ($ordinalArray->getItems() as $item) {
-                /** @var ContestProblemWrapper $contestProblemWrapper */
-                $contestProblemWrapper = $item->getItem();
-                $contestProblem        = $contestProblemWrapper->getContestProblem();
+                /** @var ContestProblemWrapper|ContestProblem $contestProblem */
+                $contestProblem = $item->getItem();
+                if ($contestProblem instanceof ContestProblemWrapper) {
+                    $contestProblem = $contestProblem->getContestProblem();
```
nickygerritsen, line 101: So we don't test jury calls (yet) I assume?

nickygerritsen (approved):

webapp/src/Controller/API/ProblemController.php
thijskh: Is indeed a good further improvement. Would need to see if that’s already possible with the BaseTest to test for multiple users?

webapp/tests/Unit/Controller/API/BaseTest.php
```diff
@@ -342,6 +342,9 @@ public function testSingleContestNotFound()
         if (($apiEndpoint = $this->apiEndpoint) === null) {
             static::markTestSkipped('No endpoint defined');
         }
+        if (in_array($apiEndpoint, static::$rootEndpoints)) {
+            static::markTestSkipped('Endpoint does not have contests prefix');
```
mvr320, line 346: $apiEndpoint('contests') is in static::$rootEndpoints, so swap this behind line 349?

webapp/src/Controller/API/ProblemController.php
nickygerritsen: Ah probably not yet, but can be an addition later on

webapp/tests/Unit/Controller/API/BaseTest.php
thijskh: Agreed that it’s now dead code after, should however probably move special case for this contests endpoint out of Base and into the Contest controller test.

mvr320: It would make both tests easier to understand.

mvr320: Not part of your PR, but can you change the typo in 353|356

thijskh: It's actually even simpler. The code was redundant since it's actually for a contest the thing that is tested by testSingleNotFound() which is already ran.

webapp/src/Controller/API/ProblemController.php
thijskh: It's not fully trivial. You could do something like this:

make $apiUser an array which can maximally contain [null, demo, admin]. When unset defaults to [null]. Give expectedObjects and expectedAbsent an extra dimension which specifies for each user the expected objects or absent objects.

Within the BaseTest tests, loop over the users in the array and repeat each test for each of the users and compare it to the corresponding specified objects.

thijskh: I've devised a different solution for now.

nickygerritsen (approved):

webapp/tests/Unit/Controller/API/ExecutableControllerTest.php
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Test that a non logged in user can not access the executables
+     */
+    public function testExecutablesNoAnonAccess()
+    {
+        $this->verifyApiResponse('GET', "/executables/compare", 401);
+    }
+
+    /**
+     * Test that a non logged in user can not access the executables
```
mvr320, line 17:
```suggestion
     * Test that a team user can not access the executables
```

webapp/tests/Unit/Controller/API/ExecutableControllerTest.php
```diff
@@ -0,0 +1,66 @@
...
+        $response = $this->verifyApiJsonResponse('GET', "/executables/compare", 200, 'admin');
+        $contents = $this->base64unzip($response);
+
+        static::assertArrayHasKey('build', $contents);
+        static::assertStringContainsString('g++ -g', $contents['build']);
+        static::assertArrayHasKey('compare.cc', $contents);
+        static::assertStringContainsString("Space!  Can't live with it, can't live without it", $contents['compare.cc']);
+    }
+
+    protected function base64unzip(string $content): array
```
mvr320, line 46: Should this be moved to Utils or Basetest?

thijskh: It can be as soon as something else wants to use it?

thijskh: It is 'testsuite' level code btw, so could be moved to BaseTest when we add tests for another endpoint that requires it. Not something to add into the main code.

webapp/tests/Unit/Controller/API/ProblemControllerTest.php
```diff
@@ -0,0 +1,47 @@
+<?php declare(strict_types=1);
+
+namespace App\Tests\Unit\Controller\API;
+
+class ProblemControllerTest extends BaseTest
```
mvr320, line 5: Maybe add PHPdoc/something that the admin case is explicit handled in another class? Or alternative can you have both classes in the same file, not sure if I'm a fan of that (or if its even possible).

mvr320 (approved):

webapp/tests/Unit/Controller/API/ExecutableControllerTest.php
mvr320: Ok makes sense, and I agree about the main code.

PR #1149, opened 2021-08-19
Use bats-assert and bats-support for easier checks.

meisterT:
This will also show what the original output was and make it easier to
debug things when they go wrong.

### DIFF
--- a/submit/assert.bash
+++ b/submit/assert.bash
@@ -1,0 +1,72 @@
+# Inspired by https://github.com/ztombol/bats-assert.
+
+fail()
+{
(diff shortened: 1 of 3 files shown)


meisterT:
Note that this fail on CI (hopefully) and give us an indication what to fix.

thijskh (changes requested):
Excellent. Just to make it explicit, the tests now pass because the previous string check has been replaced with a regex check also in this commit.

Only thing to change is that the indentation of assert.bash is inconsistent.

PR #1151, opened 2021-08-20
Show logon of user as action

mvr320:
Also enable searching for those actions, in case of auditing scrolling
through all pages makes little sense.

I'm not sure if this will give a performance penalty?


nickygerritsen (approved):

thijskh:
I don't think the client side searching makes much sense / is confusing because the table is paginated?

mvr320:
> 
> 
> I don't think the client side searching makes much sense / is confusing because the table is paginated?

Ok, I might have wrongly assumed that I indirect also removed the paginated part. I'll take it out of this PR.

thijskh (approved):

mvr320:
Changed URLs:
jury/auditlog

PR #1152, opened 2021-08-20
Redirect users away when they have no active selected contest

mvr320:


webapp/src/Twig/TwigExtension.php
```diff
@@ -206,6 +206,7 @@ public function getGlobals()
                 }, Countries::getAlpha3Codes())
             ),
             'show_shadow_differences' => $this->tokenStorage->getToken() &&
+                                         !is_null($this->dj->getCurrentContest()) &&
```
nickygerritsen, line 209:
```suggestion
                                         $this->dj->getCurrentContest() &&
```
sounds shorter

webapp/src/Controller/Jury/JuryMiscController.php
```diff
@@ -62,7 +62,7 @@ public function __construct(EntityManagerInterface $entityManager, DOMJudgeServi
     public function indexAction(Request $request): Response
     {
         $errors = [];
-        return $this->render('jury/index.html.twig', ['errors' => $errors]);
+        return $this->render('jury/index.html.twig', ['errors' => $errors, 'contest_selected' => !is_null($this->dj->getCurrentContest())]);
```
nickygerritsen, line 65: Are we using this?
Also: while we are at it, maybe hide the analytics and scoreboard links on the jury interface? Should be able to check `current_contest` in twig to see if there is one

webapp/src/Controller/Jury/ShadowDifferencesController.php
```diff
@@ -83,6 +83,11 @@ public function indexAction(Request $request)
             return $this->redirectToRoute('jury_index');
         }
 
+        if (is_null($this->dj->getCurrentContest())) {
```
nickygerritsen, line 86:
```suggestion
        if (!$this->dj->getCurrentContest()) {
```
here as well?

webapp/src/Controller/Jury/JuryMiscController.php
mvr320: I thought about that and my argumentation to leave it like this was that buttons not being there is very annoying. We could make them gray but UX wise I think people finding out they should press the button [change contest] or not showing the button has the same level of frustration.

So I'm fine with your option, if you indeed prefer it over the redirects.

nickygerritsen: OK fair point, buttons and links moving around is also a mess. Let's keep them

nickygerritsen (approved):

PR #1155, opened 2021-08-21
Fix importfeed breaking on problems without color

mvr320:


nickygerritsen (approved):

thijskh (approved):
This is actually a broader bugfix since color is nullable you must be able to unset it (e.g. remove it in the contestproblem list in the admin interface which is now also broken)

thijskh:
Why does it fail?

Issue #1054, opened 2021-04-17
Remove judgehost field


mvr320:
As reported by @meisterT in https://github.com/DOMjudge/domjudge/pull/1046

```We right now set the judgehost only when we pop back the compilation result on compile error. We should probably change that so that we remove that field from the table ...```

PR #1158, opened 2021-08-22
Fix giving back judgings after a judgehost crashes.

### LINKED ISSUE #1153: Giving back judgings is broken for submissions that are judged on mulitple machines
Both `judging `and `judgetask` have a `judgehost` associated with them.
Now if a submission gets judged by 2 machines, this is stored in `judgetask` correctly (per run the correct judgehost), but not in `judging` (obviously, since we only have one field to set).
Imagine we have judgehost `A` and `B` who did a submission, where `A` had a crash or something but `B` is stored on `judging`. Now we never give back the judgetasks for `A`.

First of all I wonder why we have the `judgehost` stored on a `judging`. Looking in the code we have quite some queries for `j.judgehost` still left, so we can't easily drop it. We also use it when we need to disable a judgehost because it had a different compilation outcome than another judgehost on the same submission.

Should we somehow change all `j.judgehost` queries to join on judgetask and use that? And then drop `judgehost` on `judgimg`?
We probably should at least do this in `JudgehostController::createJudgehostAction` where we load the judgings to give back, right?

meisterT:
We don't store the necessary info on judgings anymore, so we have to
collect it from other tables.

Fixes #1153.


webapp/src/Controller/API/JudgehostController.php
```diff
@@ -200,11 +200,15 @@ public function createJudgehostAction(Request $request) : array
         /** @var Judging[] $judgings */
         $judgings = $this->em->createQueryBuilder()
             ->from(Judging::class, 'j')
-            ->innerJoin('j.judgehost', 'jh')
             ->leftJoin('j.rejudging', 'r')
+            ->leftJoin('j.runs', 'jr')
+            ->leftJoin('jr.judgetask', 'jt')
```
nickygerritsen, line 205: Since we do an inner join on judgehost below these could also be inner right? Left is also fine, just mentioning it

nickygerritsen (approved):
We should still think about what to do with judging.judgehost. Do you also think it should be dropped in general?

webapp/src/Controller/API/JudgehostController.php
thijskh: > We should still think about what to do with judging.judgehost. Do you also think it should be dropped in general?

That is #1054?

thijskh:
> We should still think about what to do with judging.judgehost. Do you also think it should be dropped in general?

That is #1054?

webapp/src/Controller/API/JudgehostController.php
meisterT: Done

nickygerritsen:
> > We should still think about what to do with judging.judgehost. Do you also think it should be dropped in general?
> 
> That is #1054?

Ah yes!

PR #1159, opened 2021-08-22
Warn about identical test cases for a problem.

meisterT:
identical

### DIFF
--- a/webapp/src/Controller/Jury/ProblemController.php
+++ b/webapp/src/Controller/Jury/ProblemController.php
@@ -717,6 +717,16 @@ class ProblemController extends BaseController
             return $this->redirectToRoute('jury_problem_testcases', ['probId' => $probId]);
         }
 
+        $known_md5s = [];
+        foreach ($testcases as $rank => $testcase) {
+            $input_md5 = $testcase->getMd5sumInput();
+            if (isset($known_md5s[$input_md5])) {
+                $this->addFlash('warning',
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

webapp/src/Controller/Jury/ProblemController.php
```diff
@@ -717,6 +717,16 @@ public function testcasesAction(Request $request, int $probId) : Response
...
         }
 
+        $known_md5s = [];
+        foreach ($testcases as $rank => $testcase) {
+            $input_md5 = $testcase->getMd5sumInput();
+            if (isset($known_md5s[$input_md5])) {
+                $this->addFlash('warning',
+                    "Testcase #" . $rank . " has identical input to testcase #" . $known_md5s[$input_md5] . '.');
+            }
+            $known_md5s[$input_md5] = $rank;
```
thijskh, line 727: If you have triple testcases then it will only show the last one. Acceptable or make known_md5s[$input_md5] a list of ranks?

meisterT: Check this situation:
identical

first three testcases are identical and the last two are as well. I think the code is simpler this way but still everything is highlighted. It should be a rare case anyway.

thijskh: Fair enough. :+1:

PR #1160, opened 2021-08-22
Rejudging improvements

thijskh:


nickygerritsen (approved):

judge/judgedaemon.main.php
```diff
@@ -1057,7 +1058,6 @@ function compile(array $judgeTask, string $workdir, string $workdirpath, array $
 
         return false;
     }
-    logmsg(LOG_INFO, "  💻 Compilation: ($files[0]) '".$EXITCODES[$retval]."'");
```
meisterT, line 1060: Why are you moving this below? This could be valuable information, couldn't it?

thijskh: The `if` checks whether EXITCODES in fact the retval. If it doesn't, the message is meaningless _and_ triggers a php notice with stacktrace.

meisterT: Ah, we log the unknown exit code below in the snippet that the GitHub app thankfully collapsed for me...

meisterT (approved):

PR #1161, opened 2021-08-22
Capital nocontest

mvr320:
During the Open2021 we found some pages which broke when there was no active contest. This test should detect those pages.


nickygerritsen (approved):

mvr320:
Verified by rebasing a commit away which fixed such an issue, it correctly detected the mistake in such a case.

nickygerritsen (approved):

PR #1162, opened 2021-08-23
Get rid of judgings.judgehostid column. Fixes #1054.

### LINKED ISSUE #1054: Remove judgehost field

nickygerritsen:


webapp/src/Controller/API/JudgehostController.php
```diff
@@ -368,30 +368,41 @@ public function updateJudgingAction(Request $request, string $hostname, int $jud
...
-                        ->setDescription('Compilation results are different for j' . $judging->getJudgingid())
-                        ->setJudgehostlog(base64_encode('New compilation output: ' . $output_compile))
-                        ->setTime(Utils::now())
-                        ->setDisabled($disabled);
-                    $this->em->persist($error);
+                    // Since the other judgehosts were not successful, but we were , assume that the other judgehosts
+                    // are broken and disable it.
+                    $disableHostnames = [];
+                    /** @var JudgingRun $run */
+                    foreach ($judging->getRuns() as $run) {
```
meisterT, line 382: Unfortunately I think this won't happen here since if compilation failed we don't set the judging run, or do we?

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -428,7 +438,7 @@ public function updateJudgingAction(Request $request, string $hostname, int $jud
                         $this->em->flush();
                     } else if ($judging->getResult() !== Judging::RESULT_COMPILER_ERROR) {
                         // The new result contradicts a former one, that's not good.
-                        // Since the other judgehost was successful, but we were not, assume that the current judgehost
+                        // Since the other judgehosts were successful, but we were not, assume that the current judgehost
```
meisterT, line 441: Nit: at least one other judgehost

webapp/src/Controller/API/JudgehostController.php
nickygerritsen: Don't we? I will test but clearly we create judgetasks and aren't judgetasks linked to runs?

meisterT: You're right

nickygerritsen: I will still see if I can test this somehow

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -368,30 +368,41 @@ public function updateJudgingAction(Request $request, string $hostname, int $jud
...
+                    // Since the other judgehosts were not successful, but we were , assume that the other judgehosts
+                    // are broken and disable it.
+                    $disableHostnames = [];
+                    /** @var JudgingRun $run */
+                    foreach ($judging->getRuns() as $run) {
+                        if ($run->getJudgeTask() &&
+                            $run->getJudgeTask()->getJudgehost() &&
+                            $run->getJudgeTask()->getJudgehost()->getHostname() !== $judgehost->getHostname()) {
+                            $hostname = $run->getJudgeTask()->getJudgehost()->getHostname();
+                            $disableHostnames[$hostname] = $hostname;
```
meisterT, line 387: Nit: `disableHostnames[] = $hostname` should do it as well

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -1111,48 +1121,6 @@ private function maybeUpdateActiveJudging(Judging $judging): void
         }
     }
 
-    private function getSubmissionsToJudge(Judgehost $judgehost, $restrictJudgingOnSameJudgehost)
```
meisterT, line 1114: This is just dead code unrelated to this change? If so, please delete in a separate commit.

webapp/src/Controller/Jury/JudgehostController.php
```diff
@@ -97,9 +97,11 @@ public function indexAction(Request $request): Response
 
         $now           = Utils::now();
         $contest       = $this->dj->getCurrentContest();
-        $query         = 'SELECT judgehostid, SUM(IF(endtime, endtime, :now) - GREATEST(:from, starttime)) AS `load`
-                          FROM judging
-                          WHERE endtime > :from OR (endtime IS NULL AND (valid = 1 OR rejudgingid IS NOT NULL))
+        $query         = 'SELECT judgehostid, SUM(IF(jr.endtime, jr.endtime, :now) - GREATEST(:from, jt.starttime)) AS `load`
```
meisterT, line 100: Add a todo here that mentions the missing compilation time.

meisterT (approved):

webapp/src/Controller/API/JudgehostController.php
nickygerritsen: It will be in there multiple times and then we create multiple internal errors. Or we need `array_unique` below

meisterT: I need more ☕

webapp/src/Controller/API/JudgehostController.php
nickygerritsen: Tested this by having 2 judgehosts, having 1 force a compilation failure and sleeping before reporting this, then while this happens starting a second one that also sleeps but until a bit longer and be compilation success. It seems to work.

PR #1167, opened 2021-08-24
Fix submission page with multiple judgings after 514f1af75eea1af81b66…

meisterT:
…00435642d1ebf51cbda7.


nickygerritsen (approved):

nickygerritsen:
The commit in the commit message seems to be the wrong one btw but I get what you mean

PR #1169, opened 2021-08-25
Update boolfind and get rid of old way to do interactive problems.

meisterT:


nickygerritsen:
Upgrading from an old install will break if you have boolfind still right? I think that's totally fine but just mentioning it.

meisterT:
> Upgrading from an old install will break if you have boolfind still right? I think that's totally fine but just mentioning it.

Yep, I'll add a commit to mention it in the release notes.

nickygerritsen (approved):

nickygerritsen:
Changed URLs:
jury_s_executables
jury_s_executables_s_boolfind_run
jury_s_executables_s_boolfind_run_s_content
jury_s_executables_s_boolfind_run_s_edit-files@index=0
jury_s_executables_s_boolfind_run_s_edit-files@index=1
jury_s_executables_s_boolfind_run_s_edit
jury_s_problems_s_3
jury_s_problems_s_3_s_edit

meisterT:
I don't understand what this comment means and whether it implies that I have to change something

nickygerritsen:
> I don't understand what this comment means and whether it implies that I have to change something

This comment should make CI happy. The UI for these pages (replace `_s_` with `/`) changed and @mvr320 made it that you can say 'I know I changed these pages, accept them'. I still think the way we do this is a bit messy though.

PR #1179, opened 2021-09-02
Print the actual hash cost value and improve wording a bit.

eldering:
The current wording about 1 week is not really clear to me. Is the intent to say that if the hash cost is >7 then it's fine for accounts that are used for *more* than a week?

### DIFF
--- a/sql/files/defaultdata/swift/run
+++ b/sql/files/defaultdata/swift/run
@@ -53,5 +53,5 @@ fi
 # -O:                 Optimizations (default for speed)
 # -static-executable: Statically link the executable
 # -static-stdlib:     Statically link the Swift standard library
-swiftc -O -static-executable -static-stdlib -o "$DEST" $SOURCES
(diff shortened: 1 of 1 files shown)


thijskh:
There is also this pending PR #1067

eldering:
> There is also this pending PR #1067

Ok, I wasn't aware of that one. What's the intent/status of it @mvr320?  If that gets merged, then this can be closed indeed.

mvr320:
> 
> 
> > There is also this pending PR #1067
> 
> Ok, I wasn't aware of that one. What's the intent/status of it @mvr320? If that gets merged, then this can be closed indeed.

The intent is that it will be mergable semi-soon.

It first needs the edit tests for users (add new is there already) as the mistake I found was when you edit the user that the edit does not pass. The rest of the logic is there and seemed to be correct.

mvr320:
https://github.com/DOMjudge/domjudge/pull/1067 is now not pending anymore so I think the original problem is now gone.

PR #1180, opened 2021-09-03
Codespell issues

mvr320:
This PR should now only contain the adding of tests with codespell. In https://github.com/DOMjudge/domjudge/pull/1181 a set of spelling mistakes was already fixed and I would like to start the discussion here on adding the tests.

Currently this will both do some annotation and give a test in GitLab. I build both because I can see a discussion for both, not doing annotation would make that in a review spelling mistakes are not spotted (so the test for GitLab would be good) but having the GHA would make that the contributor gets the annotations but this would introduce a discussion on the current implementation of annotations on GitHub.

Personally I prefer to have both, I think the amount of FP is low enough and it saves us from annoying typos.

When the other PR is fixed this PR will be rebased which should lead to the diff of only 3 files.


misc-tools/judging_statistics
```diff
@@ -0,0 +1,323 @@
+#!/usr/bin/env php
```
eldering, line 1: This file and `simulate_contest` were accidentally committed.

webapp/src/Command/ImportEventFeedCommand.php
```diff
@@ -1979,7 +1979,7 @@ protected function processPendingEvents(string $type, $id)
             // Get all pending events
             $pending = $this->pendingEvents[$type][$id];
             // Mark them as non-pending. Note that they might depend on more events,
-            // but then they'll be readded automatically in the correct place
+            // but then they'll be read automatically in the correct place
```
eldering, line 1982: This *was* correct.

misc-tools/judging_statistics
mvr320: You and thijs found the same issues.

webapp/src/Command/ImportEventFeedCommand.php
mvr320: Fixed in my other PR as re-added. I wrongly assume `read` as in reading something into an array.

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+    codespell:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+            - uses: codespell-project/actions-codespell@master
+              with:
+                  check_filenames: true
+                  only_warn: 1
+                  ignore_words_list: ba,probleme,formend,compiletime,chage,hve,files,ans,inout,testin,files\',creat
```
eldering, line 17: Why is `hve` being ignored?

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,18 @@
...
+                  skip: .idea,.git,*.guess,*.png,*.zip,*.log,*.sub,Makefile*,ChangeLog,uploadcodecov.sh,codespell.yml,tmp,lib,var,public
```
eldering, line 18: Why skip especially ChangeLog, but also Makefiles? Also, does this match files only at the root of the project or also deeper down? If the latter, then skipping `lib,var,public` may have unintented consequences.

eldering:
> This PR should now only contain the adding of tests with codespell. In #1181 a set of spelling mistakes was already fixed and I would like to start the discussion here on adding the tests.
> 
> Currently this will both do some annotation and give a test in GitLab. I build both because I can see a discussion for both, not doing annotation would make that in a review spelling mistakes are not spotted (so the test for GitLab would be good) but having the GHA would make that the contributor gets the annotations but this would introduce a discussion on the current implementation of annotations on GitHub.
> 
> Personally I prefer to have both, I think the amount of FP is low enough and it saves us from annoying typos.
> 
> When the other PR is fixed this PR will be rebased which should lead to the diff of only 3 files.

What this check caught seems very reasonable. How many false positives did you see?

Anyways, seems fine to me to add as a test. For annotations in the PR, it depends a bit on the false positive ratio: 

> This PR should now only contain the adding of tests with codespell. In #1181 a set of spelling mistakes was already fixed and I would like to start the discussion here on adding the tests.
> 
> Currently this will both do some annotation and give a test in GitLab. I build both because I can see a discussion for both, not doing annotation would make that in a review spelling mistakes are not spotted (so the test for GitLab would be good) but having the GHA would make that the contributor gets the annotations but this would introduce a discussion on the current implementation of annotations on GitHub.
> 
> Personally I prefer to have both, I think the amount of FP is low enough and it saves us from annoying typos.
> 
> When the other PR is fixed this PR will be rebased which should lead to the diff of only 3 files.

This seems to catch actual mistakes, so fine with adding it as a test. For adding annotations I'd say it depends a bit on the false positive ratio; I'd say that about 30% at least should be real issues. Also, does it only add an annotation when it finds something?

.github/workflows/codespell.yml
mvr320: It wants to change (hVe -> have),

for example:
$options = getopt("dv:n:hVe:j:t:", ["diskspace-error"]);
./judge/judgedaemon.main.php

.github/workflows/codespell.yml
mvr320: We skip ChangeLog because it contains some typos but I think in the past we agreed to not change the ChangeLog file after usage of such tools.

This will also do deeper down. I think the only var we have is webapp/var which we can change to cache as that will contain upstream code which we dont want to test.

public should target webapp/public`
but will target:
./lib/vendor/zircote/swagger-php/docs/.vuepress/public
./lib/vendor/nelmio/api-doc-bundle/Resources/public
./webapp/templates/public
./webapp/public

the lib/vector is fine. the webapp/templates/public would be nice to target.

lib will target:
./lib
./lib/vendor/doctrine/data-fixtures/lib
./lib/vendor/doctrine/deprecations/lib
./lib/vendor/doctrine/orm/lib
./lib/vendor/doctrine/inflector/lib
./lib/vendor/doctrine/lexer/lib
./lib/vendor/doctrine/cache/lib
./lib/vendor/doctrine/annotations/lib
./lib/vendor/doctrine/persistence/lib
./lib/vendor/doctrine/common/lib
./lib/vendor/doctrine/migrations/lib
./lib/vendor/doctrine/collections/lib
./lib/vendor/doctrine/dbal/lib
./lib/vendor/doctrine/event-manager/lib
./lib/vendor/nikic/php-parser/lib
./lib/vendor/ircmaxell/password-compat/lib
./lib/vendor/novus/nvd3/examples/lib
./lib/vendor/mbostock/d3/lib 

the lib/vendor is intented, but now I see that there is code there which could be targeted will make a change for it.

var will currently only target:
./webapp/var

eldering: I don't think we want to include `lib/vendor` as it contains upstream code.

mvr320:
> 
> 
> [Edited as it included a duplicated response.]
> 
> > This PR should now only contain the adding of tests with codespell. In #1181 a set of spelling mistakes was already fixed and I would like to start the discussion here on adding the tests.
> > Currently this will both do some annotation and give a test in GitLab. I build both because I can see a discussion for both, not doing annotation would make that in a review spelling mistakes are not spotted (so the test for GitLab would be good) but having the GHA would make that the contributor gets the annotations but this would introduce a discussion on the current implementation of annotations on GitHub.
> > Personally I prefer to have both, I think the amount of FP is low enough and it saves us from annoying typos.
> > When the other PR is fixed this PR will be rebased which should lead to the diff of only 3 files.
> 
> This seems to catch actual mistakes, so fine with adding it as a test. For adding annotations I'd say it depends a bit on the false positive ratio; I'd say that about 30% at least should be real issues. Also, does it only add an annotation when it finds something?
With the current list I think it will be above 30%, it will only give an annotation when it finds something but it might mention mistakes which were not excluded for ever.

So if it hits a FP, you need to explicit exclude this in the config file otherwise it will fail create an annotation all the time and fail the gitlab test.

.github/workflows/codespell.yml
mvr320: what I tried to say was:
lib/vendor was intended to be skipped, the ./lib should not as it has some of our own code so I changed it to vendor which will still trigger ./lib/vendor but we will scan ./lib for spelling mistakes.

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,18 @@
...
+    codespell:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+            - uses: codespell-project/actions-codespell@master
+              with:
+                  check_filenames: true
+                  only_warn: 1
+                  ignore_words_list: cas,objext,ba,probleme,formend,compiletime,chage,hve,files,ans,inout,testin,files\',creat
+                  skip: .idea,.git,*.png,*.zip,*.gif,*.jpg,*.log,*.sub,ChangeLog,uploadcodecov.sh,codespell.yml,tmp,vendor,cache,bundles,ace,*d3.min*,bootstrap*,FileSaver*,jquery*,jscolor*,config.guess
```
eldering, line 18: The `skip` config seems to allow directory paths, so why not `./lib/vendor`? Similar for some other file/directory names.

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+    codespell:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+            - uses: codespell-project/actions-codespell@master
+              with:
+                  check_filenames: true
+                  only_warn: 1
+                  ignore_words_list: cas,objext,ba,probleme,formend,compiletime,chage,hve,files,ans,inout,testin,files\',creat
```
eldering, line 17: I think it might be better to use `ignore_words_file` as this might grow to a (very) long list.

mvr320: Should now be implemented. Had to do some extra code to do the same thing for the files to skip.

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,21 @@
+on:
+    push:
+        branches: [master]
```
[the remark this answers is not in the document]
mvr320: I think doing this check on any branch which is not a PR is not even relevant.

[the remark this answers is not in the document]
nickygerritsen: But should be main, not master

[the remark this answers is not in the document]
mvr320: Yes I understood Thijs his suggestion, but I think the fix is to not run on pushes to those `special` branches.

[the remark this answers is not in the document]
mvr320: after removal of the on: branch it still works.

eldering (approved):
LGTM

PR #1181, opened 2021-09-04
Codespell only spellingfixes

mvr320:
I was doing some testing with codespell which I think has potential. Because I suspect some discussion on the workings I'll create 2 PRs. This one will be for the fixed mistakes. Some I'm not 100% sure about so those are kept as individual commits but in the end this should al be put in 1 single commit.

### DIFF
--- a/webapp/config/packages/nelmio_api_doc.yaml
+++ b/webapp/config/packages/nelmio_api_doc.yaml
@@ -252,7 +252,7 @@ nelmio_api_doc:
                         roles:
                             type: array
                             items:
-                                type: sring
+                                type: string
                 ClarificationPost:
                     type: object
                     required: [text]
--- a/webapp/src/Command/ImportEventFeedCommand.php
+++ b/webapp/src/Command/ImportEventFeedCommand.php
@@ -1979,7 +1979,7 @@ class ImportEventFeedCommand extends Command
             // Get all pending events
             $pending = $this->pendingEvents[$type][$id];
             // Mark them as non-pending. Note that they might depend on more events,
-            // but then they'll be readded automatically in the correct place
+            // but then they'll be re-added automatically in the correct place
             unset($this->pendingEvents[$type][$id]);
             foreach ($pending as $event) {
                 $this->logger->debug(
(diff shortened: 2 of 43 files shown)


webapp/src/Command/ImportEventFeedCommand.php
```diff
@@ -1979,7 +1979,7 @@ protected function processPendingEvents(string $type, $id)
             // Get all pending events
             $pending = $this->pendingEvents[$type][$id];
             // Mark them as non-pending. Note that they might depend on more events,
-            // but then they'll be readded automatically in the correct place
+            // but then they'll be read automatically in the correct place
```
thijskh, line 1982: Indended is: "re-added" (added again) not "read"

misc-tools/judging_statistics
```diff
@@ -0,0 +1,323 @@
+#!/usr/bin/env php
+<?php declare(strict_types=1);
+/**
+ * Generated from 'judging_statistics.in' on Mon Aug 23 21:31:14 CEST 2021.
```
thijskh, line 4: Accidentally added generated file I

misc-tools/simulate_contest
```diff
@@ -0,0 +1,375 @@
+#!/usr/bin/env php
+<?php declare(strict_types=1);
+/**
+ * Generated from 'simulate_contest.in' on Mon Aug 23 21:31:14 CEST 2021.
```
thijskh, line 4: Accidentally added generated file II

webapp/config/packages/nelmio_api_doc.yaml
```diff
@@ -252,7 +252,7 @@ nelmio_api_doc:
                         roles:
                             type: array
                             items:
-                                type: sring
+                                type: string
```
thijskh, line 255: Why did this not fail in the first place?

thijskh (changes requested):
Some things to fix, everything else 👍

webapp/config/packages/nelmio_api_doc.yaml
mvr320: I suspect because you can also have custom types OpenAPI is not aware of. This typo was the reason that I started playing around with the tool.

thijskh (approved):

PR #1182, opened 2021-09-06
Add edit tests

mvr320:
Minimal tests to test the edit action for jury controllers.

### DIFF
--- a/webapp/tests/Unit/Controller/Jury/JuryControllerTest.php
+++ b/webapp/tests/Unit/Controller/Jury/JuryControllerTest.php
@@ -33,7 +33,10 @@ abstract class JuryControllerTest extends BaseTest
     protected static $deleteExtra       = null;
     protected static $addEntities       = [];
     protected static $addEntitiesCount  = [];
-
+    protected static $defaultEditEntityName  = null;
+    protected static $specialFieldOnlyUpdate = [];
+    protected static $editEntitiesSkipFields = [];
+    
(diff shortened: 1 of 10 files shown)


webapp/tests/Unit/Controller/Jury/JuryControllerTest.php
```diff
@@ -249,6 +252,71 @@ public function testCheckAddEntityAdmin(): void
...
+                }
+            }
+            $this->verifyPageResponse('GET', $editLink, 200);
+            $crawler = $this->getCurrentCrawler();
+            foreach ($formDataKeys as $id => $key) {
+                $formFields[static::$addForm . $key . "]"] = $formDataValues[$id];
+            }
+            $button = $this->client->getCrawler()->selectButton('Save');
+            $form = $button->form($formFields, 'POST');
+            $this->client->submit($form);
```
thijskh, line 289: Would we need to check that this actually went ok and is not eg an error 500 page?

thijskh (approved):

thijskh:
Needs some squashing before merging of course

webapp/tests/Unit/Controller/Jury/JuryControllerTest.php
mvr320: I can implement that,

I think that there is not a case where your POST action would give 500 but the resulting page would have the expected information but thats a bit dependant on where you fail.

PR #1185, opened 2021-09-10
Unit test for the reset command

vmcj:
Rewritten after Nicky his suggestion, this now tests the resetting of user passwords and some default symfony behaviour.

_The test normally fails as we don't finish DB transactions. This will introduce the problem again that we might affect tests by introducing a new test in the suite.

As this is based on https://github.com/DOMjudge/domjudge/pull/1067 that one should be merged first.
The uploading of additional tests works as can be seen in: https://gitlab.com/DOMjudge/domjudge/-/pipelines/368837685/test_report where we get results of both phpunit.xml.dist and phpunit.noDBLock.xml.dist_

### DIFF
--- a/webapp/tests/Unit/Controller/Jury/JuryControllerTest.php
+++ b/webapp/tests/Unit/Controller/Jury/JuryControllerTest.php
@@ -276,9 +276,6 @@ abstract class JuryControllerTest extends BaseTest
             foreach ($formDataKeys as $id => $key) {
                 $formFields[static::$addForm . $key . "]"] = $formDataValues[$id];
             }
-            foreach ($formDataKeys as $id => $key) {
-                $formFields[static::$addForm . $key . "]"] = $formDataValues[$id];
-            }
             $button = $this->client->getCrawler()->selectButton('Save');
             $form = $button->form($formFields, 'POST');
             $this->client->submit($form);
--- a/webapp/tests/Unit/Command/ResetUserPasswordCommandTest.php
+++ b/webapp/tests/Unit/Command/ResetUserPasswordCommandTest.php
@@ -1,0 +1,125 @@
+<?php declare(strict_types=1);
+
+namespace App\Tests\Unit\Command;
+
+use App\Command\ResetUserPasswordCommand;
+use App\Tests\Unit\BaseTest as BaseTest;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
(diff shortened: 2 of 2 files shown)


webapp/tests/NoDBLock/Command/ResetUserPasswordCommandTest.php
```diff
@@ -0,0 +1,76 @@
...
+use Symfony\Component\Process\Process;
+
+class ResetUserPasswordCommandTest extends BaseTest
+{
+
+    public static $demoPassword = 'demo';
+
+    public function testNonExistingUser(): void {
+        $username = 'NotAnUser';
+        $command  = ['webapp/bin/console','domjudge:reset-user-password',$username];
```
nickygerritsen, line 15: Instead of creating a process, can’t we get the command from the container and run it that way? Might that also solve the db lock issue?

webapp/tests/Unit/Controller/Jury/JuryControllerTest.php
```diff
@@ -284,12 +284,16 @@ public function testCheckEditEntityAdmin(string $identifier, array $formDataKeys
             foreach ($formDataKeys as $id => $key) {
                 $formFields[static::$addForm . $key . "]"] = $formDataValues[$id];
             }
+            foreach ($formDataKeys as $id => $key) {
```
nickygerritsen, line 287: Heh is this duplicate?

mvr320: It is yes. Will take it away.

webapp/tests/NoDBLock/Command/ResetUserPasswordCommandTest.php
mvr320: If you think that would make the test similar I can try it. Would be nice if it solves the DB lock issue.

nickygerritsen: Yeah it’s creating the command, an ArrayInput and some output and calling run. Should maybe fix the db lock issue, not sure

nickygerritsen: This didn’t work?

vmcj: It did, I just forgot to remove this file.

nickygerritsen (approved):

PR #1186, opened 2021-09-11
Use coffin to indicate jail

mvr320:
The checkmark was not aligned with the one used for the outcome.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -547,7 +547,7 @@ if (!empty($options['e'])) {
 umask(0022);
 
 // Check basic prerequisites for chroot at judgehost startup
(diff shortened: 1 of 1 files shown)


eldering:
I think 🔒 would be a better choice.

meisterT:
It's not a check mark, it's a square root symbol 😉

But replacing with a 🔒 sounds good

thijskh:
🛅
▶️&⏹
💂‍♀️
🥕

thijskh:
For the three occurrences in order:
🔏, 🔒, 🔓

mvr320:
> It's not a check mark, it's a square root symbol wink

I really should have seen that... still it doesn't look nice in my screen ;)

PR #1191, opened 2021-09-15
Improve internal error handling.

meisterT:
On internal errors, we disable the other outstanding judgings, or more precise judgetasks. Once the error is resolved, these would have been previously just dangling.

Now, we create a rejudging to trigger computing a verdict.

In the future, we might consider making this configurable (e.g. offer the user to judge all remaining instead of creating new judgings), but that can be done as a separate PR.

### DIFF
--- a/webapp/src/Controller/Jury/InternalErrorController.php
+++ b/webapp/src/Controller/Jury/InternalErrorController.php
@@ -6,8 +6,10 @@ use App\Controller\BaseController;
 use App\Doctrine\DBAL\Types\InternalErrorStatusType;
 use App\Entity\InternalError;
 use App\Entity\Judgehost;
+use App\Entity\JudgeTask;
 use App\Entity\Problem;
 use App\Service\DOMJudgeService;
+use App\Service\RejudgingService;
 use App\Utils\Utils;
 use Doctrine\ORM\EntityManagerInterface;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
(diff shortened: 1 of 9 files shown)


webapp/src/Controller/Jury/InternalErrorController.php
```diff
@@ -161,9 +166,21 @@ public function handleAction(int $errorId, string $action): RedirectResponse
...
                     true
                 );
-                $this->dj->auditlog('internal_error', $internalError->getErrorid(),
-                                    sprintf('internal error: %s', $status));
+                $affectedJudgings = $internalError->getAffectedJudgings();
+                if ($affectedJudgings !== null) {
+                    $skipped = array();
+                    $this->rejudgingService->createRejudging(
+                        'Internal Error ' . $internalError->getErrorid() . ' resolved',
+                        JudgeTask::PRIORITY_DEFAULT,
```
mvr320, line 174: You create new tasks right? So what happens when on t1 an executable breaks,
t2,t3 we get a lot of tasks so we have a backlog
t4 we fix the internal error,

would we first solve the tasks for t1 (created at t4) or the ones for t2?

webapp/src/Entity/Judging.php
```diff
@@ -174,6 +174,14 @@ class Judging extends BaseApiEntity implements ExternalRelationshipEntityInterfa
      */
     private $runs;
 
+    /**
+     * Rejudgings have one parent judging.
```
nickygerritsen, line 178: I don’t get this comment line

meisterT: lol, this is a copy paste error

webapp/src/Controller/Jury/InternalErrorController.php
meisterT: By default judging fairness right now, i.e. teams that didn't occupy a judgehost for the longest time come first.

But we can make this configurable in a separate PR.

PR #1196, opened 2021-09-17
Use badges for submission stats and move them to the top.

meisterT:
### **Before**:

before

### **After**:

after

I often found myself scrolling down, so I moved them to to the top and then decided to use badges instead.


thijskh:
Definitely an improvement.

Not sure about the word “total” in the first badge, is it needed?

Maybe put #submitted at the front since the others are contained in this number?

Might want to clean up this whole table header section a bit further (later) with all the buttons, filters and badges.

PR #1201, opened 2021-09-23
Only use ASCII in code, and more explicit variable names.

eldering:
Using non-ASCII characters makes it more difficult to grep, search and replace etc, and create confusion about what characters are actually used. Secondly, use a more explicit, semantic name for the constant that converts from microseconds to seconds.


mvr320 (approved):
I agree, this solves the US issue I had with it.

meisterT (approved):

nickygerritsen (approved):

thijskh (approved):

PR #1204, opened 2021-09-26
Add option to show doc links in team interface.

nickygerritsen:
Screenshot with the .dist config:
Screen Shot 2021-09-26 at 12 07 35

When merged we need to cherrypick into `wf2020`.

### DIFF
--- a/doc/manual/config-advanced.rst
+++ b/doc/manual/config-advanced.rst
@@ -33,8 +33,8 @@ path of your installation) as follows:
 
 It is also possible to load custom CSS and/or JavaScript files. To do so, place
 files ending in `.css` under `public/css/custom/` and/or files ending in `.js`
-under `public/js/custom/`. See the Config checker in the admin interfae for the
-full filesystem path of your installation). Note that there is no guaranteed
+under `public/js/custom/`. See the Config checker in the admin interface for the
+full filesystem path of your installation. Note that there is no guaranteed
 order in which the files will be loaded, but they will all be loaded after the
 main DOMjudge CSS and JavaScript files. If you have a lot of custom CSS/JavaScript
 files in these directories, the response time of DOMjudge might decrease, so it
@@ -44,6 +44,18 @@ is recommended to only place a few files there.
 
   If you add or remove any of the above files, you need to
   :ref:`clear the cache <clear-cache>` for changes to be detected.
+
+Adding links to documentation to the team interface
+---------------------------------------------------
+
+DOMjudge supports adding links to documentation to the team interface.
+First, on the DOMserver, copy the file ``etc/docs.yaml.dist`` to
+``etc/docs.yaml`` and modify its contents. The ``.dist`` file contains
+comments as to what each field in the file means and how to use it. If you
+want to link to files served by the webserver of the DOMserver, place them
+under `/public/docs/`. See the Config checker in the admin interface for
+the full filesystem path of your installation. All links open in a new
+tab / window.
 
 .. _authentication:
 
--- a/etc/docs.yaml.dist
+++ b/etc/docs.yaml.dist
@@ -1,0 +1,15 @@
+# This is an example configuration file for displaying documentation links in
+# the team interface.
+
+# This file should contain an array with items to show.
+
+  # Add a link to a local file under webapp/public/docs/
+- name: Team guide
+  link: ../docs/team.pdf
+  # Optional icon. See https://fontawesome.com/v5.15/icons?d=gallery&p=1&m=free
+  # for supported icons. Note: use full icon class, e.g. `fas fa-user`.
+  # Omit for no icon.
+  icon: fas fa-user
+  # Add a link to an external page.
+- name: C++ STL
+  link: https://www.cplusplus.com/reference/stl/
--- a/webapp/src/Twig/TwigExtension.php
+++ b/webapp/src/Twig/TwigExtension.php
@@ -208,6 +208,7 @@ class TwigExtension extends AbstractExtension implements GlobalsInterface
             'show_shadow_differences' => $this->tokenStorage->getToken() &&
                                          $this->authorizationChecker->isGranted('ROLE_ADMIN') &&
                                          $this->config->get('data_source') === DOMJudgeService::DATA_SOURCE_CONFIGURATION_AND_LIVE_EXTERNAL,
+            'doc_links' => $this->dj->getDocLinks(),
         ];
     }
 
--- a/doc/manual/shadow.rst
+++ b/doc/manual/shadow.rst
@@ -53,7 +53,7 @@ Creating the shadowing config file
 ----------------------------------
 
 On the DOMserver, copy the file ``etc/shadowing.yaml.dist`` to
-``/etc/shadowing.yaml`` and modify its contents. The ``.dist`` file contains
+``etc/shadowing.yaml`` and modify its contents. The ``.dist`` file contains
 comments as to what each field in the file means, but to summarize:
 
 - As the key pick a recognizable identifier, for example the ID of the contest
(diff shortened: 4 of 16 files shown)


doc/manual/config-advanced.rst
```diff
@@ -45,6 +45,17 @@ is recommended to only place a few files there.
   If you add or remove any of the above files, you need to
   :ref:`clear the cache <clear-cache>` for changes to be detected.
 
+Adding links to documentation to the team interface
+---------------------------------------------------
+
+DOMjudge supports adding links to documentation to the team interface.
+First, on the DOMserver, copy the file ``etc/docs.yaml.dist`` to
+``/etc/docs.yaml`` and modify its contents. The ``.dist`` file contains
```
thijskh, line 53: Probably not with a leading slash

doc/manual/config-advanced.rst
```diff
@@ -45,6 +45,17 @@ is recommended to only place a few files there.
...
+comments as to what each field in the file means and how to use it. For the
```
thijskh, line 54: s/For the/The/

doc/manual/config-advanced.rst
```diff
@@ -45,6 +45,17 @@ is recommended to only place a few files there.
...
+`file` option expects the files to be under `/public/docs/`. See the Config
+checker in the admin interfae for the full filesystem path of your installation.
+Note that all links open in a new tab / window.
```
thijskh, line 57: s/Note that all/All/

etc/docs.yaml.dist
```diff
@@ -0,0 +1,15 @@
+# This is an example configuration file for displaying documentation links in
+# the team interface.
+
+# This file should be an array with items to show.
+
+  # Add a link to a file under webapp/public/docs/
+- name: Team guide
+  file: team.pdf
```
thijskh, line 8: This distinction between file and link can be removed. If you'd specify link: team.pdf it would work the same but as a relative url.

webapp/src/Twig/TwigExtension.php
```diff
@@ -208,6 +208,7 @@ public function getGlobals()
             'show_shadow_differences' => $this->tokenStorage->getToken() &&
                                          $this->authorizationChecker->isGranted('ROLE_ADMIN') &&
                                          $this->config->get('data_source') === DOMJudgeService::DATA_SOURCE_CONFIGURATION_AND_LIVE_EXTERNAL,
+            'doc_links' => $this->dj->getDocLinks(),
```
thijskh, line 211: Why is this a twig extension and not something the controller sets for the one template that uses it?

etc/docs.yaml.dist
nickygerritsen: But not really, since the link is not /team/team.pdf but /docs/team.pdf. I can’t use a link starting with / since it might run under a subdir. Thoughts?

webapp/src/Twig/TwigExtension.php
nickygerritsen: Because the menu also uses it to decide if it should render an item, which is used on all team pages

doc/manual/config-advanced.rst
```diff
@@ -45,6 +45,17 @@ is recommended to only place a few files there.
...
 
+Adding links to documentation to the team interface
+---------------------------------------------------
+
+DOMjudge supports adding links to documentation to the team interface.
+First, on the DOMserver, copy the file ``etc/docs.yaml.dist`` to
+``/etc/docs.yaml`` and modify its contents. The ``.dist`` file contains
+comments as to what each field in the file means and how to use it. For the
+`file` option expects the files to be under `/public/docs/`. See the Config
+checker in the admin interfae for the full filesystem path of your installation.
```
eldering, line 56: typo: `interface`

etc/docs.yaml.dist
```diff
@@ -0,0 +1,15 @@
+# This is an example configuration file for displaying documentation links in
+# the team interface.
+
+# This file should be an array with items to show.
```
eldering, line 4: s/be/contain/

webapp/src/Config/Resource/OptionalFileResource.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        return $this->resource;
+    }
+
+    public function isFresh($timestamp)
+    {
+        if (file_exists($this->resource) !== $this->exists) {
+            return false;
+        }
+        return false !== ($filemtime = @filemtime($this->resource)) && $filemtime <= $timestamp;
```
eldering, line 43: The `false !== ` seems superfluous. Maybe better to write this out more, since I don't really get what this method is trying to do.

nickygerritsen: Sorry, was a copy paste from the file I linked. But I agree it’s better to write it out

thijskh: Or... we can accept that you clear the cache when changing docs.yml? Just like you need to do when adding a banner image. So we also prevent checking whether the page is modified on every page load?

etc/docs.yaml.dist
thijskh: Not if you make the route be /docs/ instead of /team/docs I think? So if you place a `<a href="team.pdf">` there it just works. And relative links are no longer special.

thijskh: If you don't want to change the route, add ../ before any relative link.

webapp/src/Config/Resource/OptionalFileResource.php
nickygerritsen: Adding a banner does *not* require clearing the cache. The main reason we add this as a parameter is that we do not have to decode yaml on every (team) page, which is kinda costly.
I can remove the resource check itself but I think we do need to set this as a parameter to not decrease performance

etc/docs.yaml.dist
nickygerritsen: This will work yes. I will change it. I don’t like changing the url because we then need to move the action to its own controller, but ../ seems fine. Is there then any reason to put the docs folder in git? You can put pdfs anywhere. Although in maintainer mode this way you don’t have files git will see as unadded

webapp/src/Config/Resource/OptionalFileResource.php
thijskh: Our docs claim that you need to do that.

nickygerritsen: Then I think the docs are wrong. Although come to think of it: I think this is true in production. In development you don’t have to.
I am fine with removing the check, but then we should remove it for custom css/js and the banner as well to make it consistent (and accept you also need to clear cache on dev mode).
Opinions?

eldering: I'm not entirely sure I follow this. This is server-side caching, right? So nothing would change client-side? And what would you need to do server side to clear the cache? Also, if this is cached, where is it stored then, and how much is the performance impact?

nickygerritsen: This is the symfony cache. It’s stored in webapp/var/cache, just like for example the routes. Impact I don’t know exactly, but parsing files is never free.
As said I’m totally fine with removing the resource check. It would still be cached but then not automatically updated in dev mode if the file changed. I can also remove the whole cache, but then we should be consistent and also do it for the custom
CSS / JavaScript and banner, which also has a performance impact: globs and file existence check on each page. Again, not sure how much the impact is, but I also wonder why we should NOT cache it? After I update the code in this file to make it more readable

nickygerritsen:
Merging for now so we can apply it to `wf2020` before going there physically and having to complicate things with USB sticks. If someone wants to change the cache thing, feel free to do so.

PR #1210, opened 2021-09-27
Clean up some things after the problem set page cleanup.

meisterT:


nickygerritsen (approved):

thijskh:
I think you can just drop your changes to the testsuite since I fixed that in main already

meisterT:
Yes, did that just now. There's another fix I want to do there, but it's not urgent.

PR #1213, opened 2021-09-29
Add an option to have judgehosts be disabled when they connect for the first time.

nickygerritsen:


etc/db-config.yaml
```diff
@@ -137,6 +137,11 @@
             default_value: true
             public: false
             description: Allows to distribute one judging to multiple judgehosts.
+        -   name: judgehost_enabled_by_default
+            type: bool
+            default_value: true
+            public: false
+            description: Enable a judgehost by default when a new one connects.
```
meisterT, line 144: s/connects/registers/

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -192,6 +192,7 @@ public function createJudgehostAction(Request $request) : array
         } else {
             $judgehost = new Judgehost();
             $judgehost->setHostname($hostname);
+            $judgehost->setActive((bool)$this->config->get('judgehost_enabled_by_default'));
```
meisterT, line 195: don't we need a `, true` here in the config get call?

meisterT (changes requested):

etc/db-config.yaml
thijskh, line 144: We call this "active" in the UI I think. Although I could see a case that that term is the thing we should change instead of this

webapp/src/Controller/API/JudgehostController.php
nickygerritsen: Nope, it uses `default`

etc/db-config.yaml
nickygerritsen: For now i will make it consistently be `active`. we can change later.

meisterT (approved):

thijskh (approved):

PR #1215, opened 2021-09-29
Fix caching behavior

meisterT:


nickygerritsen (approved):

thijskh:
Would there be drawbacks on always setting the Uuid on judgetask entity construction?

meisterT:
> Would there be drawbacks on always setting the Uuid on judgetask entity construction?

Yes, it needs to group all things of a judging together. It is not unique per judgetask. I made the comment a little bit more clear.

PR #1217, opened 2021-09-29
Put configuration in tabs.

meisterT:
The old page was way too long and unwieldy, the new one should be
easier to work with.


nickygerritsen (approved):

meisterT:
Need to fix the unit tests

PR #1221, opened 2021-09-30
Use badges with colors for problem labels.

nickygerritsen:
image

Loosely based on how the CDS shows it. Thanks @deboer-tim

### DIFF
--- a/webapp/src/Twig/TwigExtension.php
+++ b/webapp/src/Twig/TwigExtension.php
@@ -3,6 +3,7 @@
 namespace App\Twig;
 
 use App\Entity\Contest;
+use App\Entity\ContestProblem;
 use App\Entity\ExternalJudgement;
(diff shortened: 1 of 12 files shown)


webapp/src/Twig/TwigExtension.php
```diff
@@ -1157,4 +1159,23 @@ public function fileTypeIcon(string $type)
 
         return 'fas fa-file-' . $iconName;
     }
+
+    public function problemBadge(ContestProblem $problem): string
+    {
+        $rgb = Utils::convertToHex($problem->getColor());
+        $background = Utils::parseHexColor($rgb);
+        $darker = $background;
```
meisterT, line 1167: please add a comment here what this is roughly trying to do

meisterT (approved):

PR #1227, opened 2021-10-03
Return a 401 if the login fails.

meisterT:


nickygerritsen (approved):

eldering (approved):
Would be good to split out different possible errors into separate codes, but on https://symfony.com/doc/4.4/security/form_login_setup.html I can't find any information about the error message/code.

tuupke:
LGTM, I don't think there are other codes to show here without leaking info.

tuupke (approved):

PR #1235, opened 2021-10-06
Report back compilation metadata and display them in the web interface.

meisterT:
This would have been helpful during finals, where we had to log in and
retrieve the info from the judgehosts ourselves.


meisterT:
Collapsed (default):

image

----------------

Expanded:

image

PR #1244, opened 2021-10-09
Add a warning to the web submit dialog if no file is recent.

### LINKED ISSUE #1230: Provide visual indication of submitting the wrong file
Currently the cli `submit`tool alerts a team via the last change time that a team might be submitting the wrong file, it might be nice if something similar is also possible for the webgui in the form of:
- changed file size
- a single line diff with the last solution to the problem (by that user) in the same language.

meisterT:
We already had a similar warning in the command line submit client but
have seen teams getting bitten by this missing from the web client in a
recent contest.

Fixes #1230.

### DIFF
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -680,6 +680,36 @@ function pinScoreheader()
 	}
 }
 
+function humanReadableTimeDiff(seconds) {
+    var intervals = [
+        ['years', 365 * 24 * 60 * 60],
+        ['months', 30 * 24 * 60 * 60],
+        ['days', 24 * 60 * 60],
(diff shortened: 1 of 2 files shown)


meisterT:
modifiedwarning

meisterT:
> So if I file is saved 4min ago it would not be alerted?

Yep, that's the current behavior.


> I would rather see 1min because I think one saves and will than submit?

I picked 5mins since that matches the behavior of the submit client. 1min may be close to spammy. You save the file, then you compile it locally and then test it locally and only afterwards submit it.
I assume in most cases if you mistakenly chose the wrong file you haven't touched it in the last five minutes either, right?

PR #1253, opened 2021-10-10
Fix #1219: Always wrap combined run&compare scripts

### LINKED ISSUE #1219: Custom build script for interactors
When adding a custom `build` script to an interactor (output validator) for an interactive problem, judging breaks. This is because the default build script for interactive problems [creates a `run` wrapper script](https://github.com/DOMjudge/domjudge/blob/main/judge/judgedaemon.main.php#L352-L401), that now needs to be created manually.

This can be worked around either by appending that code manually to the build file, or just by creating that as a runfile directly.

It appears that the `DOMJUDGE` environment variable is not set when invoking these scripts, so currently it's a bit ugly to detect whether the code is running under DOMjudge (because for BAPCtools you'd need to just call the `exec ./runjury` directly, instead of via the wrapper script).

I think this may be fixed properly by always appending the code mentioned to the `build` file, even if it already exists.

RagnarGrootKoerkamp:
Before, they would only be wrapped if no build file was already present.
With this fix, first we (create and) execute the build file as usual,
and only then do we rename 'run' to 'runjury' and create a new 'run'
file.

This also moves the `touch($execdeploypath)` into the if-statement. In
case you actually care about touching to update the timestamp, I should
move it outside again.

I didn't actually test this yet - will do that in a few days.

Fixes #1219 (apparently putting it in the title and commit message still doesn't make it appear in `linked issues` in the sidebar.)

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -391,7 +391,7 @@ function fetch_executable_internal(
             logmsg(LOG_DEBUG, "Building executable in $execdir, under 'build/'");
             system(LIBJUDGEDIR . '/build_executable.sh ' . dj_escapeshellarg($execdir), $retval);
             if ($retval!==0) {
-                return [null, "Failed to build executable in $execdir.", "$execdir/build.log", null];
+                return [null, "Failed to build executable in $execdir.", "$execdir/build.log"];
             }
             chmod($execrunpath, 0755);
         }


meisterT:
Good news is that I like your changes and they should work. Bad news: I just merged another PR which conflicts with this one, so you need to rebase and resolve conflicts.

RagnarGrootKoerkamp:
Rebased and tested successfully.

RagnarGrootKoerkamp:
done; it's still green

PR #1256, opened 2021-10-12
Collapse contest awards in jury interface to save space. (#1250)

### LINKED ISSUE #1250: Collapse contest awards in jury interface to save space
Currently the contest awards can take up a lot of space in table on the contest page, see screenshot.
It would be nice to collapse it by default for example.
image

shuibinlong:
Feature for fixes #1250.

Using card and collapse components of bootstrap to collapse the categories of the included teams on `jury/contests/{id}` page. The page rendering is as follows.

Collapse:

Expand:

### DIFF
--- a/webapp/templates/jury/contest.html.twig
+++ b/webapp/templates/jury/contest.html.twig
@@ -115,13 +115,13 @@
                     <td>
                         {% if contest.processAwards %}
                             <div class="card">
-                                <h6 class="card-header" role="tab" id="categories">
+                                <h6 class="card-header" id="categories">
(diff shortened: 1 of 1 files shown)


shuibinlong:
Hello friends, @eldering @nickygerritsen, this is a revised version after I have referred to your good suggestions. The following is the demonstration effect. What do you think?
Befor:


After:

nickygerritsen:
I wonder if we should say instead “3 different awards”? It’s very unclear how it gets to 12 now (probably because there are nine teams and three jury members?)

shuibinlong:
Updated.

Before:


After:

PR #1260, opened 2021-10-13
latexmk appears to be required for non-debian 11 bullseye people as well

ubergeek42:
I needed this on an ubuntu 18.04 host I was testing with. Is anyone else able to confirm if this is actually the case or not?


nickygerritsen:
Haven't tested it but we indeed probably need it there as well.

eldering:
I've tested this on the 20.04 WF image, and also there `latexmk` must be installed to make this work.

PR #1266, opened 2021-10-15
Fix #1263: Filter problem filter dropdown for the current contest

### LINKED ISSUE #1263: Submissions 'filter on problems' dropdown shows problems from all contests
When typing, the dropdown shows all problems, instead of only problems from the current contest.

RagnarGrootKoerkamp:
Fix #1263


RagnarGrootKoerkamp:
With #1271, this isn't needed anymore

PR #1271, opened 2021-10-15
Pass options for submissions-filtering to the client

### LINKED ISSUE #1263: Submissions 'filter on problems' dropdown shows problems from all contests

### LINKED ISSUE #1270: Use client side options for the submissions filter form
Currently the submissions filter uses ajax to fetch server side options for the dropdown.
Client side is much more snappy and removes this annoying `please enter 1 or more characters` message.

This is already done when creating a new rejudging (`jury/rejudgings/add`), so I suppose we can just copy parts of the [formbuilder](https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Form/Type/RejudgingType.php) for this.

RagnarGrootKoerkamp:
This just copies and simplifies the Form class for adding a rejudge.

Fix #1270, Fix #1263

### DIFF
--- a/webapp/src/Controller/Jury/SubmissionController.php
+++ b/webapp/src/Controller/Jury/SubmissionController.php
@@ -19,6 +19,7 @@ use App\Entity\Submission;
 use App\Entity\SubmissionFile;
 use App\Entity\Team;
 use App\Entity\Testcase;
+use App\Form\Type\SubmissionsFilterType;
 use App\Service\BalloonService;
 use App\Service\ConfigurationService;
 use App\Service\DOMJudgeService;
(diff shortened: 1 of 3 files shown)


RagnarGrootKoerkamp:
This removes some calls to the `jury_ajax_data` endpoint. Should the handlers for this endpoint also be removed, or are these also used elsewhere?

meisterT:
> This removes some calls to the `jury_ajax_data` endpoint. Should the handlers for this endpoint also be removed, or are these also used elsewhere?

Still in use here: https://github.com/DOMjudge/domjudge/blob/main/webapp/templates/jury/balloons.html.twig#L38

nickygerritsen:
LGTM, anyone opposed to merging?

PR #1275, opened 2021-10-16
Sort clarifications table by newer clarifications first

### LINKED ISSUE #1272: Sort old clarification requests by newest first
Each time I open the clarifications page I have to search for the latest answered clarifications. Submissions are always newest first, so having the same for clarifications is more consistent I think (but feel free to disagree).

Open clarifications should probably still be sorted by oldest first, since that's where you should start answering them, so this would introduce another inconsistency anyway.

RagnarGrootKoerkamp:
Fix #1272

Also simplify some SQL.

### DIFF
--- a/webapp/templates/jury/clarifications.html.twig
+++ b/webapp/templates/jury/clarifications.html.twig
@@ -107,7 +107,7 @@
             var table = $('.data-table').DataTable({
                 "paging": false,
                 "ordering": true,
-                "order": [[0, "desc"]],
+                "order": [],
                 "info": false,
                 "autoWidth": true,
                 "language": {


thijskh:
Well probably you want to either other both sql and datatables in the same direction to avoid confusion as to the intended behaviour, or you disable datatable's default ordering by setting `"order": []` so it's derived from the SQL ordering.

RagnarGrootKoerkamp:
Hmm, now that I think about it, the SQL was already `DESC` originally (so larger ids/newer clarifications first), which is actually what I'm trying to achieve with this PR. So it was already inconsistent, and changing the default DataTables ordering should make it consistent. (Although the SQL sorting is slightly more advanced since it first sorts by time and then by ID. Not sure whether this matters in practice though?)

PR #1276, opened 2021-10-16
Replace jscolor with coloris. Fixes #1267.

### LINKED ISSUE #1267: Contest problems edit UI not optimal
Two issues I noticed while trying to edit contest problem colors, see https://github.com/DOMjudge/domjudge/issues/1264:

- I cannot enter a color by name: any word colors I type are replaced by `#ffffff` when the field gets deselected. It would be nice if HTML color names are recognized.
- When I was entering colors and then pressed <tab> a couple of times to try and navigate to the color field on the next line, the form was automatically submitted. That's annoying.

nickygerritsen:


nickygerritsen:
Not trivial but maybe it is possible. Let me investigate

thijskh:
It looks like a good replacement. I agree with Tobi that the alpha channel is confusing to have in the interface if you can't use it.

Can't we still not simply just use the HTML5 <input type=color> element?

nickygerritsen:
I don’t know how good the html5 one is supported by now. If most browsers support it we can use it

nickygerritsen:
> I don’t know how good the html5 one is supported by now. If most browsers support it we can use it

(Mobile) safari supports it (just tested). I assume chrome goed as well but I can test it and Firefox tomorrow. If it is decent, do we use it?

thijskh:
I read now that the HTML5 color element does not support color names. So that violates the use case that started this PR.

thijskh:
As for the alpha channel, I submitted PR number 1 to coloris yesterday while waiting for fast charging in the middle of the windy NOP, and it was quickly merged. So maybe if it's not possible it's an option to suggest a change to coloris?

meisterT:
See also https://github.com/mdbassit/Coloris/issues/2

nickygerritsen:
> I saw a `mode` thing which I will try out

Ah yeah no that was format. Let’s await that issue or send a PR

PR #1277, opened 2021-10-16
Make table rows instead of table cells clickable

### LINKED ISSUE #1273: Make table rows instead of table cells clickable
In many places only table cells are click targets to open the submission/clarification/... .
This is annoying, because the space in between cells isn't clickable.
Worse, when one cell has a larger height (because of lines wrapping, i.e. the green/red testcase run squares), the other cells don't grow in height and only a small part of the row is actually clickable.

One problem with making the row clickable is that buttons inside the row may be a bit more tricky, for example when hovering the `claim` button on a submission, the mouse won't change shape, but the button will still change to grey on hover.

RagnarGrootKoerkamp:
Adds a `data-link="{{ link }}"` to all table rows where (almost) all
cells have the same click target, and removes the `<a href=..` from the
individual cells.

This needs some javascript in the page footer to handle the clicks,
which is added on (I hope) all relevant pages.
It's probably better to move it to a single central place, but the `domjudge.js`
file looks like it doesn't have similar code at the moment.

There are 2 ways of handling clicks: on the `tr[data-link]` directly,
or via `$('div[data-ajax-refresh-target]').on('click', 'tr[data-link]', ...)`,
which is needed when rows are dynamically added/removed (i.e. for the
auto-refresh submissions table).
We could just use that/something similar everywhere for consistency.

Also adds CSS to change the mouse into a pointer when needed.

Fix #1273

### DIFF
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -708,4 +708,12 @@ function humanReadableBytes(bytes) {
 
 $(function () {
     $('[data-toggle="tooltip"]').tooltip();
+
+    $('tr[data-link]').on('click', function(){
+        window.location=$(this).data('link');
+    });
+
+    $('div[data-ajax-refresh-target]').on('click', 'tr[data-link]', function(){
+        window.location=$(this).data('link');
+    });
 });
--- a/webapp/public/style_jury.css
+++ b/webapp/public/style_jury.css
@@ -162,3 +162,7 @@ div.submission-summary > span {
 .importexport div.tile h3 {
     font-size: 22px;
 }
+
+tr[data-link] {
+    cursor: pointer;
+}
--- a/webapp/templates/jury/contest.html.twig
+++ b/webapp/templates/jury/contest.html.twig
@@ -297,31 +297,27 @@
                     </thead>
                     <tbody>
                     {% for problem in problems %}
-                        <tr>
-                            {% set link = path('jury_problem', {'probId': problem.probid}) %}
-                            <td><a href="{{ link }}">p{{ problem.probid }}</a></td>
-                            <td><a href="{{ link }}">{{ problem.problem.name }}</a></td>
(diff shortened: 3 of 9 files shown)


RagnarGrootKoerkamp:
The alternative solution to handling changing tables is by setting a row-callback when constructing the datatable, but that sounds like more effort. Not sure what's considered cleaner.

RagnarGrootKoerkamp:
This will need tests to be fixed as well, but please first lgtm the actual changes ;)

RagnarGrootKoerkamp:
I definitely missed some places before - I've now just added both selectors to `domjudge.js`.

RagnarGrootKoerkamp:
Ohhh right; I definitely agree that opening in a new tab should still work.

That's mostly fixed by changing the js to:
```

function openDataLink(event) {
    window.location=$(this).data('link');
};

function openDataLinkNewTab(event) {
    // Middle mouse button
    window.open($(this).data('link'), '_blank');
};


$(function () {
    $('[data-toggle="tooltip"]').tooltip();

    $('tr[data-link]').on('click', openDataLink);
    $('tr[data-link]').on('auxclick', openDataLinkNewTab);

    $('div[data-ajax-refresh-target]').on('click', 'tr[data-link]', openDataLink);
    $('div[data-ajax-refresh-target]').on('auxclick', 'tr[data-link]', openDataLinkNewTab);
});
```

but now for submissions it opens 2 new pages because each row is matched by both patterns. Also, this will need further handling for ctrl-click.
Alternatively, we could also keep all the existing `<a href..` tags, but that would probably create more chaos (if middle click works on some parts of the cell but not everywhere)

thijskh:
If you don't mind I'll close this PR since we have the open issue to track the wish and the current approach is not desired and has conflicts.

mvr320:
Discussed with Ragnar, he approves.

PR #1292, opened 2021-10-20
Add support for importing contests and problems using YAML/JSON that reflect the API structure.

nickygerritsen:
So the basis of this PR is to:
* Split importing the contest data into contest and problem data, so we can call them separately.
* Add support for importing JSON as returned by the `contest` and `problems` API endpoints.
* Add support for importing that same format, but then in YAML.

Since the API format only differs from the old YAML format in the names of some keys, the actual import function hasn't changed much. What has changed a lot:
* I added better tests for the imports, and for both formats.
* The UI now says YAML/JSON instead of JSON (which was even wrong) for the import forms.
* The documentation for importing is updated to reflect the new format for contest and problems. To not make the documentation super unreadable, I dropped the old YAML format from it.
* I created a new `import-contest-json.sh` file that imports only JSON's, except for accounts.tsv + it allows both JSON and YAML for contest and problems.

Some things I wonder:
* At some point we should drop the TSV documentation since 'we' (meaning the ICPC community) want to push people towards the JSON's for groups, teams and organizations. The ICPC Tools can easily convert from TSV's returned from the CMS to JSON's. I don't think *now* is that time though.
* At some point we should also drop the old `import-contest.sh`, allowing us to rename the new one to it.
* The new `import-contest-json.sh` requires `jq` if you use a JSON file for the problems. Doesn't seem like a big issue to me (and I suspect JSON for that file is rarely used).
* The CCS requirements still state we need to (only) support JSON for contest and problems, but @meisterT will (hopefully) send a mail and convince them that for these two files YAML make a lot of sense.
* As you might have seen, for the problems import I read both files using the YAML parser, since valid JSON is valid YAML. We can do the same for the contest import. Then we can drop one of the arguments, but if we do that it would be better to name the only argument left `data`. That would be backwards incompatible. Is that an issue?
* I needed a new route for adding problem metadata, since `POST /api/contests/<cid>/problems` was already used. I'm not sure if this is the nicest approach though.


nickygerritsen:
I think I did all comments that were actionable. I also merged the two scripts, it is not too bad. Tested it on the WF data, both in the `config` dir (i.e. TSV's + contest.yaml + problemset.yaml) as well as the `registration` dir where I manually places a `{contest,problems}.json` and `{contest,problems}.yaml` + copied problem directories. All seemed to work.

PR #1300, opened 2021-10-30
Make runpipe faster

RagnarGrootKoerkamp:
needed for interactive problems with a lot of IO (both large volume and many interactions)

### DIFF
--- a/judge/runpipe.cc
+++ b/judge/runpipe.cc
@@ -51,7 +51,7 @@
 using namespace std;
 
 /* Use the POSIX minimum for PIPE_BUF. */
-#define BUF_SIZE 512
+#define BUF_SIZE 1048576
 
 extern int errno;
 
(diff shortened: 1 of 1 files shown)


nickygerritsen:
Note that we ran this during BAPC, since it was needed for problem E and G, which had a lot of interactive data. It worked.

meisterT (approved):

judge/runpipe.cc
```diff
@@ -257,7 +257,7 @@ void pump_pipes(int *fd_out, int *fd_in, int from_val)
 	FD_SET(*fd_out, &readfds);
 
 	tv.tv_sec = 0;
-	tv.tv_usec = 1000; /* FIXME: this is just in order to not block */
+	tv.tv_usec = 20; /* FIXME: this is just in order to not block */
```
eldering, line 260: A lower timeout here means processing data with less latency, but also more polling cycles spent by the `runpipe` program, possibly increasing total wallclock of the whole judging run at a fixed amount of CPU time spent by the submission program. It would be nice to reason about or test this in a bit more detail, but given that this was used without problems at BAPC, I'm ok with this choice.

eldering (approved):

PR #1313, opened 2021-11-02
Disable processing medals by default. Fixes #1309.

### LINKED ISSUE #1309: Do not set awards (4 gold/silver/bronze) by default
It's only at ICPC and subregionals that this is consistently used. It should be a conscious decision to add awards, since this is shown in the resolver.

nickygerritsen:
Also rename awards to medals in the UI, since it better represents the content.

I wonder if we want to rename the fields in the Contest class to medals instead of awards as well. I can do that in two ways:
* Rename it in the existing migration, breaking upgrades from current main to new main.
* Adding an extra migration.

If we rename it (I would be in favour) I'd propose the first, since we have no release yet anyway. Thoughts?


meisterT (approved):

thijskh:
The first

eldering:
> Also rename awards to medals in the UI, since it better represents the content.
> 
> I wonder if we want to rename the fields in the Contest class to medals instead of awards as well. I can do that in two ways:
> 
>     * Rename it in the existing migration, breaking upgrades from current main to new main.
> 
>     * Adding an extra migration.
> 
> 
> If we rename it (I would be in favour) I'd propose the first, since we have no release yet anyway. Thoughts?

Why is medals better than awards? Does this only refer to gold, silver, bronze and never anything else?

nickygerritsen:
> > Also rename awards to medals in the UI, since it better represents the content.
> > I wonder if we want to rename the fields in the Contest class to medals instead of awards as well. I can do that in two ways:
> > ```
> > * Rename it in the existing migration, breaking upgrades from current main to new main.
> > 
> > * Adding an extra migration.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If we rename it (I would be in favour) I'd propose the first, since we have no release yet anyway. Thoughts?
> 
> Why is medals better than awards? Does this only refer to gold, silver, bronze and never anything else?

Yep that is correct. So that makes me think medals is better than awards

nickygerritsen:
I did a `s/award/medal/g` basically now, I think this is better.

webapp/src/Form/Type/ContestType.php
```diff
@@ -76,34 +76,34 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
         ]);
-        $builder->add('processAwards', ChoiceType::class, [
+        $builder->add('processMedals', ChoiceType::class, [
             'expanded' => true,
             'choices' => [
                 'Yes' => true,
                 'No' => false,
             ],
-            'help' => 'Whether to process medal (gold, silver, bronze) awards for this contest.',
+            'help' => 'Whether to process medals (gold, silver, bronze) for this contest.',
```
meisterT, line 85: nit: I don't really like the word "process" here.

I would suggest "enable medals" or "award medals" or "display medals".

nickygerritsen: I agree. I think display is even better, since I believe it is only used for the scoreboard display. I will check if that is indeed the case.
Which makes me wonder, if you disable this, shouldn't we also not expose it in the `awards` API endpoint?

meisterT: Yes, don't show medals if there are no medals

thijskh: The term "process" stems from process balloons because there it actually influences whether we keep track of it at all instead of whether we just display it.

meisterT: Thijs, is that a vote to keep the word process?

thijskh: It's just a piece of trivia

nickygerritsen: > Yes, don't show medals if there are no medals

So then we call the field 'Enable medals' I'd say. Since it fully disables it

nickygerritsen: OK it already disables the API, so renaming the field is all that is left 😄

webapp/src/Controller/Jury/ContestController.php
```diff
@@ -253,7 +253,7 @@ public function indexAction(Request $request, KernelInterface $kernel): Response
 
         $table_fields = array_merge($table_fields, [
             'process_balloons' => ['title' => 'process balloons?', 'sort' => true],
-            'process_awards'   => ['title' => 'process awards?', 'sort' => true],
+            'medals_enabled'    => ['title' => 'medals enabled?', 'sort' => true],
```
thijskh, line 256: title just 'medals'

webapp/src/Entity/Contest.php
```diff
@@ -172,49 +172,49 @@ class Contest extends BaseApiEntity
 
     /**
      * @var boolean|null
-     * @ORM\Column(type="boolean", name="process_awards",
-     *     options={"comment"="Are there awards for this contest?","default"=1},
+     * @ORM\Column(type="boolean", name="medals_enabled",
+     *     options={"comment"="Are medals eanbled for this contest?","default"=1},
```
thijskh, line 176: Besides the typo, the comment does not add any info over the name.

webapp/src/Entity/Contest.php
```diff
@@ -1211,29 +1211,29 @@ public function validate(ExecutionContextInterface $context)
...
         }
 
-        if ($this->processAwards) {
-            if ($this->goldAwards === null) {
+        if ($this->medalsEnabled) {
+            if ($this->goldMedals === null) {
                 $context
-                    ->buildViolation('This field is required when \'Process awards\' is set.')
-                    ->atPath('goldAwards')
+                    ->buildViolation('This field is required when \'Process medals\' is set.')
```
thijskh, line 1217: "Enable medals". Is repeated 3 times below so maybe an indication the code could be a bit less repetitive, at least define the string only once?

thijskh (commented):
I do get the impression that a lot of the code is doing the same thing three times for gold, silver, bronze. Can this be done in a more compact way?

webapp/src/Entity/Contest.php
nickygerritsen: I think we have many many of these comments in the database tables.

nickygerritsen:
I got rid of some duplication.

thijskh (approved):

webapp/src/Controller/Jury/ContestController.php
```diff
@@ -253,7 +253,7 @@ public function indexAction(Request $request, KernelInterface $kernel): Response
 
         $table_fields = array_merge($table_fields, [
             'process_balloons' => ['title' => 'process balloons?', 'sort' => true],
-            'process_awards'   => ['title' => 'process awards?', 'sort' => true],
+            'medals_enabled'    => ['title' => 'medals?', 'sort' => true],
```
mvr320, line 256: Remove the space?

PR #1320, opened 2021-11-05
Show all jury initiated clars as general

thijskh:


nickygerritsen (approved):

webapp/src/Controller/Jury/ClarificationController.php
```diff
@@ -110,7 +110,7 @@ public function indexAction(Request $request): Response
         $wheres            = [
             'new' => 'clar.sender IS NOT NULL AND clar.answered = 0',
             'old' => 'clar.sender IS NOT NULL AND clar.answered != 0',
-            'general' => 'clar.sender IS NULL AND clar.recipient IS NULL',
+            'general' => 'clar.sender IS NULL AND clar.in_reply_to IS NULL',
```
eldering, line 113: This makes sense: now every clarification request/reply is covered since all clarification requests from teams are covered by new/old, while a clarification from the jury is either part of a request thread in old/new as a response, or if not, it is a general clarification that is not a reply to anything.

eldering (approved):

PR #1321, opened 2021-11-06
Codeql analysis

mvr320:
Currently it only finds issues with upstream repos for javascript so this is on the very safe side.

I'm not 100% if the invocation of build-scripts is correct here as I don't know where those should be stored?

I think @meisterT @eldering @thijskh all have an opinion on how such a tool should act so choose them as reviewers.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,100 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  schedule:
+    - cron: '16 12 * * 6'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    env:
+      COMPILED: |
+        [
+          "cpp"
+        ]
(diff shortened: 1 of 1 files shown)


meisterT:
Where can I see a result of this?

mvr320:
> 
> 
> Where can I see a result of this?

Currently it does not have security items, but in the future you can see this under the security tab in GitHub. As I said its currently very on the safe side. The picture here is from an earlier run where I didn't remove the upstream code yet:
image

But it will basicly fail in case someone enters a new bug again (which is non-blocking but should be looked at)

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,100 @@
...
+        ]
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'cpp', 'java', 'javascript', 'python', 'ruby' ]
```
eldering, line 29: Not sure it makes sense to include ruby here, as we have no ruby code (except the test sources).

eldering (approved):
Looks fine to me if this indeed only shows up in a separate security tab that doesn't clutter the PR.

PR #1322, opened 2021-11-06
Conform to openAPI specification

mvr320:
Verification via 42Crunch showed that the we got the wrong resulting
type.

This PR is only to trigger the pipeline, I'll try to add all fixes in 1 go when the githubaction works.


mvr320:
Superseded by: https://github.com/DOMjudge/domjudge/pull/1329

PR #1324, opened 2021-11-06
Judge imported jury submissions with lower priority

meisterT:
This includes PR #1323, so let's decide on that one first.


mvr320:
I think this deserves/needs some tests to verify that what we expect is also really happening. When this fails a contest could take a large hit (as when you actually need this you're already very close to a Murphy's law situation)

meisterT:
I agree and I started writing tests for queuetask and prioritiies in general.

nickygerritsen (approved):

PR #1325, opened 2021-11-07
Shiftleft

mvr320:
I verified the reported issues and none of those seem to be completely false issues.

### DIFF
--- a/lib/lib.error.sh
+++ b/lib/lib.error.sh
@@ -20,7 +20,7 @@ logmsg ()
 
 	msglevel="$1"; shift
 	msec=`date '+%N'`
-	stamp="[`date '+%b %d %T'`.`printf "%.3s" $msec`] $PROGNAME[$$]:"
+	stamp="[`date '+%b %d %T'`.`printf "%.3s" $msec`] ${PROGNAME}[$$]:"
 # Expand backslash escape sequences using printf, since support for
 # echo with(out) option -e is unreliable. First use echo to put spaces
 # between arguments.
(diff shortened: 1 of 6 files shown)


lib/lib.error.sh
```diff
@@ -20,7 +20,7 @@ logmsg ()
 
 	msglevel="$1"; shift
 	msec=`date '+%N'`
-	stamp="[`date '+%b %d %T'`.`printf "%.3s" $msec`] $PROGNAME[$$]:"
+	stamp="[`date '+%b %d %T'`.`printf "%.3s" $msec`] ${PROGNAME}[$$]:"
```
mvr320, line 23: @eldering I think this is more correct but I'm not 100% sure if this has no unwanted side effects?

mvr320:
The last error:
```
Argument mixes string and array. Use * or separate argument.

if [ "$@" ]; then
		logmsg $LOG_ERR "error: $@"
```

in lib/lib.error.sh is I think already discussed in code but I'm not sure if it can currently be handled or if general support for `echo -e` is still in a bad state?

lib/lib.error.sh
eldering: No, that's fine. Actually less likely to have unwanted side effects, since it's clearer what the variable name is.

eldering:
> The last error:
> 
> ```
> Argument mixes string and array. Use * or separate argument.
> 
> if [ "$@" ]; then
> 		logmsg $LOG_ERR "error: $@"
> ```
> 
> in lib/lib.error.sh is I think already discussed in code but I'm not sure if it can currently be handled or if general support for `echo -e` is still in a bad state?

I think both of these should be `$*`, there's a subtle difference (that I also used to do wrong), see `man bash`.

nickygerritsen (approved):

PR #1331, opened 2021-11-11
Mayhem

mvr320:
Add a fuzzer for the API.

Update, because I need to add a secret for this to show results, here some results of how it looks for my fork.
image
image
image

I think this test is useful, the only thing I'm not 100% sure of if it will cleanup security items or that we will need to do this manually, because the doc file is not committed but live generated.

### DIFF
--- a/.github/workflows/mayhem-api.yml
+++ b/.github/workflows/mayhem-api.yml
@@ -1,0 +1,91 @@
+name: "Mayhem API analysis"
+
+on:
+  schedule:
+    - cron: '5 21 * * *'
+
+jobs:
+  mayhem:
+    name: Mayhem API analysis
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+    env:
+      DB_DATABASE: domjudge
+      DB_USER: user
+      DB_PASSWORD: password
+    steps:
+      - uses: actions/checkout@v2
+
+      # Install default DOMjudge dommserver apt packages
+      - name: Install domserver required apt packages
+        run: |
+          sudo apt update
+          sudo apt install -y acl zip unzip nginx php php-fpm php-gd \
+                              php-cli php-intl php-mbstring php-mysql php-curl php-json \
+                              php-xml php-zip composer ntp make sudo debootstrap \
+                              libcgroup-dev lsof php-cli php-curl php-json php-xml \
+                              php-zip procps gcc g++ default-jre-headless \
+                              default-jdk-headless ghc fp-compiler autoconf automake bats \
+                              python3-sphinx python3-sphinx-rtd-theme rst2pdf fontconfig \
+                              python3-yaml latexmk
+
+      - name: Install composer files
+        run: |
+          composer install --no-scripts
+          composer run-script package-versions-dump
+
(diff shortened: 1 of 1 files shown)


.github/workflows/mayhem-api.yml
```diff
@@ -0,0 +1,97 @@
...
+          make configure
+          ./configure --with-baseurl='http://localhost/domjudge/' --enable-doc-build=no --prefix=$HOME/domjudge
+
+      # sudo because its default in the manual
+      - name: Compile domserver
+        run: |
+          make domserver
+          sudo make install-domserver
+
+      # GitHub action sometimes doesnt start the SQL server.
```
nickygerritsen, line 59: https://github.blog/changelog/2020-02-21-github-actions-breaking-change-ubuntu-virtual-environments-will-no-longer-start-the-mysql-service-automatically/

I think 'sometimes' is 'always'

nickygerritsen (commented):
I want to look into moving more/all from GitLab CI to GitHub Actions so at some point we might want to move some of this into a shared `.sh` file?

.github/workflows/mayhem-api.yml
mvr320: I'll remove the comment

mvr320:
> 
> 
> I want to look into moving more/all from GitLab CI to GitHub Actions so at some point we might want to move some of this into a shared `.sh` file?

We could do that, as I now have this multiple times in some tests (but will make a different PR for that).

Regarding the moving the CI to GHA, I wouldn't do that yet as I think it will make the CI take longer as not all jobs are started at the same time.

I personally still find the gitlab-ci syntax much clearer and easier to debug than the GHA + documentation for GHA on how to do something still not as good as the GitLab docs.

nickygerritsen:
> > I want to look into moving more/all from GitLab CI to GitHub Actions so at some point we might want to move some of this into a shared `.sh` file?
> 
> We could do that, as I now have this multiple times in some tests (but will make a different PR for that).
> 
> Regarding the moving the CI to GHA, I wouldn't do that yet as I think it will make the CI take longer as not all jobs are started at the same time.
> 
> I personally still find the gitlab-ci syntax much clearer and easier to debug than the GHA + documentation for GHA on how to do something still not as good as the GitLab docs.

OK then let's not do this now. I do like that it is all integrated in GitHub but if it is more messy, slower and less documented this is something for later.

.github/workflows/mayhem-api.yml
```diff
@@ -0,0 +1,97 @@
+name: "Mayhem API analysis"
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
mvr320, line 7: Given that we can run 50 jobs (in my account) I think we should run this once per day just on the main branch leaving 20 jobs left unused? Also I didn't think about the secret on forks yet, so if this can even run if someone opens a PR from their fork.

mvr320:
@nickygerritsen given that my fork cannot read the secret, the test will not pass so can you look at it and approve if you believe that it will work on main?

nickygerritsen (approved):
Merge and see what happens

PR #1334, opened 2021-11-13
Crunch42 analysis

vmcj:
Given that more issues are currently about the API I think its relevant to check the API in total.

This needs a key for Crunch42 but the findings there seem to make sense and are well explained.
The findings are also added to the security tab. This does currently add quite a lot but I think that does make sense.

### DIFF
--- a/.github/workflows/crunch42-analysis.yml
+++ b/.github/workflows/crunch42-analysis.yml
@@ -1,0 +1,42 @@
+name: "42Crunch REST API Static Security Testing"
+
+# follow standard Code Scanning triggers
+on:
+  push:
+    branches: [ main ]
+  pull_request_target:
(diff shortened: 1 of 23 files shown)


nickygerritsen:
What does this exactly show/do?

mvr320:
> 
> 
> What does this exactly show/do?

It verifies the OpenAPI specification for missing parameters etc.

image

The items are added to the security tab so are easy to fix. I think due to how the workflow triggers this should now be possible to run on forks also but there is a small security issue if we don't use an additional Action due to the script which can be changed from the fork (if I understand it correctly).

https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request_target

nickygerritsen (approved):

PR #1335, opened 2021-11-15
Add support for uploading assets. Fixes #1317.

### LINKED ISSUE #1317: Support uploading team affiliations logo on the team affiliations page
If I want to add a logo to an affiliation on the current page `/jury/affiliations`, I need to manually upload the logo image file to the `domserver/webapp/public/images/affiliations` folder, which brings a burden to both management and operation and maintenance personnel.

Can we provide a quick upload interface (such as upload button in `/jury/affiliations/{id}/edit` page) on the web page? When the administrator chooses to upload the team affiliation logo, it can be automatically placed in the corresponding folder through this interface.

nickygerritsen:
To make this (more) easily implementable I had to do two things:

* Replace the single `webapp/public/images/bannner.png` with `webapp/public/images/bannners/*.png`, one per contest. I see this as an improvement over the old way.
* Drop caching these assets. It is near impossible to keep the caches up to date like this, since you can not clear the cache in a web request (believe me, I tried). It seems there is no serious performance impact anyway, so I think this is fine. If you agree, we can (later) also drop the `SetDocLinksPass`. That impact will be a bit more, but **only** if you are on the docs page.

I changed the existing tests to use the new endpoints instead of using hardcoded logic to upload images. We should also add tests for the UI forms, but I leave that to @mvr320 😅 .

Also we can later modify `import-contest.sh` to also import these images if present in a CAF 😄


doc/manual/config-advanced.rst
```diff
@@ -16,19 +16,20 @@ path of your installation) as follows:
 - *Affiliation logos*: these will be shown with the teams that are
   part of the affiliation, if the ``show_affiliation_logos`` configuration
   option is enabled. They can be placed in
-  `public/images/affiliations/1234.png` where *1234* is the numeric ID
+  `public/images/affiliations/1234.png` where *1234* is the nuermic ID
```
mvr320, line 19: Typo

doc/manual/config-advanced.rst
```diff
@@ -23,12 +23,13 @@ path of your installation) as follows:
 - *Team pictures*: a photo of the team will be shown in the team details
   page if `public/images/teams/456.jpg` exists, where *456* is the
   team's numeric ID as shown in the DOMjudge interface.
-- *Banner*: a page-wide banner can be shown on the public scoreboard
-  if that image is placed in `public/images/banner.png`.
+- *Contest Banners*: a page-wide banner can be shown on the public scoreboard
+  if that image is placed in `public/images/banners/1.png` where *1* is the
+  contest's numeric ID as shown in the DOMjudge interface.
```
meisterT, line 28: There's no file that applies to all?

doc/manual/config-advanced.rst
```diff
@@ -23,12 +23,13 @@ path of your installation) as follows:
...
 
 .. note::
 
-  The IDs for affiliations and teams need to be the *external ID*
```
meisterT, line 31: no team images?

meisterT (commented):
do we want to / need to make sure that the assets are not too large?

doc/manual/config-advanced.rst
nickygerritsen: Nope, I'd say if we improve our import-contest.sh to also read the banner while importing the contest that is not a big problem. Or do you disagree? We could add a fallback to `images/banners/global.png` or `images/banner.png`.

doc/manual/config-advanced.rst
nickygerritsen: Teams do not have an external ID for some bizarre reason. I still propose we add it at some point (since now we can't do non-numeric team ID's) but until then nope.

doc/manual/config-advanced.rst
meisterT: I actually do disagree. Using `import-contest` is just one of way setting up a contest and the banner will be the same in most situations across contests. So basically having a fallback banner is less work and trivial to implement.

eldering: +1. I've used this at our company prelim round, and I find it very convenient that I can just upload one (default) banner.

webapp/src/Controller/API/ContestController.php
```diff
@@ -200,6 +208,90 @@ public function bannerAction(Request $request, string $id): Response
...
+        if ($contest === null) {
+            throw new NotFoundHttpException(sprintf('Object with ID \'%s\' not found', $id));
+        }
+
+        /** @var UploadedFile $banner */
+        $banner = $request->files->get('banner');
+
+        if (!$banner) {
+            throw new BadRequestHttpException("Please supply a banner");
+        } else if ($banner->getMimeType() !== 'image/png') {
```
eldering, line 282: Have we always only supported PNG? Why not also support JPEG and maybe SVG too (although I understand that might be more tricky with pixel size)?

webapp/src/Controller/API/TeamController.php
```diff
@@ -124,6 +148,90 @@ public function photoAction(Request $request, string $id): Response
...
+            throw new NotFoundHttpException(sprintf('Object with ID \'%s\' not found', $id));
+        }
+
+        /** @var UploadedFile $photo */
+        $photo = $request->files->get('photo');
+
+        if (!$photo) {
+            throw new BadRequestHttpException("Please supply a photo");
+        } else if ($photo->getMimeType() !== 'image/jpeg') {
+            throw new BadRequestHttpException("Only JPG's are supported");
```
eldering, line 223: Why only JPEG here?

webapp/src/Service/AssetUpdateService.php
```diff
@@ -0,0 +1,79 @@
...
+        if (!$assetFields) {
+            return;
+        }
+
+        $fs = new Filesystem();
+
+        foreach ($assetFields as $assetField) {
+            $assetPath = $this->dj->fullAssetPath($entity, $assetField, $this->eventLog->externalIdFieldForEntity($entity) !== null);
+            $clearMethod = sprintf('isClear%s', ucfirst($assetField));
+            $shouldClear = call_user_func([$entity, $clearMethod]);
```
eldering, line 68: Can we solve this in a different way? Doing such dynamic function calls makes the code less discoverable and I'd prefer to avoid this.

doc/manual/config-advanced.rst
nickygerritsen: It was not 100% trivial (since I think we don't want to be able to delete the global banner using the contest edit form) but I added it back.

webapp/src/Controller/API/ContestController.php
nickygerritsen: We have indeed only supported PNG for these (and JPG for photo's). I'd like to add support for all three, but I want to do that in a separate PR since it involves more changes (i.e. also the GET request in the API).

webapp/src/Service/AssetUpdateService.php
nickygerritsen: I welcome ideas. I could put a `switch` statement here that loops over all types, basically getting rid of the constant. But then you have logic all over the place, also doesn't sound ideal.

Another option would be to add a method to each of the three classes (Team, TeamAffiliation and Contest) with

```
public function isClearAsset(string $field) {}
// and
public function getAssetFile(string $field) {}
```
which both have a switch statement. Maybe that is better?

webapp/src/Controller/API/TeamController.php
nickygerritsen: See remark above.

webapp/src/Controller/API/ContestController.php
nickygerritsen: > do we want to / need to make sure that the assets are not too large?

If we add SVG support (which I would like) this is already becoming hard. Or are you proposing filesize? We can add that. What do you propose in that case for size?

thijskh: svg is also dangerous so likely we'd need some validation?

nickygerritsen: Only admins can do all this right? I mean, you can also upload malicious problems in that case I guess?

thijskh:
As for the caching. Since the files are in the docroot, Symfony will already be bypassed completely if they exist.

nickygerritsen:
> As for the caching. Since the files are in the docroot, Symfony will already be bypassed completely if they exist.

I was talking about the `readdir` (used to see what files are there), which in theory can be slow but in practice we will not notice.

webapp/src/Service/AssetUpdateService.php
eldering: I don't have a perfect overview, but I think adding this latter `public function isClearAsset(string $field)` and `getAssetFile` might be nicer. I'm wondering what others think @thijskh @meisterT ?

nickygerritsen: I also think it looks better. I will give it a try tomorrow unless someone objects

eldering: The `getApiId` fix is nice BTW: that cleans things up a lot!

nickygerritsen: In the end it still has a magic method call but then only in one place. I like it as well

PR #1339, opened 2021-11-17
[8.1] Malformed CIDs do not pass URL validation.

vmcj:
Documentation needs 400 for all inputs, so we loose currently a more
specific message. Alternative is duplication of code.

### DIFF
--- a/webapp/src/Controller/API/ContestController.php
+++ b/webapp/src/Controller/API/ContestController.php
@@ -41,6 +41,7 @@ use Symfony\Component\Yaml\Yaml;
  * @OA\Tag(name="Contests")
  * @OA\Response(response="404", ref="#/components/responses/NotFound")
  * @OA\Response(response="401", ref="#/components/responses/Unauthorized")
+ * @OA\Response(response="400", ref="#/components/responses/InvalidResponse")
  */
 class ContestController extends AbstractRestController
 {
@@ -235,10 +236,6 @@ class ContestController extends AbstractRestController
      * @OA\Response(
      *     response="200",
      *     description="Contest start time changed successfully",
-     * )
-     * @OA\Response(
-     *     response="400",
-     *     description="Invalid input data"
      * )
      * @OA\Response(
      *     response="403",
--- a/webapp/src/Controller/API/AwardsController.php
+++ b/webapp/src/Controller/API/AwardsController.php
@@ -25,6 +25,7 @@ use Symfony\Component\Intl\Exception\NotImplementedException;
  * @OA\Parameter(ref="#/components/parameters/cid")
  * @OA\Response(response="404", ref="#/components/responses/NotFound")
  * @OA\Response(response="401", ref="#/components/responses/Unauthorized")
+ * @OA\Response(response="400", ref="#/components/responses/InvalidResponse")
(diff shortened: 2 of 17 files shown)


nickygerritsen:
Could we not limit what is allowed as ID's? See https://ccs-specs.icpc.io/draft/contest_api#json-attribute-types what is allowed:
> `[a-zA-Z0-9_.-]` of length at most 36 and not starting with a `-` (dash) or `.` (dot) or ending with a `.` (dot).

mvr320:
> 
> 
> Could we not limit what is allowed as ID's? See https://ccs-specs.icpc.io/draft/contest_api#json-attribute-types what is allowed:
> 
> > `[a-zA-Z0-9_.-]` of length at most 36 and not starting with a `-` (dash) or `.` (dot) or ending with a `.` (dot).

I think that is a good idea but:
- Still if someone provides "wrong" data nginx will return a 400, the problem is that the fuzzer provides this wrong data and receives a 400 which was never stated.
- The documentation should cover all the states AND declare what we expect (the regex you gave).

Overall I think this is also just a mistake with OpenAPI, I think it should always be implicit that when you provide something wrong you're going to get a 400.

nickygerritsen:
Alternative: can we write something (some event subscriber or similar) that adds 400 to all calls?  I remember I did something similar in the past

mvr320:
> 
> 
> Alternative: can we write something (some event subscriber or similar) that adds 400 to all calls? I remember I did something similar in the past

I think we can define it in 1 file, you will still lose the specific message, so is your problem with the duplication of code or with the problem that we lose a more specific error message?

nickygerritsen:
> > Alternative: can we write something (some event subscriber or similar) that adds 400 to all calls? I remember I did something similar in the past
> 
> I think we can define it in 1 file, you will still lose the specific message, so is your problem with the duplication of code or with the problem that we lose a more specific error message?

I like less copy paste. What specific messages do we have?

webapp/src/Controller/API/ContestController.php
```diff
@@ -237,10 +238,6 @@ public function bannerAction(Request $request, string $id): Response
      *     description="Contest start time changed successfully",
      * )
      * @OA\Response(
-     *     response="400",
-     *     description="Invalid input data"
```
mvr320, line 241: @nickygerritsen this one.

nickygerritsen: Yeah I don't care about that one at all 😄

thijskh:
I don’t see any change in Nginx config?

mvr320:
> I don’t see any change in Nginx config?

nginx+apache are not wrong,

URL:
domjudge/api/contests/% should throw 400, the problem is just that we don't declare that that can happen.

thijskh:
So what does the nginx in the title of this PR refer to if nginx is not involved?

mvr320:
> So what does the nginx in the title of this PR refer to if nginx is not involved?

That I initially had this on nginx, but that it should be fixed in our API spec.

vmcj:
Merged for the DJ80 release and @nickygerritsen his suggestion will be implemented for DJ81

PR #1344, opened 2021-11-19
Fix Mayhem for API invocation

swarkentin:
I saw that Mayhem for API was added, but it doesn't 
look like the right endpoints are being hit.

The spec appears to already contain the `/api` prefix
in all paths - so no need to include it in the `api-url`
argument.


mvr320 (approved):

mvr320:
@swarkentin thank you for the fix.

swarkentin:
> @swarkentin thank you for the fix.

Happy to help!

Issue #1345, opened 2021-11-19
Add 'ignored' flag to Clarifications API responses


brettreyno:
When viewing a clarification, DOMJudge offers jury users the option to click a "Set Answered" button, which moves the clarification into the "answered" section of the main clarification window and decrements the number of "open" clarifications displayed in the navigation bar across the top of the screen. However, when querying the API for clarifications, there is no data to indicate that a clarification has been ignored in this manner. Would it be possible to add a boolean 'ignored' flag or similar to the API responses to allow detection of this scenario?

thijskh:
We store this in a boolean field 'answered' with each clarification (so there's no special case for ignored ones, this is always set to true either by pressing the button you mention or by actually answering it). Would that work for you?

We can trivially add this boolean `answered` to the API output. Probably only for nonstrict mode since the spec does not define it (looking at the specs gurus here).

brettreyno:
That's perfect, yes! I'm currently identifying "answered" clarifications by scanning through the whole list for one with a matching "reply_to_id," so that would significantly reduce my processing time.

thijskh:
Question is whether it’s “bad” to expose this state to a team that sent it. I’m leaning to ‘no’.

brettreyno:
The only thing they'd gain from having access to that field (that they wouldn't already have anyway) is an answer to the question "Are the judges actively ignoring me?" It's probably best not answering that, yes, but on the other hand, the answer starts to make itself obvious if they haven't gotten a reply within a reasonable amount of time. ;-)

Is it possible to simply leave the value null for non-jury/API-reader users?

nickygerritsen:
We could also only expose it to admins or is that not ideal?

thijskh:
This is possible I think(but less straightforward)

PR #1347, opened 2021-11-20
Expose 'answered' boolean in clarifications API (api_reader only)

thijskh:
This indicates whether or not the clarification is "handled", either
actually replied to or marked has handled through the interface.

The information is only available for privileged users. We introduce
a new serialization group "Restricted" for properties that are only
released to users with a role that includes api_reader. Because this
specific attribute is also not in the spec, we do not want to expose
it when strict=true. Since groups are "or"-ed, we need to add a
combined group of Restricted and Nonstrict. To summarise:

- Set "Restricted" on attributes that need to be exposed in
  strict (and regular) mode only to restricted users.
- Set "Nonstrict" on attributes that need to be exposed to
  all users when strict is false.
- Set "RegistrictedNonstrict" on attributes that need to be
  exposed to restricted users only when strict is false.

Closes: #1345

### DIFF
--- a/webapp/src/Controller/API/AbstractRestController.php
+++ b/webapp/src/Controller/API/AbstractRestController.php
@@ -134,6 +134,12 @@ abstract class AbstractRestController extends AbstractFOSRestController
         if (!$request->query->has('strict') || !$request->query->getBoolean('strict')) {
             $groups[] = 'Nonstrict';
         }
+        if ($this->dj->checkrole('api_reader')) {
+            $groups[] = 'Restricted';
+        }
+        if (in_array('Nonstrict', $groups) && in_array('Restricted', $groups)) {
+            $groups[] = 'RestrictedNonstrict';
+        }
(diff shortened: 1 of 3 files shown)


webapp/src/Controller/API/AbstractRestController.php
```diff
@@ -134,6 +134,12 @@ protected function renderData(
         if (!$request->query->has('strict') || !$request->query->getBoolean('strict')) {
             $groups[] = 'Nonstrict';
         }
+        if ($this->dj->checkrole('api_reader')) {
+            $groups[] = 'Restricted';
```
nickygerritsen, line 138: Nice! I think we have more fields we only expose to api readers (submission files link at least) which we could also use this for. We now do (ugly) stuff with an event subscriber (see `SubmissionVisitor`)

nickygerritsen (approved):

PR #1357, opened 2021-11-23
Advice teams about creating temporary files

mvr320:


doc/manual/team.rst
```diff
@@ -319,6 +319,10 @@ memory
   If your program tries to use more memory, it will most likely abort,
   resulting in a run error.
 
+creating new files
+  You are not supposed to create temporary files. The sandbox will not allow this and
+  the submission will likely fail with a runtime error.
```
thijskh, line 324: The reference to "sandbox" here comes out of nothing since the term is used and explained for the first time in the paragraph following this one. So take those words here or move this paragraph down.

Why "likely"? Can't we be more definitive in describing our behaviour?

s/temporary/new/
s/You are not supposed to /Do not/?

eldering: The current "will fail" is not entirely correct: creating the file *will* fail. But what happens then (that is, whether it triggers a runtime error) depends on how the submission and language handles that file creation failure. For example in C/C++ this is just returning the success status of the operation, but will not abort or throw an exception.

thijskh (approved):

nickygerritsen (approved):

eldering (approved):

PR #1358, opened 2021-11-23
Remove scoreboard tsv

vmcj:


nickygerritsen (approved):

webapp/tests/Unit/Controller/Jury/ImportExportControllerTest.php
```diff
@@ -21,9 +21,6 @@ public function testIndexBasic() : void
             self::assertSelectorExists(sprintf('h2:contains("%s")', $section));
         }
         self::assertSelectorExists('small:contains(\'Create a "Web Services Token"\')');
-
-        // We've reached the end of the page.
-        self::assertSelectorExists('li:contains("scoreboard.tsv")');
```
thijskh, line 26: As the comment says the intention was so find out if we see some item near the end of the page so we have some basic certainty that the whole thing is rendered. So I would not remove it but put an assertion for a thing near it in its place.

meisterT: @vmcj I did push a commit to address this now. Let's merge this?

meisterT (approved):

webapp/tests/Unit/Controller/Jury/ImportExportControllerTest.php
vmcj: Let me fixup my things, and then I'll merge it yes.

meisterT:
Squash and go!

PR #1359, opened 2021-11-23
Send back to referrer (if local) on a lot of forms on the submission pages. Fixes #1355.

### LINKED ISSUE #1355: Send back to previous page instead of defined page on some forms
For example when claiming or verifying a problem you get send back to the submission page. Better would be to send back to the referrer

nickygerritsen:


webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -1102,7 +1106,7 @@ public function verifyAction(
             $redirect = $this->generateUrl('jury_submission_by_judging', ['jid' => $judgingId]);
         }
 
-        return $this->redirect($redirect);
+        return $this->redirectToLocalReferrer($this->router, $request, $redirect);
```
thijskh, line 1109: Old line 1096/new line 1100 above has a comment that describes the logic that follows, i think with this change the described logic/distinction is undone?

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -1213,9 +1217,13 @@ protected function processClaim($judging, Request $request, ?string &$claimWarni
...
                 if ($action === 'claim') {
-                    return $this->redirectToRoute('jury_submission', ['submitId' => $judging->getSubmission()->getSubmitid()]);
+                    return $this->redirectToLocalReferrer($this->router, $request,
+                        $this->generateUrl('jury_submission', ['submitId' => $judging->getSubmission()->getSubmitid()])
+                    );
                 } else {
-                    return $this->redirectToRoute('jury_submissions');
+                    return $this->redirectToLocalReferrer($this->router, $request,
+                        $this->generateUrl('jury_submissions')
+                    );
```
thijskh, line 1226: Maybe (?) symmetrically, if you unclaim something you're done with it and will return to the list

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -1136,7 +1140,7 @@ public function verifyShadowDifferenceAction(
             $redirect = $this->generateUrl('jury_submission_by_external_judgement', ['externalJudgement' => $extjudgementid]);
         }
 
-        return $this->redirect($redirect);
+        return $this->redirectToLocalReferrer($this->router, $request, $redirect);
```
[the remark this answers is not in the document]
nickygerritsen: I see the comment but can't seem to find where we set the `redirect` var. I think in this case it's better to always redirect to the referrer if possible.

webapp/src/Controller/Jury/SubmissionController.php
nickygerritsen: Same here

webapp/src/Controller/Jury/SubmissionController.php
nickygerritsen: Discussed with @eldering and @meisterT and we think that is not better. So I will leave this part as this PR does it.

webapp/src/Controller/Jury/SubmissionController.php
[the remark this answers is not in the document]
thijskh: Yes, so it's better now to update the comment and remove the now definitely unused code that dealt with it.

thijskh (approved):

PR #1360, opened 2021-11-23
Hide testcase boxes on submission overview page when showing more than 50 submissions. Fixes #1356.

### LINKED ISSUE #1356: Hide testcase boxes when viewing all submissions
When showing all submissions (so not recent or unjudged) we might want to hide testcase boxes for performance reasons. We should investigate first if it is (a lot) faster.

nickygerritsen:


mvr320 (approved):

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -199,11 +199,11 @@ public function indexAction(Request $request): Response
             'filteredResults' => $filteredVerdicts,
             'showExternalResult' => $this->config->get('data_source') ==
                 DOMJudgeService::DATA_SOURCE_CONFIGURATION_AND_LIVE_EXTERNAL,
+            'showTestcases' => count($submissions) <= 50,
```
thijskh, line 202: I think this magic 50 is related to the same 50 used on line 139. Please store this cutoff number in one place only.

thijskh (changes requested):

webapp/src/Controller/Jury/SubmissionController.php
nickygerritsen: Done

thijskh (approved):

PR #1361, opened 2021-11-23
Add a test for writing a local file.

eldering:
There is no clear specification on what should happen but
in DOMjudge we disallow creating files inside the chroot.
This checks that behaviour.


mvr320 (approved):
Thanks, that's indeed a working way of not getting RE although files are tried to be opened. I'll reformulate my PR to explain the expected behavior for a team.

nickygerritsen (approved):

PR #1362, opened 2021-11-23
Add option to link team to existing user upon create. Fixes #1206.

### LINKED ISSUE #1206: Make it easier to add a new team to an existing user
Currently when you add a team, you have

the option to **not** add a new user to the team: 
one
 
and the option to add a new user to the team:
two

But we don't offer the option to pick an **existing** user here. You have to create the team and then in a separate step do the association which is a bit annoying.

nickygerritsen:
I had to move some logic for creating the user from the form type to the controller but I think it makes more sense there anyway.


webapp/src/Controller/Jury/TeamController.php
```diff
@@ -421,21 +422,33 @@ public function deleteAction(Request $request, int $teamId)
...
-            } else {
-                // Otherwise, set the user's name to the team name
+            if ($team->getAddUserForTeam() === Team::CREATE_NEW_USER) {
+                // Create a user for the team
+                $user = new User();
+                $user->setUsername($team->getAddUserForTeam());
+                $team->addUser($user);
+                // Make sure the user has the team role to make validation work
+                /** @var Role $role */
+                $role = $this->em->createQueryBuilder()
```
thijskh, line 440: Is this the best way we can retrieve a Role? Just wondering if we could make some use of some helper for that maybe.

thijskh (approved):

webapp/src/Controller/Jury/TeamController.php
nickygerritsen: Lol I copied this from the previous place but it is indeed super ugly. Pushing a new version now.

meisterT (approved):

PR #1365, opened 2021-11-24
Test compiling a large static binary.

eldering:


tests/test-compile-large-binary.cc
```diff
@@ -0,0 +1,18 @@
...
+
+#include <stdio.h>
+
+const int N = 64 * 1024 * 1024;
+
+char array[N] = {1};
+
+int main()
+{
+	printf("Successfully compiled with a %d MB static array.\n", N / 1024 / 1024);
```
mvr320, line 16: Because array is never used, should it not be sizeof(array) because otherwise this can be removed by the compiler?

meisterT: it's statically initialized and thus can't be removed

eldering: For clarity I've used `sizeof` while making some other small fixes.

PR #1366, opened 2021-11-24
Document run-user-n and the corresponding isolcpus

vmcj:
As discussed at NWERC21, this should fix the documentation and explain the isolcpus.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -245,6 +245,7 @@ integration_mysql:
     MYSQL_ROOT_PASSWORD: password
     MARIADB_PORT_3306_TCP_ADDR: sqlserver
     MYSQL_REQUIRE_PRIMARY_KEY: 1
+    PIN_JUDGEDAEMON: 1
 
 integration_mariadb:
   <<: *job_integration
@@ -252,6 +253,15 @@ integration_mariadb:
   variables:
     MYSQL_ROOT_PASSWORD: password
     MARIADB_PORT_3306_TCP_ADDR: sqlserver
+    PIN_JUDGEDAEMON: 1
+
+integration_unpinned_judgehost:
+  <<: *job_integration
+  <<: *maria_db
+  variables:
+    MYSQL_ROOT_PASSWORD: password
+    MARIADB_PORT_3306_TCP_ADDR: sqlserver
+    PIN_JUDGEDAEMON: 0
 
 phpcs_compatibility:
   <<: *tiny_job
--- a/doc/manual/install-judgehost.rst
+++ b/doc/manual/install-judgehost.rst
@@ -149,6 +149,7 @@ commandline to
 Optionally the timings can be made more stable by not letting the OS schedule
 any other tasks on the same CPU core the judgedaemon is using:
 ``GRUB_CMDLINE_LINUX_DEFAULT="quiet cgroup_enable=memory swapaccount=1 isolcpus=2"``
+
 On modern distros (e.g. Debian bullseye) which have cgroup v2 enabled by
 default, you need to add ``systemd.unified_cgroup_hierarchy=0`` as well.
 Then run ``update-grub`` and reboot.
@@ -200,9 +201,10 @@ Finally start the judgedaemon::
 
 Upon its first connection to the domserver API, the judgehost will be
 auto-registered and will be by default enabled. If you wish to
-add a new judgehost but have it initially disabled, you can add it
-manually through the DOMjudge web interface and set it to disabled
-before starting the judgedaemon.
+add a new judgehost but have it initially disabled, you can change the config
+setting to automatically pause judges on first connection or manually add it
+through the DOMjudge web interface and set it to disabled before starting
+the judgedaemon.
 
 The judgedaemon can also be run as a service by running::
 
--- a/doc/manual/install-judgehost.rst
+++ b/doc/manual/install-judgehost.rst
@@ -50,15 +50,33 @@ home directory::
   make judgehost
   sudo make install-judgehost
 
+The judgedaemon can be run on various hardware configurations;
+- A virtual machine, typically these have 1 or 2 cores and no hyperthreading, because the kernel will schedule its own tasks on CPU 0, we advice CPU 1,
+- A default office machine, these sometimes have hyperthreading. Verify if the machine has hyperthreading and consider turning it off and as a rule of thumb pick CPU 2 as CPU 1 could be a hyperthreading core, be on the same die as CPU 0 and therefore share memory with that CPU. If more cores available as a rule of thumb moving to the highest CPU should be considered.
+- Multiple on a single high-grade server with multiple CPUs or a CPU with multiple cores. Check for hyperthreading and if possible run the judgedaemons on separate CPU packages/dies both from each other and when possible different from CPU 0. See the section :ref:`multiple-judgedaemons` for running multiple judgedaemons on a single host.
+
+For the next section we assume a machine with possibly hyperthreading and 3 or more CPUs. This can be checked with::
+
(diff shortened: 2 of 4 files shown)


doc/manual/install-judgehost.rst
```diff
@@ -53,12 +53,13 @@ home directory::
 For running solution programs under a non-privileged user, a user and group have
 to be added to the system that acts as judgehost. This user does not
 need a home-directory or password, so the following command would
-suffice to add a user and group ``domjudge-run-1`` with minimal privileges::
+suffice to add a user and group ``domjudge-run-2`` with minimal privileges
+with the judgedaemon restricted to CPU core 2::
```
thijskh, line 57: Do we need to say somehwere that you need a machine with at least 3 cores to run a judgehost?

vmcj: I think I always assume someone has more than 3 cores but the 2 is now chosen because 1 might be the `hyperthreading` core which is not clearly explained, the person who wouldn't have hyperthreading should have a CPU(1) and if you run a single core machine I think you're aware of what you're doing?

thijskh: I think it's not uncommon in a virtualised environment to have 'only' 1 or 2 cores right?

vmcj: I did not think of that, and that is indeed a "normal" usecase.

doc/manual/install-judgehost.rst
```diff
@@ -128,6 +144,10 @@ Edit grub config to add cgroup memory and swap accounting to the boot
 options. Edit ``/etc/default/grub`` and change the default
 commandline to
 ``GRUB_CMDLINE_LINUX_DEFAULT="quiet cgroup_enable=memory swapaccount=1"``
+Optionally the timings can be more stable by denying the OS to schedule
+other tasks to same CPU the judgedaemon is using:
```
eldering, line 148: Better wording:
"Optionally the timings can be made more stable by not letting the OS schedule
any other tasks on the same CPU core the judgedaemon is using:"

.gitlab-ci.yml
```diff
@@ -245,13 +245,23 @@ integration_mysql:
     MYSQL_ROOT_PASSWORD: password
     MARIADB_PORT_3306_TCP_ADDR: sqlserver
     MYSQL_REQUIRE_PRIMARY_KEY: 1
+    RESTRICTED: 1
```
eldering, line 248: Can we use a more explicit variable? For example `PIN_JUDGEDAEMON`.

doc/manual/install-judgehost.rst
```diff
@@ -50,15 +50,31 @@ home directory::
   make judgehost
   sudo make install-judgehost
 
+For the next section we assume a machine with possibly hyperthreading and more
+than 3 CPUs. This can be checked with::
```
eldering, line 54: Nitpick: "3 or more"

doc/manual/install-judgehost.rst
```diff
@@ -50,15 +50,31 @@ home directory::
...
+
+  lscpu | grep "Thread(s) per core"
+
+having a value above 1 indicates hyperthreading or::
+
+  cat /sys/devices/system/cpu/smt/active
+
+a value of `1` or `on`. The target CPU for the next session should be in the range of::
```
eldering, line 62: I suggest to replace "target CPU for the next session" with "target CPU core to restrict the judgedaemon to below" as "session" is unclear.

doc/manual/install-judgehost.rst
```diff
@@ -50,15 +50,31 @@ home directory::
...
+
+  cat /sys/devices/system/cpu/online
+
+if the machine has less than 3 cores and has no hyperthreading change the value to `1`.
```
eldering, line 66: Nitpick: this is a new sentence, so should start with a capitalized `If`.
Better to explain in more detail here how to possibly pick a CPU core and why we picked `2` as default. For example if the machine has only a single core then `1` will obviously not work.

eldering (changes requested):

doc/manual/install-judgehost.rst
vmcj: session is a typo and should be section. But even when the typo was not there your wording is nicer.

doc/manual/install-judgehost.rst
```diff
@@ -50,15 +50,34 @@ home directory::
   make judgehost
   sudo make install-judgehost
 
+The judgedaemon can be run on various hardware configurations;
+- A virtual machine, typicly these have 1 or 2 cores and no hyperthreading, because the kernel will schedule its own tasks on CPU 0, we advice CPU 1,
```
eldering, line 54: Typo: typically

doc/manual/install-judgehost.rst
```diff
@@ -50,15 +50,34 @@ home directory::
...
+- A default office machine, these sometimes have hyperthreading. Verify if the machine has hyperthreading and consider turning it off and as a rule of thumb pick CPU 2 as CPU 1 could be a hyperthreading core, be on the same die as CPU 0 and therefore share memory with that CPU. If more cores available as a rule of thumb moving to the highest CPU should be considered.
+- Multiple on a single high-grade server with multiple CPUs or a CPU with multiple cores. Check for hyperthreading and if possible run the judgedaemon on another CPU package or CPU die than the CPU 0. See the section :ref:`multiple-judgedaemons` for running multiple
```
eldering, line 56: Fixes: "run the judgedaemon*s* on separate CPU packages/dies and different from CPU 0".
I think in this case you mostly want the different judgedaemons to not interfere with each other. The load of the OS itself is less relevant.

vmcj: The section was meant with only 1 judgedaemon, but I think your interpretation works better.

vmcj:
After approval I'll reorder the commits some more.

doc/manual/config-advanced.rst
```diff
@@ -346,6 +346,10 @@ Although each judgedaemon process will be bound to one single CPU
 core, shared use of other resources such as disk I/O might
 still have effect on run times.
 
+Real world usage shows that the amount of judgedaemons a machine can run
+does not scale with the number of cores. As a general advice verify the
+timelimits again when more active judgedaemons are added.
```
eldering, line 351: I think this statement requires some more explanation. The problem is not that it doesn't scale, but that while scaling to multiple judgedaemons, runtimes become less consistent. That's also already partially mentioned in the previous paragraph, so I think it's better to either leave this out or merge it in with that, and e.g. mention that performance with mulitple judgedaemons can fluctuate up to 20-40%.

eldering (changes requested):

doc/manual/config-advanced.rst
vmcj: I've removed the section.

eldering (approved):

doc/manual/install-judgehost.rst
```diff
@@ -175,15 +197,15 @@ Starting the judgedaemon
...
+add a new judgehost but have it initially disabled, you can change the config
+setting to automatically pause judges on first connection or manually add it
+through the DOMjudge web interface and set it to disabled before starting
+the judgedaemon.
 
 The judgedaemon can also be run as a service by running::
 
-  sudo systemctl enable domjudge-judgehost
-  sudo systemctl start  domjudge-judgehost
+  sudo systemctl enable --now domjudge-judgehost
```
meisterT, line 211: Should this be `systemctl start`?

vmcj: `enable --now` is shorthand for `systemctl {enable,start}`, (or it was).

`enable` creates the symlink, the --now is telling that you also want to directly start.

PR #1371, opened 2021-11-25
Advice less to new contestants

vmcj:
I think all people who do not have `less` on their machine know that `more` is an alternative. And I think `less` is more intuitive for scrolling through the help in this specific case.

### DIFF
--- a/doc/manual/team.rst
+++ b/doc/manual/team.rst
@@ -108,7 +108,7 @@ These guesses can be overruled with the options
 ``-e entry_point``.
 
 See ``submit --help`` for a complete list of all options,
-extensions and some examples.  Use ``submit --help | more``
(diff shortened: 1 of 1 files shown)


thijskh:
Do you even still need this in 2021 or do new contestants just use their scrollback buffer?

vmcj:
> Do you even still need this in 2021 or do new contestants just use their scrollback buffer?

Thats a good question, I think the scrollback buffer is good enough in this usecase.

eldering:
FYI, it said `more`, since `submit` used to be also compilable under windows and as far as I know that doesn't have `less`. Since now submit is a python script, that might apply again. But I'm fine with just leaving it out.

eldering (approved):

meisterT (approved):

PR #1372, opened 2021-11-25
Minor team manual fixes

vmcj:


doc/manual/team.rst
```diff
@@ -225,6 +225,9 @@ TOO-LATE
 
 The judges may have prepared multiple test files for each problem.
 DOMjudge will report back the first non-correct result as verdict.
+*DOMjudge will always output the verdict on the first ordered testcase.
```
vmcj, line 228: I'm not very happy about this wording, so feel free to fully rewrite to actually convey what I'm trying to say.

nickygerritsen: This is also not true if you change the results prio

doc/manual/team.rst
```diff
@@ -225,6 +225,9 @@ TOO-LATE
...
+This means that when the submission is deterministic DOMjudge will give back
+the same output on resubmission.*
```
thijskh, line 230: Why is this important to know for the team and why does it need to be made so explicit?

doc/manual/team.rst
vmcj: Just to be clear you mean:

RTE: prio 99
WA: prio 98
CORRECT: prio 1

or during the contest changing the ordering of the testcases? The latter is to specific I think... the first I did not think about.

nickygerritsen: The former

doc/manual/team.rst
vmcj: I've had teams submitting multiple times in the hope to get another outcome, now that we have parallel judging a team could think that they can have more information with a small penalty.

If I would know that I only have WA I might fix it whereas if I also got TLE I might skip this problem and focus on another one.

thijskh: Not sure the manual is the place to protect people from penalty while trying to game the system

vmcj: Ok, I'll rip it out and only address @nickygerritsen his comment.

doc/manual/team.rst
```diff
@@ -224,6 +225,7 @@ TOO-LATE
 
 The judges may have prepared multiple test files for each problem.
 DOMjudge will report back the first non-correct result as verdict.
+*DOMjudge will always output the highest priority verdict.*
```
nickygerritsen, line 228: Maybe add something like “if there are multiple verdicts with the same priority, the first encountered one will be returned”?

doc/manual/team.rst
```diff
@@ -223,6 +224,7 @@ TOO-LATE
 
 The judges may have prepared multiple test files for each problem.
 DOMjudge will report back the first non-correct result as verdict.
+*If there are multiple verdicts with the same priority, the first encountered one will be returned.*
```
thijskh, line 227: Is this essentially a repeat of the line before?

vmcj: I think it can be fixed as:
DOMjudge will report back the first highest priority non-correct result as verdict. 

So adding priority to the line before.

thijskh: Will this actually clarify something for teams since the word 'priority' is never used before in the manual?

PR #1374, opened 2021-11-25
Prevent auditlog spam from queuetask deletion when no actual queuetas…

### LINKED ISSUE #1369: Many `queuetask xx deleted` audit logs will be generated during the empty judge queue

I deployed 4 judgehosts by docker.

When there is no judge task, in the audit log, a log such as `queuetask xx deleted` will be continuously generated.

Just like the screenshot below:

image

I think these audit logs are too many and useless.

- 8.0.0DEV
- Ubuntu 20.04
- Nginx

I found the relevant code to generate this log.

https://github.com/DOMjudge/domjudge/blob/1827899b8d331e03df93e2c199da91b3149e0d4b/webapp/src/Controller/API/JudgehostController.php#L1576-L1588

nickygerritsen:
…k is deleted. Fixes #1369.

This happened because [in this case](http://github.com/DOMjudge/domjudge/blob/88fc65b6a24ba63221e71dd3addf05270e9b7171/webapp/src/Controller/API/JudgehostController.php#L1382-L1394) we loaded the last judging for a judgehost. That might be with a queuetask that is already deleted: it *is* always this when a judgehost is idle. So we add a check if something actually got deleted.

### DIFF
--- a/webapp/src/Controller/API/JudgehostController.php
+++ b/webapp/src/Controller/API/JudgehostController.php
@@ -1381,6 +1381,10 @@ class JudgehostController extends AbstractFOSRestController
         // This is case 1) from above: continue what we have started.
         $lastJobId = $this->em->createQueryBuilder()
             ->from(JudgeTask::class, 'jt')
+            // Note: we are joining on queue tasks here since if there is no more queue task, there is also no more
+            // work to be done. If we would not do this join, the getJudgetasks would try to delete the queue task,
+            // which is both slow and results in spamming the auditlog
+            ->innerJoin(QueueTask::class, 'qt', Join::WITH, 'qt.jobid = jt.jobid')
             ->select('jt.jobid')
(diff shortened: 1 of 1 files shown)


nickygerritsen:
cc @dup4 FYI

thijskh:
Is it not a bit expensive (lockingwise or other) to continuously run DELETEs on the table? Better check if there's something to delete at all and only then delete it?

nickygerritsen:
@meisterT what do you think? If we first load the queuetask (with `findOneBy`) and then delete it, will that give a race condition?

meisterT:
I think this commit just hides the symptoms but doesn't fix the issue.

I assume you can fix it by using left join queuetasks in the query that you link to.

meisterT (commented):
Please add a comment there why we are doing this join because it is not obvious.

meisterT (approved):

PR #1377, opened 2021-11-27
Fix race condition for repeated rejudgings.

### LINKED ISSUE #1352: Repeated rejudgings don't always set endtime for all instances
If you repeat a rejudging say 10 times, all intermediate rejudgings should have an endtime set whenever this instance is done. Then it will also create the next repetition. Sometimes, one of the instances doesn't have an endtime set, but it's not reproducible.

Since we only show the latest instance of a repeated rejudging, this is not visible in the list of rejudgings, but it is shown in the menu bar as [1]

meisterT:
Fixes #1352.

When two judgehosts at the same time realized that there is no work left
in the current rejudging, we didn't properly synchronize before which
potentially lead to both creating a new rejudging.


nickygerritsen (approved):
How did this before set the end time of one but not the other?

meisterT:
> How did this before set the end time of one but not the other?

While both rejudgings were created the submission was only added to one of them, so it was never completed.

Issue #1379, opened 2021-11-27
Verify no 500s in integration jobs


vmcj:
Check if the symfony logs don't show HTTP500 codes for normal usage.

PR #1380, opened 2021-12-01
Check for CRITICAL and ERROR in symfony log.

meisterT:
This should inform us how much work needs to be done for #1379.


vmcj:
Should we also just fail if we get these errors?

meisterT:
Maybe - I didn't check that CI run yet whether it would be a lot of work to clean up

vmcj:
> Maybe - I didn't check that CI run yet whether it would be a lot of work to clean up

I'm now introducing your commit in other branches and then fixing the issues, currently that would be only 1 left which I think is a real problem as we try to freeze a contest which cannot be frozen.

vmcj:
This was now contained in: https://github.com/DOMjudge/domjudge/pull/1442

PR #1381, opened 2021-12-04
Stress tests error

vmcj:
Currently we submit files which should throw an 400. This PR tries to remove these log lines from the symfony integration log so we can assume that every HTTP code above 299 should be handled.


nickygerritsen (approved):

tests/Makefile
```diff
@@ -50,7 +50,9 @@ $(SUBST_FILES): %: %.in $(TOPDIR)/paths.mk
 check-syntax:
 	$(TOPDIR)/tests/syntax
 
-check: test-normal test-fltcmp test-boolfind test-bad-expected-results
+check-problems: test-normal test-fltcmp test-boolfind
```
eldering, line 53: `check-problems` should also be added to the `.PHONY` targets as it's not a real file that gets created.

eldering (changes requested):
See comment about `check-problems` as PHONY target.

eldering (approved):

PR #1383, opened 2021-12-04
Fix CodeQL

vmcj:
For some reason codeQL now broke when any process got a non-zero exit code. Because we miss a build file for the bool_find we need to remove this from the build-scripts as otherwise codeQL will stop the build.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -80,6 +80,7 @@ jobs:
 
     - name: Compile the build scripts for languages
       run: |
+        rm -rf sql/files/examples/boolfind_run*
         make build-scripts
 
     - name: Compile judgehost


vmcj:
It also discovers problems (as it should) but those are not created by me.

eldering:
Could you explain in more detail in the commit message what the actual problem is? I don't see any build file `sql/files/boolfind_run/build`, so don't understand what the problem would be.

vmcj:
> Could you explain in more detail in the commit message what the actual problem is? I don't see any build file `sql/files/boolfind_run/build`, so don't understand what the problem would be.

Thats the problem, because build does not exist, we get a return != 0 for [ -x build ] so the script fails.

eldering (approved):

PR #1390, opened 2021-12-13
Fix outdated CCS specs links

edomora97:
Around ~22 days ago they switched the layout of the spec website, changing the URLs. All the previous URLs now return 404s, since they don't include the specs version.

This find-and-replaces all the links from [https://ccs-specs.icpc.io/](https://ccs-specs.icpc.io/) to [https://ccs-specs.icpc.io/2020-03/](https://ccs-specs.icpc.io/2020-03/). The available versions, at the time of writing, are 2020-03, 2021-11 and master.

**I'm not sure which is the correct version**, and I've just assumed it's the older one. As you can see, some of the occurrences involve publicly visible links, including one in the home page of the jury interface.

### DIFF
--- a/doc/manual/config-basic.rst
+++ b/doc/manual/config-basic.rst
@@ -140,4 +140,4 @@ it need not have started yet). You can verify whether the submissions
 gave the expected answer in the Judging Verifier, available from
 the jury index page.
 
-.. _ICPC-compatible teams.tsv files: https://ccs-specs.icpc.io/ccs_system_requirements#teamstsv
+.. _ICPC-compatible teams.tsv files: https://ccs-specs.icpc.io/2021-11/ccs_system_requirements#teamstsv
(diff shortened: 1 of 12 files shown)


nickygerritsen:
This is a good question. I think for 7.3 we should clearly point to `2020-03`. For the upcoming 8.0 we should probably point to `2021-11`, since that is what we tested at NWERC.

So maybe we update this PR to `2021-11` and then when we do the cherrypick to 7.3 (if we want it) we update it there to `2020-03`?

edomora97:
Fine for me, I'm new to this repo (first contribution!), should I force-push this or make a new commit?

nickygerritsen:
Let's await @eldering his opinion but if he agrees I'd say a force push sounds like the way to go.

eldering:
I think @nickygerritsen knows better what our state is of implementing the various details in the versions of the spec, but it makes sense to me that the 7.3 branch aims at the 2020-03 version and main at 2020-11 (at least for now).

eldering (approved):

nickygerritsen (approved):

nickygerritsen:
Thanks! Looks good and merged 🎉

PR #1400, opened 2021-12-29
Run unit tests against all currently supported PHP versions

vmcj:
The author of the PPA (https://deb.sury.org/) packages for debian so I suspect this is up to debian standards.

If we allow this it will need a rewrite to put all of them in the image which I'll do in another PR.

### DIFF
--- a/misc-tools/dj_make_chroot_docker.in
+++ b/misc-tools/dj_make_chroot_docker.in
@@ -12,8 +12,20 @@
 # Abort when a single command fails(or uninitialized variable is accessed):
 set -eu
 
+cleanup() {
+    # Unmount things on cleanup
+    umount -f "$CHROOTDIR/proc" >/dev/null 2>&1  || /bin/true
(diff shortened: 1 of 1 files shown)


gitlab/unit-tests.sh
```diff
@@ -2,6 +2,16 @@
 
 set -euxo pipefail
 
+add-apt-repository ppa:ondrej/php -y
+apt-get update
+
+PACKAGES=$(dpkg-query -f '${binary:Package}\n' -W)
```
nickygerritsen, line 8: Shouldn't we check here that we only do this for php packages? i.e. aren't we now also uninstalling for example `curl` or `nginx`?

vmcj: Yes, I already fixed that, also I'm not sure if for example php8.0-pear will actually exist so I'm still investigating if this does work.

nickygerritsen (approved):

PR #1403, opened 2021-12-31
Only unit phpunit

vmcj:


vmcj:
For now I choose to accept:
```
Remaining direct deprecation notices (1)                                                                                                                                                                                                                                                                                                                                                                                                1x: The "PHPUnit\Framework\TestCase::__construct()" method is considered internal This method is not covered by the backward compatibility promise for PHPUnit. It may change without further notice. You should not extend it from "App\Tests\Unit\Controller\Jury\JuryControllerTest".                                                                                                                                                1x in AwardsControllerTest::testListWithIds from App\Tests\Unit\Controller\API
```

As its not really clear how to fix this in a nice way, so this might result in issues when upgrading to other versions of PHPUnit but as this is only a dependency for us developers and not in the end product I think its worth it compared to not testing for PHP72-80.

nickygerritsen (approved):

PR #1404, opened 2021-12-31
Assume php version of Debian stable

vmcj:
I'm not sure if those are the only places where these checks are used.


nickygerritsen (approved):

thijskh (approved):

thijskh:
In the ideal world we'd find a way to make these things version-agnostic... but I don't know how and this improves things indeed.

PR #1413, opened 2022-01-07
Fix missing PHP strict mode

vmcj:


thijskh:
Apparently we lost the automatic check for this?

vmcj:
> Apparently we lost the automatic check for this?

Well I got this as an error in another PR, so in that PR I'll introduce the check for it again. I think the `syntax check` job should be extended with the `set -e` but that's something for another PR.

This PR was also to check if the system still works if strict mode is introduced for these files.

Issue #922, opened 2020-11-14
Allow user to switch to dark theme


mpsijm:
In light of the recent additions in PR #909 (issue #908), we might just as well add a default dark theme :smile: 

As discussed in the BAPC slack, this issue is to track interest :slightly_smiling_face: 

Ideally, we'd use something similar to what YouTube uses to switch between themes:
![image](https://user-images.githubusercontent.com/9739541/99143660-90526680-265f-11eb-8b45-b36267d695c9.png)

thijskh:
I second the idea of a dark theme, but should we let users select it? Why not just follow device configuration?

nickygerritsen:
> I second the idea of a dark theme, but should we let users select it? Why not just follow device configuration?

It might be that some people have a preference to have a dark theme globally but not on DJ or the other way around? I'm not sure but at least I have some websites where I use the light theme even though my device is set to dark.

mpsijm:
Also, Chrome does not properly detect dark theme yet. The screenshot I sent was from FireFox (which does properly handle dark theme detection), while in Chrome this does not work (yet) :stuck_out_tongue:

JobDoesburg:
If there's going to be both a dark theme and light theme, there should be a way to have a dark-banner en light-banner (and affiliation logos) too

vmcj:
Because this would make the UI much harder to test if everything is still visible etc. and we have little time to work on this I think its more work to implement+maintain than the worth. I'll close this now but when someone wants to work on this we can reconsider.

PR #1419, opened 2022-01-09
Docs: Add logo to README

Dup4:
Maybe consider adding a logo to the README?

Look like this:

image

You can check by this [link](https://github.com/Dup4/domjudge).


vmcj:
When you set GitHub to dark mode this doesn't look nice.

Dup4:
> When you set GitHub to dark mode this doesn't look nice.

Yes, you are right.

nickygerritsen:
We already have the logo's in Git: https://github.com/DOMjudge/domjudge/tree/main/doc/logos

Dup4:
> We already have the logo's in Git: https://github.com/DOMjudge/domjudge/tree/main/doc/logos

Yes, I saw them but they are all transparent logos.

When we turn on dark mode, it becomes unrecognizable, like below:

image

But in light mode, it will be much better, like below:

image

You can check by this [link](https://github.com/Dup4/domjudge/tree/9b1e5e3407a595a5fc7c4a6e881fa1b1b6428f3c)

So my idea is, we can add a logo with a white background, like [this](https://github.com/Dup4/domjudge/blob/main/doc/logos/DOMjudgelogo-with-white-background.png).

Then in dark mode, the displayed effect will be slightly better.

image

I'm just making a suggestion, it's up to you to accept this PR or not.

thijskh (changes requested):
I prefer using an existing image in the repo. Also there must be an alt tag in an image.

thijskh:
You could choose not to display the logo when in dark mode (DOMjudgelogo.svg#gh-light-mode-only) but that's also not a very nice solution.

I can't think of a better solution now than just to add the logo with the white background next to the logos we have. But probably the real solution would be to redesign our logo (files) so we have a higher quality svg and two variants based on background. We'd also need that if we would ever implement a dark mode in DOMjudge itself (#922).

PR #1420, opened 2022-01-09
Fix public scoreboard banner check without active contest. Fixes #1417.

### LINKED ISSUE #1417: If there is no active contest at this time, accessing the public page will return `500 Internal Server Error`

> Replace this line with a short description.

When all contests are finished and all contests are inactive status, then go to the `public page(/domjudge/public)`, it will show `500 Internal Server Error`.

Shouldn't the expected result be that there are currently no active contests or something instead of showing `500 Internal Server Error`?

### Your environment

> Include details about your installation here.
> - DOMjudge version (e.g. 7.0.0 or a github commit hash)
> - Operating system / Linux distribution and version (e.g. Ubuntu 18.04)
> - Webserver (e.g. Apache or nginx)

- Ubuntu 20.04
- Nginx

nickygerritsen:


thijskh:
This keeps on breaking. So probably best to add a testcase for the public page without an active contest

nickygerritsen:
Yes we should. I can do that on this pr

nickygerritsen:
> Yes we should. I can do that on this pr

Done, was easy enough.

thijskh (approved):
Just fix the codespell issue and we should be good to go

PR #1421, opened 2022-01-09
Default behaviour will change July22

vmcj:


nickygerritsen:
This is not the best commit message ever. Better to at least link to https://getcomposer.org/doc/06-config.md#allow-plugins or something.

vmcj:
> This is not the best commit message ever. Better to at least link to https://getcomposer.org/doc/06-config.md#allow-plugins or something.
I'll update the message with something better.

More important, is this the fix we want?

nickygerritsen:
For flex for sure. I'm not sure what the other plugin does?

vmcj:
composer/package-versions-deprecated: It shows us which installed libs are out of support

Its not something I've installed myself, so i'm also fine with not enabling it.

nickygerritsen (approved):

Issue #1408, opened 2022-01-02
Detecting .R for R language does not work but lowercase works


melongist:
### Description of the problem

R extension file not linked to R language.

But, after adding r extension, R extension file linked to R language correctly.

### Your environment
- DOMjudge version : 7.3.4 stable
- Operating system : Ubuntu 20.04.3 LTS server
- Webserver : Apache

### Steps to reproduce

<img width="1440" alt="r1" src="https://user-images.githubusercontent.com/92401080/147875627-e14c708d-bbcc-47e5-af20-c8c7fda2094d.png">
<img width="1440" alt="r2" src="https://user-images.githubusercontent.com/92401080/147875632-a69cac5a-ae7b-49a7-adf7-0728d6a290be.png">
<img width="1440" alt="r3" src="https://user-images.githubusercontent.com/92401080/147875636-8e17af8f-74e1-4f4a-853a-18e29a51b9b6.png">
<img width="1440" alt="r4" src="https://user-images.githubusercontent.com/92401080/147875639-0b7788f2-7ed8-452c-8214-c78dc61e3931.png">

vmcj:
> ### Description of the problem
> 
> R extension file not linked to R language.
> 
> But, after adding r extension, R extension file linked to R language correctly.
> ### Your environment
> 
>     * DOMjudge version : 7.3.4 stable
> 
>     * Operating system : Ubuntu 20.04.3 LTS server
> 
>     * Webserver : Apache
> 
> 
> ### Steps to reproduce
> 
> <img alt="r1" width="1440" src="https://user-images.githubusercontent.com/92401080/147875627-e14c708d-bbcc-47e5-af20-c8c7fda2094d.png">
> 
> <img alt="r2" width="1440" src="https://user-images.githubusercontent.com/92401080/147875632-a69cac5a-ae7b-49a7-adf7-0728d6a290be.png">
> 
> <img alt="r3" width="1440" src="https://user-images.githubusercontent.com/92401080/147875636-8e17af8f-74e1-4f4a-853a-18e29a51b9b6.png">
> 
> <img alt="r4" width="1440" src="https://user-images.githubusercontent.com/92401080/147875639-0b7788f2-7ed8-452c-8214-c78dc61e3931.png">

Did you check if you get the same problem for other languages? I can reproduce this for the main branch, so I'll look at a fix.

melongist:
[...]

PR #1423, opened 2022-01-09
Keep casing of submission language

vmcj:
Languages like R have their extension as .R instead of .r, as this is
convention we should respect that. The lowercase for the filename itself
is kept as that will most likely lead to the shortname of the problem.

The alternative is to lowercase all extensions, but than we could have collisions when there are languages where capital extension would be something else compared to lowercase.

According to: https://en.wikipedia.org/wiki/C%2B%2B this could be the case for C->C++ and c->c

closes: https://github.com/DOMjudge/domjudge/issues/1408

As the issue is filed against 7.3 we might want to cherry-pick this.

### DIFF
--- a/webapp/templates/partials/javascript_language_detect.html.twig
+++ b/webapp/templates/partials/javascript_language_detect.html.twig
@@ -4,7 +4,7 @@
         switch (ext) {
                 {% for language in submission_languages %}
                 {% for extension in language.extensions %}
-            case '{{ extension }}':
+            case '{{ extension|lower }}':
                 return '{{ language.langid }}';
                 {% endfor %}
                 {% endfor %}


eldering:
I'm not sure what the intent of this PR is, but I'm not in favor of introducing case-sensitiveness in matching extensions to languages. Besides being too easy to make typos with, there are case-insensitve filesystems out there that would make it hard or impossible to submit a file with the right case.

vmcj:
Ok, this will create the 2nd case I mentioned.

But I'll change the other function where we check case sensitive.

vmcj:
@eldering 

This now has the behaviour which you prefer, this will create the issue that:

hello_cpp.C will be detected as the C language instead of C++ which according to [wikipedia](https://en.wikipedia.org/wiki/List_of_filename_extensions_(A%E2%80%93E)) is the only case

thijskh:
May I submit that having ".C" and ".c" refer to different languages is just stupid and we should not cater to such nonsense?

eldering:
> May I submit that having ".C" and ".c" refer to different languages is just stupid and we should not cater to such nonsense?

I strongly agree!

thijskh (approved):

vmcj:
> May I submit that having ".C" and ".c" refer to different languages is just stupid and we should not cater to such nonsense?

I personally agree that they could have picked another standard but I was not sure if it was commonly used. So I only mentioned it as a possible downside. I'll merge this later today.

nickygerritsen (approved):

PR #1426, opened 2022-01-11
Add missing `global $myhost;`

edomora97:
`$myhost` is used but was not imported from the global scope.

I've found this simply because vscode was complaining that `$myhost` was not declared few lines below. Why was this not caught by the CI?


meisterT (approved):
CI didn't complain since this case (chroot stop doesn't succeed) is not tested on CI.

meisterT:
@vmcj has been adding linters recently but I don't have a good overview for what.

PR #1427, opened 2022-01-15
Fix composer install for maintainer-config

vmcj:
I suspect that the autoload has no strict types so autoloading is possible. It was excluded in the past so seem reasonable to ignore strict types here.


thijskh:
This has been present since 7.0.0?

vmcj:
> This has been present since 7.0.0?

I agree that this should not be the solution.

PR #1428, opened 2022-01-15
Find typos

vmcj:


thijskh:
Can’t we convince codespell to check for this word instead of adding yet another job?

vmcj:
> Can’t we convince codespell to check for this word instead of adding yet another job?

Fine by me, I'll see if its possible.

vmcj:
@thijskh no this is not possible, as https://github.com/codespell-project/codespell/issues/1824 is still open and local testing also shows the same issue.

But I'll move the script to GitLabCI so it's part of another job.

PR #1429, opened 2022-01-16
Test with more PHP versions

vmcj:
This is only to check how much work this would be.


vmcj:
Another route was chosen.

PR #1432, opened 2022-01-17
Fix domjudge misspellings

vmcj:
This should be the typos found in: https://github.com/DOMjudge/domjudge/pull/1428

That PR should be rewritten, but I think renaming these files is a good moment for DJ8.0.


webapp/config/packages/security.yaml
```diff
@@ -39,9 +39,9 @@ security:
             guard:
                 # SEE NOTE ABOVE IF CHANGING ANYTHING HERE
                 authenticators:
-                    - App\Security\DOMJudgeIPAuthenticator
-                    - App\Security\DOMJudgeBasicAuthenticator
-                entry_point: App\Security\DOMJudgeIPAuthenticator
+                    - App\Security\DOMjudgeIPAuthenticator
```
nickygerritsen, line 42: For these classes I'm not sure. CamelCasing suggests it should be DOMJudge, or DomJudge, but DOMjudge sounds strange for a class.

vmcj: Hmm according to https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined it should become DomJudge, 

I'm fine with rewriting, I just read it as:
"DOMjudge"-IP-Authenticator 

but it was already another commit to make reverting this easier.

eldering: This precisely shows why camel casing sucks, besides being less readable than snake case. Also, according to the rules it should be `Ip`, not `IP`, ugh!

Anyways, I think we should follow these rules (very) liberally, and I'd prefer `DOMjudge`, but I'm also not a big fan of just willy-nilly starting to do a huge rewrite, so let's first settle on something.

vmcj: Thats the reason that I closed this, I'll bring this up in chat to see what the rest thinks of this.

Issue #1237, opened 2021-10-07
Optimize `runpipe`


RagnarGrootKoerkamp:
See https://github.com/DOMjudge/domjudge/blame/main/judge/runpipe.cc#L54. (CC @eldering @meisterT @mvr320 ).

When an interactive problem needs to pipe O(megabytes) of data (in one direction), this is done in chunks of 512 bytes. After each chunk, the stream in the other direction in checked with a [1ms timeout](https://github.com/DOMjudge/domjudge/blame/main/judge/runpipe.cc#L260). Doing this O(1000) times results in wallclock timeouts without the submission being able to do any real compute.

Also note that it [alternates](https://github.com/DOMjudge/domjudge/blob/main/judge/runpipe.cc#L502-L504) between trying the pipes in the two directions.

I see a few possible solutions. Happy to implement one but not sure what is preferred:

1. Decrease the timeout from `1000us` to `1us`. Would this cause issues otherwise? (Now the script becomes a busy wait loop polling all inputs very often, consuming more CPU.)
2. Increase the buffer size. Currently the comment says it's the POSIX minimum value of pipes, but I'm not sure how this matters. (And I think the pipes are set to the max of 1MB anyway.) Reading/writing 1MB of data should be fine anyway right? If the team-input pipe fills up, that can happen either in 512byte chunks of 1MB chunks, but if it's filling up the chunk size doesn't matter anyway.
3. When the number of bytes read (`nread`) is `512` exactly, we can retry pumping from the same input, instead of switching to the pipe in the other direction.
4. Multithread the program, and do the two (or more) streams in parallel, so that blocking code can be used instead of non-blocking code with timeouts.

General remark: The whole file is hard to parse and understand because it's all pretty low level C. Maybe it could be rewritten to something newer? (Although maybe it's inevitable because of all the system calls, file descriptors, signal handling, ...)

For reference, in BAPCtools I do:
* By default: create two pipes and connect the interactor and submission directly to them. No intercepting of the data through the pipe is happening.
* If needed: Create a total of 4 pipes:
  - interactor -> interceptor -> team
  - team -> interceptor -> interactor
  I then [launch two instances of this interceptor](https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/master/bin/interactive.py#L168-L197) and pipe both their `stderr` to the same file to get a log of what's happening. The interceptor is just some very short python code that reads 1 char at a time and copies it to stdout and stderr.

eldering:
[...]

PR #1433, opened 2022-01-17
Alternative implementation of runpipe using epoll

edomora97:
Hi all!

Yesterday I've spent some time trying to implement runpipe using epoll instead of the current one with select with a timeout. Most of the issues of the current implementation are described in #1237, and I hope my implementation doesn't suffer from them.
This is an almost complete rewrite, so to ease the review and the benchmarks I've saved the new code in `runpipe2.cc` (without touching the original code). The original implementation was very C-like, while mine tries to be more "C++" (C++11 in particular).

The main idea is that we don't need to "busy loop" to detect when a child exits. epoll doesn't support selecting on child exit, but we can use an extra pipe that is written by the SIGCHLD handler and wake up the main loop. This way the main loop doesn't use any timeout and is woke up only when either a child exits or when new data is ready.

I'm aware that @tuupke is working on an alternative implementation that uses two threads instead of "async" code. I'm not a fun of this approach since mixing forks and threads sounds like quite complex (but maybe I'm very wrong). Furthermore, in most of the cases the bottleneck is not the throughput but the latency caused by the high amount of syscalls.

Note that both my version and the current one have two modes: direct connections between the processes, or with an interceptor (called proxy in my code) which inspects the traffic and stores it to a file. The latter is enabled with the `-o` flag.

## Scenario

Other than the actual implementation I'm also providing a quick analysis of the performance of the various versions under various scenarios. In particular, I've analyzed 2 main scenarios:

-   many small messages (the first process sends queries of 10 bytes, and the second answers with 3 bytes)
-   few larger messages (1000 bytes, 3 bytes)

I think the first case is the most common one, and also the one in which most of overhead is present. In fact most of it comes from the thousands of flushes present in the communication.

Here it is the code of the two interacting programs:

<details>
  <summary>First process (gen.cpp)</summary>

```cpp
#include <iostream>
#include <signal.h>

using namespace std;

int main(int argc, char **argv) {
  if (argc != 3) {
    cerr << "Usage: " << argv[0] << " N messageSize" << endl;
    cerr << "  N            number of messages to send" << endl;
    cerr << "  messageSize  number of bytes for each message (incl. \\n)"
        << endl;
    return 1;
  }
  signal(SIGTERM, [](int sig) { cerr << "gen received SIGTERM\n" << flush; });

  const int N = atoi(argv[1]);
  const int messageSize = atoi(argv[2]);
  cout << N << endl;
  string message(messageSize - 1, 'e');
  for (int i = 0; i < N; i++) {
    cout << message << endl;
    string ans;
    cin >> ans;
    if (ans != "42") {
      return 43;
    }
  }
  return 42;
}
```

</details>

<details>
  <summary>Second process (sol.cpp)</summary>

```cpp
#include <iostream>
#include <signal.h>
#include <unistd.h>

using namespace std;

int main() {
  signal(SIGTERM, [](int) { cerr << "sol received SIGTERM\n" << flush; });
  int N;
  cin >> N;
  for (int i = 0; i < N; i++) {
    string in;
    cin >> in;
    cout << "42" << endl;
  }
}
```

</details>

## Benchmarks

To run the benchmarks yourself, you have to:

-   Compile `runpipe2` and `runpipe` using the Makefile in `judge/`
-   Compile `sol` and `gen` inside `judge/test`
    -   `g++ -Wall -Wextra -O2 gen.cpp -o gen`
    -   `g++ -Wall -Wextra -O2 sol.cpp -o sol`
- Run `./bench.py` in `judge/test`

The benchmarks measure three times for each run:

-   `real` time is from the start of `runpipe` to its end
-   `user` time is the amount of time in user space
-   `sys` time is the amount of time spend in syscalls

### Scenario 1

In these tests the messages are very small (10 bytes + 3 bytes) and there are many of them (1000, 5000, 10k, 50k, 100k, 500k). Since the messages are small, having a larger buffer makes no difference.

real-10

With direct connection the two implementations are pretty much equivalent. While with a proxy in between my implementation is quite a bit faster (up to 30%).

sys-10

user-10

The same can be said also for user and sys time.

### Scenario 2

Here the messages are larger (1000 bytes + 3 bytes) and there are less of them (1000, 5000, 10k, 50k). Being larger, the overhead caused by the flushes is less dominant.

real-1000

sys-1000

user-1000

In this case the difference between the two versions is less apparent.

## Conclusions

My implementation should be almost ready, but there is still some work that can be done:

- At the moment the pipes are not closed when a process exits (which was the case in the current implementation, but not _always_).
- I've not yet tested it within domjudge and the actual sandbox, I've just tried it from the command line.
- Add more tests and more documentation on what the program should do. My implementation should be correct according to the "specification" written in the comment at the top of the file, but the edge-cases are not well defined at the moment (what happens if the first process segfaults? if the first process exits before the second?).
- Since it uses epoll it only works on Linux, but making it work also on macos/*BSD can be done without too much trouble since kqueue has a very similar API.

Of course this PR *should not be merged* as it is right now; it's intended to be a conversation starter and can be used to make data-informed decisions on which is the best implementation to be used.


nickygerritsen:
Thanks!

We were trying to run it more within DOMjudge, to be able to compare it to a known slow submission with our old code and @tuupke his code, but it seems something is still wrong. You can easily verify this:
* Load the default data
* Submit boolfind-test-correct.cc
* Run runpipe2 like this:

```
./runpipe2 -v -o somefile [domjudge]/output/judgings/[host]/endpoint-default/executable/run/26/cc37bebed93673c9d3432e99d6c076fd/build/runjury [domjudge]/output/judgings/[host]/endpoint-default/testcase/5/90864a8759427d63b40f1f5f75e32308_6267776644f5bd2bf0edccf5a210e087.in [domjudge]/output/judgings/[host]]/endpoint-default/testcase/5/90864a8759427d63b40f1f5f75e32308_6267776644f5bd2bf0edccf5a210e087.out feedback = [domjudge]/output/judgings/[host]/endpoint-default/[submission]/[judging]/compile/program ignored
```

If you do this with the original runpipe it will exit with `42`, but your version hangs. I suspect this is because when the second program exits the pipe of the first one doesn't get closed.

meisterT:
In an attempt to collect more data, I added `exit(42);` here (I know that this is not the right behavior, but timing wise it should be close): https://github.com/DOMjudge/domjudge/pull/1433/files#diff-a343bb58038053280a972435ad9778ecf9da67aaa7d0c574483f5af1cb07da61R656

Timings on that one submission:
- 48s with 2 pipes connecting the programs directly
- 52s with @edomora97's solution
- 76s with @tuupke's solution
- ~120s with the old code

nickygerritsen:
@edomora97 any chance you could look into the issue with the closing of pipes if the second program exists? It seems your implementation is very promising so we might want to move forward with it.

edomora97:
Yes, I'm very sorry for being so unresponsive but these days I am super busy with the Italian Olympiads in Informatics school-level selection 😄 

By the end of the week I'll do my best to look again at runpipe (and maybe I'll sketch some tests as well).

edomora97:
I'm back!

I've written some tests for the most common scenarios, and then fixed `runpipe2`.

Programs:
- S: solution
- J: judge

Testcases:
- Correct interaction + J returns 42 -> exit 42
- Correct interaction + J returns 43 -> exit 43
- S exits early -> pipes should close
- J exits early -> pipes should close
- S doesn't write -> timeout
- J doesn't write -> timeout
- SIGTERM is forwarded -> print 2x SIGTERM
- J exits early, S loops -> timeout

The tests need some polishing, they're just quick-and-dirty just for debugging `runpipe2`, but later they can be made a bit more proper. To run them just `make test` inside `judge/test/testcases`. You need to have `runpipe` and `runpipe2` compiled in advance.

---

There were 2 small bug in the previous implementation:
- I've used `fcntl(fd, F_SETFL, FD_CLOEXEC)` instead of `fcntl(fd, F_SETFD, FD_CLOEXEC)` 🤦🏻‍♂️ 
- I was not closing the pipes when one of the two processes exited

The first fix was to use `pipe2` instead of `pipe` (that supports open flags). The latter is not fixed 100% properly yet, since it needs some further discussion.

If I understood the current `runpipe` correctly, when the judge exists the pipes are not closed (probably to avoid sending `SIGPIPE`) to the solution. My implementation closes the pipe. I'm not sure why it works, and if it actually work, but the test for that (`J_exists_early`, `J_exists_early_S_loops`) seems to pass. (Probably I'm not actually checking the status of the solution, and it may be crashing, or it may just crash sometimes). More investigations should be done.

I haven't tried to use `runpipe2` for an actual judging in DJ as @nickygerritsen did. But I'm confident that it is a bit more "compliant to spec" than before.

meisterT:
Thanks Eduoardo, on that single submission I confirmed that it works and finishes in a total walltime of ~51s on my test machine.

With respect to testing, can you rename your implementation to runpipe.cc so it actually is used by our tests on CI?

In general, this should be the behavior:
```
if validator_verdict == AC:
  if sandbox_verdict == RTE || sandbox_verdict == TLE:
    return sandbox_verdict
  else:
    return AC
elif validator_exited_first: # validator_verdict == WA
  return WA
else
  return sandbox_verdict
```

meisterT:
> The original implementation was very C-like, while mine tries to be more "C++" (C++11 in particular).

I think this is a good idea, but perhaps we can already do this in a separate commit / PR so that it is easier to review.

edomora97:
> With respect to testing, can you rename your implementation to runpipe.cc so it actually is used by our tests on CI?

Sure, nice idea! I've also tried to add a new pipeline job for running my tests. I'm not quite sure why the other tests are failing tho, I really can't figure out those logs 😅 

Temporarily, I've moved the original `runpipe.cc` in `runpipe1.cc`, and symlinked `runpipe2.cc` as `runpipe.cc`. This way we can still reference both implementation if we need (and run the tests on both implementations). When all the discussions will reach their end I'll clean up the commits and replace the old code with the new one (and probably remove the benchmark scripts).

> I think this is a good idea, but perhaps we can already do this in a separate commit / PR so that it is easier to review.

That's quite hard to do. This is basically a complete rewrite of the original runpipe. and it's >800 lines long (with tons of comments tho). Tell me if you have any suggestions on how I can help with the review process.

nickygerritsen:
So I have been debugging a bit with @edomora97 to see what the integration errors are.

For all 4 unexpected results we get a `run-error` instead of `wrong-answer`.
The main fix for that is to remove the `_` here: https://github.com/edomora97/domjudge/blob/perf/improve_runpipe/judge/runpipe2.cc#L790
Then CI should be happy.

But there is one more issue. We get the following error when running:

```
/home/nicky/Projects/domjudge/bin/runguard: copying data fd 1: Broken pipe
```

This results in the metadata not being saved for the submission. @edomora97 deduced that this is (after debugging a bit more) because we close `stdout` (and `process_to_proxy`) if we close pipes, which makes runpipe very unhappy. If we comment those two closes out it works, but it doesn't feel very clean maybe. The old runpipe does do something similar I believe: it only seems to close the out pipe -> in pipe pair, not the other one. So maybe we should just do that?

I also pushed a small fix that maybe hopefully fixes the other 2 failing CI jobs. Not sure what we should do with the runpipe CI job when we actually merge this PR btw.

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,16 @@
+{
```
nickygerritsen, line 1: We should probably get rid of this file.

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
nickygerritsen, line 1: Same here.

gitlab/ci/misc.yml
```diff
@@ -79,4 +79,4 @@ detect dump:
         -name "ace" -type d -prune -o
         -type f -print0 | xargs -0 grep --color "dump(" | grep -v "Yaml::dump(")
     - echo $OUT
-    - exit $(($(wc -l <<< $OUT)-1))
+    - exit $(($(wc -l <<< $OUT)-2))
```
nickygerritsen, line 82: Note: I added this. If we remove the bench.py file from this repo, we need to revert this to 1 again.

judge/runpipe2.cc
```diff
@@ -0,0 +1,848 @@
...
+  void close_fds() {
+    // First, close the pipes coming into the process (i.e. stdin).
+    close(stdin);
+    if (proxy_to_process != -1) {
+      logmsg(LOG_DEBUG, "closing fd: %d (proxy -> process) of %d",
+             proxy_to_process, pid);
+      close(proxy_to_process);
+    }
+    // If this process is the one that exited, we should also close the pipes
+    // comeing out of there. This will make sure all the pipes are closed.
```
nickygerritsen, line 255: Nit: typo: `coming`

judge/testcase_run.sh
```diff
@@ -291,13 +291,11 @@ resourceinfo="\
 runtime: ${program_cputime}s cpu, ${program_walltime}s wall
 memory used: ${memory_bytes} bytes"
 
-if [ $COMBINED_RUN_COMPARE -eq 1 ] && grep '^exitcode: 43' compare.meta > /dev/null 2>&1 ; then
+if [ $COMBINED_RUN_COMPARE -eq 1 ] && grep '^validator-exited-first: true' compare.meta > /dev/null 2>&1 && grep '^exitcode: 43' compare.meta > /dev/null 2>&1 ; then
```
nickygerritsen, line 294: For the others: I added this. @edomora97 his code always created the meta file and since we had a FIXME here already anyway, I decided to leave it like that and check it. We might want to improve on it even more in the future, but this is already better than before.

nickygerritsen (commented):
Added some small comments already.

I think we are very close to getting this final and merged, after we squash everything and do some cleanup.

judge/test/bench.py
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
edomora97: I don't know, I guess it may need some polish to be used again. During the squash I'll remove it but post it here so that it won't get lost.

judge/testcase_run.sh
nickygerritsen: Sounds good

gitlab/ci/unit.yml
```diff
@@ -26,8 +27,21 @@
...
+      - gitlabartifacts
+
 run unit tests:
   extends: [.mariadb_job,.phpsupported_job,.unit_job]
 
 run unit tests (MySQL):
   extends: [.mysql_job,.unit_job]
+
+run runpipe tests:
+  extends: [.runpipe_job]
```
vmcj, line 47: I think extending in this case doesn't make sense.

gitlab/runpipe-tests.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+. gitlab/ci_settings.sh
+
+set -euxo pipefail
+
+gitlabartifacts="$(pwd)/gitlabartifacts"
```
vmcj, line 7: Some of these were moved to ci_settings you started your PR.

gitlab/runpipe-tests.sh
```diff
@@ -0,0 +1,13 @@
...
+. gitlab/ci_settings.sh
+
+section_start compile "Compile runpipe"
+# Configure and make the runpipe binaries.
+make configure
+./configure --with-baseurl='http://localhost/domjudge/' --with-domjudge-user=domjudge --with-judgehost_chrootdir=${DIR}/chroot/domjudge |& tee "$GITLABARTIFACTS/configure.log"
+make judgehost |& tee "$GITLABARTIFACTS/make.log"
+section_end compile
+
+cd judge/test
```
nickygerritsen, line 12: CI is complaining:
```
shellcheck: gitlab/runpipe-tests.sh:12:1: warning: Use 'cd ... || exit' or 'cd ... || return' in case cd fails. [SC2164]
```

So

```suggestion
cd judge/test || exit
```

judge/test/J_doesnt_write/judge.c
```diff
@@ -0,0 +1,10 @@
+#include <assert.h>
```
meisterT, line 1: move all these files from `judge/test` to `judge/runpipe_test`?

meisterT (commented):
first quick pass, didn't do a deep review yet

judge/runpipe.cc
```diff
@@ -1,600 +1,858 @@
...
-   Program specifications:
-
-   This program will run two specified commands and connect their
-   stdin/stdout to each other.
-
-   When this program is sent a SIGTERM, this signal is passed to both
-   programs. This program will return when both programs are finished
-   and reports back the exit code of the first program.
- */
+  runpipe -- run two commands with stdin/stdout bi-directionally connected.
```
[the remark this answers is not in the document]
meisterT: ah, I see that it is the space in front.

judge/runpipe.cc
```diff
@@ -1,600 +1,858 @@
...
+        return false;
+    return true;
+  }
+
+  // The pipe connecting from -> to has some data ready. Consume it reading as
+  // much as possible, copy it to the target process and write it to the output
+  // file.
+  void pump_proxy_pipe(const process_t &from, const process_t &to,
+                       output_file_t &output_file) {
+    const size_t BUF_SIZE = 1024 * 1024;
```
meisterT, line 715: Did you try with other values here? I think it doesn't matter too much but probably 4k is best since it's the page size.

judge/runpipe.cc
```diff
@@ -1,600 +1,858 @@
...
-		fprintf(stderr,"%s: verbose: ",progname);
-		vfprintf(stderr,format,ap);
-		fprintf(stderr,"\n");
-	}
-
-	va_end(ap);
+// Set the NONBLOCK flag for a file descriptor.
+void set_non_blocking(fd_t fd) {
+  int flags = fcntl(fd, F_GETFL, 0);
+  if (fcntl(fd, F_SETFL, flags | O_NONBLOCK))
```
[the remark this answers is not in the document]
edomora97: I used the default settings of clang-format. So for now, I've added the missing brackets where there were "optional".

judge/runpipe.cc
```diff
@@ -1,600 +1,858 @@
...
+   writing to the other side, if that side doesn't consume data from the pipe.
+   See also: https://github.com/Kattis/problemtools/issues/113
+ */
+// For Linux specific fcntl F_SETPIPE_SZ command.
+#if __gnu_linux__
+#define PROC_MAX_PIPE_SIZE "/proc/sys/fs/pipe-max-size"
+#endif
 
-	va_end(ap);
+#ifdef PROC_MAX_PIPE_SIZE
```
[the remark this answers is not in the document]
edomora97: Yep, and I've also noted that the `#else` case was important as well (maybe).

judge/runpipe.cc
```diff
@@ -1,600 +1,858 @@
...
+// ------------
+//
+// Two processes have to be spawned and their stdio redirected: stdout of
+// one to stdin of the other. Furthermore, we must be able to look at their
+// traffic and write it to file.
+//
+// We called "proxy" the component that inspects the traffic and writes it to
+// the output file. The proxy is enabled only if the output file is provided.
+//
+// With #0 we refer to the first process (the main process), with #1 to the
```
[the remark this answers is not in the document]
edomora97: The exit code of `runpipe` matches the exit code of the "main process". Furthermore, `validator-exited-first` is used by `testcase_run.sh` and that value depends on which process is the "main process".

judge/runpipe.cc
```diff
@@ -1,600 +1,858 @@
...
+  vector<string> args;
+
+  // When the process has been spawned, this contains the pid of the child
+  // process.
+  pid_t pid = -1;
+
+  // Whether the process exited.
+  bool exited = false;
+  // Information about the exited process. Meaningful only if exited == true.
+  int exitInfo = -1;
```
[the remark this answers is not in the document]
edomora97: Not quite. `status` is what is given out by `waitpid` and includes more information about the exit other than the exit code (e.g. the signal number). More information at https://linux.die.net/man/2/waitpid

judge/runpipe.cc
edomora97: I haven't tried it actually. That value is the same as the previous implementation. 4k will be probably enough for most of the cases. Another option could be to keep the buffer very large, but allocate that buffer statically (this is a single-threaded program, so there shouldn't be any concurrency issues regarding that.)

judge/.gitignore
```diff
@@ -3,6 +3,8 @@
 /judgedaemon
 /runguard
 /runpipe
+/runpipe1
```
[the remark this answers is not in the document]
edomora97: Yep, you're right!

PR #1434, opened 2022-01-18
Add distinction between inplace install and maintainer-install

thijskh:
The base is now the inplace-install which is the same as the other install methods but just in the directory structure. So 'production ready'.

The 'maintainer-conf' target remains and still means that you indeed want to develop on this so will install dev dependencies, set the app in dev mode and disable production optimizations, on top of the inplace-installation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ dist: configure composer-dependencies composer-package-versions
 
 # Install PHP dependencies
 composer-dependencies:
-ifeq (, $(shell command -v composer))
+ifeq (, $(shell command -v composer 2> /dev/null))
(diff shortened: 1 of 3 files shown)


nickygerritsen (approved):

Makefile
```diff
@@ -195,14 +196,15 @@ maintainer-conf: dist composer-dependencies-dev
 
 # Run Symfony in dev mode (for maintainer-mode):
 webapp/.env.local:
-	echo "# This file was automatically created by 'make maintainer-install' to run" >> $@
+	echo "# This file was automatically created by 'make maintainer-conf' to run" >> $@
```
eldering, line 199: This was already wrong, but the first line should probably use `>` to make sure the file is recreated from scratch instead of appended to.

PR #1435, opened 2022-01-19
Get rid of @internal parent call.

nickygerritsen:


vmcj (approved):
Nice, I think https://github.com/DOMjudge/domjudge/blob/ed6de416aceec69d4e999080ee59588c1ae15963/gitlab/unit-tests.sh#L51 can now also be decreased.

nickygerritsen:
> Nice, I think
> 
> https://github.com/DOMjudge/domjudge/blob/ed6de416aceec69d4e999080ee59588c1ae15963/gitlab/unit-tests.sh#L51
> 
> can now also be decreased.

Good catch, seems it can be `2` now. Just pushed, waiting for CI.

PR #1437, opened 2022-01-22
Add view for queue tasks. Fixes #1378.

### LINKED ISSUE #1378: Create a view for queue tasks
As mentioned by @meisterT

nickygerritsen:
Screenshots:
image
image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -97,6 +97,7 @@
 		"twig/twig": "^2.11"
 	},
 	"require-dev": {
+		"ext-dom": "*",
 		"dama/doctrine-test-bundle": "^6.5",
 		"squizlabs/php_codesniffer": "*",
 		"symfony/debug-bundle": "4.4.*",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "49ee040a98938b100ff1ea54893b8421",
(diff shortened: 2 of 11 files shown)


thijskh (approved):
I have not reviewed every line but overall looks good in approach and presentation. Just wondering why the info in the judgestask table is so limited. Maybe not everything, but at least the `type` for example?

thijskh:
if you view /jury/queuetasks/1/judgetasks, then the judgedaemon handles those tasks, then reload the page you get `App\Entity\QueueTask object not found by the @ParamConverter annotation.`

Reproduce like this: Stop judgedaemon. Submit something. View judgetasks of the new queuetask. Start judgedaemon. Reload page.

thijskh:
Well, it still exists, it's just set to valid=0?

nickygerritsen:
> Well, it still exists, it's just set to valid=0? Throwing this exception seems kind of harsh.

No. The *queue task* is gone. That’s the whole purpose of that table: to never have too much data. The judge tasks might be there but not the queue task it belonged ro

thijskh:
Alright, maybe then just say so? It now looks like you've ran into a bug, getting an obscure exception for doing something that seems normal to me. It can also happen if you click something in the Queue task overview that has meanwhile been processed.

nickygerritsen:
> Alright, maybe then just say so? It now looks like you've ran into a bug, getting an obscure exception for doing something that seems normal to me. It can also happen if you click something in the Queue task overview that has meanwhile been processed.

Yea so my suggestion with the notice makes sense. I will do it

thijskh (approved):
Great, let's merge it, and improve it later if necessary

PR #1438, opened 2022-01-22
Unlock judge tasks for problem when enabling judging on contest. Fixes #1398.

### LINKED ISSUE #1398: When a problem is set to not `allow judge` in a contest, then set to `allow judge`, then the submission during this period will not be judged later

In the process of testing https://github.com/DOMjudge/domjudge/issues/1396, I found another phenomenon, I don't know if it is a feature or a bug.

https://user-images.githubusercontent.com/38343778/147307620-7194fafa-cb58-4a27-9ae5-c94ffd099ef4.mp4

In the video we can observe that we set a certain problem not `allow judge`, then set to `allow judge`, the the submission during this period will not be judged later.

But subsequent submissions will be judged.

- 8.0.0DEV
- Ubuntu20.04
- Nginx

BTW, I use docker to run DOMserver and judgehost.

nickygerritsen:
CC @Dup4

### DIFF
--- a/webapp/src/Controller/Jury/ContestController.php
+++ b/webapp/src/Controller/Jury/ContestController.php
@@ -503,6 +503,10 @@ class ContestController extends BaseController
             $problems = $contest->getProblems()->toArray();
             foreach ($problems as $problem) {
                 $problem->setContest($contest);
+
+                if ($problem->getAllowJudge()) {
+                    $this->dj->unblockJudgeTasksForProblem($problem->getProbid());
+                }
             }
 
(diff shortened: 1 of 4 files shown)


Dup4 (approved):

vmcj (approved):
Looks fine to me, 

Maybe we should also look into what happens when a contest is enabled/disabled as that's another way how you can easily fix an issue.

nickygerritsen:
You can’t submit in a disabled contest I think

PR #1441, opened 2022-01-23
Lookat freeze error

vmcj:
Currently the integration test ends with PHP errors. One of those is a freeze which cannot be applied. This PR was to get the DB in certain states to investigate why this is happening and if it should be fixed.

It now also proposes a fix to remove the error.

When this is fixed and the other errors are properly fixed we can add an explicit check for PHP errors and fail on those to handle a simple contest as that's what the integration test currently simulates.


vmcj:
Already mentioned in: https://github.com/DOMjudge/domjudge/pull/1442

Issue #841, opened 2020-06-26
Register: Error when affiliation name is longer than 32 chars


tom93:
### Description of the problem

During registration, if a new affiliation is added with a name longer than 32 characters, the server returns a 500 Internal Server Error page.

> An exception occurred while executing 'INSERT INTO team_affiliation (externalid, shortname, name, country, comments) VALUES (?, ?, ?, ?, ?)' with params [null, "A very long team affiliation name", "A very long team affiliation name", null, null]:
>
> SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'shortname' at row 1

### Environment

DOMjudge 7.1.3, 7.2.1, master @ aac01de9.

### Expected behaviour

Either registration should succeed, or there should be a friendly error message.
(I think this is reasonable to expect because the limit is so small.)

### Cause

The input from the user is used as both the name and the shortname of the new affiliation. The length limit of the name is 255, while the length limit of the shortname is only 32.

### Ideas for fix

 - Add validation to forbid names longer than 32 characters (simple, but needlessly restricts the length of the full name).
 - Truncate the input to 32 characters for the shortname (may result in duplicate shortnames; not sure if this is a problem, e.g. for exporting/importing).
 - Generate a unique shortname (e.g. first 20 characters of user input followed by an incrementing integer).

I'd be happy to implement one of these, just not sure what the preferred approach would be.

### Stack trace

<details>

```
Doctrine\DBAL\Exception\DriverException:
An exception occurred while executing 'INSERT INTO team_affiliation (externalid, shortname, name, country, comments) VALUES (?, ?, ?, ?, ?)' with params [null, "A very long team affiliation name", "A very long team affiliation name", null, null]:

SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'shortname' at row 1

  at /domjudge/lib/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php:106
  at Doctrine\DBAL\Driver\AbstractMySQLDriver->convertException()
     (/domjudge/lib/vendor/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php:169)
  at Doctrine\DBAL\DBALException::wrapException()
     (/domjudge/lib/vendor/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php:149)
  at Doctrine\DBAL\DBALException::driverExceptionDuringQuery()
[...]

PR #1443, opened 2022-01-26
Unique selfregister shortname

vmcj:
This is a possible fix for: https://github.com/DOMjudge/domjudge/issues/841

It still needs some styling cleanup but I would like some feedback if we indeed want this.

Also we don't have password restrictions for this form, which might be a problem for how other people use this where the people self registering are not spectators (this is how I use them for my prelims) but "real" contestants. But this is not relevant for this PR.

### DIFF
--- a/webapp/src/Controller/SecurityController.php
+++ b/webapp/src/Controller/SecurityController.php
@@ -146,12 +146,9 @@ class SecurityController extends AbstractController
                 switch ($registration_form->get('affiliation')->getData()) {
                     case 'new':
                         $affiliation = new TeamAffiliation();
-                        $shortName = $this->getShortName($registration_form->get('affiliationShortName')->getData());
-                        $name = $this->getShortName($registration_form->get('affiliationName')->getData());
-                        $this->logger->error($shortName, ['length' => strlen($shortName), 'bool' => strlen($shortName)>$shortNameLength]);
                         $affiliation
-                            ->setName($name)
-                            ->setShortname($shortName);
+                            ->setName($registration_form->get('affiliationName')->getData())
+                            ->setShortname($registration_form->get('affiliationShortName')->getData());
                         if ($registration_form->has('affiliationCountry')) {
                             $affiliation->setCountry($registration_form->get('affiliationCountry')->getData());
                         }
--- a/webapp/src/Controller/SecurityController.php
+++ b/webapp/src/Controller/SecurityController.php
@@ -146,9 +146,12 @@ class SecurityController extends AbstractController
                 switch ($registration_form->get('affiliation')->getData()) {
                     case 'new':
                         $affiliation = new TeamAffiliation();
+                        $shortName = $this->getShortName($registration_form->get('affiliationShortName')->getData());
+                        $name = $this->getShortName($registration_form->get('affiliationName')->getData());
+                        $this->logger->error($shortName, ['length' => strlen($shortName), 'bool' => strlen($shortName)>$shortNameLength]);
                         $affiliation
-                            ->setName($registration_form->get('affiliationName')->getData())
-                            ->setShortname($registration_form->get('affiliationName')->getData());
+                            ->setName($name)
+                            ->setShortname($shortName);
                         if ($registration_form->has('affiliationCountry')) {
                             $affiliation->setCountry($registration_form->get('affiliationCountry')->getData());
                         }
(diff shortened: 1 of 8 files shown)


nickygerritsen:
For `Eindhoven University of Technology` this would become `Technology0` right? (technology is longer than eindhoven).
That doesn't sound like the correct one. Also, why always append a number?

vmcj:
> For `Eindhoven University of Technology` this would become `Technology0` right? (technology is longer than eindhoven). That doesn't sound like the correct one. Also, why always append a number?

To make sure that we get a unique result, as the heuristic is not that strong (you prove that point). As the shortname is only internal I permitted myself a bit more freedom in this generation.

And although at first I agreed that your case should have "Eindhoven0" (or even "Eindhoven") but in case of:

Vrije Universiteit Amsterdam
Universiteit van Amsterdam
Hoge school van Amsterdam
High school van Amsterdam
Something in Amsterdam
You get the point it's in Amsterdam

I would say we would not prefer to get the city. 

So maybe combine the longest and shorter element? I would say that would almost always handle the case we want?

nickygerritsen:
> > For `Eindhoven University of Technology` this would become `Technology0` right? (technology is longer than eindhoven). That doesn't sound like the correct one. Also, why always append a number?
> 
> To make sure that we get a unique result, as the heuristic is not that strong (you prove that point). As the shortname is only internal I permitted myself a bit more freedom in this generation.
> 
> And although at first I agreed that your case should have "Eindhoven0" (or even "Eindhoven") but in case of:
> 
> Vrije Universiteit Amsterdam Universiteit van Amsterdam Hoge school van Amsterdam High school van Amsterdam Something in Amsterdam You get the point it's in Amsterdam
> 
> I would say we would not prefer to get the city.
> 
> So maybe combine the longest and shorter element? I would say that would almost always handle the case we want?

If the shortname is only used internally, can't we set a UUID in it upon registration? 🤔

vmcj:
> > > For `Eindhoven University of Technology` this would become `Technology0` right? (technology is longer than eindhoven). That doesn't sound like the correct one. Also, why always append a number?
> > 
> > 
> > To make sure that we get a unique result, as the heuristic is not that strong (you prove that point). As the shortname is only internal I permitted myself a bit more freedom in this generation.
> > And although at first I agreed that your case should have "Eindhoven0" (or even "Eindhoven") but in case of:
> > Vrije Universiteit Amsterdam Universiteit van Amsterdam Hoge school van Amsterdam High school van Amsterdam Something in Amsterdam You get the point it's in Amsterdam
> > I would say we would not prefer to get the city.
> > So maybe combine the longest and shorter element? I would say that would almost always handle the case we want?
> 
> If the shortname is only used internally, can't we set a UUID in it upon registration? 🤔

I still prefer to have something which one could remember if you care about the self-registered contestants for another reason.

thijskh:
I prefer [this solution](https://github.com/DOMjudge/domjudge/issues/841#issuecomment-650563633) over 'magic' that may some of the time, may not other times, generate a meaningful shortname.

Usually institutions already have such a commonly used short version of their official name, e.g. Technische Universiteit Eindhoven = "TUE" or "TU/e" (sic).

The suggested AI fails on Hogeschool Utrecht and also fails on Universiteit Utrecht if Hogeschool Utrecht also participates.

webapp/src/Form/Type/UserRegistrationType.php
```diff
@@ -139,6 +139,15 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
                     'mapped' => false,
+                ])
+                ->add('affiliationShortName', TextType::class, [
+                    'label' => false,
+                    'required' => false,
+                    'attr' => [
+                        'placeholder' => 'Affiliation shortname',
+                        'maxlength' => '32',
+                    ],
+                    'mapped' => false,
```
nickygerritsen, line 150: Useless spaces after the line

nickygerritsen (approved):

webapp/src/Controller/SecurityController.php
```diff
@@ -145,10 +145,11 @@ public function registerAction(
             if ($this->config->get('show_affiliations')) {
                 switch ($registration_form->get('affiliation')->getData()) {
                     case 'new':
+                        $shortNameLength = 32;
```
thijskh, line 148: This variable is introduced but unused.

thijskh: FYI, the shortname is already truncated with a hard `substr(,0,32)` in the Entity/TeamAffiliation class. So the form validation plus that should ensure it's never too long.

thijskh (approved):
Good work, great to see all those tests!

PR #1447, opened 2022-01-30
Add SECURITY.md to show committed PHP versions

vmcj:
Mainly to have the supported PHP versions somewhere. I dont know what we should do with the following questions so I left them open for the case where this happens:

## Reporting a Vulnerability

Use this section to tell people how to report a vulnerability.

Tell them where to go, how often they can expect to get an update on a
reported vulnerability, what to expect if the vulnerability is accepted or
declined, etc.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,12 +2,12 @@
 
 ## Supported Versions
 
-| Version | Supported          | PHP version supported |
-| ------- | ------------------ | --------------------- |
-| 8.x.x   | :warning:          | 7.4-8.1               |
-| 8.0.x   | :white_check_mark: | 7.2-8.0               |
-| 7.3.x   | :white_check_mark: | 7.2-7.4               |
-| < 7.3   | :x:                | :x:                   |
+| DOMjudge Version | Supported          | PHP version supported |
+| ---------------- | ------------------ | --------------------- |
+| 8.x.x            | :warning:          | 7.4-8.1               |
+| 8.0.x            | :white_check_mark: | 7.2-8.0               |
+| 7.3.x            | :white_check_mark: | 7.2-7.4               |
+| < 7.3            | :x:                | :x:                   |
 
 ## Reporting a Vulnerability
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,17 @@
+# Security Policy
+
+## Supported Versions
+
+| Version | Supported          | PHP version supported |
+| ------- | ------------------ | --------------------- |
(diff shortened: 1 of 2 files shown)


thijskh:
Not sure how this relates to security per se. Is it not something more to elaborate on with our system requirements?

vmcj:
> Not sure how this relates to security per se. Is it not something more to elaborate on with our system requirements?

Well it does also list the supported DJ versions, I just added the PHP versions because it was convenient to also have them there.

thijskh:
And I guess we will then not ship this file in the tarball?

thijskh:
I'm not sure about the check mark behind the current development version. Is that actually "supported"?

vmcj:
> I'm not sure about the check mark behind the current development version. Is that actually "supported"?

I think indeed only supporting after releasing is better.

thijskh (approved):

thijskh:
Or we could put ⚠️ for main. Since I do believe we'd more or less offer support to people using it.

vmcj:
> Or we could put ⚠️ for main. Since I do believe we'd more or less offer support to people using it. And it's a bit in our interest that people do use it if they dare.

Yes, that makes sense also because we will fix security issues in that release.

SECURITY.md
```diff
@@ -0,0 +1,22 @@
...
+## Supported Versions
+
+| DOMjudge Version | Supported          | PHP version supported |
+| ---------------- | ------------------ | --------------------- |
+| 8.x.x            | :warning:          | 7.4-8.1               |
+| 8.0.x            | :white_check_mark: | 7.2-8.0               |
+| 7.3.x            | :white_check_mark: | 7.2-7.4               |
+| < 7.3            | :x:                | :x:                   |
+
+## Reporting a Vulnerability
```
meisterT, line 12: Move this up?

SECURITY.md
```diff
@@ -0,0 +1,22 @@
...
+
+If you want to report a vulnerability, please do not use the issue tracker
+or pull request, but instead contact security@domjudge.org so we can assess
```
meisterT, line 15: I assume that mailing list / alias exists already?

SECURITY.md
vmcj: Its the default template, so I rather keep it default.

SECURITY.md
vmcj: According to the domjudge.org server yes,

SECURITY.md
meisterT: I didn't know.

meisterT (approved):

PR #1448, opened 2022-01-31
Upgrade to newer versions of our dependencies.

nickygerritsen:
This updates all our dependencies to newer versions. For Symfony we are on the latest LTS, for PHPUnit we have upgraded one major version and for some other packages I also updated.

All code changes you see can be put in these categories:

- Adding return types or `@return` typehints to methods. This is because Symfony will see these as deprecations otherwise. Most are done by a tool supplied by Symfony, but some are manual.
- Upgrade the security system to not use Guard anymore, as Symfony suggests.
- Fixing the tests with a new login method and clearing the right container.
- Minor changes for more deprecations.
- Moved the migrations to the new suggested directory.

Note that we can also add type-hints to properties, but I did not do that yet, except for the authentication classes. I think it's best to do this at a later point. Either in one big PR or gradually while working on things. I'd prefer the former. We can then get rid of a lot of docblocks that only specify the type of properties, thus reducing the number of code lines quite a bit.

I tested in PHP 7.4, 8.0 and 8.1 locally and it seemed to all work, but I did not go through every page. Unit tests worked, but I am totally expecting CI to still fail for a bit. I'll fix it.


webapp/src/Controller/Jury/RejudgingController.php
```diff
@@ -37,10 +37,10 @@
 use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\Routing\Annotation\Route;
 use Symfony\Component\Routing\RouterInterface;
+use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 
 /**
  * @Route("/jury/rejudgings")
- * @IsGranted("ROLE_JURY")
```
vmcj, line 43: Is this with intent?

nickygerritsen: I think this is an old commit, right? It should be back again.

webapp/src/Controller/API/AbstractRestController.php
```diff
@@ -268,10 +268,7 @@ protected function listActionHelper(Request $request)
         $queryBuilder = $this->getQueryBuilder($request);
 
         if ($request->query->has('ids')) {
-            $ids = $request->query->get('ids', []);
-            if (!is_array($ids)) {
-                throw new BadRequestHttpException('\'ids\' should be an array of ID\'s to fetch');
```
vmcj, line 273: Why can this not happen anymore?

webapp/src/Controller/API/MetricsController.php
```diff
@@ -101,7 +101,7 @@ public function prometheusAction(Request $request): Response
             /** @var Submission[] $submissions */
             list($submissions, $submissionCounts) = $this->submissionService->getSubmissionList([$contest->getCid() => $contest], [], 0);
             foreach ($submissionCounts as $kind => $count) {
-                $m['submissions_' . $kind]->set($count, $labels);
+                $m['submissions_' . $kind]->set((float)$count, $labels);
```
vmcj, line 104: Why is this a float?

webapp/src/Controller/Jury/RejudgingController.php
```diff
@@ -70,24 +71,24 @@ class RejudgingController extends BaseController
     protected $router;
 
     /**
-     * @var SessionInterface
+     * @var RequestStack
      */
-    protected $session;
+    protected $requestStack;
```
vmcj, line 76: Why can you not add the type to it now?

webapp/src/Service/ScoreboardService.php
```diff
@@ -800,7 +800,7 @@ public function initializeScoreboardFilter(Request $request, Response $response)
      * @param Contest $contest
      * @return array
```
vmcj, line 801: Can be removed?

webapp/src/Service/ScoreboardService.php
```diff
@@ -761,7 +761,7 @@ public function refreshCache(Contest $contest, $progressReporter = null)
      * @param Response|null $response
      * @return Filter
```
vmcj, line 762: Can be removed?

webapp/src/Service/ImportExportService.php
```diff
@@ -400,7 +400,7 @@ public function getTeamData(): array
      * @return array
```
vmcj, line 400: Can be removed?

webapp/src/Service/ImportExportService.php
```diff
@@ -85,7 +85,7 @@ public function __construct(
...
```
vmcj, line 85: Can be removed?

webapp/src/Service/ICPCCmsService.php
```diff
@@ -184,7 +184,7 @@ public function importTeams(string $token, string $contest, string &$message = n
      * @param string|null $message
      * @return bool
```
vmcj, line 185: Can be removed?

vmcj (commented):
I really like the return types.

webapp/src/Controller/Jury/RejudgingController.php
vmcj: Its outdated, so probably you're right.

webapp/phpunit.xml.dist
```diff
@@ -1,46 +1,36 @@
...
-    <extensions>
-        <extension class="DAMA\DoctrineTestBundle\PHPUnit\PHPUnitExtension"/>
-    </extensions>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd" backupGlobals="false" colors="true" bootstrap="config/bootstrap.php">
+  <coverage processUncoveredFiles="true">
+    <include>
+      <directory suffix=".php">src</directory>
+    </include>
+    <exclude>
+      <directory suffix=".php">src/Migrations</directory>
```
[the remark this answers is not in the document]
nickygerritsen: Yup

webapp/src/Controller/API/AbstractRestController.php
nickygerritsen: ->get with arrays is deprecated. Using ->all returns an array. Errors are handled by Symfony

webapp/src/Controller/API/MetricsController.php
nickygerritsen: Int is better. I think it is string currently

tests/syntax
```diff
@@ -22,11 +22,11 @@ if [ ! -x /usr/bin/shellcheck ]; then
 	exit -1
 fi
 
-find .. \( -path ../lib/vendor            -prune \
-	-o -path ../webapp/var            -prune \
-	-o -path ../output                -prune \
-	-o -path ../.git                  -prune \
-	-o -path ../webapp/src/Migrations -prune \
+find .. \( -path ../lib/vendor        -prune \
```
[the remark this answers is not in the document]
nickygerritsen: Yeah because of tabs vs spaces. Not going to change it now.

vmcj (approved):
To me this looks fine.

PR #1449, opened 2022-02-04
Fix composer installation for baseinstall.sh

swarkentin:
* `apt`-sourced `composer` is out of date -- install using composer setup script
  instead.


nickygerritsen (approved):
Ah yeah. When I did my update last week I also “updated” to a composer 2.x requirement. 

I think this is totally fine, especially since you don’t need it during install, only when doing development or building packages.

So this fix seems good to me but if others want to stick to the old and slow composer 1.x we could downgrade some packages.

Our package build scripts might also need updates, will look into it.

PR #1451, opened 2022-02-05
Get rid of PHP 8.1 deprecations

nickygerritsen:
This does a few things:

* Update monolog and ramsey/collection to get rid of some external deprecations related to (un)serialize and DateTime.
* Use the new way of (un)serializing in the User class.
* Explicitly cast a variable to int somewhere for one less deprecation.
* Replace `strftime` with `date`.

Most are trivial changes, except for the last one. To quote PHP, strftime

> has been DEPRECATED as of PHP 8.1.0. Relying on this function is highly discouraged.

So we use the suggested (at least on the internet by multiple people) replacement `date`. That does change the format arguments, so I hope I did all of them correctly. People who have a custom `time_format` config setting need to change it, but I don't know where/how we could mention that.

After this we lost thousands of deprecations warnings in the test suite, where the 5 deprecations that we do have left are now equal between PHP 7.4, 8.0 and 8.1.

### DIFF
--- a/etc/db-config.yaml
+++ b/etc/db-config.yaml
@@ -223,7 +223,7 @@
             description: Print times of contest events relative to contest start (instead of absolute).
         -   name: time_format
             type: string
-            default_value: "%H:%M"
+            default_value: "H:i"
             public: false
             description: The format used to print times. For formatting options see the PHP 'strftime' function.
         -   name: thumbnail_size
--- a/etc/db-config.yaml
+++ b/etc/db-config.yaml
@@ -225,7 +225,7 @@
             type: string
             default_value: "H:i"
             public: false
-            description: The format used to print times. For formatting options see the PHP 'strftime' function.
+            description: The format used to print times. For formatting options see the PHP 'date' function.
         -   name: thumbnail_size
             type: int
             default_value: 200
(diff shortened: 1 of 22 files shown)


thijskh (approved):

etc/db-config.yaml
```diff
@@ -223,7 +223,7 @@
             description: Print times of contest events relative to contest start (instead of absolute).
         -   name: time_format
             type: string
-            default_value: "%H:%M"
+            default_value: "H:i"
             public: false
             description: The format used to print times. For formatting options see the PHP 'strftime' function.
```
vmcj, line 228:
```suggestion
            description: The format used to print times. For formatting options see the PHP 'date' function.
```

Not sure if this is the correct suggestion, but you implied some formatting has changed.

nickygerritsen: Yup. Thanks

eldering:
At least mention this as a backward incompatible change in the ChangeLog clearly.

nickygerritsen:
> At least mention this as a backward incompatible change in the ChangeLog clearly.

Yes that was my plan. I can push that to this PR

meisterT (approved):

PR #1452, opened 2022-02-06
Improve w3cvalue

vmcj:
The only one which cannot be removed is the one about <option> without text in it. We use this to have a dropdown menu with nothing selected.

We can either:
- cheat by adding a <style display:none>, 
- add a label with a default text but that will show up in the UI.
- add a default text and leave the attribute value empty.

none of these are better IMO.

### DIFF
--- a/gitlab/webstandard.sh
+++ b/gitlab/webstandard.sh
@@ -77,7 +77,12 @@ if [ "$TEST" = "w3cval" ]; then
     rm -rf localhost/domjudge/bundles/nelmioapidoc*
     rm -f localhost/domjudge/css/bootstrap.min.css*
     rm -f localhost/domjudge/css/select2-bootstrap.min.css*
+    rm -f localhost/domjudge/jury/config/check/phpinfo*
     section_end upstream_problems
+
+    section_start_collap upstream_icpc "Remove icpc only features"
+    rm -rf localhost/domjudge/jury/import-export/export/results-icpc.html*
+    section_end upstream_icpc
 
     section_start_collap test_suite "Install testsuite"
     cd $DIR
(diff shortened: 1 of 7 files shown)


webapp/src/Entity/ContestProblem.php
```diff
@@ -46,12 +46,13 @@ class ContestProblem
     private string $shortname;
 
     /**
-     * @ORM\Column(type="integer", name="points", length=4,
+     * @ORM\Column(type="integer", name="points",
```
nickygerritsen, line 49: This will result in a migration update if we would run `webapp/bin/console doctrine:migrations:diff`. Why do we want this?

vmcj: Its a bit ugly but:

login here: admin/admin
https://www.domjudge.org/demoweb/jury/contests/7/edit

Search for: #contest_problems_0_points
see the attribute maxlength which is not valid for input['number']

I think the migration was not needed here but I might have overlooked this, but as it works in CI I think this is also ignored for doctrine when using integers?

nickygerritsen: afaik length is used for integers, it makes it into a `int(4)` in the database, but I might be wrong.
Migrations aren't enforced, so it makes sense CI doesn't complain.
Also we have `length=x` for many many integers in our entity, so if Doctirne indeed ignores it (which I think it doesn't) we need to get rid of them all.

nickygerritsen: Update: adding `, IntegerType::class` to the builder in `ContestProblemType.php` is the better fix

vmcj: And is now committed

nickygerritsen (approved):

PR #1455, opened 2022-02-07
Refresh whole page when contest changed from other tab. Fixes #1261.

### LINKED ISSUE #1261: Auto-refreshed page uses contest ID updated in a different tab, but doesn't update menu

I'm using DOMjudge current main and have multiple browser tabs open, at least one on the jury interface and one team interface (same logged in user). When I switch the contest in a first tab, and then in a second tab the page auto refreshed (e.g. the jury submissions page), it loads data for the new contest that was selected in the first tab, but still shows the old contest ID in the menu. Only when I actively reload the page, does it show the newly selected contest.

I'm not sure we'd want to support a contest per tab, but at least when a tab auto-refreshes, the contest shown in the menu should be in sync with what's displayed on the page.

nickygerritsen:


webapp/public/js/domjudge.js
```diff
@@ -425,6 +425,15 @@ function processAjaxResponse(jqXHR, data) {
     if (jqXHR.getResponseHeader('X-Login-Page')) {
         window.location = jqXHR.getResponseHeader('X-Login-Page');
     } else {
+        var newCurrentContest = parseInt(jqXHR.getResponseHeader('X-Current-Contest'));
+        var currentContest = parseInt($('[data-current-contest]').data('current-contest'));
+
+        // If the contest ID changed from another tab, reload or whole page
+        if (newCurrentContest !== currentContest) {
+            window.location.reload();
```
thijskh, line 433: We use `location.reload()` in the same file, should probably be consistent either way.

webapp/src/Service/DOMJudgeService.php
```diff
@@ -143,6 +143,11 @@ public function getCurrentContests(?int $onlyofteam = null, bool $alsofuture = f
         return $qb->getQuery()->getResult();
     }
 
+    public function getCurrrentContestId(): ?int
```
thijskh, line 146: I would name this more specifically as to relating to the cookie. Now it "sounds" to me related to getCurrentContest() and getCurrentContests() while in fact it's slightly different.

thijskh (approved):

webapp/public/js/domjudge.js
nickygerritsen: Seems we use window.location way more often so updated the one place we didn't to use it.

PR #1457, opened 2022-02-08
Use .ans instead of .out for testcase output. Fixes #1302.

### LINKED ISSUE #1302: Sample set zips use the .out extension instead of .ans

Sample set zipfile contains .out files instead of .ans for answers

### Your environment
-

### Steps to reproduce
Download the sample zip file

### Expected behaviour
Expected answers are in files with the .ans extension

### Actual behaviour
Expected answers are in files with the .out extension

-

nickygerritsen:


vmcj (approved):
So for all jury provided answers either in upload or download this would be .ans

and for participants we have .out?

nickygerritsen:
Indeed. My idea is “official” answers are .ans, team output is .out

webapp/templates/jury/problem_testcases.html.twig
```diff
@@ -94,7 +94,7 @@
                 <tr>
                     <td class="filename">
                         <a href="{{ path('jury_problem_testcase_fetch', {'probId': problem.probid, 'rank': testcase.rank, 'type': 'output'}) }}">
-                            {{ testcase.downloadName }}.out
+                            {{ testcase.downloadName }}.ans
```
thijskh, line 97: Here we don't use the class constant anymore?

nickygerritsen: I can change that. Maybe set it in the controller as var since constants I’m twig look kinda ugly

thijskh: Makes sense to me because having one other place where it's hardcoded that will definitely be overlooked

nickygerritsen: Agreed. Will do it

PR #1461, opened 2022-02-08
Extend CrawlAll with riskier endpoints.

vmcj:
Some endpoints change the state of the system such that it can affect of next pages. This PR tries to introduce these endpoints and the needed fixes.


nickygerritsen (approved):

vmcj:
The problem why this fails is that he resulting PHPUnit report becomes invalid XML because of the binary data.

Issue #1349, opened 2021-11-21
Shadow import improvements


nickygerritsen:
The shadow import command (`ImportEventFeedCommand`) can use some improvements:
* We should not import static data (problems, teams, organizations, contests, judgement types, languages, groups) because we can import them from the CAF. What we probably should do is validate that the data we get matches what we already have.
* Currently we output all log entries to stdout, including info. Which is useful to see if the thing is actually running but not if there is an error or warning. We should probably write the log to a file instead, at least the non-warning/error entries.
* It would be nice if we would show information in the DOMjudge UI, especially:
  * If the import is still running
  * Any actionable warning / error entries

This last point means that we need to make sure that as soon as a warning / error is not a warning / error anymore, we remove it. For example, if we receive a judgement with an unknown submission ID currently we log a warning. If we later DO get that submission, we process the judgement at that point. Thus the warning should go away.

I was thinking that maybe we should add two tables/entities to support this:
* Something like `ExternalContestSource` that contains the data we currently put in [the shadowing.yaml file](https://github.com/DOMjudge/domjudge/blob/main/etc/shadowing.yaml.dist). This means we don't need a config file anymore.
* Something like `ShadowErrors` that contain these errors, linked to an `ExternalContestSource`.

If we get an event which misses dependant events (like the judgement example above) we currently store this in memory. If the import script crashes, this would get lost and you need to read the feed from the start again. We could also store this data in a table instead to not have this issue. I'm not sure if we should though.

Any more ideas and thoughts are very welcome. Let's define first what we want and then implement stuff.

nickygerritsen:
This is implemented now and will be tested at SWERC.

PR #1464, opened 2022-02-10
Add separate ICPC ID field to team affiliation table.

nickygerritsen:
This PR makes it such that:
* All endpoints we support from the CLICS spec that have an ICPC ID, have such a field in our database as well.
* We support external ID's for teams and team categories. We didn't in the past, which meant they were limited to (globally unique) integers.
* Cleans up some logic with respect to the event logging: our external ID field is always called `externalid` so no need to have that as a config variable.
* Sets default external ID's for our teams, categories and affiliations.

I have split up the code in 3 commits, one each for each entity.

I did this PR first, since I want to work on #1349 but this touches the same code.

### DIFF
--- a/doc/manual/import.rst
+++ b/doc/manual/import.rst
@@ -86,6 +86,7 @@ It should be a JSON array with objects, each object should contain the following
 fields:
 
 - ``id``: the external affiliation ID. Must be unique
+- ``icpc_id`` (optional): an external ID, e.g. from the ICPC CMS, may be empty
 - ``name``: the affiliation short name as used in the jury interface and certain
   exports
 - ``formal_name``: the affiliation name as used on the scoreboard
@@ -95,11 +96,13 @@ Example ``organizations.json``::
 
   [{
     "id": "INST-42",
+    "icpc_id": "42",
     "name": "LU",
     "formal_name": "Lund University",
     "country": "SWE"
   }, {
     "id": "INST-43",
+    "icpc_id": "43",
     "name": "FAU",
     "formal_name": "Friedrich-Alexander-University Erlangen-Nuremberg",
     "country": "DEU"
--- a/doc/manual/import.rst
+++ b/doc/manual/import.rst
@@ -22,7 +22,8 @@ Prepare a file called ``groups.json`` which contains the team categories.
 It should be a JSON array with objects, each object should contain the following
 fields:
 
-- ``id``: the (integer) category ID to use. Must be unique
+- ``id``: the category ID to use. Must be unique
+- ``icpc_id`` (optional): an external ID, e.g. from the ICPC CMS, may be empty
 - ``name``: the name of the team category as shown on the scoreboard
 - ``hidden`` (defaults to ``false``): if ``true``, teams in this category will
   not be shown on the scoreboard
@@ -33,6 +34,7 @@ Example ``groups.json``::
 
   [{
     "id": "13337",
+    "icpc_id": "123",
     "name": "Companies",
     "hidden": true
   }, {
@@ -57,7 +59,7 @@ Prepare a file called ``groups.tsv`` which contains the team categories.
 The first line should contain ``File_Version 1`` (tab-separated).
 Each of the following lines must contain the following elements separated by tabs:
 
-- the (integer) )category ID. Must be unique
+- the category ID. Must be unique
 - the name of the team category as shown on the scoreboard
 
 Example ``groups.tsv``::
--- a/doc/manual/import.rst
+++ b/doc/manual/import.rst
@@ -129,7 +129,7 @@ Prepare a file called ``teams.json`` which contains the teams.
 It should be a JSON array with objects, each object should contain the following
 fields:
 
-- ``id``: the (integer) team ID. Must be unique
+- ``id``: the team ID. Must be unique
 - ``icpc_id`` (optional): an external ID, e.g. from the ICPC CMS, may be empty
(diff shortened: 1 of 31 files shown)


webapp/src/Command/ImportEventFeedCommand.php
```diff
@@ -1006,7 +996,7 @@ protected function importGroup(array $event): void
         $this->eventLogService->log('groups', $category->getCategoryid(), $action,
                                     $this->contestId);
 
-        $this->processPendingEvents('group', $category->getCategoryid());
+        $this->processPendingEvents('group', $category->getExternalid());
```
vmcj, line 999: Can this be null?

webapp/migrations/Version20220210155918.php
```diff
@@ -0,0 +1,37 @@
...
+    {
+        return 'Add external and ICPC ID fields to team categories';
+    }
+
+    public function up(Schema $schema) : void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE team_category ADD externalid VARCHAR(255) DEFAULT NULL COLLATE `utf8mb4_bin` COMMENT \'Team category ID in an external system\' AFTER `categoryid`, ADD icpcid VARCHAR(255) DEFAULT NULL COLLATE `utf8mb4_bin` COMMENT \'External identifier from ICPC CMS\' AFTER `externalid`');
```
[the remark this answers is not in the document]
nickygerritsen: We don’t have it for other external
ID field so I’d like to not add it only here now

webapp/src/Command/ImportEventFeedCommand.php
nickygerritsen: We assume you run in datasource>=1 mode when importing. We even check for this. So then it’s always set

nickygerritsen: I'm thinking we should even block running this with datasource=0. Currently we still have the `allow-import-as-primary` flag but it can only break stuff, so why should we?

vmcj: What was the usecase for it?

webapp/tests/Unit/Controller/Jury/TeamControllerTest.php
```diff
@@ -17,7 +17,7 @@ class TeamControllerTest extends JuryControllerTest
     protected static string  $className                = Team::class;
     protected static array   $DOM_elements             = ['h1' => ['Teams']];
     protected static string  $addForm                  = 'team[';
-    protected static array   $addEntitiesShown         = ['name', 'icpcid', 'displayName', 'room'];
```
vmcj, line 20: Why dont we display the name anymore?

webapp/src/Command/ImportEventFeedCommand.php
nickygerritsen: 🤷‍♂️ I think it is never used. If we want to get rid of it I will do it as part of #1349

webapp/src/Controller/Jury/TeamController.php
```diff
@@ -109,8 +109,7 @@ public function indexAction(): Response
         $table_fields = [
             'teamid' => ['title' => 'ID', 'sort' => true,],
             'icpcid' => ['title' => 'ICPC ID', 'sort' => true,],
-            'name' => ['title' => 'teamname', 'sort' => true, 'default_sort' => true],
-            'display_name' => ['title' => 'display name', 'sort' => true, 'default_sort' => true],
+            'effective_name' => ['title' => 'name', 'sort' => true, 'default_sort' => true],
```
meisterT, line 112: you are still using `displayName` below, but here `effective_name`. Is that intentional?

doc/manual/import.rst
```diff
@@ -129,7 +129,7 @@ Prepare a file called ``teams.json`` which contains the teams.
 It should be a JSON array with objects, each object should contain the following
 fields:
 
-- ``id``: the (integer) team ID. Must be unique
+- ``id``: the team ID. Must be unique
```
meisterT, line 132: when is this no longer an int?

nickygerritsen: When you import from something that is not the icpc cms it can be. Fact is the api spec says it’s a string, we just only supported ints because that is what we used. I don’t see a reason not to allow other strings

webapp/src/Controller/Jury/TeamController.php
nickygerritsen: Where exactly? In the form? Then yes, since effectiveName is the method that checks displayNamw and otherwise name

doc/manual/import.rst
meisterT: so we are no longer using our own auto-increment ids then? In places where we display currently `t42` (as example) we display in the future `t<whateverexternalsystem decided to enter :-)>`?

webapp/src/Controller/Jury/TeamController.php
meisterT: Perhaps I am misunderstanding the distinction between effectiveName and displayName then. What is the difference and why do we need both?

doc/manual/import.rst
nickygerritsen: Noooo only in the api. Just like we already did with contests, problems, etc. This just unifies it

PR #1468, opened 2022-02-20
Shadow import improvements.

nickygerritsen:
This drastically improves our shadowing import script.

This PR does the following:

- Some small bugfixes for importing since my recent PR.
- Fix submission filter form remembering choices (unrelated but I needed something similar and it was broken).
- Created an external contest source entity to keep track of sources to import from. Get rid of `shadowing.yml`.
- Drop never used options from the event feed import script.
- Created an entity for keeping track of warnings during import. Warnings go away after they are resolved.
- Removed memory leaks from import script in dev mode.
- Updated the documentation.
- Show badges for shadow differences and import warnings.

Some screenshots:
image
image
image
image
image

This should fix all things that are noted in #1349.


nickygerritsen:
Note that I did separate commits to maybe make reviewing a bit better. The biggest single change codewise is moving everything to the service. I have not done much more in that commit besides adding the external contest source entity, so hopefully it's readable enough.

nickygerritsen:
BTW, the fact that the import itself is now in a service makes it that we *could* add tests for this. But they are big and cumbersome and I spent enough time on this, so I leave that for the future.

webapp/src/Command/ImportEventFeedCommand.php
```diff
@@ -177,11 +177,11 @@ protected function configure(): void
...
-        // Disable SQL logging if we do not run explicitly in debug mode.
-        // This would cause a serious memory leak otherwise since this is a
-        // long running process.
-        if (!$this->debug) {
-            $this->em->getConnection()->getConfiguration()->setSQLLogger(null);
+        // Disable SQL logging and profiling. This would cause a serious memory leak otherwise
+        // since this is a long running process.
+        $this->em->getConnection()->getConfiguration()->setSQLLogger();
+        if ($this->profiler) {
+            $this->profiler->disable();
```
vmcj, line 111: This is only for the php process running the import command?

nickygerritsen: It is for the current PHP process yes, just like we do in the judgehost API.

webapp/src/Command/ImportEventFeedCommand.php
```diff
@@ -916,127 +905,88 @@ protected function validateLanguage(array $event): void
...
-                $this->em->remove($category);
-                $this->em->flush();
-                $this->eventLogService->log('groups', $category->getCategoryid(),
-                                            EventLogService::ACTION_DELETE,
-                                            $this->contestId, null, $category->getCategoryid());
-                return;
-            } else {
-                $this->logger->error('Cannot delete nonexistent group %s', [ $groupId ]);
+                $this->logger->warning(
+                    'Group with external ID %s deleted in feed but found in DOMjudge',
```
vmcj: Where do we try to delete the category?

webapp/src/Command/ImportEventFeedCommand.php
```diff
@@ -1046,15 +996,18 @@ protected function importProblem(array $event): void
...
 
         $problemId = $event['data']['id'];
 
         // First, load the problem
         /** @var Problem $problem */
         $problem = $this->em->getRepository(Problem::class)->findOneBy(['externalid' => $problemId]);
         if (!$problem) {
-            $this->logger->error('Problem %s not found, cannot import', [ $problemId ]);
+            $this->logger->warning(
+                'Problem with external ID %s nout found in DOMjudge',
```
vmcj: typo.

nickygerritsen: Is this still in the last commit? If not, let's ignore it 🙈

webapp/src/Command/ImportEventFeedCommand.php
```diff
@@ -1064,161 +1017,64 @@ protected function importProblem(array $event): void
...
-        }
-
-        if ($contestProblem->getColor() !== ($event['data']['rgb'] ?? null)) {
-            $this->logger->warning(
-                'Contest problem color from DOMjudge (%s) does not match feed (%s)',
-                [ $contestProblem->getColor(), $event['data']['rgb'] ]
-            );
-            $contestProblem->setColor($event['data']['rgb'] ?? null);
+        if (!$contestProblem) {
+            $this->logger->warning("No contest problem found in DOMjudge yet");
```
vmcj: I read this as that there is no contest problem at all, whereas the code should say something like `Read contest problem does not exist in DOMjudge`?

nickygerritsen: This is also resolved in a later commit.

etc/db-config.yaml
```diff
@@ -303,3 +303,8 @@
             public: false
             description: URL of an ICPC iCAT instance if such is available; will be linked to from the submission verification box.
             docdescription: See https://github.com/icpc-live/autoanalyst
+        -   name: external_contest_source_critical
+            type: int
+            default_value: 120
+            public: false
+            description: Time in seconds after an external contest source reader last checked in before showing its status as "critical".
```
vmcj, line 310: Aren't we the one checking in?

nickygerritsen: I don't follow? The 'external contest source reader' is checking in.

webapp/src/Service/DOMJudgeService.php
```diff
@@ -286,11 +287,14 @@ public function getUpdates(): array
...
-        $rejudgings      = [];
-        $internal_errors = [];
-        $balloons        = [];
+        $clarifications           = [];
+        $judgehosts               = [];
+        $rejudgings               = [];
+        $internal_errors          = [];
+        $balloons                 = [];
+        $shadow_difference_count  = 0;
+        $external_contest_sources = [];
```
vmcj, line 296: These don't align with the one below.

nickygerritsen: wait, only the bottom one doesnt align right?

etc/db-config.yaml
nickygerritsen: No? I still don't follow. This number is used like this:
If the daemon running the feed importer didn't report itself for more than this value, alert like crazy.

doc/manual/shadow.rst
```diff
@@ -35,59 +36,53 @@ use DOMjudge as system to shadow from, you should enter
...
+configuration data, like teams, team categories and team affiliations should also
+exist. The easiest way to create the contest and configuration data is to import
+it. For this you need JSON files as described in the :doc:`Import <import>` chapter.
+Furthermore you need problem directories as described on the that same page. Place
+all the files and all directories together in one directory and use the
+``misc-tools/import-contest.sh`` script to import them. Note that
 it will complain that no team is associated with your account so it can't import
-jury submissions. That is fine since these submissions will be read using the
-event feed later on.
+jury submissions if you have no team linked to your account. That is fine since
```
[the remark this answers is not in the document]
nickygerritsen: Not really but reworded it

doc/manual/shadow.rst
```diff
@@ -35,59 +36,53 @@ use DOMjudge as system to shadow from, you should enter
...
-- Set ``id`` to the (database) ID of the contest in DOMjudge that you created.
-- Set the ``feed-url``, ``username`` and ``password`` fields to the correct data
-  as provided by the primary system you will be shadowing.
-- Check if you need any of the other options like ``allow-insecure-ssl``. All
-  other options should only be used if you know what you are doing.
+The contest needs to exist in DOMjudge *and* have all problems loaded. All other
+configuration data, like teams, team categories and team affiliations should also
+exist. The easiest way to create the contest and configuration data is to import
+it. For this you need JSON files as described in the :doc:`Import <import>` chapter.
+Furthermore you need problem directories as described on the that same page. Place
```
vmcj, line 43:
```suggestion
Furthermore you need problem directories as described on that same page. Place
```

vmcj (commented):
I dont think I know enough about this approve, but I cannot find mistakes with it.

meisterT (approved):

PR #1470, opened 2022-02-21
Font balloon

vmcj:
Although the icon we want is behind a paywall and I cannot find a free pro license for open source projects upgrading is always better IMO.


nickygerritsen (approved):

vmcj:
Looking at the UI diffs its only pixel differences.

PR #1473, opened 2022-02-22
Use javascript to make config settings page better.

nickygerritsen:
In the past we needed to submit the page every time we wanted to add a row to collections on the config page. Now we just have add/remove buttons.

Diff looks big but if you ignore whitespace it is not so big,

Screenshots:
image
image

### DIFF
--- a/webapp/templates/jury/config.html.twig
+++ b/webapp/templates/jury/config.html.twig
@@ -80,99 +80,112 @@
                                             {% endif %}
                                         {% elseif option.type == 'array_keyval' %}
                                             <br/>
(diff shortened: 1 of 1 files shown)


thijskh (approved):
Yes, works good and a lot better. I hope people do not expect that the delete is instant though. As you still need to press 'save changes' at the bottom.

nickygerritsen:
> Yes, works good and a lot better. I hope people do not expect that the delete is instant though. As you still need to press 'save changes' at the bottom.

This is the same as other places we have this delete button (problems in a contest, extensions of a language) so at least it's consistent

PR #1474, opened 2022-02-22
Remove superfluous method descriptions that don't add any real explan…

meisterT:
…ation.


webapp/src/Twig/TwigExtension.php
```diff
@@ -1007,25 +936,19 @@ public function descriptionExpand(?string $description = null): string
     }
 
     /**
-     * Whether to show the external ID for the given entity
+     * Whether to show the external ID for the given entity.
```
vmcj, line 939: Due to the return type I think this is also implied.

vmcj (approved):

nickygerritsen (approved):

Issue #1418, opened 2022-01-09
Whether to consider adding Git Commit ID to System Informaion


Dup4:
<p align="center">

<img width="320px" src="https://user-images.githubusercontent.com/38343778/148670702-f93ed7b2-7ebc-4182-8497-f1d6985e1673.png">

</p>

Let's take a look at the current `System information`.

It contains a version of DOMjudge.

But does not contain the **Git Commit ID**.

This is actually not friendly to users who directly pull the code from Git and install it.

For example, we are using `8.0.0DEV`, but when reporting the problem, you cannot know exactly which commit we are based on.

May be some problems have been solved, but we have not re-pulled the code to update it.

thijskh:
Good idea, make it so

vmcj:
@Dup4 will you make a PR for this?

Dup4:
> @Dup4 will you make a PR for this?

Okay, these days I'll finish my work first, then try it out.

vmcj:
I'll close this issue as you added the feature in https://github.com/DOMjudge/domjudge/pull/1476.

PR #1476, opened 2022-02-24
Add Git Commit ID

Dup4:
Implementation for https://github.com/DOMjudge/domjudge/issues/1418.

BTW, if build by directly downloading the code, will not be able to get the `Git Commit ID`, like the we get the code [in the current docker build](https://github.com/DOMjudge/domjudge-packaging/blob/2d440b10c0f4d168f325e460ba5c775836722fc0/docker/build.sh#L23).

Preview:

image

image

image

image

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
 # Process this file with autoconf to produce a configure script.
 AC_INIT([DOMjudge],
-        m4_esyscmd([grep 'version' README.md | sed -n '1s/^.*version //p' | tr -d '\n']),
+        m4_esyscmd([grep 'version' README.md | sed -n '1s/^.*version //p' | sed "s/$/$(command -v git >/dev/null 2>&1 && git rev-parse --is-inside-work-tree >/dev/null 2>&1 && git rev-parse --short HEAD | sed 's|^|\\/|' || echo '' )/" | tr -d '\n']),
         [domjudge-devel@domjudge.org])
 AC_PREREQ([2.64])
 AC_CONFIG_HEADERS([etc/config.h])


nickygerritsen:
I wonder if we an make the `make dist` (which is what is used to build the downloads) add the git commit. Then in that case for new .tar.gz's we will have the commit in there. Maybe that even already works with your change?

Dup4:
> I wonder if we an make the `make dist` (which is what is used to build the downloads) add the git commit. Then in that case for new .tar.gz's we will have the commit in there. Maybe that even already works with your change?

I think maybe it won't work.

after executing `make dist`, will generate `configure`.

image

We can see that `GIT_COMMIT` is assigned by calling `git` in `configure`.

nickygerritsen:
So think when running `make dist` we should write the git commit hash in some file that we package with the rest, and have `configure` read that file if it can't run `git`.

Dup4:
what you said makes sense.

Maybe we should use `m4_esyscmd` to assign a value to `GIT_COMMIT` if needed.

like below:

image

after call `make dist`:

image

nickygerritsen:
Looks good in general. What do you think about this patch, to make it 'live' in debug mode when we have a `.git` directory:

```diff
diff --git a/webapp/src/Controller/API/GeneralInfoController.php b/webapp/src/Controller/API/GeneralInfoController.php
index c462635ec..33e28c055 100644
--- a/webapp/src/Controller/API/GeneralInfoController.php
+++ b/webapp/src/Controller/API/GeneralInfoController.php
@@ -99,7 +99,7 @@ class GeneralInfoController extends AbstractFOSRestController
         return [
             'api_version' => static::API_VERSION,
             'domjudge_version' => $this->getParameter('domjudge.version'),
-            'domjudge_gitcommit' => $this->getParameter('domjudge.gitcommit'),
+            'domjudge_gitcommit' => $this->dj->getGitCommit(),
             'environment' => $this->getParameter('kernel.environment'),
             'doc_url' => $this->router->generate('app.swagger_ui', [], RouterInterface::ABSOLUTE_URL),
         ];
diff --git a/webapp/src/Controller/Jury/ConfigController.php b/webapp/src/Controller/Jury/ConfigController.php
index a43e8bbe8..fbfa0dbc7 100644
--- a/webapp/src/Controller/Jury/ConfigController.php
+++ b/webapp/src/Controller/Jury/ConfigController.php
@@ -126,6 +126,7 @@ class ConfigController extends AbstractController
         $results = $this->checkConfigService->runAll();
         return $this->render('jury/config_check.html.twig', [
             'results' => $results,
+            'gitCommit' => $this->dj->getGitCommit(),
             'dir' => [
                     'project' => dirname($projectDir),
                     'log' => $logsDir,
diff --git a/webapp/src/Service/DOMJudgeService.php b/webapp/src/Service/DOMJudgeService.php
index d6b37c984..bab499383 100644
--- a/webapp/src/Service/DOMJudgeService.php
+++ b/webapp/src/Service/DOMJudgeService.php
@@ -63,6 +63,7 @@ class DOMJudgeService
     protected HttpKernelInterface $httpKernel;
     protected ConfigurationService $config;
     protected RouterInterface $router;
+    protected bool $debug;
     protected ?Executable $defaultCompareExecutable = null;
     protected ?Executable $defaultRunExecutable = null;
 
@@ -90,7 +91,8 @@ class DOMJudgeService
         TokenStorageInterface $tokenStorage,
         HttpKernelInterface $httpKernel,
         ConfigurationService $config,
-        RouterInterface $router
+        RouterInterface $router,
+        bool $debug
     ) {
         $this->em                   = $em;
         $this->logger               = $logger;
@@ -101,6 +103,7 @@ class DOMJudgeService
         $this->httpKernel           = $httpKernel;
         $this->config               = $config;
         $this->router               = $router;
+        $this->debug                = $debug;
     }
 
     public function getEntityManager(): EntityManagerInterface
@@ -1312,4 +1315,17 @@ class DOMJudgeService
 
         return $res;
     }
+
+    public function getGitCommit(): string
+    {
+        $gitCommit = GIT_COMMIT;
+        if ($this->debug) {
+            // In debug, check if we have a .git directory and get it from there
+            $webappParentDir = dirname($this->params->get('domjudge.webappdir'));
+            if (is_dir($webappParentDir) && is_dir($webappParentDir . '/.git')) {
+                $gitCommit = exec(sprintf('cd %s && git rev-parse --short HEAD', escapeshellarg($webappParentDir)));
+            }
+        }
+        return $gitCommit;
+    }
 }
diff --git a/webapp/templates/jury/config_check.html.twig b/webapp/templates/jury/config_check.html.twig
index 4c018b900..68a57cf3a 100644
--- a/webapp/templates/jury/config_check.html.twig
+++ b/webapp/templates/jury/config_check.html.twig
@@ -9,7 +9,7 @@
   <h5 class="card-title">System information</h5>
   <table class="w-100">
   <tr><td>DOMjudge:</td><td>{{ DOMJUDGE_VERSION }}</td></tr>
-  <tr><td>Git Commit:</td><td>{{ GIT_COMMIT }}</td></tr>
+  <tr><td>Git Commit:</td><td>{{ gitCommit }}</td></tr>
   <tr><td>Environment:</td><td>{{ app.environment }}</td></tr>
   <tr><td>Debug:</td><td>{{ app.debug|printYesNo }}</td></tr>
   <tr><td colspan="2">&nbsp;</td></tr>
```

thijskh:
@nickygerritsen We then update it on the config checker/API, but not on the other places where it is shown?

nickygerritsen:
> @nickygerritsen We then update it on the config checker/API, but not on the other places where it is shown?

Fair point. Any idea to make this better?

thijskh:
One solution is to display it in less places... not sure how useful it is to put it in the HTML header. Or how useful it is nowadays to put the version into the HTML header at all.

nickygerritsen:
🤷‍♂️ . I'm also fine with not doing my patch but for 'developers' that don't run the makefile stuff every time it is then out of date.

eldering:
I don't really like having a separate "git commit" variable and then it can be "Unknown" in various common cases like if you built from a release or snapshot tarball. So I would propose to just add the git commit to the version string in case of building from a DEV version on the main branch. So we'd have something like
`DOMJUDGE_VERSION=8.0.0DEV/a0b1c2d3`. I think that provides everything you want, since for a release, the git commit is irrelevant and given by the git tag of the release, and for maintainer mode, I think one should be able to figure out what you're using. Also, in that case a commit might not even be available in the main repo, nor might it accurately reflect code with uncommitted changes.

Dup4:
> I don't really like having a separate "git commit" variable and then it can be "Unknown" in various common cases like if you built from a release or snapshot tarball. So I would propose to just add the git commit to the version string in case of building from a DEV version on the main branch. So we'd have something like `DOMJUDGE_VERSION=8.0.0DEV/a0b1c2d3`. I think that provides everything you want, since for a release, the git commit is irrelevant and given by the git tag of the release, and for maintainer mode, I think one should be able to figure out what you're using. Also, in that case a commit might not even be available in the main repo, nor might it accurately reflect code with uncommitted changes.

Fair point. 

- For release builds we don't need git commit.

Maybe we don't need to add a new `GIT_COMMIT` variable, we just need the following patch?

```patch
diff --git a/configure.ac b/configure.ac
index 9f232b1..3a30145 100644
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
 # Process this file with autoconf to produce a configure script.
 AC_INIT([DOMjudge],
-        m4_esyscmd([grep 'version' README.md | sed -n '1s/^.*version //p' | tr -d '\n']),
+        m4_esyscmd([grep 'version' README.md | sed -n '1s/^.*version //p' | sed "s/$/$(git rev-parse --short HEAD | sed 's|^|\\/|' || echo '' )/" | tr -d '\n']),
         [domjudge-devel@domjudge.org])
 AC_PREREQ([2.64])
 AC_CONFIG_HEADERS([etc/config.h])
```

Preview:

Has `.git` directory:

image

`.git` directory does not exist:

image

nickygerritsen:
I like this. @thijskh what do you think?

thijskh:
Yes, this seems ok. Still it means that you need to regenerate `configure` and then run it after a git pull. But I'm not sure there's a better solution...

I think it's not very neat that you repeatedly will get `sh: 1: git: not found` when running configure on a

Dup4:
> Yes, this seems ok. Still it means that you need to regenerate `configure` and then run it after a git pull. But I'm not sure there's a better solution...
> 
> I think it's not very neat that you repeatedly will get `sh: 1: git: not found` when running configure on a system without git.

Maybe we can add some checks.

like the following patch:

```diff
diff --git a/configure.ac b/configure.ac
index 9f232b1..899a556 100644
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
 # Process this file with autoconf to produce a configure script.
 AC_INIT([DOMjudge],
-        m4_esyscmd([grep 'version' README.md | sed -n '1s/^.*version //p' | tr -d '\n']),
+        m4_esyscmd([grep 'version' README.md | sed -n '1s/^.*version //p' | sed "s/$/$(command -v git >/dev/null 2>&1 && git rev-parse --is-inside-work-tree >/dev/null 2>&1 && git rev-parse --short HEAD | sed 's|^|\\/|' || echo '' )/" | tr -d '\n']),
         [domjudge-devel@domjudge.org])
 AC_PREREQ([2.64])
 AC_CONFIG_HEADERS([etc/config.h])
```

vmcj:
> > Yes, this seems ok. Still it means that you need to regenerate `configure` and then run it after a git pull. But I'm not sure there's a better solution...
> > I think it's not very neat that you repeatedly will get `sh: 1: git: not found` when running configure on a system without git.
> 
> Maybe we can add some checks.
> 
> like the following patch:
> 
> ```diff
> diff --git a/configure.ac b/configure.ac
> index 9f232b1..899a556 100644
> --- a/configure.ac
> +++ b/configure.ac
> @@ -1,6 +1,6 @@
>  # Process this file with autoconf to produce a configure script.
>  AC_INIT([DOMjudge],
> -        m4_esyscmd([grep 'version' README.md | sed -n '1s/^.*version //p' | tr -d '\n']),
> +        m4_esyscmd([grep 'version' README.md | sed -n '1s/^.*version //p' | sed "s/$/$(command -v git >/dev/null 2>&1 && git rev-parse --is-inside-work-tree >/dev/null 2>&1 && git rev-parse --short HEAD | sed 's|^|\\/|' || echo '' )/" | tr -d '\n']),
>          [domjudge-devel@domjudge.org])
>  AC_PREREQ([2.64])
>  AC_CONFIG_HEADERS([etc/config.h])
> ```

Hi @Dup4, will you implement your fix so we can see if we can test it at our next contest?

Dup4:
> > > Yes, this seems ok. Still it means that you need to regenerate `configure` and then run it after a git pull. But I'm not sure there's a better solution...
> > > I think it's not very neat that you repeatedly will get `sh: 1: git: not found` when running configure on a system without git.
> > 
> > 
> > Maybe we can add some checks.
> > like the following patch:
> > ```diff
> > diff --git a/configure.ac b/configure.ac
> > index 9f232b1..899a556 100644
> > --- a/configure.ac
> > +++ b/configure.ac
> > @@ -1,6 +1,6 @@
> >  # Process this file with autoconf to produce a configure script.
> >  AC_INIT([DOMjudge],
> > -        m4_esyscmd([grep 'version' README.md | sed -n '1s/^.*version //p' | tr -d '\n']),
> > +        m4_esyscmd([grep 'version' README.md | sed -n '1s/^.*version //p' | sed "s/$/$(command -v git >/dev/null 2>&1 && git rev-parse --is-inside-work-tree >/dev/null 2>&1 && git rev-parse --short HEAD | sed 's|^|\\/|' || echo '' )/" | tr -d '\n']),
> >          [domjudge-devel@domjudge.org])
> >  AC_PREREQ([2.64])
> >  AC_CONFIG_HEADERS([etc/config.h])
> > ```
> 
> Hi @Dup4, will you implement your fix so we can see if we can test it at our next contest?

Done.

And I simply tested it.

run `make configure` with `.git`:

image

run `make configure` without `.git`:

image

vmcj:
It now has a merge commit, can you get rid of that?

eldering (approved):
Indeed, please remove the merge commit, but other than that looks fine to me.

Dup4:
> Indeed, please remove the merge commit, but other than that looks fine to me.

Sorry, I fixed it.

vmcj:
I've merged it, thanks for your contribution.

PR #1478, opened 2022-03-02
Register endpoint does not exist when no categories are available

vmcj:
According to https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.4
HTTP404 is cachable so I chose the less standard 409 as the problem is
on the server side.


thijskh:
I'm not sure 400 was bad per se since the URL is not linked to in this case so the client requests something which is not advertised as being available. 409 is quite esoteric but also suggests the user might be able to do something about it which is not the case here. If you want to avoid 400, I think 403 is also good match and customary for functionality that is disabled.

thijskh:
The commit message is off now

vmcj:
> The commit message is off now

You are right, but I don't think this is worth force pushing to fix it.

PR #1479, opened 2022-03-02
Try to be more explicit when wrong file type is imported.

thijskh:
This is to help users who upload the wrong thing in the wrong box.

* For YAML (and JSON which is YAML): catch a parse error and make explict that we're parsing YAML/JSON here in the error message.

* For TSV: parse error already handled (checks file header), make message explict that we're trying TSV's here to avoid misunderstanding.

* For Problem package: would stumble on for any zip file that is uploaded and create a half-baked problem. ZIP als file format is already filtered in the UI. The spec is not 100% clear on what the minimum is to know if something is an actual problem package. I assumed that it needs at least one of `domjudge-problem.ini`,`problem.yaml` to be present. If it's not, error out early and explicitly.


vmcj (commented):
Looks good to me, but in the future we might want to look at also the encoding of the textfiles. I think that has also failed for people in the past.

nickygerritsen (approved):

Issue #1289, opened 2021-10-19
No scoreboard for balloon runners (non-public)


iMylene:
For a non-public contest, the scoreboard does not show up in the header for at least balloon runners.

nickygerritsen:
OK so this raises the question: should balloon runners have access to the jury scoreboard?

mvr320:
> OK so this raises the question: should balloon runners have access to the jury scoreboard?

I think there should be 2 roles:
_intrapublic_scoreboard
_jury_scoreboard

and the balloon runner should get all semi-public info which is intrapublic, so not the unfrozen standing during the contest but more information than the outside world gets, similar to teams.

nickygerritsen:
Wouldn’t the solution then not just be to make a hidden team for the balloon user?

nickygerritsen:
Yeah good question. We could maybe make /jury/scoreboard return the 'public' scoreboard for balloon users. What do others think?

nickygerritsen:
For the roles: I don't like adding all these roles...

nickygerritsen:
> > For the roles: I don't like adding all these roles...
> 
> I meant an internal role not one in the GUI, so instead of check for ROLE_JURY we check for which scoreboard can actually be accessed, similar to how the ROLE_ADMIN also has ROLE_JURY.

Ah so you actually want the same idea I have? Allow both ballon and jury to view the scoreboard but when you are not jury, only show the 'public' one?

nickygerritsen:
That is not what I'm suggesting. What I want to do is in this method: http://github.com/DOMjudge/domjudge/blob/main/webapp/src/Controller/Jury/ScoreboardController.php#L42-L42
set the `$jury` param only to true if you have the `ROLE_JURY`, otherwise set it to false. Then in that same controller change the security to also have balloon in it. That should be all to make balloon users see the 'public' (what you call intrapublic...) scoreboard.

eldering:
> That is not what I'm suggesting. What I want to do is in this method: [http://github.com/DOMjudge/domjudge/blob/main/webapp/src/Controller/Jury/ScoreboardController.php#L42-L42](https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Controller/Jury/ScoreboardController.php#L42-L42) set the `$jury` param only to true if you have the `ROLE_JURY`, otherwise set it to false. Then in that same controller change the security to also have balloon in it. That should be all to make balloon users see the 'public' (what you call intrapublic...) scoreboard.

[...]

PR #1481, opened 2022-03-05
Balloon nonpublic scoreboard

vmcj:
Based on the suggestions of @nickygerritsen in https://github.com/DOMjudge/domjudge/issues/1289.

This does not close that issue as discussion there created more functionality questions.

This discloses a bit more information to balloon runners as in the past they knew nothing in default setting during a frozen scoreboard about the amount of submissions which are now disclosed as pending.

### DIFF
--- a/webapp/src/DataFixtures/Test/DemoPreDeactivateContestFixture.php
+++ b/webapp/src/DataFixtures/Test/DemoPreDeactivateContestFixture.php
@@ -1,0 +1,32 @@
+<?php declare(strict_types=1);
+
+namespace App\DataFixtures\Test;
+
+use App\Entity\Contest;
+use Doctrine\Persistence\ObjectManager;
+
+class DemoPreDeactivateContestFixture extends AbstractTestDataFixture
+{
+    public function load(ObjectManager $manager): void
+    {
+        $demoContest = $manager->getRepository(Contest::class)->findOneBy(['shortname' => 'demo']);
+        $demoContest->setEndtimeString(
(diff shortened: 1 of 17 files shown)


nickygerritsen (approved):

nickygerritsen:
I think this is a good solution for now to see at least something useful

vmcj:
I'll try to make some extra tests before merging.

vmcj:
@nickygerritsen can you look at the tests if you agree?

webapp/src/DataFixtures/Test/DemoPreDeactivateContestFixture.php
```diff
@@ -0,0 +1,32 @@
...
+
+class DemoPreDeactivateContestFixture extends AbstractTestDataFixture
+{
+    public function load(ObjectManager $manager): void
+    {
+        $demoContest = $manager->getRepository(Contest::class)->findOneBy(['shortname' => 'demo']);
+        $demoContest->setEndtimeString(
+            sprintf(
+                '%s-06-06 09:00:00 Europe/Amsterdam',
+                date('Y') -1
```
nickygerritsen, line 16: Nit: missing space

nickygerritsen (approved):

webapp/src/DataFixtures/Test/DemoPreDeactivateContestFixture.php
vmcj: Fixed.

Issue #872, opened 2020-10-02
Make it possible to see team members from the web interface


ykazakov:
### Description of the problem

Currently it is not possible to see from the Jury interface, which users belong to a particular team.
There appears to be the field Members in the form, but is always empty, unless some text is entered there manually.

> Replace this line with a short description.

### Your environment
- DOMjudge 7.3.0

### Steps to reproduce
- install Domserver and initialise the database according to the [manual](https://www.domjudge.org/docs/manual/7.3/index.html)
- login as admin with the initial password
- under the Users link (Before contest) observe that the dummy user is a member of "Example tamname" (press the edit symbol to open the form)

![Screenshot 2020-10-02 at 15 27 32](https://user-images.githubusercontent.com/2140361/94928629-46089200-04c4-11eb-9130-6d0d485074b8.png)

- under the Teams link (Before contest) observe that the "Example tamname" does not have any member:

![Screenshot 2020-10-02 at 15 29 43](https://user-images.githubusercontent.com/2140361/94928641-486aec00-04c4-11eb-86c9-4d945062b58b.png)

### Expected behaviour
I think that team members should be shown in this form, and it should be possible to add / remove users to / from teams.

### Actual behaviour
Team members are not visible.

### Any other information that you want to share?

In database, there appears to be two columns of the team table: "users" which stores the list of users having this team, and "members", which as far as I see is not used anywhere in the code.

I think it should not be difficult to display / add / remove users from the team form, similar to Contests field in the bottom of the form:

![Screenshot 2020-10-02 at 15 34 55](https://user-images.githubusercontent.com/2140361/94929034-d3e47d00-04c4-11eb-8c24-3c08561d31da.png)

Maybe the only question is when adding users to teams, should all users be shown or only users that do not belong a team. I think the latter is a better option.

nickygerritsen:
In a lot of setups there is only one user per team but multiple people accessing that one user.

That is the reason we have this free form members field.

However I'd say it would be useful to have the users belonging to a team **besides** the free form members.

Selecting users when editing a team can also be a feature.

ykazakov:
@nickygerritsen is it really a good practice to use one account per team? Maybe that should be discouraged especially if an option for self-registration with an existing team can be provided? See #871 

To avoid confusion with members/users, should the free form field be renamed to something neutral, like "Description"?

nickygerritsen:
During ICPC contests where people are on site the rules are simple:
[...]

PR #1482, opened 2022-03-11
Clarify team* freeform fields in jury interface

vmcj:
This is the first of the 2 items mentioned by Nicky in https://github.com/DOMjudge/domjudge/issues/872

Mainly to clarify the difference in intended usage between the 2 fields.


webapp/config/load_db_secrets.php
```diff
@@ -34,7 +34,7 @@ function get_db_url(): string
         break;
     }
 
-    return sprintf('mysql://%s:%s@%s:%d/%s?serverVersion=5.7.0', $user, $pass, $host, $port ?? 3306, $db);
+    return sprintf('mysql://%s:%s@%s:%d/%s?serverVersion=mariadb-10.7.3', $user, $pass, $host, $port ?? 3306, $db);
```
nickygerritsen, line 37: Do we want this?

webapp/templates/jury/team.html.twig
```diff
@@ -150,17 +150,17 @@
...
-                <tr>
-                    <th>Comments</th>
-                    <td>
-                        {{ team.comments | nl2br }}
-                    </td>
-                </tr>
+                    <tr>
+                        <th>Comments</th>
+                        <td>
+                            {{ team.comments | nl2br }}
```
nickygerritsen, line 156:
```suggestion
                            {{ team.internalComments | nl2br }}
```

Right?

nickygerritsen (commented):
The name of this PR is not really correct, it has nothing to do with members (anymore)

webapp/config/load_db_secrets.php
vmcj: It was for me to check if that would make it easier to generate doctrine:migrations:diff.

But I still needed to manually fix the migrations so the commit message already reminds me to later remove it.

webapp/templates/jury/team.html.twig
vmcj: Yes

webapp/config/load_db_secrets.php
nickygerritsen: I have a similar thing on my Linux machine but think I didn't have it on my mac with Docker (even after adding stuff to `.env.local`). We might want to investigate later.

nickygerritsen (approved):

PR #1492, opened 2022-03-13
Add option to delete existing testcases/attachments/other data when importing a problem. Fixes #1485.

### LINKED ISSUE #1485: Problem re-import does not overwrite/delete old testcases
When a problem is re-imported (either from the admin UI, or from the API) the old testcases which are not present in the new zip are not deleted. Also, if a testcase is updated the old version is not removed.

### Your environment
Latest main revision, started using `docker-compose up` from this repository.

### Steps to reproduce
Attached to this issue there is [tests.zip](https://github.com/DOMjudge/domjudge/files/8237658/tests.zip) which contains two versions of the same task:

- `first/test.zip` has 3 testcases: 1, 2, 3
- `second/test.zip` has 2 testcases: 1 (same as before), 2 (modified)

Import the first zip, and check that 3 testcases have been imported.

> `curl 'http://localhost:12345/api/v4/contests/2/problems' -H 'Authorization: Basic <redacted>' -H 'accept: application/json' -X POST -F zip=@tests/first/test.zip | jq`

Then, import the second task overwriting the first:

> `curl 'http://localhost:12345/api/v4/contests/2/problems' -H 'Authorization: Basic <redacted>' -H 'accept: application/json' -X POST -F zip=@tests/second/test.zip -F problem=4 | jq`

This is what it outputs to me:
```json
{
  "problem_id": "4",
  "messages": [
    "Added problem statement from: problem.pdf",
    "Skipped sample testcase 1: already exists",
    "Added 1 secret testcase(s): {2}.{in,ans}",
    "Added/updated 0 attachment(s).",
    "No jury solutions added: must associate team with your user first.",
    "Saved problem 4"
  ]
}
```

image

### Expected behaviour
I expect that when a problem is reimported, the new version is exactly the version I've just uploaded. This is especially important when testing a task.

### Actual behaviour
A mix of old data and new data is present after the import, which doesn't make much sense from an API for importing a task.

Reported to me by @dario2994

nickygerritsen:


webapp/src/Form/Type/ProblemUploadType.php
```diff
@@ -28,6 +29,10 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
                 'accept' => 'application/zip',
             ],
         ]);
+        $builder->add('delete_data_first', CheckboxType::class, [
+            'help' => 'If checked, all existing testcases and attachment for the problem will be deleted, if any. Only useful for existing problems.',
```
vmcj, line 33:
```suggestion
            'help' => 'If checked, all existing testcases and attachments for the problem will be deleted, if any. Only useful for existing problems.',
```

webapp/src/Controller/API/ProblemController.php
```diff
@@ -213,6 +218,7 @@ public function addProblemAction(Request $request): array
         set_time_limit(120);
 
         $probId = $request->request->get('problem');
+        $deleteDataFirst = $request->request->getBoolean('delete_data_first');
```
vmcj, line 221: I think we should have a test for this

webapp/src/Controller/API/ProblemController.php
```diff
@@ -180,6 +180,11 @@ public function listAction(Request $request): Response
      *                 property="problem",
      *                 description="Optional: problem id to update.",
      *                 type="string"
+     *             ),
+     *             @OA\Property(
+     *                 property="delete_data_first",
```
meisterT, line 185: nit: I would call this `delete_old_data`

webapp/src/Form/Type/ProblemUploadType.php
```diff
@@ -28,6 +29,10 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
                 'accept' => 'application/zip',
             ],
         ]);
+        $builder->add('delete_data_first', CheckboxType::class, [
+            'help' => 'If checked, all existing testcases and attachments for the problem will be deleted, if any. Only useful for existing problems.',
```
meisterT, line 33: what about resetting other things like validators/validator args, and limits?

nickygerritsen: Aren’t they always overwritten? If not this makes sense to do and I will do ir. Any more we want to do?

meisterT: I don't think so. You should give this a try. And perhaps add a test :-D

nickygerritsen: 🤣 we should really do the tests

webapp/src/Form/Type/ProblemUploadType.php
```diff
@@ -28,6 +29,10 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
                 'accept' => 'application/zip',
             ],
         ]);
+        $builder->add('delete_data_first', CheckboxType::class, [
+            'help' => 'If checked, all old data for the problem will be deleted, if any. Only useful for existing problems.',
```
vmcj, line 33:
```suggestion
            'help' => 'If checked, any old data for the problem will be deleted. Only useful for existing problems.',
```

```suggestion
            'help' => 'Only useful for existing problems. If checked, any old data for the problem will be deleted.',
```
Second is just to know how it looks, it makes reading slightly easier as people directly know that 9/10 times this is not relevant for their usecase.

nickygerritsen: I'm not sure if I like this

vmcj: Both are only suggestions, feel free to ignore.

meisterT: What about `If checked, old data will be deleted.`?

The second half "Only useful for existing problems" can be omitted if we only show the box for existing problems, right?

webapp/src/Service/ImportProblemService.php
```diff
@@ -190,6 +191,62 @@ public function importZippedProblem(
             }
         }
 
+        if ($deleteOldData && $problem->getProbid()) {
+            if (abs($problem->getTimelimit() - $defaultTimelimit) > 0.001) {
```
meisterT, line 195: Is there really no better way to check for an already set timelimit?

webapp/src/Service/ImportProblemService.php
```diff
@@ -190,6 +191,62 @@ public function importZippedProblem(
...
+            }
+            if ($problem->getMemlimit()) {
+                $problem->setMemlimit(null);
+                $messages[] = 'Clearing memory limit.';
+            }
+            if ($problem->getOutputlimit()) {
+                $problem->setOutputlimit(null);
+                $messages[] = 'Clearing output limit.';
+            }
+            if ($problem->getProblemtext()) {
```
meisterT, line 223: .... or problemtexttype is null?

webapp/src/Service/ImportProblemService.php
nickygerritsen: Reply to the comment above but I cant:
> The second half "Only useful for existing problems" can be omitted if we only show the box for existing problems, right?

But we show it on the general import screen, where we don't know if it updates a problem, since we find an existing one by basename of the ZIP. #1496 makes that a bit awkward anyway, so maybe we should drop that feature?

Reply to:
> Is there really no better way to check for an already set timelimit?

Like how? I can't think of any.

webapp/src/Service/ImportProblemService.php
meisterT: Yep, not null. I mean it should never happen, but do you want to rely here on the contract between the two?

nickygerritsen: Doesn't hurt to add it I guess

webapp/src/Service/ImportProblemService.php
vmcj: Why does `$problem->getTimelimit() !== $defaultTimelimit` not work?

nickygerritsen: Rounding might make this 2.0000000001 vs 2

webapp/src/Controller/API/ProblemController.php
```diff
@@ -180,6 +180,11 @@ public function listAction(Request $request): Response
      *                 property="problem",
      *                 description="Optional: problem id to update.",
      *                 type="string"
+     *             ),
+     *             @OA\Property(
+     *                 property="delete_old_data",
```
thijskh, line 185: `data` (and `old`) (also in other places in this PR, is a bit vague, maybe something like 'clear existing testcases'?

thijskh (approved):

webapp/src/Controller/API/ProblemController.php
nickygerritsen: I don't like 'testcases' since it also deletes attachments and other stuff like limits and problem texts.

thijskh: Alright, let's not fess over it too much then

Issue #1495, opened 2022-03-21
Import testing


vmcj:
Currently we don't test all API endpoints for imports but we should.

This issue is to track the progress.

meisterT:
https://github.com/DOMjudge/domjudge/pull/1690 improved the test situation a little bit, since we do run `import-contest` in the integration test now.

PR #1502, opened 2022-03-23
Allow to set external ID and shortname for problem during import. Fixes #1496.

### LINKED ISSUE #1496: Add support for setting problem.externalid and contestproblem.shortname to domjudge-problem.ini
Currently by default we use the basename of the ZIP for both when importing, unless the problem already exists.
It makes sense to me to allow a user to overwrite both if they want that. Also see the discussion on Slack with @dario2994

nickygerritsen:
Only thing I am wondering is whether we should call the property in the .ini `shortname` as well, making the code cleaner.
However, this then differs from `ImportExportService::getContestYamlData` and `ImportExportService::importProblemsData`, both of which we could also change if we want. Opinions?

### DIFF
--- a/webapp/src/Service/ImportProblemService.php
+++ b/webapp/src/Service/ImportProblemService.php
@@ -83,8 +83,8 @@ class ImportProblemService
         $submission_file = 'submissions.json';
         $problemIsNew    = $problem === null;
 
-        $iniKeysProblem        = ['name', 'timelimit', 'special_run', 'special_compare'];
-        $iniKeysContestProblem = ['allow_submit', 'allow_judge', 'points', 'color'];
+        $iniKeysProblem        = ['name', 'timelimit', 'special_run', 'special_compare', 'externalid'];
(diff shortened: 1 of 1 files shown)


nickygerritsen:
Yes we need tests. Will be done in #1495.

meisterT (approved):

meisterT:
I approved but then started thinking again. What happens if you specify this but import it to a specific problem like we do at WF?

nickygerritsen:
That is a good question. I think it will update the external ID. But then again, it already did: if you uploaded `tobi.zip` to problem `nicky`, it would update the external ID.

Shall I change in this PR for that to be an error? Or do we want to NOT change external ID's for existing problems and only display a warning message? Either should be simple.

nickygerritsen:
> I approved but then started thinking again. What happens if you specify this but import it to a specific problem like we do at WF?

This is now fixed in #1492 so I will merge this one.

PR #1503, opened 2022-03-23
Add API endpoint to import problem without a contest. Fixes #1498.

### LINKED ISSUE #1498: Add an API endpoint to import a problem without a contest
Currently you can only import a problem into a contest, but a problem can also exist outside of any contest.
We should add `POST /api/problems` to create a new problem (or update an existing one) without any contest associated to it.

nickygerritsen:


nickygerritsen:
Yes we need tests. Will be done in #1495.

webapp/src/Service/ImportProblemService.php
```diff
@@ -793,4 +796,67 @@ public function importZippedProblem(
...
+
+        $contest = null;
+        if ($contestId) {
+            /** @var Contest $contest */
+            $contest = $this->em->getRepository(Contest::class)->find($contestId);
+        }
+        $allMessages = [];
+
+        // Only timeout after 2 minutes, since importing may take a while.
+        set_time_limit(120);
```
vmcj, line 815: Do we really have problems which need more than 60 seconds?

vmcj (commented):
Looks fine to me.

webapp/src/Service/ImportProblemService.php
nickygerritsen: For sure

meisterT (approved):

PR #1504, opened 2022-03-23
Use custom CSS files

vmcj:
Although a simple fix, I'm not sure if this is actually the correct solution.

### DIFF
--- a/webapp/src/Service/DOMJudgeService.php
+++ b/webapp/src/Service/DOMJudgeService.php
@@ -1228,7 +1228,7 @@ class DOMJudgeService
 
         $results = [];
         foreach (scandir($customDir) as $file) {
-            foreach (static::MIMETYPE_TO_EXTENSION as $extension) {
+            foreach (array_merge(['css','js'], static::MIMETYPE_TO_EXTENSION) as $extension) {
                 if (strpos($file, '.' . $extension) !== false) {
                     $results[] = $file;
                 }


webapp/src/Service/DOMJudgeService.php
```diff
@@ -1228,13 +1228,15 @@ public function getAssetFiles(string $path): array
 
         $results = [];
         foreach (scandir($customDir) as $file) {
-            foreach (static::MIMETYPE_TO_EXTENSION as $extension) {
+            foreach (array_merge(['css'],static::MIMETYPE_TO_EXTENSION) as $extension) {
```
nickygerritsen, line 1231: We also need .js in that case

webapp/src/Service/DOMJudgeService.php
```diff
@@ -1228,13 +1228,15 @@ public function getAssetFiles(string $path): array
...
                 if (strpos($file, '.' . $extension) !== false) {
                     $results[] = $file;
                 }
             }
         }
 
+        //dump($results);
```
nickygerritsen, line 1238: Dumpiedump

webapp/src/Service/DOMJudgeService.php
nickygerritsen: Also missing space after ,

vmcj: But where did this stop working.... and why?

nickygerritsen: I think when I rewrote everything to not use the symfony DI container detection

PR #1505, opened 2022-03-23
Add API endpoints to link and unlink problems to/from contests. Fixes #1497.

### LINKED ISSUE #1497: Add API endpoints to link and unlink problems to/from a contest
A problem can exist in multiple contests. We should add two API endpoints:
* `DELETE api/contests/<id>/problems/<id>`: delete the problem from the contest
* `POST api/contests/<id>/problems/<id>`: add the problem to the contest. Body should contain the fields needed in ContestProblem.

nickygerritsen:
Woohooo with tests :tada: .

### DIFF
--- a/webapp/src/Controller/API/ProblemController.php
+++ b/webapp/src/Controller/API/ProblemController.php
@@ -208,6 +208,137 @@ class ProblemController extends AbstractRestController implements QueryObjectTra
     }
 
     /**
+     * Unlink a problem from this contest.
+     * @Rest\Delete("/{id}")
+     * @IsGranted("ROLE_ADMIN")
+     * @OA\Response(response="204", description="Problem unlinked from contest succeeded")
+     * @OA\Parameter(ref="#/components/parameters/id")
+     */
+    public function unlinkProblemAction(Request $request, string $id): Response
+    {
+        $problem = $this->em->createQueryBuilder()
+                            ->from(Problem::class, 'p')
+                            ->select('p')
+                            ->andWhere(sprintf('%s = :id', $this->getIdField()))
+                            ->setParameter('id', $id)
+                            ->getQuery()
+                            ->getOneOrNullResult();
+
+        if (empty($problem)) {
+            throw new NotFoundHttpException(sprintf('Object with ID \'%s\' not found', $id));
+        }
+
+        $cid = $this->getContestId($request);
+
+        /** @var ContestProblem|null $contestProblem */
+        $contestProblem = $this->em->createQueryBuilder()
+                                   ->from(ContestProblem::class, 'cp')
+                                   ->select('cp')
+                                   ->andWhere('cp.contest = :contest')
+                                   ->andWhere('cp.problem = :problem')
+                                   ->setParameter('contest', $cid)
+                                   ->setParameter('problem', $problem)
+                                   ->getQuery()
+                                   ->getOneOrNullResult();
+
+        if (empty($contestProblem)) {
+            throw new NotFoundHttpException(sprintf('Object with ID \'%s\' not found', $id));
+        }
+
+        $this->em->remove($contestProblem);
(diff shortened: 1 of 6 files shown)


nickygerritsen:
Note: if we would add support for `allow_submit` this would break: we don't return the problem in the list/single action if it is not submittable, so the add action will return a 404. So I decided to not add that or `allow_judge` for now.

webapp/src/Controller/API/ProblemController.php
```diff
@@ -258,6 +258,137 @@ public function addProblemAction(Request $request): array
...
+
+        if (empty($problem)) {
+            throw new NotFoundHttpException(sprintf('Object with ID \'%s\' not found', $id));
+        }
+
+        $cid = $this->getContestId($request);
+
+        /** @var ContestProblem|null $contestProblem */
+        $contestProblem = $this->em->createQueryBuilder()
+                            ->from(ContestProblem::class, 'cp')
```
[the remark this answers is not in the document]
nickygerritsen: Yup, done

webapp/src/Controller/API/ProblemController.php
```diff
@@ -258,6 +258,137 @@ public function addProblemAction(Request $request): array
         ];
     }
 
+    /**
+     * Delete a problem from this contest.
```
meisterT, line 262: I wonder if we should call this unlink instead of delete?

nickygerritsen: Sure, and call the other one link 😄

nickygerritsen: Done.

webapp/src/Controller/API/ProblemController.php
```diff
@@ -258,6 +258,137 @@ public function addProblemAction(Request $request): array
...
     }
 
+    /**
+     * Unlink a problem from this contest.
+     * @Rest\Delete("/{id}")
+     * @IsGranted("ROLE_ADMIN")
+     * @OA\Response(response="204", description="Deleting problem succeeded")
+     * @OA\Parameter(ref="#/components/parameters/id")
+     */
+    public function deleteProblemAction(Request $request, string $id): Response
```
thijskh, line 268: Maybe the contest context is not obvious to everyone. So it might be helpful to name it `deleteContestProblem`?

webapp/tests/Unit/Controller/API/ProblemControllerTest.php
```diff
@@ -47,4 +49,20 @@ class ProblemControllerTest extends BaseTest
     ];
 
     protected array $expectedAbsent = ['4242', 'nonexistent'];
+
+    public function testDeleteNotAllowed(): void
+    {
+        if ($this->apiUser) {
```
thijskh, line 55: Probably more robust to check if the apiUser is not admin?

thijskh (approved):

webapp/src/Controller/API/ProblemController.php
nickygerritsen: But should it then also be listInContestAction, singleInContestAction, etc? With that I mean: we have many more actions without problem in the name.

webapp/tests/Unit/Controller/API/ProblemControllerTest.php
nickygerritsen: Done, also added a 401 test for add

webapp/src/Controller/API/ProblemController.php
thijskh: Perhaps we should? I would personally also get nervous from a message like "Deleting problem succeeded" while it was only unlinked. Maybe that's just me

nickygerritsen: Of course that message is not what you get back, thats the description of the API doc, where it already says "Unlink a problem from this contest.":
image

I do see we say 'the ID of the entity to get' there btw, which is not right.

PR #1506, opened 2022-03-24
Custom CSS can be used on the login page

vmcj:


webapp/templates/base.html.twig
```diff
@@ -16,14 +16,13 @@
     {% for file in customAssetFiles('js') %}
         <script src="{{ asset('js/custom/' ~ file) }}"></script>
     {% endfor %}
+    {% for file in customAssetFiles('css') %}
+        <link rel="stylesheet" href="{{ asset('css/custom/' ~ file) }}">
+    {% endfor %}
```
nickygerritsen, line 21: Should we move this to BELOW the extrahead block? Now things in style_domjudge.css will overwrite the custom CSS instead of the other way around

vmcj: That does make more sense, also because that's what is mentioned in the documentation.

nickygerritsen (approved):

Issue #1353, opened 2021-11-21
Make judge remaining possible for a whole problem/contest


meisterT:
It would be nice to be able to do after the contest (or in a slow period) to judge all remaining test cases of incorrect submissions.

We already can do that on single submissions, but there is no way to do it for the whole contest or a problem.

Suggested by @jsannemo

vmcj:
When we would run this for the problem it should be for all judgings with the problem or with the contestproblem? Or is there a usecase for both so should there be 2 buttons?

nickygerritsen:
With all judgings you mean also for the ones of rejudgings that are not active anymore? I'm not sure if we want that (immediately)? Maybe first focus on the 'all current for contest/problem'.

vmcj:
> With all judgings you mean also for the ones of rejudgings that are not active anymore? I'm not sure if we want that (immediately)? Maybe first focus on the 'all current for contest/problem'.

That was not what I meant but I'll specify a bit further:
Current contest is 1, and I go to the /jury/contests and press the "judge remaining" (new button)
In scope should be (IMO) all judgings which are currently valid (or in other words, the current verdict is based on this judging)

On that same page we also list the contest problems, which should also get a button, in scope should be the same judgings but now only if they are for this problem.

If I would visit /jury/problems (so the general problem page) and problem 3 is in contest 1 and 2, should a judge remaining have all judgings of this problem (so all judgings where problem==3 && (contest==1 or contest==2)
or only the judgings for this problem IN the current contest (so the same as above)

My proposal is to have 2 buttons, namely 1 for only current contest (99/100 times this is what you want) and 1 for all judgings of this problem regardless in which contest this happened (which might even be in a deactivated contest, thats something to think about later for now I would say everything which is activated)

nickygerritsen:
> If I would visit /jury/problems (so the general problem page) and problem 3 is in contest 1 and 2, should a judge remaining have all judgings of this problem (so all judgings where problem==3 && (contest==1 or contest==2) or only the judgings for this problem IN the current contest (so the same as above)
> 
> My proposal is to have 2 buttons, namely 1 for only current contest (99/100 times this is what you want) and 1 for all judgings of this problem regardless in which contest this happened (which might even be in a deactivated contest, thats something to think about later for now I would say everything which is activated)

I'd say if you have a current contest selected, only the ones in the contest. If you have no selected, do all. We can change the label of the button based on that.

PR #1507, opened 2022-03-25
Feat contest judge remaining

vmcj:
This is the simple stuff for https://github.com/DOMjudge/domjudge/issues/1353.

I think its quite easy to allow this also for TeamCategories, (where I now ignore hidden categories)
languages (maybe to verify that a certain fix indeed works) etc. but if we don't have an usecase for it yet it would just clutter the UI.

In a next PR I'll address the actual Problem if this is the accepted solution.

### DIFF
--- a/webapp/src/Controller/BaseController.php
+++ b/webapp/src/Controller/BaseController.php
@@ -4,11 +4,14 @@ namespace App\Controller;
 
 use App\Entity\BaseApiEntity;
 use App\Entity\Contest;
+use App\Entity\JudgeTask;
+use App\Entity\Judging;
 use App\Entity\Problem;
 use App\Entity\RankCache;
 use App\Entity\ScoreCache;
 use App\Entity\Team;
 use App\Entity\TeamCategory;
+use App\Entity\QueueTask;
 use App\Service\DOMJudgeService;
 use App\Service\EventLogService;
 use App\Utils\Utils;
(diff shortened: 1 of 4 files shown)


webapp/src/Controller/Jury/ContestController.php
```diff
@@ -855,4 +855,65 @@
             'form' => $form->createView(),
         ]);
     }
+
+    /**
+     * @Route("/{contestId<\d+>}/request-remaining", name="jury_contest_request_remaining")
+     */
+    public function requestRemainingRunsWholeContest(int $contestId): RedirectResponse
```
nickygerritsen, line 862: I think we end most actions with `action`

webapp/src/Controller/Jury/ContestController.php
```diff
@@ -855,4 +855,65 @@
...
+    {
+        /** @var Contest $contest */
+        $contest = $this->em->getRepository(Contest::class)->find($contestId);
+        if (!$contest) {
+            throw new NotFoundHttpException(sprintf('Contest with ID %s not found', $contestId));
+        }
+        $judgings = [];
+        foreach ($contest->getProblems() as $problem) { // This should get the ContestProblem
+            foreach ($problem->getSubmissions() as $submission) { // So this has only submissions in this contest
```
nickygerritsen, line 871: These loops will result in 1 (contest) + N (problems) + M (submissions) + T (teams) + J (judgings) queries. Maybe we should, at line 865, use a query builder to fetch these rows in one query?

webapp/src/Controller/Jury/ContestController.php
```diff
@@ -855,4 +855,65 @@
...
+                if (!$submission->getTeam()->getCategory()->getVisible()) { // Ignore teams not on the scoreboard
+                    continue;
+                }
+                foreach ($submission->getJudgings() as $judging) { // So these judgings are in scope
+                    if ($judging === null) {
```
nickygerritsen, line 876: When does this happen? And do we also want non active judgings (i.e. old ones for rejudgings)? Since that is what you are doing now right?

webapp/src/Controller/Jury/ContestController.php
```diff
@@ -855,4 +855,65 @@
...
+            }
+        }
+        $this->judgeRemaining($judgings);
+        return $this->redirect($this->generateUrl('jury_contest', ['contestId' => $contestId]));
+    }
+
+    /**
+     * @Route("/{contestId<\d+>}/problems/{probId<\d+>}/request-remaining", name="jury_contest_problem_request_remaining")
+     */
+    public function requestRemainingRunsContestProblem(int $contestId, int $probId): RedirectResponse
```
nickygerritsen, line 890: `Action` again

webapp/src/Controller/Jury/ContestController.php
```diff
@@ -855,4 +855,65 @@
...
+        ]);
+        if (!$contestProblem) {
+            throw new NotFoundHttpException(
+                sprintf('Contest problem with contest ID %s and problem ID %s not found',
+                        $contestId, $probId)
+            );
+        }
+
+        $judgings = [];
+        foreach ($contestProblem->getSubmissions() as $submission) {
```
nickygerritsen, line 905: Same remark here about huge amount of queries.

webapp/src/Controller/BaseController.php
```diff
@@ -488,4 +491,68 @@
...
+        foreach ($judgings as $judging) {
+            $judgingId = $judging->getJudgingid();
+            if ($judging->getResult() === null) {
+                $inProgress[] = $judgingId;
+            } elseif ($judging->getJudgeCompletely()) {
+                $alreadyRequested[] = $judgingId;
+            } elseif (!$judging->getValid()) {
+                $invalidJudgings[] = $judgingId;
+            }
+            else {
```
nickygerritsen, line 510: This should go after the `}` on the same line

webapp/src/Controller/BaseController.php
```diff
@@ -488,4 +491,68 @@
...
+        if (count($judgings) === 1) {
+            if ($inProgress !== []) {
+                $this->addFlash('warning', 'Please be patient, this judging is still in progress.');
+            }
+            if ($alreadyRequested !== []) {
+                $this->addFlash('warning', 'This judging was already requested to be judged completely.');
+            }
+        } else {
+            if ($inProgress !== []) {
+                $this->addFlash('warning', sprintf('Please be patient, these judgings are still in progress: %s',implode(',', $inProgress)));
```
nickygerritsen, line 543: Nit: space after , before implode

webapp/templates/jury/contest.html.twig
```diff
@@ -356,6 +357,13 @@
                                     </a>
                                 {% endif %}
                             </td>
+                            <td>
+                                {% if is_granted('ROLE_ADMIN') %}
+                                    <a href="{{ path('jury_contest_problem_request_remaining', {'contestId': contest.cid, 'probId': problem.probid}) }}">
```
nickygerritsen, line 362: We should probably add a `title="Judge remaining` here (and similar with the delete button above) to have a hover text.

vmcj: Why does the W3||WCAG test not flag this. I'll fix it.

webapp/src/Controller/Jury/ContestController.php
vmcj: It should never happen, I'll replace it with the check for valid judgings.

webapp/src/Controller/BaseController.php
```diff
@@ -488,4 +491,67 @@ protected function streamResponse(callable $callback): StreamedResponse
...
+                $submission = $judging->getSubmission();
+
+                $queueTask = new QueueTask();
+                $queueTask->setJobId($judging->getJudgingid())
+                    ->setPriority(JudgeTask::PRIORITY_LOW)
+                    ->setTeam($submission->getTeam())
+                    ->setTeamPriority((int)$submission->getSubmittime())
+                    ->setStartTime(null);
+                $this->em->persist($queueTask);
+                $this->em->flush();
```
nickygerritsen, line 530: We could do the flush after the loop. I think it’s faster to only flush once

nickygerritsen (approved):

webapp/src/Controller/BaseController.php
vmcj: There were to flushed, both are now removed but I'm not sure if the 1st should actually be removed? And if we don't remove the first the amount of flushes is only halved (+1).

nickygerritsen: One flush at the end should be enough I guess

PR #1511, opened 2022-03-26
Results.html fixes

nickygerritsen:


webapp/src/Controller/Jury/ImportExportController.php
```diff
@@ -436,6 +436,13 @@ protected function getResultsHtml(Request $request, bool $useIcpcLayout): Respon
 
         $collator = new Collator('en_US');
         $collator->sort($honorable);
```
vmcj, line 438: What does this do if we afterwards sort it again?

nickygerritsen: Do we? This is honorable and below is ranked.

vmcj: Ok, and rank is decided by time instead of name.

vmcj (approved):

PR #1512, opened 2022-03-26
CCS Specs API information in root endpoint

nickygerritsen:
The API spec wants us to expose these two fields in the root, see [here](https://ccs-specs.icpc.io/master/contest_api#api-version).

I put the existing DOMjudge specific fields in a `domjudge` object since IMHO otherwise the root keys are a bit strange.

### DIFF
--- a/webapp/src/Controller/API/GeneralInfoController.php
+++ b/webapp/src/Controller/API/GeneralInfoController.php
@@ -39,8 +39,8 @@ class GeneralInfoController extends AbstractFOSRestController
 {
     protected const API_VERSION = 4;
 
-    public const CCS_SPEC_API_VERSION = '2021-11';
-    public const CCS_SPEC_API_URL = 'https://ccs-specs.icpc.io/2021-11/contest_api';
+    public const CCS_SPEC_API_VERSION = 'draft';
+    public const CCS_SPEC_API_URL = 'https://ccs-specs.icpc.io/master/contest_api';
 
     protected EntityManagerInterface $em;
     protected DOMJudgeService $dj;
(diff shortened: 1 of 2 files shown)


meisterT:
This change LGTM in general. I am not sure if there are clients out there which check DOMjudge version etc right now. If so, this would be a breaking change then.

nickygerritsen:
I am aware it would be a breaking change. I expect more of them in the 'near' future w.r.t. the CLICS API (at least in the event feed, of which the format will change). So we might want to think of how to handle that. We could put this change (as well as the later event feed changes) behind some config flag to have a way to get the old behaviour back, but I'm not sure if I like that.

webapp/src/Controller/API/GeneralInfoController.php
```diff
@@ -91,20 +94,33 @@ public function getVersionAction(): array
      *     description="Information about the API and DOMjudge",
      *     @OA\JsonContent(
      *         type="object",
-     *         @OA\Property(property="api_version", type="integer"),
-     *         @OA\Property(property="domjudge_version", type="string"),
-     *         @OA\Property(property="environment", type="string"),
-     *         @OA\Property(property="doc_url", type="string")
+     *         @OA\Property(property="version", type="string", description="Version of the CCS Specs Contest API the API adheres to"),
+     *         @OA\Property(property="version_url", type="string", description="URL with the speification of the Contest API"),
```
vmcj, line 98: Why did we not catch a typo here?
```suggestion
     *         @OA\Property(property="version_url", type="string", description="URL with the specification of the Contest API"),
```

PR #1513, opened 2022-03-26
Implement account endpoints

nickygerritsen:
This implements the account endpoints as specified [here](https://ccs-specs.icpc.io/master/contest_api#accounts) based on our existing user endpoints.

As the comment in the AccountController says, it's a bit of a copy paste from the UserController. I don't really see how we could fix that except when we move the user calls to the account controller as well and give them the "Accounts" tag instead of "Users". Some controller actions will then have two API endpoints (for example /users and /api/contests/<cid>/accounts). What do you guys think?

### DIFF
--- a/webapp/src/Controller/API/AccountController.php
+++ b/webapp/src/Controller/API/AccountController.php
@@ -119,6 +119,6 @@ class AccountController extends AbstractRestController
 
     protected function getIdField(): string
     {
-        return 'u.userid';
+        return sprintf('u.%s', $this->eventLogService->externalIdFieldForEntity(User::class) ?? 'userid');
     }
 }
--- a/webapp/src/Controller/API/AccountController.php
+++ b/webapp/src/Controller/API/AccountController.php
@@ -1,0 +1,124 @@
+<?php declare(strict_types=1);
+
+namespace App\Controller\API;
+
+use App\Entity\User;
+use Doctrine\ORM\NonUniqueResultException;
+use Doctrine\ORM\QueryBuilder;
+use FOS\RestBundle\Controller\Annotations as Rest;
+use Nelmio\ApiDocBundle\Annotation\Model;
+use OpenApi\Annotations as OA;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+/**
+ * @Rest\Route("/contests/{cid}")
+ * @OA\Tag(name="Accounts")
+ * @OA\Parameter(ref="#/components/parameters/cid")
+ * @OA\Response(response="404", ref="#/components/responses/NotFound")
+ * @OA\Response(response="401", ref="#/components/responses/Unauthorized")
+ * @OA\Response(response="400", ref="#/components/responses/InvalidResponse")
+ */
+class AccountController extends AbstractRestController
+{
+    // Note: this controller is basically a copy of the UserController but then with account endpoints.
+    // There are two differences that make it impossible to add it to that controller or extend it:
+    // - The /api/contests/<cid> prefix to load the contest
+    // - The fact that we have /api/contests/<cid>/account
+    // Also it seems to not be possible to overwrite the OA\Tag or the base controller route when
+    // extending a controller
+
+    /**
+     * Get all the accounts
+     * @Rest\Get("/accounts")
+     * @Security("is_granted('ROLE_ADMIN') or is_granted('ROLE_API_READER')")
+     * @OA\Response(
+     *     response="200",
+     *     description="Returns all the accounts for this contest",
+     *     @OA\JsonContent(
+     *         type="array",
+     *         @OA\Items(ref=@Model(type=User::class))
+     *     )
+     * )
+     * @OA\Parameter(ref="#/components/parameters/idlist")
+     * @OA\Parameter(ref="#/components/parameters/strict")
+     * @OA\Parameter(
+     *     name="team_id",
+     *     in="query",
+     *     description="Only show accounts for the given team",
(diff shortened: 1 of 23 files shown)


nickygerritsen:
Ah and I did not expose the `password` field for obvious reasons (we can't and we shouldn't). That field in the spec is mainly there for when we use this same endpoint to import accounts.

eldering:
This currently doesn't do anything special when the contest is private and only has a limited set of teams associated with it. In that case we might want to filter out accounts that are associated only with teams that are not part of the contest. But then we should also not return such an account in `GET contests/<id>/account`.

nickygerritsen:
Ah yes that both sounds like a good idea. For the latter: if Austria is not admin or jury, return 404?

nickygerritsen:
> This currently doesn't do anything special when the contest is private and only has a limited set of teams associated with it. In that case we might want to filter out accounts that are associated only with teams that are not part of the contest. But then we should also not return such an account in `GET contests/<id>/account`.

This is now both fixed. Well the latter because only admins can currently request account information at all. When we add other user types to have access, we should think about what to return.

nickygerritsen:
Updated this PR with support for JSON/YAML import, accounts in the event feed and validating accounts during event feed import.

nickygerritsen:
CI fails because accounts is not in the JSON spec code from @eldering yet

webapp/src/Controller/API/AccountController.php
```diff
@@ -0,0 +1,124 @@
...
+     * @Security("is_granted('ROLE_ADMIN') or is_granted('ROLE_API_READER')")
+     * @OA\Response(
+     *     response="200",
+     *     description="Returns the given account",
+     *     @Model(type=User::class)
+     * )
+     * @OA\Parameter(ref="#/components/parameters/id")
+     * @OA\Parameter(ref="#/components/parameters/strict")
+     */
+    public function singleAction(Request $request, string $id): Response
```
vmcj, line 75: Where is the strict used?

nickygerritsen: Automatically just like other calls: it filters all non strict fields

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -871,6 +875,45 @@ protected function validateTeam(array $event): void
...
+            ->getRepository(User::class)
+            ->findOneBy(['externalid' => $userId]);
+
+        if ($event['op'] === EventLogService::ACTION_DELETE) {
+            // We need to check if the user is known
+            if ($user) {
+                $this->addOrUpdateWarning($event, ExternalSourceWarning::TYPE_ENTITY_SHOULD_NOT_EXIST);
+            }
+
+            $this->removeWarning($event, ExternalSourceWarning::TYPE_ENTITY_SHOULD_NOT_EXIST);
```
vmcj, line 893: Why is the warning directly removed?

webapp/src/Service/ImportExportService.php
```diff
@@ -543,10 +545,14 @@ public function importJson(string $type, UploadedFile $file, ?string &$message =
...
-        catch (\JsonException $e) {
-            $message = "File contents is not valid JSON: " . $e->getMessage();
-            return -1;
+        } catch (\JsonException $e) {
+            // Check if we can parse it as YAML
+            try {
+                $data = Yaml::parse($content, Yaml::PARSE_DATETIME);
+            } catch (ParseException $e) {
+                $message = "File contents is not valid JSON or YAML: " . $e->getMessage();
+                return -1;
```
vmcj, line 554: We only need to return numeric exit codes in Commands right?

webapp/src/Service/ExternalContestSourceService.php
nickygerritsen: I don’t follow.  We remove the warning that the user should not exist if it is stil there but the user is deleted externally since we are in sync again. But it misses a return two lines above. Which is copy pasted so it misses somewhere else as well

webapp/src/Service/ImportExportService.php
```diff
@@ -813,6 +821,73 @@ public function importTeamsJson(array $data, ?string &$message = null, ?array &$
...
+                        'externalid'        => $account['externalid'] ?? $account['username'],
+                        'category'          => $juryCategory,
+                        'publicdescription' => $account['name'] ?? $account['username'],
+                    ];
+                    break;
+                case 'team':
+                    $roles[] = $teamRole;
+                    break;
+                case 'analyst':
+                case 'staff':
```
vmcj, line 867: I think staff should be clarification reader or balloon runner?

webapp/src/Service/ImportExportService.php
```diff
@@ -813,6 +821,73 @@ public function importTeamsJson(array $data, ?string &$message = null, ?array &$
...
+                    // Ignore type analyst and staff for now. We don't have a useful mapping yet.
+                    continue 2;
```
vmcj, line 869: Should there be a warning that those were not imported?

webapp/src/Service/ImportExportService.php
nickygerritsen: This is based on other imports here. Id like to be consistent so let’s not change it now

webapp/src/Service/ImportExportService.php
nickygerritsen: We didn’t do anything with staff before so let’s keep it that way for now. I think no one ever uses it but if they do we can improve

webapp/src/Service/ImportExportService.php
nickygerritsen: I don’t think so. We don’t do it for the tsv either. We also don’t really have a way to report it currently

webapp/src/Service/ExternalContestSourceService.php
vmcj: That explains why you directly remove the Warning from 890.

PR #1515, opened 2022-03-28
Allow judge remaining for problems, languages and categories

vmcj:
IMO this closes https://github.com/DOMjudge/domjudge/issues/1353.

Reason for the Team categories is that we default only look at judgings for visible categories and otherwise we can never get the jury solutions (apart from 1 by 1). Language is because it could be used as a soft rejudge where you want to know if a certain change leads to errors (for example)


nickygerritsen (approved):
I'm not sure about the usefulness of all these but then again they dont hurt.

PR #1527, opened 2022-04-15
Update dependencies before SWERC

vmcj:


nickygerritsen (approved):
Most are only minor updates, but still we should test this properly BEFORE SWERC.
If CI succeeds and someone does some minor local testing this should be fine?

vmcj:
Alternative is only deploying this to the shadow system.

nickygerritsen:
We can do that but I expect us to make more changes, which then becomes harder to deploy (we need a branch and keep cherry picking)? Plus shadow clearly doesn't test everything.

PR #1528, opened 2022-04-16
Add an option to connect to mysql via socket (instead of port/TCP).

meisterT:


meisterT:
@eldering I have no idea how to do this is in a more elegant way that is still POSIX compliant. Can you help?

eldering:
> @eldering I have no idea how to do this is in a more elegant way that is still POSIX compliant. Can you help?

I think you can use alternate value expansion here, something like:
```sh
[ -z "$USE_SOCKET ] && port="-P$DBPORT"
command mysql $user ${pass:+"$pass"} -h "$DBHOST" ${port:+"$port"} --silent --skip-column-names "$@"
```
A bit clunky, but should work (as long as the variables are undefined otherwise, not just empty. See https://newbedev.com/index.php/how-can-i-expand-a-quoted-variable-to-nothing-if-it-s-empty

nickygerritsen (approved):

eldering (approved):

PR #1531, opened 2022-04-16
Remove unnecessary cast to `string`.

meisterT:
`getApiId` returns a `string` (and is type checked).


nickygerritsen (approved):
Makes sense

Issue #1526, opened 2022-04-12
Possibility of removing/adding new file to executable


baziorek:
I'm using DOMjudge for academic teaching. I'm creating new executable for each problem, because I'm checking submissions with Unit Tests in C++.
Unfortunately it is really hard to make its many times, because:

1. sometimes I add file by accident - the only way to remove the file is to attach entire executable again, then correct language to use the new script
2. sometimes I forgot to add some file - the only way as above
3. sometimes in files are national characters e.g. "ąśćżźć" and the text file is detected as binary and I can't edit

Possibility of adding/removing file would solve all the problems above.

I'm using DOMjudge: 7.3.4

baziorek:
Hi,
I see that it was done in commits:
1. https://github.com/DOMjudge/domjudge/pull/1539
2. https://github.com/DOMjudge/domjudge/pull/1535
3. https://github.com/DOMjudge/domjudge/pull/1536

Thanks really much for making easier to edit executables!

What should I do now - it looks that the issue is implemented, so should I close its?

meisterT:
I'm not fully done yet improving and simplifying the interface

meisterT:
Also what I just noticed when running another contest: we drop binary files (such as executables) when editing.

eldering:
See also #1841

meisterT:
This is how the executables overview looks like now:
![image](https://github.com/DOMjudge/domjudge/assets/418721/04694433-bde6-4825-a401-06dc3381b2e2)

Possible improvements:
- put the id in `code` tags
- separate out used from unused executables
- either use some fontawesome icon or similar to visually distinguish the executable type
- perhaps up for debate: for the currently selected contest, add problem badges where appropriate

nickygerritsen:
I like all these. Maybe even a table per type?

vmcj:
> I like all these. Maybe even a table per type?

If we do the table, would we still want the icon?

PR #1539, opened 2022-04-18
Allow deletion of individual files in executable.

meisterT:
Part of #1526.

### DIFF
--- a/webapp/templates/jury/executable_edit_content.html.twig
+++ b/webapp/templates/jury/executable_edit_content.html.twig
@@ -33,6 +33,12 @@
                        href="{{ path('jury_executable_download_single', {execId: executable.execid, rank: ranks[idx]}) }}">
                         <i class="fas fa-download"></i> Download
                     </a>
+                    {% if is_granted('ROLE_ADMIN') %}
+                        <a class="btn btn-secondary btn-sm"
+                           href="{{ path('jury_executable_delete_single', {execId: executable.execid, rank: ranks[idx]}) }}">
+                            <i class="fas fa-trash"></i> Delete
+                        </a>
+                    {% endif %}
                 </div>
 
                 {{ files[idx] | codeEditor(idx, null, is_granted('ROLE_ADMIN'), 'form_source' ~ idx, aceFilenames[idx]) }}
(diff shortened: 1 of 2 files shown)


webapp/templates/jury/executable_edit_content.html.twig
```diff
@@ -33,6 +33,12 @@
                        href="{{ path('jury_executable_download_single', {execId: executable.execid, rank: ranks[idx]}) }}">
                         <i class="fas fa-download"></i> Download
                     </a>
+                    {% if is_granted('ROLE_ADMIN') %}
+                        <a class="btn btn-secondary btn-sm"
```
nickygerritsen, line 37: No confirmation?

meisterT: Good call, what's the default way to accomplish that shiny modal dialog?

nickygerritsen: * Add `data-ajax-modal` the the link (so `<a href="..." data-ajax-modal ...>` to make it use AJAX.
* Make the delete action respond to both GET and POST. Get will render a form with 'are you sure' or something and POST will do the actual delete.

The form can be simple, just a submit button to the same action. Note that it does need some logic to switch between ajax and not, like [here](https://github.com/DOMjudge/domjudge/blob/37ff2f0b1fa8061e6f910de1041841573cb57c1d/webapp/src/Controller/BaseController.php#L407-L411) because you can still enter the delete URL in your browser yourself.

If you are stuck let me know and I can push to your branch.

meisterT: I added the modal dialog and it sort of works now. Cancel cancels, delete deletes, but .... it somehow redirects to URLs like `http://localhost/domjudge/jury/executables/awk/undefined` and I have no clue why.
One additional data point: if I remove `data-ajax-modal` I am redirected to the right page (but of course a modal dialog is nicer).

nickygerritsen: Redirect after deletion? That’s odd. Let me investigate

nickygerritsen: This is now fixed, see my commit.

nickygerritsen (approved):

Issue #540, opened 2019-04-06
Add 'super lazy' mode when running in shadow mode to only judge submissions upon request


nickygerritsen:
As requested by @SteinNorheim it would be great if we had a mode where we would not judge submissions automatically, but only when someone presses some button per submission.

We could do this by making the lazy judging mode a 'troolean' and modify the query to get the next judging to not return submissions we don't want.

This is related to #539 because this seems only useful when running in shadow mode.

meisterT:
With the new judgehost API this basically means to create judgetasks on demand, similar to when a problem is not allowed to judge.

vmcj:
So I have something in #1540 for this, but I assume that we also want to inverse this flag for the category the analysts belong to? Otherwise they have to press the button all the time to get their own submissions.

nickygerritsen:
Maybe only do it for external judgements? The analyst once aren't external.

vmcj:
I propose to close this as I think this is now implemented (#1540)

vmcj:
@SuprDewd do you think the feature is now useful for analysts during the WF?

PR #1540, opened 2022-04-19
Troolean judging

vmcj:
I think this would be a valid implementation for: https://github.com/DOMjudge/domjudge/issues/540

I'll merge this myself as I want to rewrite the history after approval

### DIFF
--- a/etc/db-config.yaml
+++ b/etc/db-config.yaml
@@ -98,10 +98,14 @@
             public: false
             description: Maximum size of run/diff/error/system output shown in the jury interface (in bytes); use `-1` to disable any limits.
         -   name: lazy_eval_results
-            type: bool
-            default_value: true
+            type: int
+            default_value: 1
             public: false
-            description: Lazy evaluation of results? If enabled, stops judging as soon as a highest priority result is found, otherwise always all testcases will be judged.
+            description: Lazy evaluation of results? If enabled, stops judging as soon as a highest priority result is found, otherwise always all testcases will be judged. On request will not auto-start judging and is typically used when running as analyst system.
+            options:
+                1: Lazy
+                2: Full judging
+                3: Only on request
         -   name: judgehost_warning
             type: int
             default_value: 30
--- a/webapp/migrations/Version20221004135409.php
+++ b/webapp/migrations/Version20221004135409.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20221004135409 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Allow super lazy evaluation where nothing is judged unless requested.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE contestproblem CHANGE lazy_eval_results lazy_eval_results INT UNSIGNED DEFAULT NULL COMMENT \'Whether to do lazy evaluation for this problem; if set this overrides the global configuration setting\'');
+        $this->addSql('UPDATE contestproblem SET lazy_eval_results=2 WHERE lazy_eval_results=0');
+        $this->addSql('UPDATE configuration SET value=2 WHERE value=0 AND name="lazy_eval_results"');
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('UPDATE configuration SET value=0 WHERE value=2 AND name="lazy_eval_results"');
+        $this->addSql('UPDATE contestproblem SET lazy_eval_results=0 WHERE lazy_eval_results=2');
+        $this->addSql('ALTER TABLE contestproblem CHANGE lazy_eval_results lazy_eval_results TINYINT(1) DEFAULT NULL COMMENT \'Whether to do lazy evaluation for this problem; if set this overrides the global configuration setting\'');
+    }
+
+    public function isTransactional(): bool
+    {
+        return false;
+    }
+}
--- a/webapp/src/Controller/Jury/ConfigController.php
+++ b/webapp/src/Controller/Jury/ConfigController.php
@@ -75,6 +75,9 @@ class ConfigController extends AbstractController
                         }
                     }
                     $data[substr($key, strlen('config_'))] = $valueToUse;
+                    if ($key === 'config_lazy_eval_results' && $value !== DOMJudgeService::EVAL_DEMAND) {
+                        $this->dj->unblockJudgeTasks();
+                    }
                 }
             }
             $this->config->saveChanges($data, $eventLogService, $this->dj);
--- a/webapp/src/Controller/Jury/SubmissionController.php
+++ b/webapp/src/Controller/Jury/SubmissionController.php
@@ -1129,4 +1129,13 @@ class SubmissionController extends BaseController
 
         return null;
     }
+
+    /**
+     * @Route("/{submitId<\d+>}/create-tasks", name="jury_submission_create_tasks")
+     */
+    public function createJudgeTasks(string $submitId): RedirectResponse {
+        $this->dj->unblockJudgeTasksForSubmission($submitId);
+        $this->addFlash('info', "Started judging for submission: $submitId");
+        return $this->redirectToRoute('jury_submission', ['submitId' => $submitId]);
+    }
 }
--- a/webapp/src/Entity/ContestProblem.php
+++ b/webapp/src/Entity/ContestProblem.php
@@ -82,13 +82,13 @@ class ContestProblem
     private ?string $color = null;
 
     /**
-     * @ORM\Column(type="boolean", name="lazy_eval_results",
+     * @ORM\Column(type="integer", name="lazy_eval_results",
      *     options={"comment"="Whether to do lazy evaluation for this problem; if set this overrides the global configuration setting",
      *              "unsigned"="true"},
      *     nullable=true)
      * @Serializer\Exclude()
      */
-    private ?bool $lazyEvalResults;
+    private ?int $lazyEvalResults = null;
 
     /**
      * @var Contest|int
@@ -189,13 +189,13 @@ class ContestProblem
         return $this->color;
     }
 
-    public function setLazyEvalResults(?bool $lazyEvalResults): ContestProblem
+    public function setLazyEvalResults(?int $lazyEvalResults): ContestProblem
     {
-        $this->lazyEvalResults = $lazyEvalResults;
+        $this->lazyEvalResults = $lazyEvalResults === 0 ? null : $lazyEvalResults;
         return $this;
     }
 
-    public function getLazyEvalResults(): ?bool
+    public function getLazyEvalResults(): ?int
     {
         return $this->lazyEvalResults;
     }
--- a/webapp/src/Form/Type/ContestProblemType.php
+++ b/webapp/src/Form/Type/ContestProblemType.php
@@ -4,6 +4,7 @@ namespace App\Form\Type;
 
 use App\Entity\ContestProblem;
 use App\Entity\Problem;
+use App\Service\DOMJudgeService as DJS;
 use Doctrine\ORM\EntityRepository;
 use Symfony\Bridge\Doctrine\Form\Type\EntityType;
 use Symfony\Component\Form\AbstractType;
@@ -53,8 +54,9 @@ class ContestProblemType extends AbstractType
             'label' => 'Lazy eval',
             'choices' => [
                 'Default' => null,
-                'Yes' => true,
-                'No' => false,
+                'Yes' => DJS::EVAL_LAZY,
+                'No' => DJS::EVAL_FULL,
+                'On demand' => DJS::EVAL_DEMAND,
             ],
         ]);
     }
--- a/webapp/src/Service/DOMJudgeService.php
+++ b/webapp/src/Service/DOMJudgeService.php
@@ -71,6 +71,10 @@ class DOMJudgeService
     const DATA_SOURCE_LOCAL = 0;
     const DATA_SOURCE_CONFIGURATION_EXTERNAL = 1;
     const DATA_SOURCE_CONFIGURATION_AND_LIVE_EXTERNAL = 2;
+    const EVAL_DEFAULT = null;
+    const EVAL_LAZY = 1;
+    const EVAL_FULL = 2;
+    const EVAL_DEMAND = 3;
 
     // Regex external identifiers must adhere to. Note that we are not checking whether it
     // does not start with a dot or dash or ends with a dot. We could but it would make the
@@ -961,7 +965,6 @@ class DOMJudgeService
             ->select('j')
             ->from(Judging::class, 'j')
             ->leftJoin(JudgeTask::class, 'jt', Join::WITH, 'j.judgingid = jt.jobid')
-            ->join(Submission::class, 's', Join::WITH, 'j.submission = s.submitid')
             ->where('jt.jobid IS NULL');
     }
 
@@ -969,6 +972,7 @@ class DOMJudgeService
     {
         // These are all the judgings that don't have associated judgetasks yet. Check whether we unblocked them.
         $judgings = $this->helperUnblockJudgeTasks()
+            ->join(Submission::class, 's', Join::WITH, 'j.submission = s.submitid')
             ->join(Language::class, 'l', Join::WITH, 's.language = l.langid')
             ->andWhere('l.langid = :langid')
             ->setParameter('langid', $langId)
@@ -983,6 +987,7 @@ class DOMJudgeService
     {
         // These are all the judgings that don't have associated judgetasks yet. Check whether we unblocked them.
         $judgings = $this->helperUnblockJudgeTasks()
+            ->join(Submission::class, 's', Join::WITH, 'j.submission = s.submitid')
             ->join(Problem::class, 'p', Join::WITH, 's.problem = p.probid')
             ->andWhere('p.probid = :probid')
             ->setParameter('probid', $probId)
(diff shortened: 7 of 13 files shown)


webapp/src/Controller/Jury/ConfigController.php
```diff
@@ -75,6 +75,12 @@ public function indexAction(EventLogService $eventLogService, Request $request):
                         }
                     }
                     $data[substr($key, strlen('config_'))] = $valueToUse;
+                    if (strpos($key, 'config_lazy_eval_results') === 0 && $value !== '2') {
+                        $this->dj->unblockJudgeTasks();
+                        var_dump("unlock");
```
vmcj, line 80: Will take this out.

webapp/src/Controller/Jury/ConfigController.php
```diff
@@ -75,6 +75,12 @@ public function indexAction(EventLogService $eventLogService, Request $request):
...
+                    } else {
+                        var_dump("lock");
```
vmcj, line 82: Same for here.

etc/db-config.yaml
```diff
@@ -98,10 +98,14 @@
             public: false
             description: Maximum size of run/diff/error/system output shown in the jury interface (in bytes); use "-1" to disable any limits.
         -   name: lazy_eval_results
-            type: bool
-            default_value: true
+            type: int
+            default_value: 0
             public: false
-            description: Lazy evaluation of results? If enabled, stops judging as soon as a highest priority result is found, otherwise always all testcases will be judged.
+            description: Lazy evaluation of results? If enabled, stops judging as soon as a highest priority result is found, otherwise always all testcases will be judged. On request will not auto-start judging and is typically used when running as analyst sytem.
```
nickygerritsen, line 104:
```suggestion
            description: Lazy evaluation of results? If enabled, stops judging as soon as a highest priority result is found, otherwise always all testcases will be judged. On request will not auto-start judging and is typically used when running as analyst system.
```

webapp/src/Controller/Jury/ConfigController.php
```diff
@@ -75,6 +75,12 @@ public function indexAction(EventLogService $eventLogService, Request $request):
                         }
                     }
                     $data[substr($key, strlen('config_'))] = $valueToUse;
+                    if (strpos($key, 'config_lazy_eval_results') === 0 && $value !== '2') {
```
nickygerritsen, line 78: Why can't we check the full key here? i.e. `if ($key === 'config_lazy_eval_results')`?

webapp/src/Controller/Jury/ConfigController.php
nickygerritsen, line 82: Do we want to lock something? Probably finishing the current set is fine?

webapp/src/Controller/Jury/ConfigController.php
```diff
@@ -111,6 +117,9 @@ public function indexAction(EventLogService $eventLogService, Request $request):
                 'data' => $data
             ];
         }
+        if ($allData[5]['data'][1]['value']!==2) {
+            array_splice($allData[1]['data'],12,1);
```
nickygerritsen, line 121: What...is...this...? 🤣

webapp/src/Service/DOMJudgeService.php
```diff
@@ -985,13 +985,45 @@ public function unblockJudgeTasksForProblem(int $probId): void
...
+        }
+    }
+
+    public function unblockJudgeTasks(): void
+    {
+        // These are all the judgings that don't have associated judgetasks yet. Check whether we unblocked them.
+        $judgings = $this->em->createQueryBuilder()
+            ->select('j')
+            ->from(Judging::class, 'j')
+            ->leftJoin(JudgeTask::class, 'jt', Join::WITH, 'j.judgingid = jt.jobid')
```
nickygerritsen, line 1011: This query is very similar to above? Maybe try to put it in one?

webapp/src/Controller/Jury/ConfigController.php
vmcj: This was debug for what happens if you change the lazy from 2 to something else. In this case we create all missing queue tasks but as I before started with an extra boolean which sometimes was not sent on post (so it can be easily hidden when you're not shadow to prevent people from selecting the `on demand` when they don't know what it does)

But all that is now removed so this was already wrong debugging stuff.

webapp/src/Controller/Jury/ConfigController.php
vmcj: This is a hard lock that we NEVER change the options... also it can be removed because of different implementation ;)

webapp/src/Service/DOMJudgeService.php
vmcj: I thought about this but as the original author hadn't merged those I wanted to discuss this in person what was the nicest way.

etc/db-config.yaml
```diff
@@ -98,10 +98,14 @@
             public: false
             description: Maximum size of run/diff/error/system output shown in the jury interface (in bytes); use "-1" to disable any limits.
         -   name: lazy_eval_results
-            type: bool
-            default_value: true
+            type: int
+            default_value: 0
```
meisterT, line 102: we need a DB migration for this

webapp/src/Controller/Jury/ConfigController.php
```diff
@@ -75,6 +75,9 @@ public function indexAction(EventLogService $eventLogService, Request $request):
                         }
                     }
                     $data[substr($key, strlen('config_'))] = $valueToUse;
+                    if ($key === 'config_lazy_eval_results' && $value !== '2') {
```
meisterT, line 78: please introduce an enum instead of magic string constants

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -1129,4 +1129,15 @@ protected function processClaim($judging, Request $request, ?string &$claimWarni
...
+
+    /**
+     * @Route("/{submitId<\d+>}/create-tasks", name="jury_submission_create_tasks")
+     */
+    public function createJudgeTasks(
+        string $submitId
+    ): RedirectResponse {
+        $this->dj->unblockJudgeTasksForSubmission($submitId);
+        $this->addFlash('info', "Started judging for submission: $submitId");
+        return $this->redirectToRoute('jury_submissions');
```
meisterT, line 1141: I would think this should redirect to the submission itself since you specifically requested it, right?

webapp/src/Service/DOMJudgeService.php
```diff
@@ -985,13 +985,45 @@ public function unblockJudgeTasksForProblem(int $probId): void
...
+            ->where('jt.jobid IS NULL')
+            ->getQuery()
+            ->getResult();
+        foreach ($judgings as $judging) {
+            $this->maybeCreateJudgeTasks($judging);
+        }
+    }
+
+    public function maybeCreateJudgeTasks(Judging $judging, int $priority = JudgeTask::PRIORITY_DEFAULT, bool $forceCreation = False): void
```
meisterT, line 1020: nit: `forceCreation` still respects (as it should) `allowJudge`, so perhaps rename to `manualRequest` or similar?

webapp/src/Service/DOMJudgeService.php
```diff
@@ -985,13 +985,45 @@ public function unblockJudgeTasksForProblem(int $probId): void
...
     {
         $submission = $judging->getSubmission();
         $problem    = $submission->getContestProblem();
         $language   = $submission->getLanguage();
 
-        if (!$problem->getAllowJudge() || !$language->getAllowJudge()) {
+        if ((!$problem->getAllowJudge() || !$language->getAllowJudge() || $this->config->get('lazy_eval_results') === 2) && !$forceCreation) {
```
meisterT, line 1026: use the enum here as well once created

meisterT: The `ContestProblem` also has the lazy property - so we probably want to update that entity as well.

meisterT (commented):
Since this touches the judging internals I think the best approach here is to check this into a branch and use that branch for the analyst system instead of using `main`. WDYT?

vmcj:
> Since this touches the judging internals I think the best approach here is to check this into a branch and use that branch for the analyst system instead of using `main`. WDYT?

Fine by me, that would remove the need for the quick merge also.

etc/db-config.yaml
nickygerritsen: Don’t we store bools as int already? If so we don’t

webapp/src/Service/DOMJudgeService.php
vmcj: So it can be that only some problems are judged on demand? I'm fine with implementing it but when would that be used?

etc/db-config.yaml
```diff
@@ -98,10 +98,14 @@
...
             public: false
-            description: Lazy evaluation of results? If enabled, stops judging as soon as a highest priority result is found, otherwise always all testcases will be judged.
+            description: Lazy evaluation of results? If enabled, stops judging as soon as a highest priority result is found, otherwise always all testcases will be judged. On request will not auto-start judging and is typically used when running as analyst system.
+            options:
+                0: Lazy
```
vmcj, line 106: I think we should rewrite those to 1,2,3 so the numbering works with whats in contestProblem as that one also has false/0 because of the "default"

webapp/templates/jury/contest.html.twig
```diff
@@ -356,7 +356,13 @@
                             <td>
                                 <a href="{{ link }}">
                                     {% if problem.lazyEvalResults is not null %}
-                                        {{ problem.lazyEvalResults | printYesNo }}
+                                        {% if problem.lazyEvalResults == 1 %}
```
vmcj, line 359: @nickygerritsen how can I use the constants here also?

nickygerritsen: `constant(‘FQDN'`) which is horrible

vmcj:
Assuming this gets approved I'll fixup some commits to keep history clean.

vmcj:
> Since this touches the judging internals I think the best approach here is to check this into a branch and use that branch for the analyst system instead of using `main`. WDYT?

I helped someone on 8.1 where not all queue items where generated and I think because of the error there it makes sense to have this as backupplan in main. When someone has missing queue items they can toggle between superlazy and normal lazy which will as side effect also create the missing queue items. So I propose to merge this to main.

webapp/src/Form/Type/ContestProblemType.php
```diff
@@ -4,6 +4,7 @@
 
 use App\Entity\ContestProblem;
 use App\Entity\Problem;
+use App\Service\DOMJudgeService as C;
```
nickygerritsen, line 7: I would prefer `DJ`

webapp/src/Form/Type/ContestProblemType.php
```diff
@@ -52,9 +53,10 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
         $builder->add('lazyEvalResults', ChoiceType::class, [
             'label' => 'Lazy eval',
             'choices' => [
-                'Default' => null,
-                'Yes' => true,
-                'No' => false,
+                'Default' => false,
```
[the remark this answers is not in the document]
nickygerritsen: or `C:: EVAL_DEFAULT`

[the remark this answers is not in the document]
vmcj: EVAL_DEFAULT = 0, and when we set no value it should be null. I already have another check which explicit sets 0 to null but I think I prefer to not depend on lines of code in another file. Will set it to null.

meisterT:
> Assuming this gets approved I'll fixup some commits to keep history clean.

Can you do this before merging? That makes it easier to review

webapp/migrations/Version20221004135409.php
```diff
@@ -0,0 +1,31 @@
...
+final class Version20221004135409 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Allow super lazy evaluation where nothing is judged unless requested.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE contestproblem CHANGE lazy_eval_results lazy_eval_results INT UNSIGNED DEFAULT NULL COMMENT \'Whether to do lazy evaluation for this problem; if set this overrides the global configuration setting\'');
```
meisterT, line 19: does this do the right thing on conversion? As in, I assume `true` gets changed to a `1`, and `false` to a `0`, right? It should probably be `2` and `1` instead?

webapp/src/Entity/ContestProblem.php
```diff
@@ -189,13 +189,13 @@ public function getColor(): ?string
         return $this->color;
     }
 
-    public function setLazyEvalResults(?bool $lazyEvalResults): ContestProblem
+    public function setLazyEvalResults(?int $lazyEvalResults): ContestProblem
     {
-        $this->lazyEvalResults = $lazyEvalResults;
+        $this->lazyEvalResults = $lazyEvalResults === 0 ? null : $lazyEvalResults;
```
meisterT, line 194: Why do we need to check for `=== 0` here?

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -1129,4 +1129,13 @@ protected function processClaim($judging, Request $request, ?string &$claimWarni
 
         return null;
     }
+
+    /**
+     * @Route("/{submitId<\d+>}/create-tasks", name="jury_submission_create_tasks")
+     */
+    public function createJudgeTasks(string $submitId): RedirectResponse {
```
meisterT, line 1136: This promises to return a `RedirectResponse` but doesn't do so.

webapp/src/Service/DOMJudgeService.php
```diff
@@ -993,13 +998,50 @@ public function unblockJudgeTasksForProblem(int $probId): void
...
+    public function maybeCreateJudgeTasks(Judging $judging, int $priority = JudgeTask::PRIORITY_DEFAULT, bool $manualRequest = False): void
     {
         $submission = $judging->getSubmission();
         $problem    = $submission->getContestProblem();
         $language   = $submission->getLanguage();
 
-        if (!$problem->getAllowJudge() || !$language->getAllowJudge()) {
+        $evalOnDemand = False;
+        // We have 2 cases, the problem picks the global value or the value is set.
+        if ( ((int)$problem->getLazyEvalResults() === 0 && $this->config->get('lazy_eval_results') === static::EVAL_DEMAND)
```
meisterT, line 1034: instead of 0 and 3 can we use human readable constants here?

webapp/src/Service/DOMJudgeService.php
```diff
@@ -993,13 +998,50 @@ public function unblockJudgeTasksForProblem(int $probId): void
...
+             || $problem->getLazyEvalResults() === 3) {
+            $evalOnDemand = True;
+        }
+        // Special case, we're shadow and someone submits on our side in that case
+        // we're not super lazy.
+        if ($this->config->get('data_source') === DOMJudgeService::DATA_SOURCE_CONFIGURATION_AND_LIVE_EXTERNAL
+            && $submission->getExternalid() === null) {
+                $evalOnDemand = False;
+        }
+        if ((!$problem->getAllowJudge() || !$language->getAllowJudge() || $evalOnDemand) && !$manualRequest) {
```
meisterT, line 1044: I think we should not judge this problem if the language or the problem are disabled.

Basically I would remove the `&& $!manualRequest` and in the line before do something like:
```
// Submission was requested explicitly to be judged.
if ($manualRequest) {
  $evalOnDemand = false;
}
```

(please check my logic)

webapp/migrations/Version20221004135409.php
vmcj: This migration indeed does not work anymore since I changed the constants. Will think about how to do this properly.

webapp/src/Entity/ContestProblem.php
vmcj: The POST sends a 0, but I wanted to store this as a null to have a clear difference between a default value and a value which is explicit set. I can change this to just be 0 now that everything is annotated with constants so it is easier to read in code if that is preferred.

webapp/src/Service/DOMJudgeService.php
vmcj: Yes, I thought I've done all the cases but apparently I had forgotten those.

webapp/src/Service/DOMJudgeService.php
vmcj: So I'm not 100% sure who's wrong here, but I'll just try to rewrite this to a form where it is easier to understand who is correct as those negations combined with `OR` is difficult to read.

webapp/src/Controller/Jury/SubmissionController.php
vmcj: I assume `return $this->redirectToRoute` would return that response, how did you find this mistake?

webapp/migrations/Version20221004135409.php
vmcj: Implemented the solution to convert the "false" to a 2 as the boolean yes (lazy is true) is already done correctly.

webapp/src/Service/DOMJudgeService.php
vmcj: Implemented

webapp/migrations/Version20221004135409.php
```diff
@@ -0,0 +1,33 @@
...
+        $this->addSql('UPDATE contestproblem SET lazy_eval_results=2 WHERE lazy_eval_results=0');
```
meisterT, line 20: do we have to update configuration as well if it exists?

vmcj: Probably yes, I'll check.

nickygerritsen (approved):

meisterT (approved):
Ready to be merged after some commit clean up

PR #1543, opened 2022-04-23
Add autologin and last used IP to users view.

### LINKED ISSUE #1542: Team ip displayed is the last used and this is confusing
On the jury teams view, the `ip` column shows the associated user's `last_ip_address`. This is confusing. Should we even show this in the teams view instead of the users view? We should also show the `ip_address` field and make it clearer what is displayed.

eldering:
The autologin IP is only shown when the ipaddress authentication method is enabled. Also display "IP" in uppercase.

Fixes #1542


nickygerritsen (approved):

thijskh (approved):
Yes, good to make it explicit

PR #1551, opened 2022-04-26
Fix main 260422

vmcj:
I'm not sure what is the correct string we should check for, @meisterT I assume you know.

Also when we change such strings we should grep if they exist in the tests I think.

### DIFF
--- a/webapp/tests/Unit/Controller/Jury/ConfigControllerTest.php
+++ b/webapp/tests/Unit/Controller/Jury/ConfigControllerTest.php
@@ -43,7 +43,7 @@ class ConfigControllerTest extends BaseTest
         self::assertSelectorExists('a.nav-link:contains("Scoring")');
 
         self::assertSelectorExists('label:contains("Penalty time:")');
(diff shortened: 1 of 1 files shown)


meisterT:
Your commit is not changing anything else than indentation, right?

-----

> Also when we change such strings we should grep if they exist in the tests I think.

I was trusting CI. Interesting enough the PR seems to be missing from https://gitlab.com/DOMjudge/domjudge/-/pipelines?page=1&scope=all

Not sure what I did wrong that CI was not triggered (and I misread the other checks that displayed as :heavy_check_mark: in GitHub as CI results).

meisterT:
The correct string is defined here: https://github.com/DOMjudge/domjudge/blob/main/etc/db-config.yaml#L19

Looks the same to me as in https://github.com/DOMjudge/domjudge/blame/main/webapp/tests/Unit/Controller/Jury/ConfigControllerTest.php#L46

vmcj:
Yes after rebasing etc. I got in an "interesting" state. I've now manually triggered the CI to see if main is in a correct state.

PR #1555, opened 2022-04-26
Use problem badges on balloon page. Fixes #1546.

### LINKED ISSUE #1546: Use problem badges instead of balloon character on balloons page
We should use the problem badges on the balloons page, like we do on the scoreboard. This prevents balloons from being difficult to spot, for example, when they are white on white background:
image

nickygerritsen:
With the sample data it's not too great (see first screenshot) but with 1 letter badges it's fine (see second screenshot).
image
image

### DIFF
--- a/webapp/src/Service/BalloonService.php
+++ b/webapp/src/Service/BalloonService.php
@@ -124,7 +124,7 @@ class BalloonService
                 continue;
             }
 
-            $TOTAL_BALLOONS[$balloonsData['teamid']][$balloonsData['probshortname']] = $balloonsData['color'];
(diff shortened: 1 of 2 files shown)


nickygerritsen:
> Should we than change the default problems to B,F,H?

I don't disagree, but this can be a different PR.
What do others think?

eldering:
> > Should we than change the default problems to B,F,H?
> 
> I don't disagree, but this can be a different PR. What do others think?

I think we should only do that if we want to enforce single-letter problem labels, and I'm not sure we'd want to do that.

PR #1556, opened 2022-04-27
Set external ID also on user during self registration.

### LINKED ISSUE #1544: Set external ID of affiliation (and team?) during self registration
Otherwise the scoreboard breaks.

nickygerritsen:
Also add test. Fixes #1544.

### DIFF
--- a/webapp/tests/Unit/Controller/PublicControllerTest.php
+++ b/webapp/tests/Unit/Controller/PublicControllerTest.php
@@ -6,6 +6,7 @@ use App\DataFixtures\Test\EnableSelfregisterFixture;
 use App\DataFixtures\Test\EnableSelfregisterSecondCategoryFixture;
 use App\DataFixtures\Test\SelfRegisteredUserFixture;
 use App\Entity\Contest;
+use App\Entity\User;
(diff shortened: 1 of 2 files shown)


webapp/tests/Unit/Controller/PublicControllerTest.php
```diff
@@ -95,6 +96,19 @@ public function testSelfRegister(array $inputs, string $password, array $fixture
...
+
+        /** @var User $user */
+        $user = static::getContainer()
+            ->get(EntityManagerInterface::class)
+            ->getRepository(User::class)
+            ->findOneBy(['username' => $inputs['username']]);
+
+        self::assertNotNull($user);
+        self::assertNotEmpty($user->getExternalid());
+        self::assertNotEmpty($user->getTeam()->getExternalid());
```
vmcj, line 108: Should the category also be checked?

nickygerritsen: The category is not created during register but by an admin. So that already has an external id

vmcj (approved):

PR #1569, opened 2022-05-05
Exclude base access denied exception from sentry.

nickygerritsen:


webapp/config/packages/sentry.yaml
```diff
@@ -1,2 +1,15 @@
...
+    options:
+        integrations:
+        - 'Sentry\Integration\IgnoreErrorsIntegration'
+
+services:
+    Sentry\Integration\IgnoreErrorsIntegration:
+        arguments:
+            $options:
+                ignore_exceptions:
+                    # Ignore this exeption. It is thrown before being redirected
```
vmcj, line 12:
```suggestion
                    # Ignore this exception. It is thrown before being redirected
```

vmcj (approved):

PR #1570, opened 2022-05-05
Bump flex to ^1.17

em411:
https://symfony.com/blog/the-old-flex-infrastructure-is-shutting-down


nickygerritsen (commented):
Shouldn’t this also change composer.lock? I will check and do it

nickygerritsen:
Ah the version itself was already 1.17 but still the lock needed a new hash. Committed in e44e84df8bc8f8dbb06c22a6dd253d12ad2f6701.

PR #1576, opened 2022-05-12
Fix yaml import whitespace

vmcj:


nickygerritsen (approved):

meisterT (approved):
We should really rewrite this in python

nickygerritsen:
> We should really rewrite this in python

Yes I'm planning to do this when I have time.

PR #1577, opened 2022-05-12
Add team admin api import

vmcj:
This is to trigger the CI, still need to check if it also solves my issue I had yesterday.

### DIFF
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -851,6 +851,7 @@ class ImportExportService
                 case 'admin':
                     $roles[] = $adminRole;
                     break;
+                case 'jury': // We don't break to let non existing role jury be interpret as role judge
                 case 'judge':
                     $roles[]  = $juryRole;
                     $roles[]  = $teamRole;
@@ -1154,6 +1155,7 @@ class ImportExportService
                 case 'admin':
                     $roles[] = $adminRole;
                     break;
+                case 'jury': // We don't break to let non existing role jury be interpret as role judge
                 case 'judge':
                     $roles[]  = $juryRole;
                     $roles[]  = $teamRole;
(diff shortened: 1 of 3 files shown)


webapp/src/Service/ImportExportService.php
```diff
@@ -850,9 +850,11 @@ public function importAccountsJson(array $data, ?string &$message = null, ?array
             switch ($account['type']) {
                 case 'admin':
                     $roles[] = $adminRole;
-                    break;
+                case 'jury':
                 case 'judge':
-                    $roles[]  = $juryRole;
+		    if ($account['type'] === 'admin') {
```
nickygerritsen, line 855: The formatting is a bit wrong. And maybe we should add a comment instead of the removed break as to why we do this?

vmcj: its also the wrong if statement.

webapp/src/Service/ImportExportService.php
```diff
@@ -849,10 +849,13 @@ public function importAccountsJson(array $data, ?string &$message = null, ?array
             $roles    = [];
             switch ($account['type']) {
                 case 'admin':
-                    $roles[] = $adminRole;
-                    break;
+			$roles[] = $adminRole;
```
eldering, line 852: Indentation seems off here.

webapp/src/Service/ImportExportService.php
```diff
@@ -849,10 +849,13 @@ public function importAccountsJson(array $data, ?string &$message = null, ?array
...
+			// Don't break so we can also add the jury features
+                case 'jury':
                 case 'judge':
-                    $roles[]  = $juryRole;
+		    if ($account['type'] === 'judge' || $account['type'] === 'jury') {
+		        $roles[]  = $juryRole;
```
eldering, line 857: Nitpick: double space before `=`.

nickygerritsen (approved):

Issue #1396, opened 2021-12-22
When starting parallel judging, it will report an error `deleting cgroup: libcgroup: Device or resource busy`


Dup4:
NOTE: If this is a _discussion starter_, you need any _installation help_ or have a _question_ on how to accomplish something,
rather post at [our discussion channel](https://www.domjudge.org/chat) or send an email to our
[DOMjudge-devel mailinglist](https://www.domjudge.org/mailman/listinfo/domjudge-devel)
instead of filing an issue here.

### Description of the problem
> Replace this line with a short description.

![image](https://user-images.githubusercontent.com/38343778/147025013-85547a12-1959-4de5-b3c1-87497593b158.png)

When I turn on parallel judgeing.

There will be many errors on judgehost.

![image](https://user-images.githubusercontent.com/38343778/147025113-d63283a5-0216-4a9f-8f35-07b7ff0eb27d.png)

But when I trun off parallel judging, these errors do **not occur**.

The following is about my deployment architecture:

![image](https://user-images.githubusercontent.com/38343778/147025372-b62d174b-11be-4d10-96f3-3c9b81516f3a.png)

![image](https://user-images.githubusercontent.com/38343778/147025412-88e3aa2b-648a-4972-b62e-0834417fa173.png)

I run 12 judgehosts on the same physical machine through docker.

```bash
#! /bin/bash

ID="${1}"
DAEMON_ID="${2}"
PASSWORD="${3}"

sudo docker run \
    -d \
    --restart=always \
    --privileged \
    -v /sys/fs/cgroup:/sys/fs/cgroup:ro \
    --name judgehost-"${ID}" \
    --hostname judgedaemon-"${ID}" \
    -e DAEMON_ID="${DAEMON_ID}" \
    -e JUDGEDAEMON_PASSWORD="${PASSWORD}" \
    -e DOMSERVER_BASEURL=http://172.x.x.x:8101/domjudge/ \
    -e CONTAINER_TIMEZONE=Asia/Shanghai \
    dup4/domjudge-judgehost
```

This is the script I used to start `judgehost`(Some sensitive information has been desensitized).

[`dup4/domjudge-judgehost`](https://hub.docker.com/layers/dup4/domjudge-judgehost/latest/images/sha256-7d9630d88174594f36316537bf98f38f729da2fb24c9e67e017c7184250ff85b?context=explore) is the judgehost image that I built by myself by pulling the latest code of `github.com/domjudge/domjudge`.

[...]

PR #1581, opened 2022-05-17
Add more randomness to cgroupname.

eldering:
This should help prevent collisions of cgroups, which might be the cause for #1396. Either way, it doesn't hurt.


nickygerritsen (approved):
Might the clash be because the pid is the same for all docker containers (pid 1)? Then if you do parallel judging it might start at the same time.

eldering:
> Might the clash be because the pid is the same for all docker containers (pid 1)? Then if you do parallel judging it might start at the same time.

This is the pid of `runguard`, although that could still be the same. However, that doesn't explain that this only happens when parallel judging... But maybe it's just more likely in that case.

PR #1585, opened 2022-05-26
Fix pdo

vmcj:
Since the introduction of Sentry we got on some systems issues related to this:
https://github.com/getsentry/sentry-symfony/issues/488

This was fixed in doctrine but sentry changed the class so that fix didn't work, instead of reverting the sentry introduction I've now marked all migrations as non transactional to fix the issue.

After some discussion with @nickygerritsen we've decided to add the fix per migration instead of introducing another class to inherit from for the following reasons:
- The amount of lines actually gets higher with that class
- There is now duplication on generated code which is a lot less of an issue
- This is a lot easier to follow
- Future migrations don't need this fix but would for consistency get that same baseclass or there is a difference between migrations which makes it a lot harder to work and debug why something is an issu

### DIFF
--- a/webapp/config/packages/doctrine_migrations.yaml
+++ b/webapp/config/packages/doctrine_migrations.yaml
@@ -3,3 +3,4 @@ doctrine_migrations:
         # namespace is arbitrary but should be different from App\Migrations
         # as migrations classes should NOT be autoloaded
         'DoctrineMigrations': '%kernel.project_dir%/migrations'
+    transactional: false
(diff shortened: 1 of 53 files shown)


nickygerritsen:
Can you squash the commits? And update the commit message explaining that we opted to not create a base migration class because new migrations will have the transactional=false option by default so it makes more sense to also do that for the old ones? (At least that is my opinion)

vmcj:
> Can you squash the commits? And update the commit message explaining that we opted to not create a base migration class because new migrations will have the transactional=false option by default so it makes more sense to also do that for the old ones? (At least that is my opinion)

I've added that now,

PR #1587, opened 2022-05-27
Drop presentation-error

vmcj:
Keeping this makes that it shows up different in the shadowing process
as the other unknown verdicts.

I think we should have more discussion on what is the correct process (IMO is to use another verdicts.php for shadowing where we can easily add all the unknown verdicts to have them show up in the shadowing page)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -8,6 +8,8 @@ Version 8.1.0 - DEV
  - Replace calls to deprecated `strftime` with `date`. Note: if you have
    configured a custom `time_format`, you need to update it to the format
    supported by `date`.
+ - Dropped the presentation-error verdict, databases from before DOMjudge 5.0
+ will be altered to replace the verdict with wrong-answer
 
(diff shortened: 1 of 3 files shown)


meisterT:
I think we need to add a database migration with it, right?

vmcj:
That would remove it also from existing installations right? Looking at the code we read the db values from verdicts.php so for new installations we wouldn't add the verdict.

nickygerritsen:
But old installs would indeed break. But I do wonder how many old installs we have, since this got dropped in 5.0, which is a long time ago

vmcj:
Where do we actually store this in the database? For me the configuration table is empty.

meisterT:
The issue is if you have judgings or judging_runs with that removed verdict and then in some places (eg. rejudgings) DOMjudge will not know how to handle this unknown verdict.

nickygerritsen (approved):

PR #1589, opened 2022-05-27
Since 2204LTS PHP81 is a normal standard

vmcj:
On main we added support for PHP81 but we couldn't add support for PHP81
and keep support for PHP72


nickygerritsen (approved):

thijskh:
I think you mean < 8.1 instead of <= 8.0?

vmcj:
> I think you mean < 8.1 instead of <= 8.0?

I'll fix it.

PR #1596, opened 2022-06-07
Rewrite contest import script in Python. Fixes #1594.

### LINKED ISSUE #1594: Rewrite import-contest script in python
Currently the bash script is hard to debug since we also support reading YAML/JSON. It was discussed that the script should be rewritten in python to make this easier to read/debug/extend.

nickygerritsen:
This makes the code so much more readable. It also allows us to more easily make changes in the future.


misc-tools/import-contest.py
```diff
@@ -0,0 +1,221 @@
+#!/usr/bin/env python3
+
+#
+# import-contest.py -- Convenience script to import a contest (including metadata, teams and
+# problems) via the command line. Supports reading credentials from ~/.netrc. See also
+# https://www.domjudge.org/docs/manual/main/import.html
```
vmcj, line 6: It would be nice to have the DOMjudge version here as this is depending on the API version

misc-tools/import-contest.py
```diff
@@ -0,0 +1,221 @@
...
+#
+# Part of the DOMjudge Programming Contest Jury System and licensed
+# under the GNU GPL. See README and COPYING for details.
+#
```
vmcj, line 10:
```suggestion
'''
import-contest.py -- Convenience script to import a contest (including metadata, teams and
problems) via the command line. Supports reading credentials from ~/.netrc. See also
https://www.domjudge.org/docs/manual/main/import.html

Part of the DOMjudge Programming Contest Jury System and licensed
under the GNU GPL. See README and COPYING for details.
'''
```

vmcj: Might be personal preference

misc-tools/import-contest.py
```diff
@@ -0,0 +1,221 @@
...
+headers = {
+    "user-agent": f"domjudge-import-contest ({requests.utils.default_user_agent()})"}
+
+api_url = None
+cid = None
+
+
+def confirm(message: str, default: bool) -> bool:
+    answer = 'x'
+    while answer not in ["", "y", "n"]:
```
vmcj, line 30:
```suggestion
    while answer.lower() not in ["", "y", "n"]:
```

misc-tools/import-contest.py
```diff
@@ -0,0 +1,221 @@
...
+        raise RuntimeError(e)
+
+    if response.status_code >= 300:
+        print(response.text)
+        if response.status_code == 401:
+            raise RuntimeError(
+                'Authentication failed, please check your DOMjudge credentials in ~/.netrc.')
+        else:
+            raise RuntimeError(
+                f'API request {name} failed (code {response.status_code}).')
```
vmcj, line 70: This is duplicated

misc-tools/import-contest.py
nickygerritsen: I copied this as is. We could add the version but we will probably forget to update it…

misc-tools/import-contest.py
nickygerritsen: Then we should remove the lower on line 32?

misc-tools/import-contest.py
nickygerritsen: I will put the “parsing” in a shared method

misc-tools/import-contest.py
eldering: You should *not* put the version here literally. You could either import it from a file like `etc/config.h` (which is written from `etc/config.h.in` by configure) or create `misc-tools/import-contest.py.in` and use a substitution rule, see `misc-tools/Makefile` and the `*.in` files there.

nickygerritsen: Ah yes that makes sense although we need something to strip the patch from if

nickygerritsen: Any clue how to do this? We have [this](https://github.com/DOMjudge/domjudge/blob/05f7af95e4a292f91a4dba443514e29a2a00e693/doc/manual/version.py.in) in the docs, but that is not run at Makefile time. I basically want something like this but then at Make time, so we can link to `https://www.domjudge.org/docs/manual/8.1/index.html`.
Also, how do we do this for `main`? We can't link to `https://www.domjudge.org/docs/manual/8.2/index.html` since that doesn't exist.

doc/manual/import.rst
```diff
@@ -360,11 +360,11 @@ Putting all API imports together
 
 If you prepare your contest configuration as we described in the previous
 subsections, you can also use the script that we provide in
-`misc-tools/import-contest.sh`.
+`misc-tools/import-contest.py`.
```
thijskh, line 363: What's with the extensions. Why do we do that?

nickygerritsen: I just 'copied' it from the old one. I'm fine dropping it, just like our submit client also doesn't have it. What do the others think?

vmcj: I vote in favor of dropping

misc-tools/import-contest.py
eldering: I'm not sure I follow: what information do you need? The domjudge version and the git branch?

That information should be collected at configure time and then when running make it would get injected into the script, see for example `misc-tools/Makefile` and the `subst-config` rule for `misc-tools/import-contest.py.in`.

misc-tools/import-contest
```diff
@@ -0,0 +1,214 @@
...
+                problemData = json.load(problemFile)
+                problems = [problem["id"] for problem in problemData]
+        else:  # problemset.json
+            with open("problemset.yaml") as problemFile:
+                problemData = yaml.safe_load(problemFile)
+                problems = [problem["short-name"]
+                            for problem in problemData["problems"]]
+
+        for problem in problems:
+            print(F"Preparing problem '{problem}'.")
```
vmcj, line 198:
```suggestion
            print(f"Preparing problem '{problem}'.")
```

misc-tools/import-contest
```diff
@@ -0,0 +1,214 @@
+#!/usr/bin/env python3
+
+'''
+import-contest -- Convenience script to import a contest (including metadata, teams and
+problems) via the command line. Supports reading credentials from ~/.netrc. See also
```
meisterT, line 5: super-nit: put the "See also " on a new line

misc-tools/import-contest
```diff
@@ -0,0 +1,214 @@
...
+import json
+import os.path
+import requests
+import requests.utils
+import sys
+from typing import List
+import yaml
+
+headers = {
+    "user-agent": f"domjudge-import-contest ({requests.utils.default_user_agent()})"}
```
meisterT, line 22: nit: be consistent about using `'` or `"` for all strings (I would use the former)

misc-tools/import-contest
```diff
@@ -0,0 +1,214 @@
...
+
+api_url = None
+cid = None
+
+
+def confirm(message: str, default: bool) -> bool:
+    answer = 'x'
+    while answer not in ["", "y", "n"]:
```
meisterT, line 30: the empty string here never happens

misc-tools/import-contest
```diff
@@ -0,0 +1,214 @@
...
+            last = f"'{files[-1]}'"
+            print(
+                f"Neither {prefix_joined} nor {last} found, skipping {basename}s import.")
+
+
+if len(sys.argv) < 2:
+    usage()
+
+api_url = sys.argv[1]
+
```
meisterT, line 130: As a first step here, we might want to check whether we have the right credentials and permissions by calling `{api_url}/user`?

misc-tools/import-contest
```diff
@@ -0,0 +1,214 @@
...
+    files = [(apifilename, open(file, "rb"))]
+
+    url = f"{api_url}/{name}"
+
+    response = requests.post(url, files=files, headers=headers, data=data)
+
+    return parse_api_response(name, response)
+
+
+def import_file(basename: str, files: List[str]):
```
meisterT, line 103: instead of adding the `s` at the end every time you are using `basename`, consider passing in plural versions right away?

misc-tools/import-contest
```diff
@@ -0,0 +1,214 @@
...
+    any_matched = False
+    for file in files:
+        if os.path.exists(file):
+            if confirm(f"Import {basename}s from {file}?", False):
+                print(f"Importing {basename}s.")
+                type = file.split(".")[1]
+                response = upload_file(f"/users/{basename}s", type, file)
+                print(json.dumps(response, indent=4))
```
meisterT, line 111: Would it make sense to return here? If there are multiple files and I import the first one, I probably don't want to get prompted for the other ones.

misc-tools/import-contest
```diff
@@ -0,0 +1,214 @@
...
+
+import_file("group", ["groups.json", "groups.tsv"])
+import_file("organization", ["organizations.json"])
+import_file("team", ["teams.json", "teams2.tsv"])
+import_file("account", ["accounts.json", "accounts.yaml", "accounts.tsv"])
+
+# Contest import is a special case: we can have a contest.yaml and problemset.yaml and then we need
+# to import the combined file. We can also gather the contest ID.
+if os.path.exists("contest.yaml"):
+    combine = False
```
meisterT, line 139: You may simplify the code here a little by doing something like

```
input_files = ['contest.yaml']
if os.path.exists...:
  input_files.append('problemset.yaml')
question = f'Import contest metadata (from {' and '.join(input_files)})'
if confirm(...)":
  if len(input_files) > 1:
    ...
```

misc-tools/import-contest
```diff
@@ -0,0 +1,214 @@
...
+if os.path.exists("problems.yaml") or os.path.exists("problems.json") or os.path.exists("problemset.yaml"):
+    if confirm("Import problems?", True):
+        # Check if our user is linked to a team
+        user_data = do_api_request("user")
+        if not "team" in user_data and not confirm("No team associated with your account. Jury submissions won't be imported. Really continue?", False):
+            exit(2)
+
+        print("Importing problems.")
+
+        if cid is None:
```
meisterT, line 175: I would move this up to line 162

misc-tools/import-contest
nickygerritsen: I thought I used `"` everywhere but apparently not. If you prefer `'` I will change them.

misc-tools/import-contest
nickygerritsen: Ah yeah it used to be until I introduced `default`.

misc-tools/import-contest
nickygerritsen: This makes sense, yes

misc-tools/import-contest
nickygerritsen: We didn't before but I do like it. Better to fail early. Will add.

misc-tools/import-contest
nickygerritsen: Bellow 162 I assume? And isn't it strange to ask for a contest ID if we DON'T have problems(et).{json,yaml}?

misc-tools/import-contest.py
nickygerritsen: I gave it a go but as you can see I needed to modify the configure.ac and paths.mk.in file for the actual thing I needed. I'm not sure if this is what we want so feel free to modify it.

nickygerritsen:
Fixed all comments except the last one from @meisterT (since it has a question open) and I'm not sure about the substitution logic during `make`.

meisterT (approved):

configure.ac
```diff
@@ -31,6 +31,23 @@ AC_SUBST(DOMJUDGE_VERSION, $PACKAGE_VERSION)
 AC_DEFINE_UNQUOTED(DOMJUDGE_VERSION, "$PACKAGE_VERSION",
 [DOMjudge version number (alias for PACKAGE_VERSION)])
 
+DOMJUDGE_VERSION_MIMOR=m4_esyscmd([grep 'version' README.md | sed -n '1s/^.*version //p' | sed 's/DEV$//' | sed 's/\.[[:digit:]]*$//'])
+DOMJUDGE_IS_DEV=m4_esyscmd([grep 'version' README.md | sed -n '1s/^.*version //p' | sed 's/[[:digit:]]//g;s/\.//g'])
+
+AC_SUBST(DOMJUDGE_VERSION_MIMOR, $DOMJUDGE_VERSION_MIMOR)
```
eldering, line 37: There's a typo `MIMOR`.

But besides that, I think it's better to not put all this logic here, but in the import-script instead: just parse the full `DOMJUDGE_VERSION` string there and determine the URL from it. Then the changes in `configure.ac` and `paths.mk.in` are not needed.

nickygerritsen: How do I do that? It’s a comment in python so i can’t really do that? Or does subst have some magic?

misc-tools/import-contest.py
eldering: Oh, I didn't look at the python part. I don't know whether you can create dynamic doc strings there.

But if all of this is just about getting that doc string to point to the exact URL, then maybe this is not worth it?

configure.ac
eldering: Oh, I didn't look at the python part. I don't know whether you can create dynamic doc strings there.

But if all of this is just about getting that doc string to point to the exact URL, then maybe this is not worth it?

nickygerritsen: That is good enough yeah

eldering (approved):

PR #1600, opened 2022-06-18
Sentry update

vmcj:
When working on another PR it seems that updating sentry removes some deprecations, this PR is to see what is the before/after difference


nickygerritsen (approved):

nickygerritsen (approved):
This doesn't actually update the sentry-symfony bundle. But it does update Sentry and others.

vmcj:
> This doesn't actually update the sentry-symfony bundle. But it does update Sentry and others.

Yes, there is not a newer version for that one. So most issues still persist.

PR #1602, opened 2022-06-19
Fix mayhem API

vmcj:
We now also trigger on push to show that it works, this will be reverted after approval

To fix the composer packages we now set the environment to development.
This should be fine as we only care for the API output and not the other
workings.

The logs are now changed to be foldable and more information is shown
about the state of the github actions runner. When this works correctly
we could speed this up by removing these.

Mayhem does find issues with the API definition but I'm not 100% sure yet how to solve those yet.


swarkentin:
👋 Hi there - I am a developer with Mayhem for API and saw this come up. I tried resolving this a while back, but was running into issues generating the swagger/OpenAPI spec from the application. I'm happy to help you get this intergrated back into your pipelines if you are stuck.

> Mayhem does find issues with the API definition but I'm not 100% sure yet how to solve those yet.

Last I looked, DOMjudge specification is auto-generated. This can lead to a number of Invalid Spec warnings (ie: a missing response code for some endpoint). If this is too noisy, or cannot be resolved, it is possible to ignore these warnings in the GH action with the `ignore-rule` argument. Here is an example of where we have done that in another project:

https://github.com/mayhemheroes/keycloak/blob/main/.github/workflows/mayhem.yml#L127

```yaml
-  name: Mayhem for API
        uses: ForAllSecure/mapi-action@v1
        continue-on-error: true
        with:
          mapi-token: ${{ secrets.MAPI_TOKEN }}
          run-args: |
            --ignore-rule                       << Ignore spec warnings
            InvalidResponseSpec

 ```

❓  How would I go about getting a swagger/OpenAPI spec DOMjudge once it is built?

**Comment by vmcj** [comment]
> 👋 Hi there - I am a developer with Mayhem for API and saw this come up. I tried resolving this a while back, but was running into issues generating the swagger/OpenAPI spec from the application. I'm happy to help you get this intergrated back into your pipelines if you are stuck.
> 
Hi @swarkentin, thank you for your help! I indeed got stuck.

> > Mayhem does find issues with the API definition but I'm not 100% sure yet how to solve those yet.
> 
> Last I looked, DOMjudge specification is auto-generated. This can lead to a number of Invalid Spec warnings (ie: a missing response code for some endpoint). If this is too noisy, or cannot be resolved, it is possible to ignore these warnings in the GH action with the `ignore-rule` argument. Here is an example of where we have done that in another project:
> 
> https://github.com/mayhemheroes/keycloak/blob/main/.github/workflows/mayhem.yml#L127
> 
> ```yaml
> -  name: Mayhem for API
>         uses: ForAllSecure/mapi-action@v1
>         continue-on-error: true
>         with:
>           mapi-token: ${{ secrets.MAPI_TOKEN }}
>           run-args: |
>             --ignore-rule                       << Ignore spec warnings
>             InvalidResponseSpec
> ```
> 
> ❓ How would I go about getting a swagger/OpenAPI spec DOMjudge once it is built?

You can find a run here:
https://github.com/vmcj/domjudge/runs/6973736975?check_suite_focus=true) and the definition can also be found at domjudge.org/demoweb/doc.json.

**Comment by swarkentin** [comment]
> The main problem I currently have is this:
> 
> ```
> error: Failed to parse specification - invalid OAS3.1 spec: Error("data did not match any variant of untagged enum Schema", line: 0, column: 0)
> ```

I've got this reproducing locally. We recently added support for OpenAPI 3.1 and it looks some assumptions between the spes changed between 3.0 and 3.1. 

I believe this is something we can resolve on the Mayhem ofr API side of things. I will share with the team and get back to this MR once we have a solution!

**Comment by vmcj** [comment]
I think the run I gave was from my other branch where I tried to remove some parts to see if I could make it valid again. The current issues (https://github.com/vmcj/domjudge/actions/runs/2533224993) are actually also:
```we noticed the following issues:
/components/parameters/balloonId: Additional properties are not allowed ('examples' was unexpected)
/components/parameters/cid: Additional properties are not allowed ('examples' was unexpected)
/components/parameters/id: Additional properties are not allowed ('examples' was unexpected)
/components/parameters/idlist: Additional properties are not allowed ('explode' was unexpected)
/components/securitySchemes/basicAuth: Additional properties are not allowed ('scheme' was unexpected)```

looking at the:
`/components/parameters/balloonId: Additional properties are not allowed ('examples' was unexpected)`
in the spec for 3.1.0 (https://spec.openapis.org/oas/v3.1.0#fixed-fields-9) examples should be allowed as property.
This could be a misinterpretation on my side or is this a bug in mayhem?

**Comment by swarkentin** [comment]
> I think the run I gave was from my other branch where I tried to remove some parts to see if I could make it valid again. The current issues (https://github.com/vmcj/domjudge/actions/runs/2533224993) are actually also:
> 
> ```
> /components/parameters/balloonId: Additional properties are not allowed ('examples' was unexpected)
> /components/parameters/cid: Additional properties are not allowed ('examples' was unexpected)
> /components/parameters/id: Additional properties are not allowed ('examples' was unexpected)
> /components/parameters/idlist: Additional properties are not allowed ('explode' was unexpected)
> /components/securitySchemes/basicAuth: Additional properties are not allowed ('scheme' was unexpected)```
> 
> looking at the:
> `/components/parameters/balloonId: Additional properties are not allowed ('examples' was unexpected)`
> in the spec for 3.1.0 (https://spec.openapis.org/oas/v3.1.0#fixed-fields-9) examples should be allowed as property.
> This could be a misinterpretation on my side or is this a bug in mayhem?
> ```

Thank you for the additional logs! We have identified the issue and are working on a fix now. It appears to in the OpenAPI spec deserialization code. 

I will update you as soon as the new version is ready - at which point, all you should need to do is re-run the action.

**Comment by swarkentin** [comment]
> > @vmcj We have published a new version of the CLI (2.11.1) which should accept the DOMJudge spec now. The fixes should apply if you re run the action now.
> 
> Hi @swarkentin, I'm not sure if the problem is fixed. When I check locally with the latest `mapi` I still get the problem about OAS3.1 although the openapi is still version 3.0.0. For an example run see: https://github.com/vmcj/domjudge/runs/6990827269?check_suite_focus=true

It looks like the change to declaring the OAS spec as "3.1.0" needed to be handled differently on our end (we were testing with "3.0.0"). We have deployed the fix and it should be good to go! Re-running the action once more should do it. 

```
nelmio_api_doc:
    documentation:
        openapi: 3.1.0  <<< this update
        info:
```

vmcj:
> > > @vmcj We have published a new version of the CLI (2.11.1) which should accept the DOMJudge spec now. The fixes should apply if you re run the action now.
> > 
> > 
> > Hi @swarkentin, I'm not sure if the problem is fixed. When I check locally with the latest `mapi` I still get the problem about OAS3.1 although the openapi is still version 3.0.0. For an example run see: https://github.com/vmcj/domjudge/runs/6990827269?check_suite_focus=true
> 
> It looks like the change to declaring the OAS spec as "3.1.0" needed to be handled differently on our end (we were testing with "3.0.0"). We have deployed the fix and it should be good to go! Re-running the action once more should do it.
> 
> ```
> nelmio_api_doc:
>     documentation:
>         openapi: 3.1.0  <<< this update
>         info:
> ```

I indeed forgot to take that out for now. After reverting that commit everything seems to work. Thank you for the help!

nickygerritsen (approved):

PR #1610, opened 2022-06-25
Mayhem nullable property

vmcj:
Currently we had some properties which according to the api doc are always a string. But when these have no value they were rendered as null. This PR makes this explicit so a client can first check if something is null and base its rendering on that instead of having to check if a string is empty which could be an actual choice.

Looking at https://github.com/DOMjudge/domjudge/commit/7f0b5a479a20e2870029f92d5b0482c718ca3de2 I'm not sure if the AssertNotBlank had another use but it seems to work this way.

### DIFF
--- a/webapp/src/Entity/Contest.php
+++ b/webapp/src/Entity/Contest.php
@@ -76,7 +76,6 @@ class Contest extends BaseApiEntity implements AssetEntityInterface
      *                            nullable=false)
      * @Serializer\Groups({"Nonstrict"})
      * @Identifier()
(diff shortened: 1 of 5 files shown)


nickygerritsen (approved):
I was hoping swagger would be smart enough to understand `?string` means `string or null` but I guess not.

vmcj:
> I was hoping swagger would be smart enough to understand `?string` means `string or null` but I guess not.

This actually breaks Unit tests, so I need to think of another solution (for example ignoring this and fixing it in a new API version)

PR #1613, opened 2022-06-26
When creating an user for a new team read the correct field

vmcj:


vmcj:
I still need to make a test for this.

nickygerritsen (approved):

nickygerritsen (approved):

PR #1616, opened 2022-07-04
Fix post user bug

vmcj:
We should consider making more tests but we already have that issue and I rather first release 8.1.1.

### DIFF
--- a/webapp/src/Controller/API/UserController.php
+++ b/webapp/src/Controller/API/UserController.php
@@ -367,6 +367,9 @@ class UserController extends AbstractRestController
 
         $roles = (array)$request->request->get('roles');
         foreach ($roles as $djRole) {
+            if ($djRole === '') {
+                continue;
+            }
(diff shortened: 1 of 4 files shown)


vmcj:
It seems we can backport this to 8.{0,1},
https://github.com/DOMjudge/domjudge/blob/8.0/webapp/src/Entity/User.php

7.3 seems to not have this code.

webapp/src/Controller/API/UserController.php
```diff
@@ -348,7 +348,7 @@ public function addAction(Request $request): Response
             ->setPlainPassword($request->request->get('password'))
             ->setEnabled($request->request->getBoolean('enabled', true));
 
-        if ($request->request->has('team_id')) {
+        if ($request->request->has('team_id') && $request->request->get('team_id') !== '') {
```
nickygerritsen, line 351: Why not:
```suggestion
        if ($request->request->get('team_id')) {
```
Then?

nickygerritsen (approved):

Issue #229, opened 2016-06-04
Implement web UI to delete multiple of a kind


meisterT:
Currently it's only possible to delete one team, problem, ...
It would be nice to delete e.g. multiple teams at once.

mvr320:
Am I allowed to help with this?

mvr320:
I'll work on this

atiabjobayer:
Any possibility of this getting released soon?

meisterT:
@atiabjobayer a fix for this has not been merged yet - what entities do you care the most about?

atiabjobayer:
> [@atiabjobayer](https://github.com/atiabjobayer) a fix for this has not been merged yet - what entities do you care the most about?

In the order of preference:
1. Users
2. Teams
3. Affiliations


It'd be really helpful if you could please merge it quickly. Thanks in advance.

vmcj:
@meisterT I think we can close this?

meisterT:
I think so, if there are entities missing, it should be relatively easy to add

PR #1627, opened 2022-07-16
First part for multidelete

vmcj:
This adds the functionality to delete multiple:
contests
problems

It can be extended to also do other types of entities but I'm not sure if it makes sense for everything and there are some pages that don't have the primaryKey rendered in the correct part. This makes extending need a rewrite. I prefer to first fix the easy things and extend on this in the future when this passes codereview.

This might also close https://github.com/DOMjudge/domjudge/issues/229 if we agree that no more endpoints are needed.

### DIFF
--- a/webapp/src/Controller/BaseController.php
+++ b/webapp/src/Controller/BaseController.php
@@ -356,6 +356,9 @@ abstract class BaseController extends AbstractController
         array $entities,
         string $redirectUrl
     ) : Response {
+        if (count($entities) === 0) {
+            throw new BadRequestHttpException('No entities provided.');
+        }
         // Assume that we only delete entities of the same class.
         foreach ($entities as $entity) {
             assert(get_class($entities[0]) === get_class($entity));
--- a/webapp/src/Controller/BaseController.php
+++ b/webapp/src/Controller/BaseController.php
@@ -241,9 +241,17 @@ abstract class BaseController extends AbstractController
                 );
                 $entityManager->clear();
                 $entity = $entityManager->getRepository(Problem::class)->find($entity->getProbid());
+            } elseif ($entity instanceof Team) {
+                $users = $entity->getUsers();
             }
+
             $entityManager->remove($entity);
-            $entityManager->flush();
+
+            if ($entity instanceof Team) {
+                foreach($users as $user) {
+                    $entityManager->refresh($user);
+                }
+            }
         });
 
         if ($entity instanceof Team) {
--- a/webapp/src/Controller/BaseController.php
+++ b/webapp/src/Controller/BaseController.php
@@ -408,7 +408,7 @@ abstract class BaseController extends AbstractController
         }
 
         if (count($entities) > 1) {
-            $readableType = (substr($readableType, -1) !== 'y') ? $readableType.'s' : substr($readableType, 0, -1).'ies';
+            $readableType = (substr($readableType, -1) === 'y') ? (substr($readableType, 0, -1).'ies') : ($readableType.'s');
         }
 
         $data = [
--- a/webapp/src/Controller/BaseController.php
+++ b/webapp/src/Controller/BaseController.php
@@ -28,6 +28,7 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\HttpKernel\KernelInterface;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\Routing\Exception\ResourceNotFoundException;
--- a/webapp/src/Controller/BaseController.php
+++ b/webapp/src/Controller/BaseController.php
@@ -415,6 +415,7 @@ abstract class BaseController extends AbstractController
             'showModalSubmit' => !$isError,
             'modalUrl' => $request->getRequestUri(),
             'redirectUrl' => $redirectUrl,
+            'primaryKeyItems' => $primaryKeyData,
         ];
         if ($request->isXmlHttpRequest()) {
             return $this->render('jury/delete_modal.html.twig', $data);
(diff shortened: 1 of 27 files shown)


webapp/src/Controller/BaseController.php
```diff
@@ -364,23 +368,27 @@ protected function deleteEntities(
 
         [$isError, $primaryKeyData, $deleteTreeMessages] = $this->buildDeleteTree($entities, $relations, $entityManager);
         if (!empty($deleteTreeMessages)) {
-            $messages = $deleteTreeMessages;
+            $messages = array_unique($deleteTreeMessages);
         }
 
-        if ($request->isMethod('POST')) {
+        // Check if we handle a list of deletable entities.
+        $removeList = $request->request->has('confirmation') ? $request->request->get('confirmation') === '1' : true;
```
nickygerritsen, line 375:
```suggestion
        $removeList = $request->request->get('confirmation', true);
```

webapp/src/Controller/BaseController.php
```diff
@@ -364,23 +368,27 @@ protected function deleteEntities(
...
+
+        if ($request->isMethod('POST') && $removeList ){
```
nickygerritsen, line 377:
```suggestion
        if ($request->isMethod('POST') && $removeList) {
```

webapp/src/Controller/BaseController.php
```diff
@@ -364,23 +368,27 @@ protected function deleteEntities(
...
             if ($isError) {
                 throw new BadRequestHttpException(reset($messages));
             }
 
             $msgList = [];
-            foreach ($entities as $entity) {
+            foreach ($entities as $id=>$entity) {
```
nickygerritsen, line 383:
```suggestion
            foreach ($entities as $id => $entity) {
```

webapp/src/Controller/BaseController.php
```diff
@@ -394,6 +402,10 @@ protected function deleteEntities(
             $descriptions[] = $entity->getShortDescription();
         }
 
+        if (count($entities) > 1) {
+            $readableType = (substr($readableType, -1) !== 'y') ? $readableType.'s' : substr($readableType, 0, -1).'ies';
```
nickygerritsen, line 406: Maybe swap the two parts and change the `!==` to `===`?
Also maybe add some braces like `($readableType.'s')` for readability?
Also spaces.

webapp/src/Controller/BaseController.php
```diff
@@ -556,4 +570,29 @@ protected function judgeRemaining(array $judgings): void
...
+            }
+            $entityId = substr($key, strlen($checkboxPrefix));
+            $entity = $this->em->getRepository($class)->find($entityId);
+            if (!$entity) {
+                throw new NotFoundHttpException(sprintf('%s with ID %s not found', ucfirst($classHumanReadable), $entityId));
+            }
+            $entitiesToDelete[] = $entity;
+        }
+
+        if (count($entitiesToDelete)===0) {
```
nickygerritsen, line 590:
```suggestion
        if (count($entitiesToDelete) === 0) {
```

webapp/src/Controller/Jury/ContestController.php
```diff
@@ -907,4 +907,13 @@ private function checkTimezones(FormInterface $form): ?Response
...
         return null;
     }
+
+    /**
+     * @Route("/deleteList", name="jury_contests_delete")
+     * @IsGranted("ROLE_ADMIN")
+     */
+    public function deleteListAction(Request $request): Response
+    {
+        return parent::deleteListActionHelper($request, Contest::class, 'jury_contests', 'contest');
```
nickygerritsen, line 917: Why not:

```suggestion
        return $this->deleteListActionHelper($request, Contest::class, 'jury_contests', 'contest');
```

? I think that is more common?

webapp/templates/jury/jury_macros.twig
```diff
@@ -71,85 +71,100 @@
     </style>
 {% endmacro %}
 
-{% macro table(data, fields, num_actions, options) %}
+{% macro table(data, fields, num_actions, options, delete_list_endpoint, multidelete=true) %}
```
nickygerritsen, line 74: Maybe we make `multidelete` false by dfault for now? Or even better: make `delete_list_endpoint` null by default, but if it is not `null`, we know we have multi delete?

webapp/templates/jury/contests.html.twig
```diff
@@ -84,12 +84,15 @@
...
     <h3>All available contests</h3>
 
-    {{ macros.table(contests_table, table_fields, num_actions) }}
+    {{ macros.table(contests_table, table_fields, num_actions, null, "jury_contests_delete") }}
 
     {% if is_granted('ROLE_ADMIN') %}
         <p>
             {{ button(path('jury_contest_add'), 'Add new contest', 'primary', 'plus') }}
             {{ button(path('jury_import_export', {'_fragment':'contests'}), 'Import contest', 'primary', 'upload') }}
+            <button type="submit" form="multidelete" class="btn btn-danger" name="multidelete_submit">
```
nickygerritsen, line 93: Should we add some javascript to disable this button if nothing is selected? We can probably do that in the macro?

webapp/src/Controller/Jury/ContestController.php
vmcj: In an earlier version the function was overwritten, I agree this makes more sense with the current code.

webapp/src/Controller/BaseController.php
```diff
@@ -240,8 +241,17 @@ protected function commitDeleteEntity(
                 );
                 $entityManager->clear();
                 $entity = $entityManager->getRepository(Problem::class)->find($entity->getProbid());
+            } elseif ($entity instanceof Team) {
+                $users = $entity->getUsers();
```
nickygerritsen, line 245: Maybe add a comment as to why we do this? But I think it would be better even if we do the eventlogservice->log() at the end with *all* ID's at once, which will mean we don't have to do this and the below change. Probably adding a `bool $eventlog = true` as argument to this method and setting it to false for multi delete should work.

webapp/src/Controller/BaseController.php
```diff
@@ -369,24 +382,28 @@ protected function deleteEntities(
...
+        $removeList = $request->request->get('confirmation', true);
+
+        if ($request->isMethod('POST') && $removeList) {
             if ($isError) {
                 throw new BadRequestHttpException(reset($messages));
             }
 
-            $msgList = [];
             foreach ($entities as $id => $entity) {
+                $msgList = [];
```
nickygerritsen, line 397: So basically we do the eventlog after this loop with array_keys($id) as argument. That will also speed up deletion

PR #1629, opened 2022-07-22
Remove special chroot for java8

vmcj:
If I understand correctly this was to install java8 via oracle, as far as I'm aware Oracle has changed their licensing and OpenJDK is now the adviced method for our case.


vmcj:
I'm not sure if the reparation of the Makefile is correct, will ask @eldering tomorrow for help.

nickygerritsen (approved):

eldering:
Please wait for @ubergeek42 to review this as I think he added it.

ubergeek42 (approved):
Yep, totally fine to delete this, I don't think I've used it in ages

PR #1641, opened 2022-07-23
Add button to download static scoreboard ZIP. Fixes #1595.

### LINKED ISSUE #1595: Add way to download static scoreboard with all linked files
Add an export to export the static scoreboard together with:
* All CSS
* All Javascript
* All images
* A way to view team pages, either creating separate pages or doing something smart with some JSON. The latter is probably cleaner

Requested by @verwoerd and I like the idea.

nickygerritsen:
Team detail pages are not here (yet), since that is more difficult. If really wanted, we could add it, but this should already be a good start.

### DIFF
--- a/webapp/src/Controller/PublicController.php
+++ b/webapp/src/Controller/PublicController.php
@@ -176,7 +176,7 @@ class PublicController extends BaseController
                 $contest      = null;
                 $activateTime = null;
                 foreach ($this->dj->getCurrentContests(-1) as $possibleContest) {
-                    if (!$possibleContest->getPublic() || !$possibleContest->getEnabled()) {
+                    if (!($possibleContest->getPublic() && $possibleContest->getEnabled())) {
                         continue;
                     }
                     if ($activateTime === null || $activateTime < $possibleContest->getActivatetime()) {
--- a/webapp/src/Controller/PublicController.php
+++ b/webapp/src/Controller/PublicController.php
@@ -5,20 +5,26 @@ namespace App\Controller;
 use App\Entity\Contest;
 use App\Entity\ContestProblem;
 use App\Entity\Team;
+use App\Helpers\RequestBasePathModifier;
 use App\Service\ConfigurationService;
 use App\Service\DOMJudgeService;
 use App\Service\ScoreboardService;
 use App\Service\StatisticsService;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\NonUniqueResultException;
+use ReflectionClass;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\HttpKernel\Exception\ServiceUnavailableHttpException;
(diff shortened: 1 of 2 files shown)


webapp/src/Controller/PublicController.php
```diff
@@ -116,6 +95,114 @@ public function scoreboardAction(Request $request): Response
...
+                $contest = null;
+                foreach ($this->dj->getCurrentContests(-1) as $possibleContest) {
+                    if ($possibleContest->getCid() == $contestId || $possibleContest->getExternalid() == $contestId) {
+                        $contest = $possibleContest;
+                        break;
+                    }
+                }
+
+                if (!$contest) {
+                    throw new NotFoundHttpException('Specified contest not found');
```
vmcj, line 198:
```suggestion
                    throw new NotFoundHttpException('Specified contest not found.');
```

You also add it for the error in https://github.com/DOMjudge/domjudge/pull/1641/files#diff-cc71aa6a9b851ee37894e1943a11fd1d6676ae347fac6e088ec93325e4a90e1aR130

nickygerritsen: Well no, I copied it, but it makes sense.

webapp/src/Controller/PublicController.php
```diff
@@ -116,6 +95,114 @@ public function scoreboardAction(Request $request): Response
...
+    {
+        $contest = null;
+        // For static scoreboards, allow to pass a contest= param.
+        if ($contestId = $request->query->get('contest')) {
+            if ($contestId === 'auto') {
+                // Automatically detect the contest that is activated the latest.
+                $contest      = null;
+                $activateTime = null;
+                foreach ($this->dj->getCurrentContests(-1) as $possibleContest) {
+                    if (!$possibleContest->getPublic() || !$possibleContest->getEnabled()) {
```
vmcj, line 179:
```suggestion
                    if (!($possibleContest->getPublic() && $possibleContest->getEnabled())) {
```
I find this easier to read (I think you just copied this from what was there already)

webapp/src/Controller/PublicController.php
```diff
@@ -116,6 +95,114 @@ public function scoreboardAction(Request $request): Response
...
+                        continue;
+                    }
+                    if ($activateTime === null || $activateTime < $possibleContest->getActivatetime()) {
+                        $activateTime = $possibleContest->getActivatetime();
+                        $contest      = $possibleContest;
+                    }
+                }
+            } else {
+                // Find the contest with the given ID.
+                $contest = null;
```
vmcj, line 189: I think you already set this to null on L171

vmcj (approved):
Looks good to me, assuming you also tested it.

There are some variable initializations which can be moved outside of the if/else to save some lines.

webapp/src/Controller/PublicController.php
```diff
@@ -116,6 +95,112 @@ public function scoreboardAction(Request $request): Response
         return $this->render('public/scoreboard.html.twig', $data, $response);
     }
 
+    /**
+     * @Route("/scoreboard-data.zip", name="public_scoreboard_data_zip")
+     */
+    public function scoreboardDataZipAction(RequestStack $requestStack, string $projectDir, Request $request): Response
+    {
+        $contest = $this->getContestFromRequest($request) ?? $this->dj->getCurrentContest(-1);
```
meisterT, line 103: no access check required or is this implicit?

nickygerritsen: The only check is for non public scoreboards and that should be implicit, but I will verify

nickygerritsen: Indeed, `getCurrentContest` (which is also used in `getContestFromRequest`) checks whether you have access.

Issue #1558, opened 2022-04-27
Add "Protected" flag to a contest


dario2994:
I have a feature to suggest. 
A contest should have a flag "Protected" that, when set, makes it impossible to modify the properties of the contest (modify/add/remove problems, change title, starting or ending time, ...).

Why do I suggest to add such a feature? During SWERC I was running scripts to import the contest on different instances (for example the mirror, the shadow contest...) and if I had done a mistake (wrong url) I would have modified (with rather bad consequences I believe) the content of the ongoing contest.

A "normal" use of the flag should be to set it some ~5 minutes after the beginning of the contest and keep it on unless something very bad happens (i.e., a problem must be fixed because the official solution is wrong). 

It shall be impossible to modify the value of the flag through the API (an active human intervention should always be needed to flip the value of the flag).

It would also be very nice if, when the flag is off and the contest started by > 15 minutes, there were a red alert in the web-interface telling to jury/admin that the contest is not protected.

PR #1642, opened 2022-07-23
Adding option to lock a contest.

meisterT:
This is an attempt to address https://github.com/DOMjudge/domjudge/issues/1558.

I have likely forgotten quite a few places, and want to go through the usability.

### DIFF
--- a/webapp/src/Controller/API/ContestController.php
+++ b/webapp/src/Controller/API/ContestController.php
@@ -278,7 +278,7 @@ class ContestController extends AbstractRestController
             throw new AccessDeniedHttpException('Contest is locked, go to ' . $contestUrl . ' to unlock it.');
         }
 
-    /** @var UploadedFile $banner */
+        /** @var UploadedFile $banner */
         $banner = $request->files->get('banner');
 
         if (!$banner) {
@@ -345,9 +345,6 @@ class ContestController extends AbstractRestController
             $response = new JsonResponse('Missing "start_time" in request.', Response::HTTP_BAD_REQUEST);
         } elseif ($request->request->get('id') != $contest->getApiId($this->eventLogService)) {
             $response = new JsonResponse('Invalid "id" in request.', Response::HTTP_BAD_REQUEST);
-        } else if ($contest->isLocked()) {
-            $contestUrl = $this->generateUrl('jury_contest', ['contestId' => $cid], UrlGeneratorInterface::ABSOLUTE_URL);
-            throw new AccessDeniedHttpException('Contest is locked, go to ' . $contestUrl . ' to unlock it.');
         } elseif (!$request->request->has('force') &&
             $contest->getStarttime() != null &&
             $contest->getStarttime() < $now + 30) {
--- a/webapp/src/Controller/API/ContestController.php
+++ b/webapp/src/Controller/API/ContestController.php
@@ -32,6 +32,7 @@ use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\HttpKernel\KernelInterface;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 use Symfony\Component\Yaml\Yaml;
 
@@ -143,7 +144,7 @@ class ContestController extends AbstractRestController
      */
     public function listAction(Request $request): Response
     {
-        
+
         return parent::performListAction($request);
     }
 
@@ -222,6 +223,11 @@ class ContestController extends AbstractRestController
             throw new NotFoundHttpException(sprintf('Object with ID \'%s\' not found', $id));
         }
 
+        if ($contest->isLocked()) {
+            $contestUrl = $this->generateUrl('jury_contest', ['contestId' => $id], UrlGeneratorInterface::ABSOLUTE_URL);
+            throw new AccessDeniedHttpException('Contest is locked, go to ' . $contestUrl . ' to unlock it.');
+        }
+
         $contest->setClearBanner(true);
 
         $this->assetUpdater->updateAssets($contest);
@@ -267,7 +273,12 @@ class ContestController extends AbstractRestController
             throw new NotFoundHttpException(sprintf('Object with ID \'%s\' not found', $id));
         }
 
-        /** @var UploadedFile $banner */
+        if ($contest->isLocked()) {
+            $contestUrl = $this->generateUrl('jury_contest', ['contestId' => $id], UrlGeneratorInterface::ABSOLUTE_URL);
+            throw new AccessDeniedHttpException('Contest is locked, go to ' . $contestUrl . ' to unlock it.');
+        }
+
+    /** @var UploadedFile $banner */
         $banner = $request->files->get('banner');
 
         if (!$banner) {
@@ -334,6 +345,9 @@ class ContestController extends AbstractRestController
             $response = new JsonResponse('Missing "start_time" in request.', Response::HTTP_BAD_REQUEST);
         } elseif ($request->request->get('id') != $contest->getApiId($this->eventLogService)) {
             $response = new JsonResponse('Invalid "id" in request.', Response::HTTP_BAD_REQUEST);
+        } else if ($contest->isLocked()) {
+            $contestUrl = $this->generateUrl('jury_contest', ['contestId' => $cid], UrlGeneratorInterface::ABSOLUTE_URL);
+            throw new AccessDeniedHttpException('Contest is locked, go to ' . $contestUrl . ' to unlock it.');
         } elseif (!$request->request->has('force') &&
             $contest->getStarttime() != null &&
             $contest->getStarttime() < $now + 30) {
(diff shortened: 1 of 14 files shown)


webapp/src/Controller/API/ContestController.php
```diff
@@ -335,6 +335,8 @@ public function changeStartTimeAction(Request $request, string $cid): Response
             $response = new JsonResponse('Missing "start_time" in request.', Response::HTTP_BAD_REQUEST);
         } elseif ($request->request->get('id') != $contest->getApiId($this->eventLogService)) {
             $response = new JsonResponse('Invalid "id" in request.', Response::HTTP_BAD_REQUEST);
+        } else if ($contest->isLocked()) {
+            $response = new JsonResponse('Contest is locked.', Response::HTTP_BAD_REQUEST);
```
vmcj, line 339:
```suggestion
            $response = new JsonResponse('Contest is locked, go to /jury/somewhere to unlock to make modifications to this running contest.', Response::HTTP_BAD_REQUEST);
```

meisterT: Done.

webapp/src/Controller/API/ContestController.php
```diff
@@ -267,7 +273,12 @@ public function setBannerAction(Request $request, string $id, ValidatorInterface
             throw new NotFoundHttpException(sprintf('Object with ID \'%s\' not found', $id));
         }
 
-        /** @var UploadedFile $banner */
+        if ($contest->isLocked()) {
+            $contestUrl = $this->generateUrl('jury_contest', ['contestId' => $id], UrlGeneratorInterface::ABSOLUTE_URL);
+            throw new AccessDeniedHttpException('Contest is locked, go to ' . $contestUrl . ' to unlock it.');
+        }
+
+    /** @var UploadedFile $banner */
```
vmcj, line 281: nitpick: indentation is wrong.

webapp/src/Controller/API/ProblemController.php
```diff
@@ -87,7 +89,12 @@ public function addProblemsAction(Request $request): array
         /** @var Contest $contest */
         $contest = $this->em->getRepository(Contest::class)->find($contestId);
 
-        /** @var UploadedFile $file */
+        if ($contest->isLocked()) {
+            $contestUrl = $this->generateUrl('jury_contest', ['contestId' => $contestId], UrlGeneratorInterface::ABSOLUTE_URL);
+            throw new AccessDeniedHttpException('Contest is locked, go to ' . $contestUrl . ' to unlock it.');
+        }
+
+    /** @var UploadedFile $file */
```
vmcj, line 97: nitpick: indentation.

webapp/src/Controller/Jury/ContestController.php
```diff
@@ -95,6 +96,10 @@ public function indexAction(Request $request): Response
             if (!$contest) {
                 throw new NotFoundHttpException('Contest not found');
             }
+            if ($contest->isLocked()) {
```
vmcj, line 99: I think unfreeze, finalize and delay_start should all be possible when its locked? Or we should get a popup 'do you want to unlock and do $action` as the user is already in the UI where the risk is much lower.

webapp/src/Controller/API/ProblemController.php
meisterT: Done

webapp/src/Controller/Jury/ContestController.php
meisterT: Yep, you are right. Locking a contest is mostly meant to prevent accidental edits to problem data and such. We *need* to allow the start time interface to allow pausing/resuming contest countdown via the CDS anyway. So I relaxed things a little.

webapp/src/Controller/API/ContestController.php
meisterT: Done

webapp/tests/Unit/Controller/API/ProblemControllerAdminTest.php
```diff
@@ -170,4 +171,34 @@ public function testAddExisting(): void
...
+
+    public function testAddToLocked(): void
+    {
+        $this->loadFixture(LockedContestFixture::class);
+        $this->loadFixture(DummyProblemFixture::class);
+
+        $body = [
+            'label'        => 'newproblem',
+            'points'       => 3,
+            'rgb'        => '#013370',
```
vmcj, line 183:
```suggestion
            'rgb'          => '#013370',
```

webapp/templates/jury/contest.html.twig
```diff
@@ -27,6 +27,13 @@
             This contest is final.
         </div>
     {% endif %}
+    {%- if is_granted('ROLE_ADMIN') -%}
+        {% if contest.isLocked %}
```
vmcj, line 31: why don't we use an and?

webapp/src/Entity/Contest.php
```diff
@@ -314,6 +314,15 @@ class Contest extends BaseApiEntity implements AssetEntityInterface
      */
     private bool $openToAllTeams = true;
 
+    /**
+     * @ORM\Column(type="boolean", name="is_locked",
+     *     options={"comment"="Is this contest locked for modifications?",
+     *              "default"=0},
+     *     nullable=false)
+     * @Serializer\Exclude()
```
vmcj, line 322: I might be wrong, but I think if we exclude the serializer it does not show up in the API response. I think we want it to show up there.

vmcj:
I think the alert is still missing that when the contest has started an admin gets an alert that the contest is not locked yet.

webapp/src/Entity/Contest.php
meisterT: I have followed the other attributes, e.g. public, enabled, openToAllTeams.

webapp/templates/jury/contest.html.twig
meisterT: Done.

meisterT:
> I think the alert is still missing that when the contest has started an admin gets an alert that the contest is not locked yet.

Intentionally so. I would first want to play with the mechanics of this manually in a real contest before pushing this onto people.

webapp/src/Controller/Jury/ContestController.php
```diff
@@ -259,7 +260,7 @@ public function indexAction(Request $request): Response
                 }
             }
 
-            if ($this->isGranted('ROLE_ADMIN')) {
+            if ($this->isGranted('ROLE_ADMIN') && !$contest->isLocked()) {
```
thijskh, line 263: Might be cooler to grey them out in this case?

webapp/src/Entity/Contest.php
```diff
@@ -770,6 +779,16 @@ public function isOpenToAllTeams(): bool
...
     }
 
+    public function isLocked(): bool
+    {
+        return $this->isLocked;
+    }
+
+    public function setIsLocked(bool $isLocked): bool
+    {
+        return $this->isLocked = $isLocked;
```
thijskh, line 789: Should probably return `self`, see other setters of boolean flags in the file

webapp/templates/jury/contest.html.twig
```diff
@@ -27,6 +27,11 @@
             This contest is final.
         </div>
     {% endif %}
+    {%- if is_granted('ROLE_ADMIN') and contest.isLocked -%}
+        <div class="alert alert-danger">
+            This contest is locked. Unlock it to edit contest data.
```
thijskh, line 32: As you're not actually editing yet, and this is supposed to be a 'normal operation situation', maybe `danger` is too high a class and warning or info would suffice?

webapp/templates/jury/problem.html.twig
```diff
@@ -117,8 +120,10 @@
 
     <div class="button-row">
         {%- if is_granted('ROLE_ADMIN') -%}
-            {{ button(path('jury_problem_edit', {'probId': problem.probid}), 'Edit', 'primary', 'edit') }}
-            {{ button(path('jury_problem_delete', {'probId': problem.probid}), 'Delete', 'danger', 'trash-alt', true) }}
+            {%- if not lockedProblem -%}
+                {{ button(path('jury_problem_edit', {'probId': problem.probid}), 'Edit', 'primary', 'edit') }}
+                {{ button(path('jury_problem_delete', {'probId': problem.probid}), 'Delete', 'danger', 'trash-alt', true) }}
+            {% endif %}
```
thijskh, line 126: Use disabled here instead of not showing?

thijskh (approved):

webapp/src/Controller/Jury/ContestController.php
meisterT: I actually have a slight preference to not show buttons you can't click.

webapp/templates/jury/problem.html.twig
meisterT: see above

webapp/src/Controller/Jury/ContestController.php
vmcj: I have the same preference, but:
In case that we auto lock contests (in the future) and someone doesn't know about this new feature, that person would never find the edit buttons but when you grey it out it's clear that its currently not possible (and we could even have a remark when you hover over it)

meisterT: That's a good point, let's keep it in mind when we add the auto lock configuration

PR #1644, opened 2022-07-25
Trigger delete event before actually deleting entity.

nickygerritsen:
@vmcj was this already in this (wrong) order before? Or maybe this did happen during your multi delete changes?


vmcj (approved):
If I broke it it was a while back when refactoring deleteEntit**y** to deleteEntit**ies** and moving the logic to separate functions. But your change makes sense and should not hurt.

PR #1646, opened 2022-07-26
Remove commented code

vmcj:
I think this was added when make_chroot was copied and refactored as
run_chroot.


nickygerritsen (approved):

meisterT (approved):

eldering:
I'm wondering whether in certain circumstances overwriting `/etc/resolv.conf` with the latest version from the host might make sense, but I don't remember, so fine to delete this.

Issue #1631, opened 2022-07-23
Implement new event feed format


nickygerritsen:
The spec that will be used at the world finals in Dhaka specifies a new, slightly different event feed format, where the `op` field is dropped in favour of sending the object (for creates and updates) or `null` (when deleting). Also it supports sending a whole collection of objects. Also the event `id` and therefor the `since_id` parameter has been dropped. This should still be replaced with something else, although it's not sure yet why.

We should implement this new format in two places:

- When writing the event feed. I was thinking about making a configuration flag that switches between the new and old format, so we can still do the old format for any tool that doesn't support the new one. I think we should not bother sending whole collections, to make the change easy.
- When reading event feeds we should support both formats. This should also not be too hard with a little refactor of the logic, since in the end they both contain the data to import.

For the dropped `since_id` parameter we should wait to see what we end up with in the end.

nickygerritsen:
This is now implemented fully.

PR #1647, opened 2022-07-26
Add option to use new event feed format. Part of #1631.

nickygerritsen:
CI will fail until we merge domjudge/domjudge-scripts#56, but this will add the new event feed:
* It adds an option to switch between the old and new format.
* It adds reconnection tokens as described in icpc/ccs-specs#108 which should soon be merged.
* It supports reading both versions of the feed as shadow.

This last thing required quite some changes to the reading of the feed, because we need to separate the event data from the event metadata (operation, optional event ID and endpoint type). But in the end I think this makes it better, since it is now a better separation.

### DIFF
--- a/etc/db-config.yaml
+++ b/etc/db-config.yaml
@@ -281,6 +281,15 @@
             public: true
             description: If set, enable teams and jury to send source code to this command. See admin manual for allowed arguments.
             docdescription: See :ref:`printing` for more information.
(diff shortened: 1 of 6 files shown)


vmcj:
I've made https://github.com/DOMjudge/domjudge/pull/1668 which fixes the pipeline.

nickygerritsen:
Superseded

PR #1648, opened 2022-07-29
Use chroot arch in mirror selection for ubuntu

vmcj:
It's already explained in the comment but:
- Ubuntu has only i386 & amd64 in the archive mirrors and the other
archs are in ports.ubuntu, when selecting where to download packages for
the chroot we should pick the mirror selected for the chroot arch
instead of the host arch as we might install a arm64 chroot in a amd64
machine (why is another question but the script allowes for this).


vmcj:
The next question is we should actually allow a chroot with a different arch in chroot compared to host?

nickygerritsen (approved):

PR #1652, opened 2022-08-04
Mention alternative constant used by GRUB

vmcj:
One of the two will most likely work. This was needed to help someone in Slack.

### DIFF
--- a/doc/manual/install-judgehost.rst
+++ b/doc/manual/install-judgehost.rst
@@ -157,7 +157,8 @@ distributions may be different (send us your feedback!).
 Edit grub config to add cgroup memory and swap accounting to the boot
 options. Edit ``/etc/default/grub`` and change the default
 commandline to
-``GRUB_CMDLINE_LINUX_DEFAULT="quiet cgroup_enable=memory swapaccount=1"``
+``GRUB_CMDLINE_LINUX_DEFAULT="quiet cgroup_enable=memory swapaccount=1"``, if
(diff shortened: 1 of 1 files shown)


doc/manual/install-judgehost.rst
```diff
@@ -157,7 +157,8 @@ distributions may be different (send us your feedback!).
 Edit grub config to add cgroup memory and swap accounting to the boot
 options. Edit ``/etc/default/grub`` and change the default
 commandline to
-``GRUB_CMDLINE_LINUX_DEFAULT="quiet cgroup_enable=memory swapaccount=1"``
+``GRUB_CMDLINE_LINUX_DEFAULT="quiet cgroup_enable=memory swapaccount=1"``, if
+available also do the same for ``GRUB_CMDLINE_LINUX``.
```
eldering, line 161: Do *both* really have to be changed if they exist, or does only one of these exist?

vmcj: According to https://askubuntu.com/a/780545 both can exist, GRUB_CMDLINE_LINUX will just always work but will also work when someone starts recovery mode. Maybe this user his setup was just to different so I'm fine with not merging this PR.

vmcj:
Looking at the grub file again (https://domjudge-org.slack.com/archives/CHHURFUM7/p1659600364948419?thread_ts=1659249244.654389&cid=CHHURFUM7) the user duplicated the GRUB_CMDLINE_LINUX_DEFAULT so I suspect that was the issue and this is not needed.

PR #1654, opened 2022-08-17
Store sessions in database, allowing to load balance web servers.

nickygerritsen:


webapp/config/packages/framework.yaml
```diff
@@ -4,7 +4,7 @@ framework:
     fragments: ~
 
     session:
-        handler_id: ~
+        handler_id: "%env(DATABASE_URL)%"
```
thijskh, line 7: The handler_id needs to be a class name, right? Not the dsn?

nickygerritsen: It can also be a DSN since https://github.com/symfony/symfony/commit/de9c61f423b52e022ffa8caa344a83029e7e60a5
And this is shorter than making a service that we pass the DSN to.

PR #1656, opened 2022-08-18
Add endpoint to get problem statement through API following CLICS spec. Closes #1655.

nickygerritsen:


webapp/src/Controller/API/ProblemController.php
```diff
@@ -371,6 +372,40 @@ public function singleAction(Request $request, string $id): Response
         return $this->renderData($request, $object);
     }
 
+    /**
+     * Get the statement for given problem for this contest.
+     * @throws NonUniqueResultException
+     * @Rest\Get("/{id}/statement")
+     * @OA\Response(
+     *     response="200",
+     *     description="Returns the given problem for this contest",
```
vmcj, line 381:
```suggestion
     *     description="Returns the given problem text for this contest",
```

webapp/src/Controller/API/ProblemController.php
```diff
@@ -371,6 +372,40 @@ public function singleAction(Request $request, string $id): Response
...
+            ->andWhere(sprintf('%s = :id', $this->getIdField()));
+
+        $contestProblems = $queryBuilder->getQuery()->getResult();
+
+        if (empty($contestProblems)) {
+            throw new NotFoundHttpException(sprintf('Problem with ID \'%s\' not found', $id));
+        }
+
+        /** @var ContestProblem $contestProblems */
+        $contestProblems = $contestProblems[0][0];
```
vmcj, line 400: Why don't we need the handle the case where you get multiple of these problems? We should consider adding a comment to explain why as its currently unclear to me.

webapp/tests/Unit/Controller/API/ProblemControllerTest.php
```diff
@@ -13,39 +13,60 @@ class ProblemControllerTest extends BaseTest
...
+            "rgb"        => "#87CEEB",
+            "color"      => "skyblue",
+            "statement" => [
+                [
+                    'href'     => 'contests/2/problems/1/statement',
+                    'mime'     => 'application/pdf',
+                    'filename' => 'statement.pdf',
+                ],
+            ],
+        ],
```
vmcj, line 69: We should consider adding a test to show that you cannot ask for the statement of a problem not in this contest.

webapp/src/Controller/API/ProblemController.php
```diff
@@ -371,6 +372,42 @@ public function singleAction(Request $request, string $id): Response
...
+            ->andWhere(sprintf('%s = :id', $this->getIdField()));
+
+        // Get the one result; we know it's only one since we filter on ID
+        $contestProblemData = $queryBuilder->getQuery()->getOneOrNullResult();
+
+        if (empty($contestProblemData)) {
+            throw new NotFoundHttpException(sprintf('Problem with ID \'%s\' not found', $id));
+        }
+
+        // The result contains the contest problem as well as the test data count. So get only the priblem
```
vmcj, line 400:
```suggestion
        // The result contains the contest problem as well as the test data count. So get only the problem
```

webapp/src/Controller/API/ProblemController.php
```diff
@@ -381,7 +418,7 @@ protected function getQueryBuilder(Request $request): QueryBuilder
             ->from(ContestProblem::class, 'cp')
             ->join('cp.problem', 'p')
             ->leftJoin('p.testcases', 'tc')
-            ->select('cp, partial p.{probid,externalid,name,timelimit,memlimit}, COUNT(tc.testcaseid) AS testdatacount')
+            ->select('cp, partial p.{probid,externalid,name,timelimit,memlimit,problemtext_type,problemtext}, COUNT(tc.testcaseid) AS testdatacount')
```
thijskh, line 421: So does this mean that now any problem related API call will download the full binary pdf from the database or for all problems for a list action) when this is only needed in the case that one requests one single statement explicitly?

nickygerritsen: Hmmm yes. let me change that to only loading the type

nickygerritsen: Note we do lease the feature that we expose the `hash`, but that's fine (and if we want this, we should store the hash as a column)

webapp/src/Serializer/ContestProblemVisitor.php
```diff
@@ -46,5 +59,28 @@ public function onPostSerialize(ObjectEvent $event): void
...
+            $property = new StaticPropertyMetadata(
+                ContestProblem::class,
+                'statement',
+                null
+            );
+            $visitor->visitProperty($property, [
+                [
+                    'href'     => $route,
+                    'mime'     => 'application/pdf',
+                    'filename' => 'statement.pdf',
```
meisterT, line 81: Can we use the letter here, say A.pdf, B.pdf and so on?

webapp/tests/Unit/Controller/API/ProblemControllerTest.php
nickygerritsen: I think I already replied in Slack, but this specific test is not for the statements, but more general: you should not be able to access a problem from another contest in some contest. I did add some other tests though.

webapp/src/Serializer/ContestProblemVisitor.php
nickygerritsen: Done. Note that the filename is mainly used for people who want to create a Contest Package from an API and the spec says files should be placed in `<endpoint>/<id>/<filename>` so in this case `problems/A/A.pdf`. But I think thats fine, even better.

webapp/src/Controller/API/ProblemController.php
```diff
@@ -371,6 +372,43 @@ public function singleAction(Request $request, string $id): Response
...
+            ->andWhere(sprintf('%s = :id', $this->getIdField()));
+
+        // Get the one result; we know it's only one since we filter on ID
+        $contestProblemData = $queryBuilder->getQuery()->getOneOrNullResult();
+
+        if (empty($contestProblemData)) {
+            throw new NotFoundHttpException(sprintf('Problem with ID \'%s\' not found', $id));
+        }
+
+        // The result contains the contest problem as well as the test data count. So get only the problem
```
meisterT, line 401: nit: make it more explicit that we don't want to leak any data here

meisterT (approved):

webapp/src/Controller/API/ProblemController.php
nickygerritsen: @vmcj i uh ik you offered to merge this. Could you also fix this nit? And the conflict

vmcj: Yes, will do it later today/tomorrow.

meisterT: Since I made the proposal I went ahead and pushed a nit fix. Will wait for CI to run and then merge.

PR #1657, opened 2022-08-18
Add filename to file objects in API in accordance with CLICS spec

nickygerritsen:
… spec.

### DIFF
--- a/webapp/src/Serializer/TeamAffiliationVisitor.php
+++ b/webapp/src/Serializer/TeamAffiliationVisitor.php
@@ -76,7 +76,6 @@ class TeamAffiliationVisitor implements EventSubscriberInterface
                     'mime'     => 'image/svg+xml',
                     'width'    => $viewBoxSize[0],
                     'height'   => $viewBoxSize[1],
-                    'hash'     => md5_file($flagFile),
                     'filename' => 'country-flag-' . $size . '.svg',
                 ];
             }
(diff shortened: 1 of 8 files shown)


webapp/src/Serializer/TeamAffiliationVisitor.php
```diff
@@ -99,7 +107,16 @@ public function onPostSerialize(ObjectEvent $event): void
...
                 null
             );
-            $visitor->visitProperty($property, [['href' => $route, 'mime' => mime_content_type($affiliationLogo), 'width' => $imageSize[0], 'height' => $imageSize[1]]]);
+            $visitor->visitProperty($property, [
+                [
+                    'href'     => $route,
+                    'mime'     => mime_content_type($affiliationLogo),
+                    'width'    => $imageSize[0],
+                    'height'   => $imageSize[1],
+                    'hash'     => md5_file($affiliationLogo),
```
meisterT, line 116: how expensive is it to do this again and again?

nickygerritsen: 🤷‍♂️ files should not be gigantic right? So I think it’s not too bad

nickygerritsen: Having said that, this field is optional so if you feel it might be (too) expensive, we can drop it.
Alternatively we can use Symfony's `CacheInterface` to cache it, so that we only calculate it once. However, that does mean people need to clear the Symfony cache when they change an image. We can automatically do this when you upload one in the jury interface, but not when manually placing a file in the directory. Unless we can come up with a cache key that identifies the file uniquely (mtime might work but not ideal).

So maybe we just drop it?

meisterT: I'm in favor of dropping

nickygerritsen: OK let me drop it while it's hot

nickygerritsen: Aaaannnnndddd it's gone.

PR #1658, opened 2022-08-18
Contest archive is now called contest package.

nickygerritsen:
New spec renames this to a better name, so we should as well.


vmcj (approved):
With grep I can not find other usages of this and CI seems happy.

PR #1666, opened 2022-09-02
API fixes

vmcj:
This fixes some of the issues reported here:
https://github.com/DOMjudge/domjudge/security/code-scanning

Regarding the null properties, we probably need to add this to most files but I first want to see if this decreases the findings to a more reasonable number.

### DIFF
--- a/webapp/src/Controller/API/RunController.php
+++ b/webapp/src/Controller/API/RunController.php
@@ -153,7 +153,11 @@ class RunController extends AbstractRestController implements QueryObjectTransfo
         }
 
         if ($request->query->has('limit')) {
-            $queryBuilder->setMaxResults($request->query->getInt('limit'));
+            $limit = $request->query->getInt('limit');
+            if ($limit<0) {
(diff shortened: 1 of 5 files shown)


webapp/src/Controller/API/RunController.php
```diff
@@ -152,8 +152,10 @@ protected function getQueryBuilder(Request $request): QueryBuilder
                 ->setParameter('judging_id', $request->query->get('judging_id'));
         }
 
-        if ($request->query->has('limit')) {
-            $queryBuilder->setMaxResults($request->query->getInt('limit'));
+	if ($request->query->has('limit')) {
+	    $limit = $request->query->getInt('limit');
+	    $limit = ($limit>0 ? $limit : 0);
```
meisterT, line 157: alternatively error out when `$limit < 0`?

vmcj: That might be even better it explains why you would get 0 results.

meisterT (approved):

PR #1668, opened 2022-09-04
Proposal to fix new event feed format

vmcj:


nickygerritsen (approved):

etc/db-config.yaml
```diff
@@ -281,6 +281,15 @@
             public: true
             description: If set, enable teams and jury to send source code to this command. See admin manual for allowed arguments.
             docdescription: See :ref:`printing` for more information.
+        -   name: event_feed_format
+            type: int
+            default_value: 0
+            public: false
+            description: "Format of the event feed to use"
+            options:
+                0: New format in use since the `2022-07` version # TODO: when spec is finalized, update version
```
eldering, line 290:
```suggestion
                0: New format in use since the `2022-07` version
```

etc/db-config.yaml
```diff
@@ -281,6 +281,15 @@
...
+                1: Legacy format in use until the `2020-03` version
```
eldering, line 291: I think it makes more sense to renumber them from `1` and incrementing, so that if we add a new version, it becomes `2` instead of having to do other weird numbering schemes.

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -287,7 +289,13 @@ protected function importFromCcsApi(array $eventsToSkip, ?callable $progressRepo
                         $event = $this->dj->jsonDecode($line);
                         $this->importEvent($event, $eventsToSkip);
 
-                        $this->setLastEvent($event['id']);
+                        $newFormat = !isset($event['op']);
```
eldering, line 292: Maybe use a slightly more explicit variable name (since this is used far away in the code), for example `$event_format_202207`?
In the future we might even want to read/detect the exact version and store it in a variable with the same constants https://github.com/DOMjudge/domjudge/pull/1668/files#diff-efa9e2dd8b5fccf408219fb062d6e943b7c49fa715163a4a600fbb23c941bae1R51 ...(?)

etc/db-config.yaml
vmcj: I propose to than swap the 0 & 1, datasource is already 0-indexed so I prefer to not have this configuration to be 1-indexed.

nickygerritsen: I’m fine with either

eldering: I think the more important question is whether the `0` value is a good default. Because, if there is no good default, then `0` can mean undefined. I guess in this case `0 = legacy format` seems reasonable to me, so fine with this. But in general, I wouldn't always stick with zero-based indexing, just because we've done it once before.

nickygerritsen (approved):

etc/db-config.yaml
```diff
@@ -281,6 +281,15 @@
             public: true
             description: If set, enable teams and jury to send source code to this command. See admin manual for allowed arguments.
             docdescription: See :ref:`printing` for more information.
+        -   name: event_feed_format
+            type: int
+            default_value: 0
+            public: false
```
nickygerritsen, line 287: @vmcj make 1 the default maybe?

nickygerritsen (approved):

etc/db-config.yaml
```diff
@@ -281,6 +281,15 @@
             public: true
             description: If set, enable teams and jury to send source code to this command. See admin manual for allowed arguments.
             docdescription: See :ref:`printing` for more information.
+        -   name: event_feed_format
+            type: int
+            default_value: 0
```
vmcj, line 286:
```suggestion
            default_value: 1
```

eldering (approved):

PR #1676, opened 2022-09-22
Implement (hardcoded) access endpoint.

nickygerritsen:


webapp/tests/Unit/Controller/API/AccessControllerTest.php
```diff
@@ -0,0 +1,163 @@
...
+        $access = $this->verifyApiJsonResponse('GET', $url, 200, 'admin');
+        self::assertArrayHasKey('capabilities', $access);
+        self::assertSame(
+            ['contest_start', 'team_submit', 'team_clar', 'proxy_submit', 'proxy_clar', 'admin_submit', 'admin_clar'],
+            $access['capabilities']
+        );
+
+        self::assertArrayHasKey('endpoints', $access);
+
+        $expectedEndpoints = [
```
eldering, line 25: Is there really value in having the same hardcoded list here again?

nickygerritsen: I’m not sure. I can check only for the endpoints maybe?

PR #1679, opened 2022-09-28
Drop py3 from extensions.

meisterT:


nickygerritsen:
Tests will fail, since we have test submissions ending in .py3 which will now not submit, so we need to change the number of successful submissions in the integration test somewhere.

vmcj:
So why don't we fail?

meisterT:
Because it is already rejected before: https://gitlab.com/DOMjudge/domjudge/-/jobs/3106311834#L3149

meisterT:
(and the number of "unverified" submissions is what we check and that doesn't change)

meisterT:
I have added another commit to delete the old test files with the old extensions.

vmcj (approved):

nickygerritsen (approved):

PR #1680, opened 2022-09-29
Don't give back judgings which are already marked as compiler error.

### LINKED ISSUE #1678: Judgehost information missing for judging with compile-error

When a submission gets a compiler error, there is no information in the jury interface about on which judgehost it was compiled, making it difficult to trace issues, as I need to find out manually on which judgehost it was compiled.

### Your environment

DOMjudge 8.0 release branch

meisterT:
Fixes #1678.

### DIFF
--- a/webapp/src/Controller/API/JudgehostController.php
+++ b/webapp/src/Controller/API/JudgehostController.php
@@ -185,7 +185,9 @@ class JudgehostController extends AbstractFOSRestController
             ->andWhere('j.judgingid = jt.jobid')
             ->andWhere('jr.runresult IS NULL')
             ->andWhere('j.valid = 1 OR r.valid = 1')
+            ->andWhere('j.result != :compiler_error')
             ->setParameter('hostname', $hostname)
+            ->setParameter('compiler_error', 'compiler-error')
             ->getQuery()
             ->getResult();
 


nickygerritsen (approved):

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -185,7 +185,9 @@ public function createJudgehostAction(Request $request): array
             ->andWhere('j.judgingid = jt.jobid')
             ->andWhere('jr.runresult IS NULL')
             ->andWhere('j.valid = 1 OR r.valid = 1')
+            ->andWhere('j.result != :compiler_error')
```
eldering, line 188: Shouldn't this just check for `result NOT IS NULL`?

meisterT: I assume you mean `j.result IS NULL` (so no `NOT`).

This doesn't work when we judge remaining testcases of a problem after the final verdict is known and crash while doing so.

eldering: Indeed, I meant `j.result IS NULL`. To be honest, I think that we have to think better about how to do this kind of logic: as a query or code, do we want to move certain logic away in common functions, completely rethinnk this, etc. as I feel that these are places that contain critical code, but are getting too complex.

PR #1682, opened 2022-09-30
Upgrade dependencies for Twig security issue

vmcj:


nickygerritsen (approved):

vmcj:
> If we merge this we should immediately text it on the current test setup

We could even use it before merging..,

vmcj:
I redeployed the analyst instance with this and that seems to work so I think this is fine.

PR #1684, opened 2022-10-01
The hidden property should be on teams since the 2021-11 spec.

nickygerritsen:


vmcj (approved):
We still make teams hidden by adding them to a hidden category right. So this is only on how those are exposed in the API.

nickygerritsen:
> We still make teams hidden by adding them to a hidden category right. So this is only on how those are exposed in the API.

Yes

PR #1686, opened 2022-10-02
Import problems1confirmation

vmcj:
In case that one has multiple instances the pausing is annoying. Default we will still keep the old behavior to confirm for every problem.

Also fixed a bug where we crash if someone doesn't want to import the problemset.yaml

### DIFF
--- a/misc-tools/import-contest
+++ b/misc-tools/import-contest
@@ -200,6 +200,7 @@ if os.path.exists('problems.yaml') or os.path.exists('problems.json') or os.path
                 problems = [problem['short-name']
                             for problem in problemData['problems']]
 
+        confirmIndividually = confirm("Confirm individually for every problem", False)
         for problem in problems:
             print(f'Preparing problem \'{problem}\'.')
             if os.path.exists(f'{problem}.zip'):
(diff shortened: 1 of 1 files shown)


misc-tools/import-contest
```diff
@@ -189,8 +189,11 @@ if os.path.exists('problems.yaml') or os.path.exists('problems.json') or os.path
         if not problems_imported:
             if os.path.exists('problems.yaml'):
                 problems_file = 'problems.yaml'
-            else:
+            elif os.path.exists('problems.json'):
                 problems_file = 'problems.json'
+            else:
+                print("Neither problems.{yaml,json} found.")
```
meisterT, line 195: This is basically where it crashed for me as reported in Slack. The condition in line 177 also allows `problemset.yaml`.

misc-tools/import-contest
```diff
@@ -212,6 +215,8 @@ if os.path.exists('problems.yaml') or os.path.exists('problems.json') or os.path
                 problems = [problem['short-name']
                             for problem in problemData['problems']]
 
+        askConfirmation = confirm("Confirm for every problem", True)
+        print(askConfirmation)
```
meisterT, line 219: debug print?

misc-tools/import-contest
```diff
@@ -212,6 +215,8 @@ if os.path.exists('problems.yaml') or os.path.exists('problems.json') or os.path
                 problems = [problem['short-name']
                             for problem in problemData['problems']]
 
+        askConfirmation = confirm("Confirm for every problem", True)
```
meisterT, line 218: nit: I think we should make this a little bit more clear, e.g.

`Confirm individually for every problem` and then set it default to `False` and rename the variable to `confirmIndividually`

meisterT:
I like this

misc-tools/import-contest
vmcj: Just to make it explicit:
Set it default to False would make that we just import everything unless someone really wants the old behavior.

misc-tools/import-contest
vmcj: Should I take it out so it can be addressed in another PR as this will now be discussed during CLICS?

nickygerritsen: I will fix this

meisterT (approved):

nickygerritsen (approved):

PR #1687, opened 2022-10-02
Don't check type for non-team accounts

vmcj:
Quoting the spec: "The team that this account is for. Required iff type is team.", in the analyst instance we can read our own account but we don't have a team there.

It might be that we should add the team/jury for analysts so they can actually submit but I assume we give these people personal accounts in that instance.


nickygerritsen:
IMHO we should just drop importing accounts

PR #1688, opened 2022-10-02
Ccs spec contest start guideline

vmcj:
I think we did the right thing but I think this implementation will lead to less discussions and is not harder to read and I think this interpretation is also the correct one.

### DIFF
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -673,7 +673,7 @@ class ExternalContestSourceService
 
         // The timezones are given in ISO 8601 and we only support names.
         // This is why we will use the platform default timezone and just verify it matches.
-        $startTime = $data['start_time'] === null ? null : new DateTime($data['start_time']);
+        $startTime = isset($data['start_time']) ? new DateTime($data['start_time']) : null;
         if ($startTime !== null) {
             // We prefer to use our default timezone, since that is a timezone name
             // The feed only has timezone offset, so we will only use it if the offset


webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -673,7 +673,7 @@ protected function validateAndUpdateContest(string $entityType, ?string $eventId
 
         // The timezones are given in ISO 8601 and we only support names.
         // This is why we will use the platform default timezone and just verify it matches.
-        $startTime = $data['start_time'] === null ? null : new DateTime($data['start_time']);
```
meisterT, line 676: What problem does this solve?

meisterT:
> in the analyst instance we can read our own account but we don't have a team there.

Wouldn't we import the same set of teams there? And perhaps add some more for the analysts?

webapp/src/Service/ExternalContestSourceService.php
nickygerritsen: If start time is omitted from the json instead of being null

nickygerritsen (approved):

webapp/src/Service/ExternalContestSourceService.php
meisterT: Do we know of a tool which does that?

nickygerritsen: We do. The cds removes null fields

vmcj:
> > in the analyst instance we can read our own account but we don't have a team there.
> 
> Wouldn't we import the same set of teams there? And perhaps add some more for the analysts?

Yes we would import all team teams.

And I don't know,
To me it makes sense that at one point we get all the WF teams, and that we add to every instance our accounts as admin.

How do we know which accounts to add for the jury is that from the config repo or do we add these manually? If the first does it matter if we also give analysts an account in the primary so we just have 1 set of accounts to import?

More interesting is what type of account would an analyst have as we currently don't do anything with those but it would make sense to than import them as just jury (so they get a team also to do test submissions)

But this was more about that there are now account types which will not have an team_id, and we could decide to generate one later (but only as internal..) for that specific account type. Also I agree with Nicky that importing accounts via the feed has no real good usecase.

PR #1689, opened 2022-10-02
Initialize testcase runs as 0

vmcj:
The runobject should have the run_time according to (https://ccs-specs.icpc.io/2022-07/contest_api?highlight=event-feed#runs) but as the example does already not list it this might be ambigious.

We don't allow runs to be 0, so initializing as 0.0 as that is the most common mistake is a good backup IMO.

### DIFF
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -1716,7 +1716,7 @@ class ExternalContestSourceService
         }
 
         $time    = Utils::toEpochFloat($data['time']);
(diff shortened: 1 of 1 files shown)


meisterT:
Why don't we allow runs to be 0?

nickygerritsen:
> Why don't we allow runs to be 0?

Michael meant null I think

meisterT:
Ah, I think I understand now. Then this change looks good to me

nickygerritsen:
The change makes sense indeed

nickygerritsen (approved):

PR #1693, opened 2022-10-03
Refactoring problem import: move validator handling into own method.

meisterT:


webapp/src/Service/ImportProblemService.php
```diff
@@ -971,4 +879,100 @@ public function importProblemFromRequest(Request $request, ?int $contestId = nul
             'messages'   => $allMessages,
         ];
     }
+
+    private function searchAndAddValidator(ZipArchive $zip, ?array &$messages, $externalId, $validationMode, ?Problem $problem)
```
nickygerritsen, line 883: Nit:
```suggestion
    private function searchAndAddValidator(ZipArchive $zip, ?array &$messages, $externalId, $validationMode, ?Problem $problem): void
```

nickygerritsen (approved):

PR #1694, opened 2022-10-03
Split problem import messages in info/warning/danger.

meisterT:
This is helpful for both web import and import via the API where the warnings previously were drowned in info messages.


vmcj (approved):
Looks fine to me, and makes sense to not flood these.

nickygerritsen (approved):
> Nicky liked this

PR #1696, opened 2022-10-03
fix: handle transient network errors with exponential backoff

### LINKED ISSUE #1673: Judgehost dies on `judgehosts/fetch-work` API call, while it is marked as `$failonerror=false`
[DOMjudge-devel mailinglist](https://www.domjudge.org/mailman/postorius/lists/domjudge-devel.domjudge.org/)

- We run domserver behind a proxy(AWS ALB in this case), and judgehost connects to the domserver through the proxy.
- Proxy can send erroneous responses instead of the actual domserver if any trouble happens.
- But we cannot guarantee that the responses originated from the proxy is JSON-formatted.
- `dj_json_decode` function in `lib/lib.wrapper.php` throws an error because the data from a proxy has a JSON Syntax Error.
- The error is never handled, thus judgehost is killed, while it passed `$failonerror` parameter as `false` while calling request function.

### Your environment
- DOMJudge 8.1.2 / Ubuntu 20.04.5 LTS / Multiple instances which run domserver behind apache2, and one AWS ALB that forwards requests to apache2s. (ALB - [Apache2 - domserver] * n)

- Run a proxy that responds with 5xx errors sometimes (e.g. 502 error)
- Run a judgehost with the randomly-failing proxy.

### Expected behaviour
- `judgehosts/fetch-work` API call failure should not exit(1).

### Actual behaviour
- judgehost dies.

- One small tip for any contest operator who runs their domserver behind an AWS ALB: review your Idle Connection Timeout of ALB(default=120s) and Keep-Alive Timeout (Apache default=5s), or your judgehosts die randomly. [posts](https://aws.amazon.com/premiumsupport/knowledge-center/elb-alb-troubleshoot-502-errors/)

Suckzoo:
Implemented two enhancements.
1. Retries curl_exec on transient errors (e.g. temporary network error or temporary HTTP error from a proxy in front of DOMservers)
2. Does not throw error on non-JSON error response. This fixes #1673.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -98,8 +98,7 @@ function close_curl_handles(): void
  * Every error except HTTP 401, 500 is considered transient.
  */
 $lastrequest = '';
-function request(
-    string $url, string $verb = 'GET', $data = '', bool $failonerror = true, int $jitter = 200, $factor = 2, $steps = 3)
+function request(string $url, string $verb = 'GET', $data = '', bool $failonerror = true)
 {
     global $endpoints, $endpointID, $lastrequest;
 
@@ -143,7 +142,7 @@ function request(
     $response = null;
     $errstr = null;
 
-    for ($trial = 1; $trial <= $steps; $trial++) {
+    for ($trial = 1; $trial <= BACKOFF_STEPS; $trial++) {
         $response = curl_exec($curl_handle);
         if ($response === false) {
             $errstr = "Error while executing curl $verb to url " . $url . ": " . curl_error($curl_handle);
@@ -153,7 +152,7 @@ function request(
                 $errstr = "Authentication failed (error $status) while contacting $url. " .
                     "Check credentials in restapi.secret.";
                 break;
-            } else if ($status != 200) {
+            } else if ($status < 200 || $status >= 300) {
                 $json = dj_json_try_decode($response);
                 if ($json !== null) {
                     $response = var_export($json, true);
@@ -170,14 +169,14 @@ function request(
                 break;
             }
         }
-        if ($trial == $steps) {
+        if ($trial == BACKOFF_STEPS) {
             $errstr = $errstr . " Retry limit reached.";
         } else {
             $warnstr = $errstr . " This request will be retried after about " .
-                $delay_in_ms . "ms... (" . $trial . "/" . $steps . ")";
+                $delay_in_ms . "ms... (" . $trial . "/" . BACKOFF_STEPS . ")";
             warning($warnstr);
-            usleep($delay_in_ms + random_int(0, $jitter));
-            $delay_in_ms = $delay_in_ms * $factor;
+            usleep($delay_in_ms + random_int(0, BACKOFF_JITTER_MS));
+            $delay_in_ms = $delay_in_ms * BACKOFF_FACTOR;
         }
     }
     if (!$succeeded) {
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -62,7 +62,7 @@ function read_credentials(): void
 }
 
 /**
- * @return resource|false
+ * @return CurlHandle|false
  */
 function setup_curl_handle(string $restuser, string $restpass)
 {
@@ -92,9 +92,14 @@ function close_curl_handles(): void
  * $data is the urlencoded data passed as GET or POST parameters.
  * When $failonerror is set to false, any error will be turned into a
  * warning and null is returned.
+ * This function retries request on transient network errors.
+ * To deal with the transient errors while avoiding overloads,
+ * this function uses exponential backoff algorithm.
+ * Every error except HTTP 401, 500 is considered transient.
  */
 $lastrequest = '';
-function request(string $url, string $verb = 'GET', $data = '', bool $failonerror = true)
+function request(
+    string $url, string $verb = 'GET', $data = '', bool $failonerror = true, int $jitter = 200, $factor = 2, $steps = 3)
 {
     global $endpoints, $endpointID, $lastrequest;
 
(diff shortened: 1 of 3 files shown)


judge/judgedaemon.main.php
```diff
@@ -92,9 +92,14 @@ function close_curl_handles(): void
...
  * warning and null is returned.
+ * This function retries request on transient network errors.
+ * To deal with the transient errors while avoiding overloads,
+ * this function uses exponential backoff algorithm.
+ * Every error except HTTP 401, 500 is considered transient.
  */
 $lastrequest = '';
-function request(string $url, string $verb = 'GET', $data = '', bool $failonerror = true)
+function request(
+    string $url, string $verb = 'GET', $data = '', bool $failonerror = true, int $jitter = 200, $factor = 2, $steps = 3)
```
eldering, line 102: I'm not sure it makes sense to add these as 3 extra optional parameters. In that case the parameter list becomes very long and unclear what each parameter means.

lib/lib.wrappers.php
```diff
@@ -14,6 +14,15 @@ function dj_json_decode(string $str)
     return json_decode($str, true, 512, JSON_THROW_ON_ERROR);
 }
 
+/**
+ * Try decode a JSON string with our preferred settings.
+ * Does not throw error, but errors can be obtained via json_last_error().
+ */
+function dj_json_try_decode(string $str)
+{
+    return json_decode($str, true, 512);
```
eldering, line 23: The `512` argument was only there to be able to specify the next, so might be better left out. Or maybe make `JSON_THROW_ON_ERROR` dependent on an optional parameter to `dj_json_decode`. @meisterT , @thijskh WDYT?

judge/judgedaemon.main.php
```diff
@@ -133,32 +138,49 @@ function request(string $url, string $verb = 'GET', $data = '', bool $failonerro
...
         } else {
-            warning($errstr);
-            $endpoints[$endpointID]['errorred'] = true;
-            return null;
+            $status = curl_getinfo($curl_handle, CURLINFO_HTTP_CODE);
+            if ($status == 401) {
+                $errstr = "Authentication failed (error $status) while contacting $url. " .
+                    "Check credentials in restapi.secret.";
+                break;
+            } else if ($status != 200) {
```
eldering, line 156: I think this should maybe any `2xx` code, and the same below for `5xx`.

judge/judgedaemon.main.php
Suckzoo: Thanks. Moved them into etc/judgehost-config.php.

lib/lib.wrappers.php
Suckzoo: Removed `512`. IMO, `dj_json_try_decode` seems clearer than `dj_json_try_decode($str, 0)`.

judge/judgedaemon.main.php
Suckzoo: Fixed code to accept any 2xx code.
For 5xx, I think 500 error only matters. 500 is an error returned from DOMJudge, that implies there is an actual error in contest system. Other 5xx error usually occurs when the proxy in front of the DOMserver is busy.

judge/judgedaemon.main.php
```diff
@@ -133,32 +137,49 @@ function request(string $url, string $verb = 'GET', $data = '', bool $failonerro
...
         } else {
-            warning($errstr);
-            $endpoints[$endpointID]['errorred'] = true;
-            return null;
+            $status = curl_getinfo($curl_handle, CURLINFO_HTTP_CODE);
+            if ($status == 401) {
+                $errstr = "Authentication failed (error $status) while contacting $url. " .
+                    "Check credentials in restapi.secret.";
+                break;
+            } else if ($status < 200 || $status >= 300) {
```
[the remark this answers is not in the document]
vmcj: nvm, the original code also didn't have this so I'll add this later.

judge/judgedaemon.main.php
```diff
@@ -133,32 +137,49 @@ function request(string $url, string $verb = 'GET', $data = '', bool $failonerro
         curl_setopt($curl_handle, CURLOPT_POSTFIELDS, null);
     }
 
-    $response = curl_exec($curl_handle);
-    if ($response === false) {
-        $errstr = "Error while executing curl $verb to url " . $url . ": " . curl_error($curl_handle);
-        if ($failonerror) {
-            error($errstr);
+    $delay_in_ms = 1000;
```
meisterT, line 140: nit: let's also move this constant to the other constants?

judge/judgedaemon.main.php
```diff
@@ -133,32 +137,49 @@ function request(string $url, string $verb = 'GET', $data = '', bool $failonerro
...
-                $response = var_export($json, true);
-            }
-            $errstr = "Error while executing curl $verb to url " . $url .
-                ": http status code: " . $status .
-                ", request size = " . strlen(print_r($data, true)) .
-                ", response: " . $response;
+            $warnstr = $errstr . " This request will be retried after about " .
+                $delay_in_ms . "ms... (" . $trial . "/" . BACKOFF_STEPS . ")";
+            warning($warnstr);
+            usleep($delay_in_ms + random_int(0, BACKOFF_JITTER_MS));
```
meisterT, line 178: This is now sleeping after the last unsuccessful attempt as well, isn't it?

lib/lib.wrappers.php
```diff
@@ -14,6 +14,15 @@ function dj_json_decode(string $str)
     return json_decode($str, true, 512, JSON_THROW_ON_ERROR);
 }
 
+/**
+ * Try decode a JSON string with our preferred settings.
```
eldering, line 18:
```suggestion
 * Try to decode a JSON string with our preferred settings.
```

judge/judgedaemon.main.php
Suckzoo: I don't think so. The function does not sleep after the last unsuccessful attempt because `$trial` is equal to `BACKOFF_STEPS` on the last repetition.

meisterT: You are right, I missed the condition above.

meisterT (approved):
Thanks for your contribution!

PR #1700, opened 2022-10-06
organization_id is not required

vmcj:
I think it is required according to https://ccs-specs.icpc.io/2022-07/ccs_system_requirements#contest-api (teams endpoint)
The team we failed on is the default DOMjudge team, which has no affiliation in the primary, the CDS removes all fields with a null value and we expected this to be there.

### DIFF
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -1012,7 +1012,7 @@ class ExternalContestSourceService
             'externalid'             => $data['id'],
             'name'                   => $data['formal_name'] ?? $data['name'],
             'display_name'           => $data['display_name'] ?? null,
-            'affiliation.externalid' => $data['organization_id'],
+            'affiliation.externalid' => $data['organization_id'] ?? null,
             'category.externalid'    => $data['group_ids'][0] ?? null,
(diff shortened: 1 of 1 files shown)


meisterT:
For completeness (and it shouldn't stop this change), I think in the past few finals we did *not* import the DOMjudge team but individual teams for everyone of us.

vmcj:
> For completeness (and it shouldn't stop this change), I think in the past few finals we did _not_ import the DOMjudge team but individual teams for everyone of us.

That is also what I discussed with Nicky, so I don't think we need this change so if there is any reason to close it just close it.

I only found it when I started the event-feed as admin instead of analytics to make sure that if the access policy would change we could handle that without real issues.

nickygerritsen:
> > For completeness (and it shouldn't stop this change), I think in the past few finals we did _not_ import the DOMjudge team but individual teams for everyone of us.
> 
> That is also what I discussed with Nicky, so I don't think we need this change so if there is any reason to close it just close it.
> 
> I only found it when I started the event-feed as admin instead of analytics to make sure that if the access policy would change we could handle that without real issues.

The change still makes sense in general imho

nickygerritsen (approved):

PR #1702, opened 2022-10-08
Add suggestion on OS and CPU architecture

vmcj:


nickygerritsen (approved):
Personally I agree with this change, unless someone wants to actively test 32 bit OS'es.

eldering (approved):

nickygerritsen (approved):

PR #1704, opened 2022-10-10
Limit ci minutes

vmcj:
We now run the tests on the main branch, so we could more often end up with a broken CI but thats better than not having CI for 50% of the month IMO.


nickygerritsen (approved):
This makes sense to me

PR #1706, opened 2022-10-13
Fixes for trivial PSR violations

vmcj:
No functional changes intended

### DIFF
--- a/webapp/src/Controller/API/AwardsController.php
+++ b/webapp/src/Controller/API/AwardsController.php
@@ -110,7 +110,7 @@ class AwardsController extends AbstractRestController
                 $group_winners[$team->getCategory()->getCategoryId()][] = $teamid;
                 $groups[$team->getCategory()->getCategoryid()] = $team->getCategory()->getName();
             }
-            foreach($scoreboard->getProblems() as $problem) {
+            foreach ($scoreboard->getProblems() as $problem) {
                 $probid = $problem->getApiId($this->eventLogService);
                 if ($scoreboard->solvedFirst($team, $problem)) {
                     $problem_winners[$probid][] = $teamid;
--- a/webapp/src/Controller/API/ClarificationController.php
+++ b/webapp/src/Controller/API/ClarificationController.php
@@ -290,8 +290,7 @@ class ClarificationController extends AbstractRestController
             ->setParameter('cid', $this->getContestId($request));
 
         if (!$this->dj->checkrole('api_reader') &&
-            !$this->dj->checkrole('judgehost'))
-        {
+            !$this->dj->checkrole('judgehost')) {
             if ($this->dj->checkrole('team')) {
                 $queryBuilder
                     ->andWhere('clar.sender = :team OR clar.recipient = :team OR (clar.sender IS NULL AND clar.recipient IS NULL)')
--- a/webapp/src/Controller/API/JudgehostController.php
+++ b/webapp/src/Controller/API/JudgehostController.php
@@ -1201,7 +1201,7 @@ class JudgehostController extends AbstractFOSRestController
      */
     public function getFilesAction(string $type, string $id): array
     {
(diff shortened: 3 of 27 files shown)


webapp/config/lib_load_db_secrets.php
```diff
@@ -0,0 +1,45 @@
+<?php declare(strict_types=1);
```
nickygerritsen, line 1: Why is this put in a separate file?

vmcj:
I think https://github.com/DOMjudge/domjudge/pull/1706/commits/a349a34962bebc22217e1f102df77e8c1814746c is the one which might need the most discussion. I don't think we ever change the file and this is one of the corner cases where the original file also made sense.

webapp/config/lib_load_db_secrets.php
vmcj: Reasoning is this:
```
--------------------------------------------------------------------------------------------------------FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE                                                           -------------------------------------------------------------------------------------------------------- 1 | WARNING | A file should declare new symbols (classes, functions, constants, etc.) and cause no        |         | other side effects, or it should execute logic with side effects, but should not do         |         | both. The first symbol is defined on line 9 and the first side effect is on line 51.     --------------------------------------------------------------------------------------------------------                                                                                                        Time: 56ms; Memory: 6MB
```

nickygerritsen:
> I think [a349a34](https://github.com/DOMjudge/domjudge/commit/a349a34962bebc22217e1f102df77e8c1814746c) is the one which might need the most discussion. I don't think we ever change the file and this is one of the corner cases where the original file also made sense.

What was the error we were getting? Would it make sense to ignore this file for checks? I'm not really happy with splitting it just to make CI happy.

vmcj:
> > I think [a349a34](https://github.com/DOMjudge/domjudge/commit/a349a34962bebc22217e1f102df77e8c1814746c) is the one which might need the most discussion. I don't think we ever change the file and this is one of the corner cases where the original file also made sense.
> 
> What was the error we were getting? Would it make sense to ignore this file for checks? I'm not really happy with splitting it just to make CI happy.

I've posted it inline in your original comment,

And we could ignore the file,
Or could remove the PSR1 rule, 
Or accept the annotation if someone ever changes something in that file as we only annotate the changed files.

nickygerritsen:
> > > I think [a349a34](https://github.com/DOMjudge/domjudge/commit/a349a34962bebc22217e1f102df77e8c1814746c) is the one which might need the most discussion. I don't think we ever change the file and this is one of the corner cases where the original file also made sense.
> > 
> > 
> > What was the error we were getting? Would it make sense to ignore this file for checks? I'm not really happy with splitting it just to make CI happy.
> 
> I've posted it inline in your original comment,
> 
> And we could ignore the file, Or could remove the PSR1 rule, Or accept the annotation if someone ever changes something in that file as we only annotate the changed files.

I would vote for any of these, since I don't think this rule will apply to any other file.

vmcj:
> > > > I think [a349a34](https://github.com/DOMjudge/domjudge/commit/a349a34962bebc22217e1f102df77e8c1814746c) is the one which might need the most discussion. I don't think we ever change the file and this is one of the corner cases where the original file also made sense.
> > > 
> > > 
> > > What was the error we were getting? Would it make sense to ignore this file for checks? I'm not really happy with splitting it just to make CI happy.
> > 
> > 
> > I've posted it inline in your original comment,
> > And we could ignore the file, Or could remove the PSR1 rule, Or accept the annotation if someone ever changes something in that file as we only annotate the changed files.
> 
> I would vote for any of these, since I don't think this rule will apply to any other file.

I think I agree, thats why I already explicitly mentioned that one. Will take it out to see if that also makes CI happy

nickygerritsen (approved):

PR #1707, opened 2022-10-14
PSR violations fixed.

vmcj:
The visibility is probably the one which needs most discussion,

Also this should make that we're very close that 80% of the work is now done and I suspect there will only be 1 PR to fix the length of the lines and 1 to fix spacing etc.

### DIFF
--- a/webapp/src/Controller/API/OrganizationController.php
+++ b/webapp/src/Controller/API/OrganizationController.php
@@ -42,8 +42,7 @@ class OrganizationController extends AbstractRestController
         ConfigurationService $config,
         EventLogService $eventLogService,
         AssetUpdateService $assetUpdater
-    )
-    {
+    ) {
         parent::__construct($entityManager, $dj, $config, $eventLogService);
         $this->assetUpdater = $assetUpdater;
     }
--- a/webapp/src/Utils/Scoreboard/Filter.php
+++ b/webapp/src/Utils/Scoreboard/Filter.php
@@ -34,10 +34,18 @@ class Filter
     public function getFilteredOn(): string
     {
         $filteredOn = [];
-        if ($this->affiliations) $filteredOn[] = 'affiliations';
-        if ($this->countries)    $filteredOn[] = 'countries';
-        if ($this->categories)   $filteredOn[] = 'categories';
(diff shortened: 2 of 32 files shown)


nickygerritsen (approved):
I like consistency so this makes me happy.

webapp/src/Controller/API/OrganizationController.php
```diff
@@ -42,8 +42,7 @@ public function __construct(
         ConfigurationService $config,
         EventLogService $eventLogService,
         AssetUpdateService $assetUpdater
-    )
-    {
+    ) {
```
eldering, line 45: This violates opening a function brace on a newline...

webapp/src/Utils/Scoreboard/Filter.php
```diff
@@ -34,10 +34,18 @@ public function __construct(
     public function getFilteredOn(): string
     {
         $filteredOn = [];
-        if ($this->affiliations) $filteredOn[] = 'affiliations';
-        if ($this->countries)    $filteredOn[] = 'countries';
-        if ($this->categories)   $filteredOn[] = 'categories';
-        if ($this->teams)        $filteredOn[] = 'teams';
+        if ($this->affiliations) {
+            $filteredOn[] = 'affiliations';
```
eldering, line 38: I think this reduces readability and makes it much harder to spot small typos between the different statements.

vmcj: I personally do prefer this, but I'm fine with changing it back.

webapp/src/Controller/API/OrganizationController.php
vmcj: I think that is indeed fixed in my last PR, if/when that one is accepted I'll verify if this is still the case and otherwise see why phpcs doesn't detect this.

webapp/src/Utils/Scoreboard/Filter.php
nickygerritsen: We could even rewrite this for even less typos:
```php
$fields = ['affiliations', 'countries', 'categories', 'teams'];
foreach ($fields as $field) {
  if ($this->$field) {
    $filteredOn[] = $field
  }
}
```

PR #1709, opened 2022-10-14
Fix runpipe propagating stdout/stdin closes to processes

edomora97:
If one of the child processes closes stdout, we need to close the stdin of the other process. This is important because, for example, the validator may send some input to the solution and signal the end of it with EOF. If we don't close the stdin of the solution it will deadlock waiting for more input.

This is done by calling close_fds() on EOF and making sure no file descriptor is leaked (otherwise read() won't return 0, but waits for more data).

This patch also adds a simple test for this scenario.


meisterT (approved):

edomora97:
I'm not sure it works under DJ. I'll need to test it a bit more.

meisterT (approved):
I did a lot of testing and think this works.

PR #1711, opened 2022-10-14
PSR fixes

vmcj:


meisterT:
Is there a simple command to bring all (changed) code into compliance?

vmcj:
> Is there a simple command to bring all (changed) code into compliance?

For other PRs? I use phpcbf for that and check if I agree with the new look. But I think now that most stuff is done that we have less annotations so it is easier to see for the one opening the PR if they agree with the comment.

nickygerritsen:
> > Is there a simple command to bring all (changed) code into compliance?
> 
> For other PRs? I use phpcbf for that and check if I agree with the new look. But I think now that most stuff is done that we have less annotations so it is easier to see for the one opening the PR if they agree with the comment.

You can set up such a tool as pre commit hook even to check the files you modified

meisterT:
What `phpcbf` command line do you use?

vmcj:
> What `phpcbf` command line do you use?

phpcbf --standard=.github/workflowscripts/phpruleset.xml webapp/tests/ webapp/src/ webapp/config webapp/public/index.php

PR #1716, opened 2022-10-17
Highlight on the submission site when a judging is outdated because

meisterT:
compile/run/compare configs have been changed.

One could imagine adding such a check for a whole problem or a whole contest as well as on the rejudging page so you could easily rejudge everything that needs to be done.

Example:
image

### DIFF
--- a/webapp/src/Controller/Jury/SubmissionController.php
+++ b/webapp/src/Controller/Jury/SubmissionController.php
@@ -524,7 +524,7 @@ class SubmissionController extends BaseController
                         $header = "Some parameters have changed since the judging was created, consider rejudging.\n\n";
                     } else {
                         $type = 'warning';
-                        $header = "Some parameters have changed since the judging was created, but this judging is no longer valid.\n\n";
+                        $header = "Some parameters have changed since the judging was created, but this judging has been superseded, please verify if that needs a rejudging.\n\n";
                     }
                     $this->addFlash($type, $header . implode("\n", $errors));
                 }


webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -500,6 +500,31 @@ public function viewAction(Request $request, int $submitId): Response
                 'after' => 15,
                 'url' => $this->generateUrl('jury_submission', ['submitId' => $submission->getSubmitid()]),
             ];
+        } else {
```
nickygerritsen, line 503: Should this not be `elseif $selectJudging != $currentJudging` (pseudo)?

nickygerritsen (approved):
I like it

webapp/src/Controller/Jury/SubmissionController.php
meisterT: Actually, I think it is still useful. But I downgraded the severity a little bit and changed the message:
image

WDYT?

nickygerritsen:
I like it!

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -500,6 +500,36 @@ public function viewAction(Request $request, int $submitId): Response
...
+                    $this->dj->getCompareConfig($contestProblem),
+                    $judgeTask->getCompareConfig(),
+                    $errors);
+                if (!empty($errors)) {
+                    if ($selectedJudging->getValid()) {
+                        $type = 'danger';
+                        $header = "Some parameters have changed since the judging was created, consider rejudging.\n\n";
+                    } else {
+                        $type = 'warning';
+                        $header = "Some parameters have changed since the judging was created, but this judging is no longer valid.\n\n";
```
vmcj, line 527:
```suggestion
                        $header = "Some parameters have changed since the judging was created, but this judging has been superseded, please verify if that needs a rejudging.\n\n";
```

So probably another wording is better, but I think this is unclear for first time users which are also the people who might change the settings during the contest.

PR #1717, opened 2022-10-18
Move jobs to github actions

vmcj:
These jobs are just for learning the syntax etc. the most benefit is to be had from the unit tests and integration jobs.


vmcj:
Before merging I'll takeout [24c1672](https://github.com/DOMjudge/domjudge/pull/1717/commits/24c1672188cfe27150fefc6ef1bf5b883af1a4b7) but after that CI will fail on main again (but hopefully with less jobs)

meisterT:
What's the free quota here? Higher than on gitlab?

vmcj:
> What's the free quota here? Higher than on gitlab?

For public project it seems to have a cost of 0, so the only disadvantage is the more difficult syntax.

vmcj:
Also I think we should think about a better naming for the underlying folder:

So do we want:
./github/{workflows,workflowscripts}
or
./github/{workflows,jobscripts,jobdata}
or
./github/{workflows,job/{scripts,data} }

nickygerritsen:
> Also I think we should think about a better naming for the underlying folder:
> 
> So do we want: ./github/{workflows,workflowscripts} or ./github/{workflows,jobscripts,jobdata} or ./github/{workflows,job/{scripts,data} }

Any of the last two are fine by me

.github/jobscripts/detect_dump.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+set -eux
```
meisterT, line 3: And consider adding `set -o pipefail`

.github/jobscripts/detect_dump.sh
```diff
@@ -0,0 +1,13 @@
...
+
+OUT=$(find ./ -name ".git*" -type d -prune -o -name "lib" -prune -o -name "var" -prune -o -name "bundles" -prune -o -name "cache" -type d -prune -o -name "ace" -type d -prune -o -type f -print0 | xargs -0 grep --color "dump(" | grep -v "Yaml::dump(") || true
```
meisterT, line 5: Can we make this more readable like this?

```
find ./ -name ".git*" -type d \
  -prune -o -name "lib" \
  -prune -o -name "var" \
  -prune -o -name "bundles" \
...
```

.github/jobscripts/detect_dump.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# Show detected debug statements
+echo "$OUT"
```
meisterT, line 8: nit: want to redirect this to stderr?

.github/workflows/codestyle.yml
```diff
@@ -0,0 +1,42 @@
+name: Codestandard
+on:
+  push:
+  #  branches:
```
meisterT, line 4: is this intentionally commented out?

meisterT:
> > Also I think we should think about a better naming for the underlying folder:
> > So do we want: ./github/{workflows,workflowscripts} or ./github/{workflows,jobscripts,jobdata} or ./github/{workflows,job/{scripts,data} }
> 
> Any of the last two are fine by me

What data do we expect?

vmcj:
> > > Also I think we should think about a better naming for the underlying folder:
> > > So do we want: ./github/{workflows,workflowscripts} or ./github/{workflows,jobscripts,jobdata} or ./github/{workflows,job/{scripts,data} }
> > 
> > 
> > Any of the last two are fine by me
> 
> What data do we expect?

In this case:
gitlab/codespellignorefiles.txt
gitlab/codespellignorewords.txt

And probably for future PRs:
gitlab/default-nginx
gitlab/assert.bash
gitlab/chroot.bats

meisterT:
> > > > Also I think we should think about a better naming for the underlying folder:
> > > > So do we want: ./github/{workflows,workflowscripts} or ./github/{workflows,jobscripts,jobdata} or ./github/{workflows,job/{scripts,data} }
> > > 
> > > 
> > > Any of the last two are fine by me
> > 
> > 
> > What data do we expect?
> 
> In this case: gitlab/codespellignorefiles.txt gitlab/codespellignorewords.txt
> 
> And probably for future PRs: gitlab/default-nginx gitlab/assert.bash gitlab/chroot.bats

What about:
```
- .github/workflows
- .github/jobs
- .github/jobs/data
```
?

.github/jobscripts/detect_dump.sh
meisterT: I don't see any downside from switching to bash.

While there is (limited) value in staying POSIX compliant for the scripts that come with DOMjudge, I don't see the value for github actions.

.github/jobscripts/detect_dump.sh
vmcj: I've added the newlines

meisterT (approved):
squash it and we're good to go

nickygerritsen (approved):

PR #1725, opened 2022-10-23
Disable the use of #pramga

yzlai:
### Disable the use of #pramga to tell the compiler to change compile flags.

The following directive will override `-O2`, which I think should be disabled.
`#pragma GCC optimize ("Ofast")`

### DIFF
--- a/sql/files/defaultdata/c/run
+++ b/sql/files/defaultdata/c/run
@@ -6,6 +6,9 @@
 DEST="$1" ; shift
 MEMLIMIT="$1" ; shift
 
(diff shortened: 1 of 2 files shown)


meisterT:
Hi @yzlai, thanks for your PR, but I don't think that we want to change this in the default scripts that come with DOMjudge. You can change these in your local installation though.

Note that the way you do it with `sed` could be made more readable if you pick another char to separate the parts:
```
sed -i 's|^#pragma|// #pragma|i' "$@"
```

PR #1730, opened 2022-10-28
Use entity ID badges for team ID's everywhere so we can show the external as well as the internal ID and make it look better.

nickygerritsen:
Also use the badges on the submission detail page for user, contest, problem and language.

With thanks to @meisterT .

Screenshots on request in Slack since data is officially still secret.

### DIFF
--- a/webapp/templates/jury/submission.html.twig
+++ b/webapp/templates/jury/submission.html.twig
@@ -95,7 +95,7 @@
         <span>
             <i class="fas fa-users" title="Team:"></i>
             <a href="{{ path('jury_team', {teamId: submission.team.teamid, cid: submission.contest.cid}) }}">
-                {{ submission.team.effectiveName }} (t{{ submission.team.teamid }})
+                {{ submission.team.effectiveName }} {{ submission.team | entityIdBadge('t') }}
             </a>
         </span>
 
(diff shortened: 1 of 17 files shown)


webapp/templates/jury/submission.html.twig
```diff
@@ -123,13 +124,15 @@
...
                 {% endif %}
+                {{ submission.problem | entityIdBadge('p') }}
             </a>
         </span>
 
         <span>
             <i class="fas fa-comments" title="Language:"></i>
             <a href="{{ path('jury_language', {'langId': submission.language.langid}) }}">
                 {{ submission.language.name }}
+                {{ submission.language | entityIdBadge }}
```
tuupke, line 135: Maybe for consistency sake we should prefix the language id too

tuupke (approved):
Looks good, one very very minor nit. The language does not having a prefix (I remember seeing it in another place, but now cannot seem to find me) bugs me. If we can agree on what to prefix it with it might be nice to include it now.

webapp/templates/jury/submission.html.twig
nickygerritsen: We never did this since language id’s are strings. lpy3 looks weird imho

tuupke: I know, that is why it is a very very minor nit.

meisterT (approved):
I think the unit test fails until you rebase this on current head of main

nickygerritsen:
> I think the unit test fails until you rebase this on current head of main

It also found one actual stupid thing I did. Should now be fixed so if CI passes I will squash and merge.

meisterT (approved):

PR #1731, opened 2022-10-29
Add first script to configure DOMjudge programmatically.

meisterT:


misc-tools/configure-domjudge
```diff
@@ -0,0 +1,137 @@
...
+    usage()
+
+api_url = sys.argv[1]
+
+user_data = do_api_request('user')
+if 'admin' not in user_data['roles']:
+    print('Your user does not have the \'admin\' role, can not import.')
+    exit(1)
+
+if os.path.exists('config.json'):
```
nickygerritsen, line 103: Could we (also) do yaml? Maybe in the future

misc-tools/configure-domjudge
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/env python3
+
+'''
+TODO
+(replace main with the DOMjudge major.minor version if you are running a released version,
```
nickygerritsen, line 5: What main?

misc-tools/configure-domjudge
meisterT: Yes, should be trivial.

I started with JSON since it's easy to retrieve (and change locally to test with).

misc-tools/configure-domjudge
meisterT: After I resolved the Todo the line above it will be clear.

misc-tools/configure-domjudge
nickygerritsen: 👍

nickygerritsen (approved):
Nice

vmcj (approved):

vmcj:
I think this is besides a start to the issue also the correct fix to close it.

meisterT:
> I think this is besides a start to the issue also the correct fix to close it.

I actually plan to also configure languages (and in an even later step, probably after my version branch is in merge-able form) to even allow configuring versions and compiler command lines.

PR #1734, opened 2022-10-29
Fix some typos and refer to the draft CCS requirements.

eldering:


meisterT (approved):
thanks!

doc/manual/import.rst
```diff
@@ -425,11 +425,11 @@ You can retrieve the current configuration via
...
     http --check-status --pretty=format "<API_URL>/config" > config.json
 
-For your convenience, we added a script to update configuration via from a file
+For your convenience, we added a script to update configuration from a file
 called `config.json` in your current directory::
 
     misc-tools/configure-domjudge <API_URL>
 
-.. _CCS specification: https://ccs-specs.icpc.io/2021-11/ccs_system_requirements#appendix-file-formats
+.. _CCS specification: https://ccs-specs.icpc.io/draft/ccs_system_requirements#appendix-file-formats
```
nickygerritsen, line 433: In our api op main we use the 2022-08 version since that makes more sense than draft. So should we here as well?

eldering: Ok, but these files don't follow the API versioning, right? (Which is BTW maybe a small issue with the current versioning of the CCS specs repo). So I'd lean towards using `draft` since that is less likely to be outdated over time. But if you think that we're then really pointing to incorrect information, I'm ok changing it to `2022-07`.

nickygerritsen: I think we are mostly closer to the release than draft for the spec like 90% of the time so I would vote for it yes

meisterT: Perhaps we should mention that you can configure the version in the configuration option?

nickygerritsen: That’s only for the event feed right

eldering: I'm not sure I understand your comment @meisterT. Note that this is not about the Contest API specification, but the CCS requirements. Although we've versioned the whole repo/docs at https://ccs-specs.icpc.io/, this was effectively only applied to the Contest API specification part, while this links to the CCS requirement.
I'll change it to `2022-07` in the mean time.

meisterT: You are both right, ignore me :-)

PR #1735, opened 2022-10-30
Do not reset color of contestproblem during problem import since it could've been set during problemset import already.

nickygerritsen:
Also add a comment as to why we do not (re)set specific properties.

After checking the logic here: https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Service/ImportExportService.php#L227-L227, it seems indeed only the color was a property we DID (re)set during problem import but shouldn't. There are some more properties (shortname of the contestproblem or timelimit of the problem) that we should not reset since they can be set in the problemset import, so I added a comment to make sure we do not do that in the future.

### DIFF
--- a/webapp/src/Service/ImportProblemService.php
+++ b/webapp/src/Service/ImportProblemService.php
@@ -213,7 +213,12 @@ class ImportProblemService
             }
         }
 
-        // Reset problem settings that might not appear in the ZIP back to default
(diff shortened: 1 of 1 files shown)


meisterT (approved):
I am sorry, but you need to rebase this on top of main because I modified this line in my PR.

PR #1738, opened 2022-10-30
Also install `dj_utils.py` after splitting out code in a3daff2892fb08…

meisterT:
…d1e3f26e78675906c04726cdac.

### DIFF
--- a/misc-tools/Makefile
+++ b/misc-tools/Makefile
@@ -25,6 +25,7 @@ $(SUBST_FILES): %: %.in $(TOPDIR)/paths.mk
 install-domserver:
 	$(INSTALL_PROG) -t $(DESTDIR)$(domserver_bindir) $(SUBST_DOMSERVER)
 	$(INSTALL_PROG) -t $(DESTDIR)$(domserver_bindir) import-contest
+	$(INSTALL_PROG) -t $(DESTDIR)$(domserver_bindir) dj_utils.py
 
 install-judgehost:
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

misc-tools/Makefile
```diff
@@ -25,6 +25,7 @@ $(SUBST_FILES): %: %.in $(TOPDIR)/paths.mk
 install-domserver:
 	$(INSTALL_PROG) -t $(DESTDIR)$(domserver_bindir) $(SUBST_DOMSERVER)
 	$(INSTALL_PROG) -t $(DESTDIR)$(domserver_bindir) import-contest
+	$(INSTALL_PROG) -t $(DESTDIR)$(domserver_bindir) dj_utils.py
```
eldering, line 28: I think this should be installed in `libdir` instead, as this file will not be called directly.

meisterT: How would that import then work in python?

eldering: Good question, I don't know. How does this work for other (standard) libraries?

eldering: I guess with something like https://stackoverflow.com/questions/135035/python-library-path inject the `domserver_libdir` path into each python binary using this library.

meisterT: It may be that one can modify the `sys.path`, if that works it would require us to make the `import-contest` and the `configure-domjudge` scripts template files.

eldering: See https://github.com/DOMjudge/domjudge/pull/1745

PR #1739, opened 2022-10-30
Use problem badge in submission list.

meisterT:
Before: 

After:


nickygerritsen (approved):
More color is more better

PR #1742, opened 2022-10-30
Fix incorrect counts on award render due to non-award participants.

shuibinlong:
I found a bug while using awards, as shown below. We set 52 gold medals, however, since there are 2 teams were non-award participants, they also in the top 52, so the gold medal should be sent to the top 54, but it isn't!

a

b

c

After reading the code, I think I need to use an additional award counter as a variable instead of using `score.rank` directly in `scoreboard_table.html.twig`.

### DIFF
--- a/webapp/templates/partials/scoreboard_table.html.twig
+++ b/webapp/templates/partials/scoreboard_table.html.twig
@@ -104,6 +104,7 @@
     {% set previousSortOrder = -1 %}
     {% set previousTeam = null %}
     {% set backgroundColors = {"#FFFFFF": 1} %}
+    {% set medalCount = 0 %}
     {% for score in scores %}
         {% set classes = [] %}
         {% if score.team.category.sortorder != previousSortOrder %}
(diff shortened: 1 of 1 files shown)


webapp/templates/partials/scoreboard_table.html.twig
```diff
@@ -119,11 +120,12 @@
         {# process medal color #}
         {% set medalColor = '' %}
         {% if showLegends and medalsEnabled and score.team.category in medalCategories and score.numPoints > 0 %}
-            {% if score.rank <= contest.goldMedals %}
+            {% set medalCount = medalCount + 1 %}
```
nickygerritsen, line 123: Should this line not also be added to two elseifs below?

shuibinlong: In the following gold, silver and bronze ifelse(line 122/124/126) is need to be added, so I mentioned this line of code to the front. I have tested this code and it works.

shuibinlong: Its logic is that as long as the award conditions are met, add 1 to the `medalCount`, and then determine which award should be put.

nickygerritsen (approved):

nickygerritsen:
Thanks. It makes sense to me

PR #1745, opened 2022-10-30
Install misc-tools/dj_utils.py as library.

eldering:
Also add inplace-(un)install to recursive Makefile rules, and install `import-contest` on domserver, not judgehost.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@
 export TOPDIR = $(shell pwd)
 
 REC_TARGETS=build domserver install-domserver judgehost install-judgehost \
-            docs install-docs
+            docs install-docs inplace-install inplace-uninstall
 
 # Global Makefile definitions
 include $(TOPDIR)/Makefile.global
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -201,7 +203,7 @@ webapp/.env.local:
 # symlinks where necessary to let it work from the source tree.
 # This stuff is a hack!
 maintainer-install: inplace-install
-inplace-install: build domserver-create-dirs judgehost-create-dirs
+inplace-install-l: build domserver-create-dirs judgehost-create-dirs
```
meisterT, line 206: what does this `-l` suffix mean?

nickygerritsen (approved):

Makefile
eldering: It stands for (directory) local. The recursive targets all call their `*-l` variant after recursing, see https://www.domjudge.org/docs/manual/8.0/develop.html#makefile-structure

Since I changed `inplace-install` to a recursive target, the original rule here is conflicting and that's why I renamed it to the local one, which is now called automatically.

meisterT: Thanks for explaining and fixing!

meisterT (approved):

PR #1747, opened 2022-10-31
Add a check for .netrc when loading example problems and offer to write it.

nickygerritsen:
I wish this file was Python 😛 .

I hope this is all POSIX compliant and a logical way to do it.


vmcj:
In case that admin password does not exist in the file, should we not also run this step or output the command how someone can do that?

nickygerritsen:
> In case that admin password does not exist in the file, should we not also run this step or output the command how someone can do that?

We could add a check that the file doesn't exist? There is no use in creating the file if it doesn't exist IMHO, then people should just create the .netrc themself.

vmcj:
Although I like this also I meant something else:
 if ! -f ntrc:
   -> Create it how you had it
 else:
   -> echo "nitric already found, please run `echo '$machine admin $password' >> .netrc`"
   || Just do it ourselves if we feel like we're allowed to write to people their files

nickygerritsen:
> Although I like this also I meant something else:
> if ! -f ntrc:
> -> Create it how you had it
> else:
> -> echo "nitric already found, please run `echo '$machine admin $password' >> .netrc`"
> || Just do it ourselves if we feel like we're allowed to write to people their files

I don't like this. This is more if a help for people to not get fully stuck. If you install DOMjudge on a machine that already has a .netrc I hope you know what you are doing... Maybe we improve the error mesage to tell you you need to fix your .netrc. There are so many ways you can still break it (one of them being: you changed the admin password and didn't update the initial_admin_password.secret because why would you) that I don't think it's worth it to try to fix all random things that people can break. Don't most people install DOMjudge on a clean machine anyway?

meisterT (approved):

vmcj (approved):

PR #1749, opened 2022-10-31
Use problem badge also on main team screen.

meisterT:
Example:

image

### DIFF
--- a/webapp/templates/team/partials/clarification_list.html.twig
+++ b/webapp/templates/team/partials/clarification_list.html.twig
@@ -44,7 +44,7 @@
                 <td>
                     <a data-ajax-modal data-ajax-modal-after="markSeen" href="{{ link }}">
                         {%- if clarification.problem -%}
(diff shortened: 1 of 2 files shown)


webapp/templates/team/partials/clarification_list.html.twig
```diff
@@ -44,7 +44,7 @@
                 <td>
                     <a data-ajax-modal data-ajax-modal-after="markSeen" href="{{ link }}">
                         {%- if clarification.problem -%}
-                            problem {{ clarification.problem.contestProblems.first.shortname -}}
+                            problem {{ clarification.problem.contestProblems.first | problemBadge -}}
```
nickygerritsen, line 47: We should dot his on the jury clar list as well then

nickygerritsen (approved):

webapp/templates/team/partials/clarification_list.html.twig
meisterT: True, I'll look into it after lunch

PR #1751, opened 2022-10-31
Allow test-fork-parent-exit.c to be a timelimit as well.

meisterT:
Whether this is actually time limit or wrong answer depends on the gitlab runner or its host.


nickygerritsen (approved):

eldering:
I think that if this code gets a timelimit, it is better to replace the `while (1) {};` loop with a sleep or so, so that it doesn't overload the CPU.

meisterT:
Good point, let's see what CI thinks of https://github.com/DOMjudge/domjudge/pull/1752

PR #1753, opened 2022-10-31
Set activate time to current time when importing via API.

meisterT:
If you import a contest, e.g. via `import-contest`, you typically want to play around right away with it.


meisterT:
I think @vmcj mentioned this recently.

vmcj:
> I think @vmcj mentioned this recently.

Maarten Sijm did.

nickygerritsen (approved):

vmcj (approved):
In the future we should also just have a way to set the property.

meisterT:
> In the future we should also just have a way to set the property.

Yeah, in the `contest.yaml` we could have a key for it.

PR #1755, opened 2022-11-01
Dont return null as judgehost when judgehost is unknown

vmcj:
As this is not a fix to the docs but to the logic done in another PR.

Fixes: https://github.com/DOMjudge/domjudge/security/code-scanning/435


nickygerritsen (approved):

vmcj:
None of the failing unit tests are because of this change.

PR #1756, opened 2022-11-02
Fix mixing string

vmcj:
Fixes: https://github.com/DOMjudge/domjudge/security/code-scanning/475

@meisterT I assume this was introduces by accident?

### DIFF
--- a/example_problems/hello/submissions/multiple/extra-binary/abc.sh
+++ b/example_problems/hello/submissions/multiple/extra-binary/abc.sh
@@ -27,7 +27,7 @@ if [ ! -x $BC ]; then
 	fi
 fi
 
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

meisterT:
It doesn't really matter, right?

We probably want to exclude these example submissions from syntax checking anyway.

Teams can do whatever (if it compiles).

vmcj:
> It doesn't really matter, right?
> 
> We probably want to exclude these example submissions from syntax checking anyway.
> 
> Teams can do whatever (if it compiles).

No it should not matter, but its a small fix to make sure we keep coverage when people move folders around.

eldering (approved):
This is indeed more correct, but in practice irrelevant.

PR #1763, opened 2022-11-04
Add Dhaka banner; also restrict its width so it doesn't get too large…

meisterT:
… (and still looks good on Mac/Retina).


nickygerritsen (approved):
Will you later also add logos?

meisterT:
> Will you later also add logos?

I think you already re-sized them (and my internet connection is spotty), so I will leave this for you.

PR #1765, opened 2022-11-05
Add support for a contest-wide scoreboard warning message

thijskh:
Useful when specific contests are used for testing and to avoid any confusion about their officialdom.

### DIFF
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -66,6 +66,9 @@ class ImportExportService
             'start-time' => Utils::absTime($contest->getStarttime(), true),
             'duration' => Utils::relTime($contest->getContestTime((float)$contest->getEndtime())),
         ];
+        if ($warnMsg = $contest->getWarningMessage()) {
+            $data['warningMessage'] = $warnMsg;
+        }
         if ($contest->getFreezetime() !== null) {
             $data['scoreboard-freeze-duration'] = Utils::relTime(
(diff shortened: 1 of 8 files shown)


thijskh:
@meisterT I moved the message down a few lines in the template vs your commit dd381242598462b911826cf98cda57d842b2385c to avoid the extra if for current-contest

meisterT (approved):

meisterT:
We probably want to add an attribute to `contest.yaml` and import it via the API if set.

webapp/src/Service/ImportExportService.php
```diff
@@ -66,6 +66,9 @@ public function getContestYamlData(Contest $contest): array
             'start-time' => Utils::absTime($contest->getStarttime(), true),
             'duration' => Utils::relTime($contest->getContestTime((float)$contest->getEndtime())),
         ];
+        if ($warnMsg = $contest->getWarningMessage()) {
+            $data['warningMessage'] = $warnMsg;
```
meisterT, line 70: nit: `warning-message`? so it matches the other attributes better?

meisterT:
If you rebase on main the unit test will pass.

thijskh:
Already done

thijskh:
I can see the advantage. If we’d do that we need some way to keep it up to date/can we install it with composer?

Of course adding this by default adds a high shoot in foot factor and also allows admins to delete their own auditlog lines. Something to consider. Maybe a feature toggle can help

meisterT:
Is this the right PR for this comment?

PR #1768, opened 2022-11-06
Always end up executable files with newlines.

meisterT:


webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -417,6 +417,11 @@ public function editFilesAction(Request $request, string $execId): Response
                     return $this->redirectToRoute('jury_executable', ['execId' => $executable->getExecid()]);
                 }
                 $newContent = str_replace("\r\n", "\n", $submittedData['source' . $idx]);
+                if (!str_ends_with($newContent, "\n")) {
```
nickygerritsen, line 420: This only works in php8+: https://www.php.net/manual/en/function.str-ends-with.php

nickygerritsen (changes requested):

webapp/src/Controller/Jury/ExecutableController.php
meisterT: Done

nickygerritsen (approved):

PR #1775, opened 2022-11-07
Add Adminer.php (for easy db fiddling)

ubergeek42:
Targeting the packaging branch so packages get built/the diff is not enormous.

This adds a link to Adminer in the admin dashboard. Adminer is a phpmyadmin type thing, basically giving a nice gui db editor without having to do a whole lot. https://www.adminer.org/

Authentication is controlled by symfony, and only available to people with the ADMIN role.

This might be worth upstreaming (?), but probably needs to be reorganized/rewritten to be a bit more generic.


nickygerritsen (approved):
I like putting this upstream

thijskh:
I can see the advantage. If we’d do that we need some way to keep it up to date/can we install it with composer?

Of course adding this by default adds a high shoot in foot factor and also allows admins to delete their own auditlog lines. Something to consider. Maybe a feature toggle can help

eldering:
I manually pushed this in ce7de3d5fad1ef9d8ebb0dfe78f5241f7f23a2e7

vmcj:
When checking something for the online branch it looks like adminer uses stuff which is not allowed since PHP7.0. That might be a reason to not do this.

meisterT:
@nickygerritsen will check whether it can be installed via composer and put behind a config flag

PR #1778, opened 2022-11-08
Add option to kill remaining processes on judgehost

eldering:


thijskh:
Naming is of course the hardest problem... but is `programs` the best command name? The key aspect seems to be the runuser(actually rungroup). So maybe "runuser-procs" although that's a bit unwieldy perhaps...

eldering:
> Naming is of course the hardest problem... but is `programs` the best command name? The key aspect seems to be the runuser(actually rungroup). So maybe "runuser-procs" although that's a bit unwieldy perhaps...

I agree the name is not perfect, but I wanted something reasonably short. I feel that we can leave out `runuser-` as that should be implied by the script name `dj_judgehost_cleanup`, so maybe `procs`?

thijskh:
Agreed and that we will probably not find the perfect name, so procs seems right

thijskh (approved):

PR #1784, opened 2022-11-09
Enable building debian packages for the wfdhaka_online branch.

eldering:
So that we automatically deploy changes to the online judge.


.github/workflows/build-packages.yml
```diff
@@ -2,7 +2,7 @@ name: "Debian Package Builder"
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, wfdhaka_online ]
   pull_request:
```
ubergeek42, line 6:
```suggestion
```
I think delete this line too so it only builds on the main branch, not when someone makes a PR

eldering: Done

PR #1789, opened 2022-11-10
Use (descriptive) name instead of shortname in HTML title on scoreboa…

meisterT:
…rds.

Reported by @jsannemo.

### DIFF
--- a/webapp/templates/jury/scoreboard.html.twig
+++ b/webapp/templates/jury/scoreboard.html.twig
@@ -1,7 +1,7 @@
 {% extends "jury/base.html.twig" %}
 {% import "jury/jury_macros.twig" as macros %}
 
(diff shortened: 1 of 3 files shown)


nickygerritsen (approved):

thijskh:
I think this was intentional as browser tabs have very limited space so a shortname is more informative there than _The International Collegiate Programming Contest North Western European Regional Contest Sponsored by International Business Machines Hosted by Nicky Gerritsen Honored by Upsilon Pi Epsilon_

meisterT:
But `dress` or `finals` is a really not descriptive name.

thijskh:
The shortname shows in other places e.g. the menu bar so it makes sense to choose the short names at least somewhat descriptively e.g. `NWERC2022`

nickygerritsen:
We do only show it in the menu bar for teams if there is >1 contest. For jury we do always show it. I think the finals contest is called `finals` for years, and I am not sure if we should change that...

PR #1790, opened 2022-11-10
Do not force hours in time diff to be at least two digits.

meisterT:
Previously, this caused ugly scoreboard free times such as +04:00:00.

Reported by @jsannemo

### DIFF
--- a/webapp/src/Utils/Utils.php
+++ b/webapp/src/Utils/Utils.php
@@ -259,7 +259,7 @@ class Utils
         $m = floor($s / 60);
         $s -= $m * 60;
 
-        return sprintf('%02d:%02d:%02d', $h, $m, $s);
+        return sprintf('%d:%02d:%02d', $h, $m, $s);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


nickygerritsen (approved):

thijskh:
Why is this better?

If you have a table now they will not all have the same length.

It's also a deviation from ISO 8601 and what 'everything' does, e.g. `date` on the command line or times in syslog files.

thijskh:
We could replace this with "time elapsed:", "remaining:" instead of the wall clock times?

Cropped screenshot of Scoreboard showing "starts: 10:53 - ends: 15:53"

I doubt we would need to have both wallclock and stopwatch times listed side by side.

meisterT:
This is about time differences not, absolute times. How does `date` do display time diffs?

thijskh:
Not sure commandline `date` supports time diferences but e.g. a timezone offset is commonly listed as `+04:00` and not `+4:00`.

ISO 8601 specifies the given duration as `P4H`.

PR #1793, opened 2022-11-10
Use a shared service for awards between the API and the scoreboard. Fixes #1792

### LINKED ISSUE #1792: Visual medals on scoreboard should use same logic as AwardsController in API
The medals in the UI are different from the API and the API seems to be (more) correct. Let's move that to a service and then use it in both places.

nickygerritsen:
Also do not hand out any (medal) winner awards for teams who did not score.

Note: at a later point I want to add tests for the service, but for now we have the ones from the API (which are limited).

### DIFF
--- a/webapp/tests/Unit/Controller/API/AwardsControllerTest.php
+++ b/webapp/tests/Unit/Controller/API/AwardsControllerTest.php
@@ -2,11 +2,19 @@
 
 namespace App\Tests\Unit\Controller\API;
 
+use App\DataFixtures\Test\SampleSubmissionsThreeTriesCorrectFixture;
+use App\Entity\Contest;
+use App\Service\ScoreboardService;
+use Doctrine\ORM\EntityManagerInterface;
+
 class AwardsControllerTest extends BaseTest
 {
     protected ?string $apiEndpoint = 'awards';
 
     protected static string $skipMessageIDs = "Filtering on IDs not implemented in this endpoint.";
+
+    // We add a submission since we are not handing out any awards without it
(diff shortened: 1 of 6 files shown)


webapp/tests/Unit/Service/AwardServiceTest.php
```diff
@@ -0,0 +1,223 @@
...
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+class AwardServiceTest extends KernelTestCase
+{
+    protected Contest $contest;
+    protected Scoreboard $scoreboard;
+
+    protected function setUp(): void
+    {
+        // THe contest with have 1 gold, 1 silver and 2 bronze medals
```
vmcj, line 24:
```suggestion
        // The contest with have 1 gold, 1 silver and 2 bronze medals
```

webapp/tests/Unit/Controller/API/AwardsControllerTest.php
```diff
@@ -18,6 +26,20 @@ class AwardsControllerTest extends BaseTest
 
     protected array $expectedAbsent = ['bronze-medal', 'first-to-solve'];
 
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        // We need to refersh the scoreboard cache since we have added a submission
```
vmcj, line 33:
```suggestion
        // We need to refresh the scoreboard cache since we added a submission
```

vmcj (approved):
I think the tests make sense and the code is understandable so is not worse than the situation before.

meisterT (approved):

PR #1804, opened 2022-11-11
Disallow changes to immutable executables or files in them. Fixes #1796.

### LINKED ISSUE #1796: Deleting a file from an executable doesn't update the immutable executable hash
Steps to reproduce:

- Write down the hash for an immutable executable
- Delete a file from an executable
- Observe that no new immutable executable is created nor are the hashes updated

This will later break judging when the judgehost compares the hash.

nickygerritsen:
I did the thing Jaap suggested and removed all writable methods on the immutable executable entity.

However, we removed executable files from the DB directly, so I also forbid that.

### DIFF
--- a/webapp/src/Controller/Jury/ExecutableController.php
+++ b/webapp/src/Controller/Jury/ExecutableController.php
@@ -228,10 +228,10 @@ class ExecutableController extends BaseController
     }
 
     /**
-     * @Route("/{execId}/delete/{rank}", name="jury_executable_delete_single")
+     * @Route("/{execId}/delete/{rankToDelete}", name="jury_executable_delete_single")
      * @IsGranted("ROLE_ADMIN")
      */
-    public function deleteSingleAction(Request $request, string $execId, int $rank): Response
+    public function deleteSingleAction(Request $request, string $execId, int $rankToDelete): Response
     {
         /** @var Executable $executable */
         $executable = $this->em->getRepository(Executable::class)->find($execId);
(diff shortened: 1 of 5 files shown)


webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -268,7 +268,26 @@ public function deleteSingleAction(Request $request, string $execId, int $rank):
 
             return $this->render('jury/delete.html.twig', $data);
         } else {
-            $this->em->remove($file);
+            // Create a copy of all files except $file
```
meisterT, line 271: The `$file` in this comment is the `$file from above` and then in the loop below is a different `$file`. While this is valid code, it is confusing.

I suggest renaming `$file` to `$fileToDelete`, and `$rank` to `$rankToDelete`.

webapp/src/Entity/ExecutableFile.php
```diff
@@ -138,4 +139,12 @@ public function isExecutable(): bool
     {
         return $this->isExecutable;
     }
+
+    /**
+     * @ORM\PreRemove()
+     */
+    public function disallowDelete(): void
+    {
+        throw new \RuntimeException('An executable file cannot be deleted');
```
meisterT, line 148: If you delete an immutable executable, can you still delete the file? Should you be able to even?

nickygerritsen: You can't delete an immutable executable either.

webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -268,7 +268,26 @@ public function deleteSingleAction(Request $request, string $execId, int $rank):
...
+            /** @var ExecutableFile $file */
+            foreach ($executable->getImmutableExecutable()->getFiles() as $file) {
+                if ($file->getRank() == $rankToDelete) {
+                    continue;
+                }
+
+                $executableFile = new ExecutableFile();
+                $executableFile
+                    ->setRank($file->getRank())
+                    ->setIsExecutable($file->isExecutable())
```
vmcj, line 282: When is an executableFile not Executable?

webapp/src/Entity/ImmutableExecutable.php
```diff
@@ -33,7 +37,7 @@ class ImmutableExecutable
      * @ORM\JoinColumn(name="userid", referencedColumnName="userid", onDelete="SET NULL")
```
vmcj, line 37: The `SET NULL` is still needed because we disconnect the user from the Executable?

nickygerritsen: I don’t know. I don’t know when we filled it

meisterT (approved):

Issue #1719, opened 2022-10-20
Allow setting contest activate time via import


mpsijm:
When importing a contest (I'm using `bt export` from BAPCtools), the activate time of the contest is set to "-24:00" (1 day) by default. But, when I'm testing the uploading of a contest, it's typically the only contest in the system, so I would like it to be active immediately.

It would be nice if we can either:
- set the activation date/time, e.g., using an extra field in `contest.yaml`
- automatically mark the contest as "active" while uploading (could also be an API parameter, perhaps)

meisterT:
I agree that this can be improved. Perhaps we should just set the activation time to the current time if a contest is created through the API?

mpsijm:
That would also be fine :smile: I asked @nickygerritsen about it, and he said that there's not really a reason to set the activation time to -24:00, except that "multiple active contests at the same time is unhandy" :stuck_out_tongue: But for my use cases, it's always desirable to activate the contest immediately on import :slightly_smiling_face:

PR #1807, opened 2022-11-14
Allow setting activation time in contest.yaml

vmcj:
Closes https://github.com/DOMjudge/domjudge/issues/1719

I've added some more times to the contest which might be useful. If we can expose a property but currently it's null, should we display as null or do we omit it from the                                response? I've now chosen to omit as that is also easier to parse for downstream clients. It's easy to crash an a missing key and fix it, but finding why you get a strange time when converting null to something else is slightly harder IMO.

### DIFF
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -138,7 +138,7 @@ class ImportExportService
         $activateTimeFields = ['activation_time','activate_time','activation-time', 'activate-time'];
         $deactivateTimeFields = preg_filter('/^/', 'de', $activateTimeFields);
         $startTimeFields = ['start_time', 'start-time'];
-        $requiredFields = [$startTimeFields, 'formal_name', ['id', 'name', 'short-name', 'short_name'], 'duration'];
+        $requiredFields = [$startTimeFields, ['name', 'formal_name'], ['id', 'short_name', 'short-name'], 'duration'];
         $missingFields = [];
         foreach ($requiredFields as $field) {
             if (is_array($field)) {
@@ -188,7 +188,7 @@ class ImportExportService
 
         $contest = new Contest();
         $contest
-            ->setName($data['formal_name'])
+            ->setName($data['name'] ?? $data['formal_name'] )
             ->setShortname(preg_replace(
                                $invalid_regex,
                                '_',
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -47,19 +47,34 @@ class ImportExportService
         // We expect contest.yaml and problemset.yaml combined into one file here.
 
         $data = [
+            'id' => $contest->getExternalid(),
             'formal_name' => $contest->getName(),
             'name' => $contest->getShortname(),
             'start_time' => Utils::absTime($contest->getStarttime(), true),
+            'end_time' => Utils::absTime($contest->getEndtime(), true),
             'duration' => Utils::relTime($contest->getContestTime((float)$contest->getEndtime())),
             'penalty_time' => $this->config->get('penalty_time'),
+            'activation_time' => Utils::absTime($contest->getActivatetime(), true),
         ];
         if ($warnMsg = $contest->getWarningMessage()) {
             $data['warning_message'] = $warnMsg;
         }
         if ($contest->getFreezetime() !== null) {
+            $data['scoreboard_freeze_time'] = Utils::absTime($contest->getFreezetime(), true);
             $data['scoreboard_freeze_duration'] = Utils::relTime(
                 $contest->getContestTime((float)$contest->getEndtime()) - $contest->getContestTime((float)$contest->getFreezetime()),
-                true);
+                true,
+            );
+            if ($contest->getUnfreezetime() !== null) {
+                $data['scoreboard_thaw_time'] = Utils::absTime($contest->getUnfreezetime(), true);
+            }
+        }
+        if ($contest->getFinalizetime() !== null) {
+            $data['finalize_time'] = Utils::absTime($contest->getFinalizetime(), true);
+        }
+
+        if ($contest->getDeactivatetime() !== null) {
+            $data['deactivate_time'] = Utils::absTime($contest->getDeactivatetime(), true);
         }
 
         if ($includeProblems) {
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -110,11 +110,12 @@ class ImportExportService
         $timeValue = null;
         $usedField = null;
         foreach ($fields as $field) {
-            if ($timeValue) {
-                continue;
-            }
             $timeValue = $data[$field] ?? null;
             $usedField = $field;
+            // We need to check as the value for the key can be null
+            if ($timeValue) {
+                break;
+            }
         }
 
         if (is_string($timeValue)) {
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -136,10 +136,10 @@ class ImportExportService
         return $time instanceof DateTime ? DateTimeImmutable::createFromMutable($time) : $time;
     }
 
-    public function importContestData($data, ?string &$message = null, string &$cid = null): bool
+    public function importContestData($data, ?string &$errorMessage = null, string &$cid = null): bool
     {
         if (empty($data) || !is_array($data)) {
-            $message = 'Error parsing YAML file.';
+            $errorMessage = 'Error parsing YAML file.';
             return false;
         }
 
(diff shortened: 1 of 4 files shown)


webapp/src/Service/ImportExportService.php
```diff
@@ -148,9 +148,20 @@ public function importContestData($data, ?string &$message = null, string &$cid
 
         $starttime->setTimezone(new DateTimeZone(date_default_timezone_get()));
         $activateTime = new DateTime();
-        if ($activateTime > $starttime) {
+        if (array_key_exists('activate-time', $data)) {
+            $activatetimeValue = $data['activate-time'];
+            if (is_string($activatetimeValue)) {
```
nickygerritsen, line 153: We have this logic somewhere else as well right? Maybe put it in Utils?

webapp/src/Service/ImportExportService.php
```diff
@@ -148,9 +148,20 @@ public function importContestData($data, ?string &$message = null, string &$cid
...
+                    // Make sure ISO 8601 but with the T replaced with a space also works.
+                    date_create_from_format('Y-m-d H:i:sO', $activatetimeValue);
+            } else {
+                /** @var DateTime $activateTime */
+                $activateTime = $activatetimeValue;
+            }
+        } elseif ($activateTime > $starttime) {
             $activateTime = $starttime;
         }
+        $activateTime->setTimezone(new DateTimeZone(date_default_timezone_get()));
```
eldering, line 164: Why this change? This is also changing the behavior of the old case, is that intentional?

webapp/src/Service/ImportExportService.php
vmcj: I've rewrote the PR, so I'll resolve this.

webapp/src/Service/ImportExportService.php
vmcj: I've rewrote the PR, so I'll resolve this.

nickygerritsen (approved):

webapp/src/Controller/API/ContestController.php
```diff
@@ -417,13 +417,25 @@ public function getContestYamlAction(Request $request, string $cid): StreamedRes
         $response->setCallback(function () use ($contest, $penalty_time) {
             echo "name:                     " . $contest->getName() . "\n";
             echo "short-name:               " . $contest->getExternalid() . "\n";
+            echo "activation-time:               " .
```
eldering, line 420: Alignment is off here and below.

webapp/src/Controller/API/ContestController.php
```diff
@@ -417,13 +417,25 @@ public function getContestYamlAction(Request $request, string $cid): StreamedRes
...
+                Utils::absTime($contest->getActivatetime(), true) . "\n";
             echo "start-time:               " .
                 Utils::absTime($contest->getStarttime(), true) . "\n";
             echo "duration:                 " .
                 Utils::relTime($contest->getEndtime() - $contest->getStarttime(), true) . "\n";
             echo "scoreboard-freeze-length: " .
```
eldering, line 426: Slightly unrelated, but we should change this to `scoreboard_freeze_duration`, see https://ccs-specs.icpc.io/draft/contest_api#contests and https://ccs-specs.icpc.io/draft/contest_package#example-yaml-files

webapp/src/Controller/API/ContestController.php
```diff
@@ -417,13 +417,25 @@ public function getContestYamlAction(Request $request, string $cid): StreamedRes
         $response->setCallback(function () use ($contest, $penalty_time) {
             echo "name:                     " . $contest->getName() . "\n";
             echo "short-name:               " . $contest->getExternalid() . "\n";
```
eldering, line 419: Here and below: we should use `_` (underscore) instead of `-` (dash).

webapp/src/Controller/API/ContestController.php
```diff
@@ -417,13 +417,25 @@ public function getContestYamlAction(Request $request, string $cid): StreamedRes
...
             echo "scoreboard-freeze-length: " .
                 Utils::relTime($contest->getEndtime() - $contest->getFreezetime(), true) . "\n";
             echo "penalty-time:             " . $penalty_time . "\n";
+            echo "freeze-time:               " .
+                Utils::absTime($contest->getFreezetime(), true) . "\n";
+            echo "end-time:               " .
+                Utils::absTime($contest->getEndtime(), true) . "\n";
+            echo "finalize-time:               " .
+                Utils::absTime($contest->getFinalizetime(), true) . "\n";
+            echo "unfreeze-time:               " .
```
eldering, line 435: We may want to call this `scoreboard_thaw_time` to already prepare for https://github.com/icpc/ccs-specs/pull/134/files#diff-353f5220dc395fec383c89d1431e3c5733913a1af93f9e8d3c0135d49b2fb5ebR735

vmcj:
I had to break the import of the old import format for this, I assume we want to support that again as we also still support the TSV files?

webapp/src/Service/ImportExportService.php
```diff
@@ -83,15 +99,47 @@ public function getContestYamlData(Contest $contest, bool $includeProblems = tru
         return $data;
     }
 
+    protected function convertImportedTime(array $fields, array $data, ?string &$message = null): ?DateTimeImmutable
+    {
+        $timeValue = null;
+        $usedField = null;
+        foreach ($fields as $field) {
+            if ($timeValue) {
+                continue;
```
eldering, line 108: Why not just break as soon as `$timeValue` gets set?

webapp/src/Service/ImportExportService.php
```diff
@@ -83,15 +99,47 @@ public function getContestYamlData(Contest $contest, bool $includeProblems = tru
         return $data;
     }
 
+    protected function convertImportedTime(array $fields, array $data, ?string &$message = null): ?DateTimeImmutable
```
eldering, line 102: Would be good to briefly comment what this does: find the first set field from `$fields` in `$data` and parse it as a date.
Also, would be good to clarify that there's a distinction betwee a return value of `null` and parse failure, and that that's only the case when `$message` is not `null`. Would be better to rename that to `$error_message` or so.

webapp/src/Service/ImportExportService.php
```diff
@@ -83,15 +99,47 @@ public function getContestYamlData(Contest $contest, bool $includeProblems = tru
...
     public function importContestData($data, ?string &$message = null, string &$cid = null): bool
     {
         if (empty($data) || !is_array($data)) {
             $message = 'Error parsing YAML file.';
             return false;
         }
 
-        $requiredFields = [['start_time', 'start-time'], 'name', ['id', 'short-name'], 'duration'];
-        $missingFields  = [];
+        $activateTimeFields = ['activate-time', 'activation-time', 'activate_time', 'activation_time'];
```
eldering, line 138: I think these should be ordered from most recent standard to oldest/most non-standard. So `activate_time` first.

webapp/src/Service/ImportExportService.php
```diff
@@ -83,15 +99,47 @@ public function getContestYamlData(Contest $contest, bool $includeProblems = tru
...
+        $deactivateTimeFields = preg_filter('/^/', 'de', $activateTimeFields);
+        $startTimeFields = ['start_time', 'start-time'];
+        $requiredFields = [$startTimeFields, 'formal_name', ['id', 'name', 'short-name', 'short_name'], 'duration'];
```
eldering, line 141: `formal_name` should not be required, we should just set our `name` to either `name` from the `contest.yaml` or if that doesn't exist the `formal_name`. So I guess it should be `['name', 'formal_name']` in there.

webapp/src/Service/ImportExportService.php
```diff
@@ -117,39 +165,43 @@ public function importContestData($data, ?string &$message = null, string &$cid
...
-            $activateTime = $starttime;
+        $deactivateTime = $this->convertImportedTime($deactivateTimeFields, $data, $message);
+        if ($message) {
+            return false;
         }
+
         $contest = new Contest();
         $contest
-            ->setName($data['name'])
+            ->setName($data['formal_name'])
```
eldering, line 191: This should use `name` and fall back to `formal_name`.

webapp/src/Service/ImportExportService.php
vmcj: Original author didn't do that, but makes sense IMO

mpsijm:
Thanks for merging this! :smile:

Question: since either `activat{e,ion}{_,-}time` is allowed as key, what is the canonical form? I would guess `activation_time`, but I also see `deactivate_time` used in [the code](https://github.com/DOMjudge/domjudge/pull/1807/files#diff-2e92e4f26fbadf9eff221ac3fa36ef1bf0bf1077b805645896e361aa5db841c9R78), so asking just to be sure :slightly_smiling_face:

eldering:
It's `(de)activate_time`, see https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Entity/Contest.php#L84

For most other properties you can refer to https://ccs-specs.icpc.io/draft/contest_api#interface-specification, but these are our own specification.

PR #1189, opened 2021-09-15
Judgedaemon sleep increase

mvr320:
Based on suggestion by @meisterT

PR #1815, opened 2022-11-26
Always use `usleep`, instead of `sleep` for >1s sleeps.

meisterT:
There is no real reason to use `sleep`, and it allows us to remove one branch.


meisterT:
https://github.com/DOMjudge/domjudge/pull/1189#issuecomment-920618471 says we shouldn't do it...

meisterT:
Perhaps https://www.php.net/manual/en/function.time-nanosleep.php is the solution

PR #1817, opened 2022-11-26
Use a wrapper function dj_sleep around time_nanosleep.

eldering:
This instead of using usleep as it is not guaranteed to handle times larger than a second and may also consume CPU cycles.

Alternative fix for #1815

### DIFF
--- a/etc/judgehost-config.php
+++ b/etc/judgehost-config.php
@@ -15,7 +15,7 @@ define('CREATE_WRITABLE_TEMP_DIR', getenv('DOMJUDGE_CREATE_WRITABLE_TEMP_DIR') ?
 // If any transient network error occurs on the nth trial,
 // the judgehost retries the HTTP request after pow(factor, trial - 1) + rand(0, jitter) sec.
 
-function define_backoff_params_from_env(string $var_name, int|float $default_value) {
+function define_backoff_params_from_env(string $var_name, $default_value) {
     if (defined($var_name)) {
         return;
     }
@@ -25,7 +25,7 @@ function define_backoff_params_from_env(string $var_name, int|float $default_val
             'min_range' => 0,
         ),
     );
-    $final_value = filter_var(getenv('DOMJUDGE_' . $var_name), FILTER_VALIDATE_INT, $options);
+    $final_value = filter_var(getenv('DOMJUDGE_' . $var_name), FILTER_VALIDATE_FLOAT, $options);
     define($var_name, $final_value);
 }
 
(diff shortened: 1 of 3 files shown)


judge/judgedaemon.main.php
```diff
@@ -620,7 +619,7 @@ function fetch_executable_internal(
             break;
         }
         logmsg(LOG_WARNING, "Failed to report $judgeTaskId in attempt #" . ($i + 1) . ".");
-        usleep(100 + random_int(200, ($i+1)*1000));
+        dj_sleep(0.0001 * random_int(3, ($i+1)*10));
```
meisterT, line 622: I think this was intended to be milliseconds.

lib/lib.wrappers.php
```diff
@@ -71,3 +71,23 @@ function dj_escapeshellarg(?string $arg) : string
...
+ * A simple function that allows to sleep for fractional seconds. Note that
+ * usleep is documented to consume CPU cycles and may not work for times
+ * larger than a second.
+ * Returns a boolean value for success or if the delay was interrupted by a
+ * signal, returns a float with the time remaining, similar to time_nanosleep.
+ */
+function dj_sleep(float $seconds) : bool
+{
+    $seconds_int = (int)$seconds;
+    $nanoseconds = (int)(1000000000*($seconds-$seconds_int));
```
meisterT, line 85: Use underscores to make it more readable?

lib/lib.wrappers.php
```diff
@@ -71,3 +71,23 @@ function dj_escapeshellarg(?string $arg) : string
...
+
+    $result = time_nanosleep($seconds_int, $nanoseconds);
+    if (is_array($result)) {
+        $result = $result['seconds'] + 0.000000001 * $result['nanoseconds'];
```
meisterT, line 89: Instead of the multiplication here, make the value above a constant and use division with the same constant?

etc/judgehost-config.php
```diff
@@ -13,9 +13,9 @@
 
 // These define HTTP request backoff related constants.
 // If any transient network error occurs on the nth trial,
-// the judgehost retries the HTTP request after (1000 * pow(factor, trial - 1) + rand(0, jitter)) ms.
+// the judgehost retries the HTTP request after pow(factor, trial - 1) + rand(0, jitter) sec.
 
-function define_backoff_params_from_env(string $var_name, int $default_value) {
+function define_backoff_params_from_env(string $var_name, int|float $default_value) {
```
meisterT, line 18: I think multiple type hints are only supported in PHP 8.0+

judge/judgedaemon.main.php
```diff
@@ -620,7 +619,7 @@ function fetch_executable_internal(
             break;
         }
         logmsg(LOG_WARNING, "Failed to report $judgeTaskId in attempt #" . ($i + 1) . ".");
-        usleep(100 + random_int(200, ($i+1)*1000));
+        dj_sleep(0.1 * random_int(3, ($i+1)*10));
```
meisterT, line 622: super-nit: I would probably write this as:
```
$sleep_millis = random_int(3000, ($i+1)*10000);
dj_sleep(0.001 * $sleep_millis);
```

meisterT (approved):

PR #1825, opened 2022-11-27
Remove entry point detection info from compile output if successful.

### LINKED ISSUE #1822: Confusing "Info: detected entry point:" message
### Steps to reproduce
- Configure python without requiring an entry point with the default compile script.
- Submit as a team
- Click on your submission in the submission list on the team page

### Actual behaviour
image

This is confusing because it is not actionable for teams. It is leaking that we need entry points for shadowing and should not be exposed at all to teams.

eldering:
This only tends to confuse contestants into thinking that this is an actual error, while it is there just for informational/debug purposes. Since it is mostly useful for admins/devs only and this is already stored in the compile.meta file, filter it out.

Closes #1822

### DIFF
--- a/judge/compile.sh
+++ b/judge/compile.sh
@@ -181,7 +181,8 @@ fi
 
 # Check if the compile script auto-detected the entry point, and if
 # so, store it in the compile.meta for later reuse, e.g. in a replay.
-grep '[Dd]etected entry_point: ' compile.tmp | sed 's/^.*etected //' >>compile.meta
+ENTRY_POINT_REGEX='[Dd]etected entry_point: '
+grep "$ENTRY_POINT_REGEX" compile.tmp | sed 's/^.*etected //' >>compile.meta
 
 logmsg $LOG_DEBUG "checking compilation exit-status"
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

meisterT (approved):

meisterT:
Integration test failure still looks related: [Nov 27 23:40:08.403] compile.sh[16716]: error: grep -v '[Dd]etected entry_point: ' failed

judge/compile.sh
```diff
@@ -199,7 +200,16 @@ if [ ! -f compile/program ] || [ ! -x compile/program ]; then
...
 	cleanexit ${E_COMPILER_ERROR:-1}
 fi
-cat compile.tmp >>compile.out
+
+# Remove any entry point detection message when compilation succeeded,
+# since we already stored it above and it only confuses contestants.
+if ! grep -v "$ENTRY_POINT_REGEX" compile.tmp >>compile.out; then
+	 exitcode=$?
+	 # Only report real errors, not when no matches were found.
+	 if [ $exitcode -ne 1 ]; then
```
thijskh, line 209: Intended was: `-gt 1`?

PR #1827, opened 2022-11-27
Delete some dead code, I assume it was leftover debug code?

meisterT:


nickygerritsen (approved):
It was used when we put accounts in the feed. It can indeed be gone now

PR #1833, opened 2022-11-28
Split access log between API and other requests.

nickygerritsen:
This is part of DOMjudge/domjudge-scripts#80


nickygerritsen:
When this looks good I will do a similar change in Ansible.

thijskh:
And for Apache?

nickygerritsen:
I am not so skilled in Apache so was hoping someone else would do it after this is merged

vmcj (approved):

Issue #1794, opened 2022-11-10
Display contest clock on scoreboard


clevengr:
### Description of the problem

The DOMjudge public scoreboard for WF Dhaka shows the contest start-time and contest end-time, but it doesn't show the contest "elapsed time" or "remaining time".  It would be nice if, while the contest was under way, there was a field at the top of the scoreboard showing at least one (maybe both?) of these values.

### Your environment

I was using Chrome Version 107.0.5304.88 (Official Build) (64-bit), but I doubt that's relevant; I'm pretty sure this is consistently (missing) in all web browsers.

### Steps to reproduce
Open the WF Dhaka web page at https://image.icpc.global/wfscoreboard/index.html.
Look everywhere on the page -- no elapsed or remaining time can be found.

### Expected behaviour
A countdown (remaining time) or count-up (elapsed time) clock should appear on the page (somewhere, hopefully near the top).

### Actual behaviour
There's no way to determine the remaining time in the contest except to look at what time it started, find another clock to see what the  current time is (which is challenging if you're not in the same time zone), and do the math.  Math should be done by computers, especially ones smart enough to run DOMjudge.

eldering:
Note that this is the static scoreboard. The contest time clock normall shows in the menu header, but that's stripped from the static scoreboard.

thijskh:
We could replace this with "time elapsed:", "remaining:" instead of the wall clock times?

![Cropped screenshot of Scoreboard showing "starts: 10:53 - ends: 15:53"](https://user-images.githubusercontent.com/3808792/201057004-1e8567d4-93d7-43ee-a03c-fdd5d638ac75.png)

I doubt we would need to have both wallclock and stopwatch times listed side by side.

eldering:
I think for the static scoreboard a time remaining doesn't make that much sense, but adding a current contest time would. Since this is a static scoreboard, the time of viewing might be different from the time of generating it. Also, timezone differences might make it difficult to calculate the scoreboard time from local wallclock time.

So I'd suggest adding a current contest time here.

thijskh:
Well, we do display the information already by means of the progress bar. So time elapsed / remaining are already represented, only not in numerical form.

PR #1836, opened 2022-11-28
On static scoreboard, display remaining time.

meisterT:
Attempt to fix https://github.com/DOMjudge/domjudge/issues/1794

Examples:
image

image

image

### DIFF
--- a/webapp/src/Twig/TwigExtension.php
+++ b/webapp/src/Twig/TwigExtension.php
@@ -85,6 +85,7 @@ class TwigExtension extends AbstractExtension implements GlobalsInterface
     {
         return [
             new TwigFilter('printtimediff', [$this, 'printtimediff']),
+            new TwigFilter('printremainingminutes', [$this, 'printremainingminutes']),
             new TwigFilter('printtime', [$this, 'printtime']),
             new TwigFilter('printtimeHover', [$this, 'printtimeHover'], ['is_safe' => ['html']]),
             new TwigFilter('printResult', [$this, 'printResult'], ['is_safe' => ['html']]),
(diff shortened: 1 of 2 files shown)


webapp/src/Twig/TwigExtension.php
```diff
@@ -174,6 +175,16 @@ public function printtimediff(float $start, ?float $end = null): string
         return Utils::printtimediff($start, $end);
     }
 
+    public function printremainingminutes(float $start, float $end): string
+    {
+        $minutesRemaining = round(($end - $start)/60, 0);
+        if ($minutesRemaining < 1) {
+            return 'less than 1 minute to go';
+        } else {
+            return $minutesRemaining . ' minutes to go';
```
nickygerritsen, line 184: How does this ever display `1 minute to go` like in your screenshot?

webapp/src/Twig/TwigExtension.php
```diff
@@ -174,6 +175,16 @@ public function printtimediff(float $start, ?float $end = null): string
         return Utils::printtimediff($start, $end);
     }
 
+    public function printremainingminutes(float $start, float $end): string
+    {
+        $minutesRemaining = round(($end - $start)/60, 0);
```
thijskh, line 180: Should this be floor instead of round?

webapp/templates/partials/scoreboard.html.twig
```diff
@@ -31,6 +31,9 @@
                     {% endif %}
                 {% elseif scoreboard.freezeData.stopped %}
                     contest over, waiting for results
+                {% elseif static %}
+                    {% set now = 'now'|date('U') %}
```
thijskh, line 35: I think {{ now }} should already be set (e.g. used by the menu bar clock)?

webapp/src/Twig/TwigExtension.php
meisterT: Urgs, that was because I can't do git properly.

nickygerritsen (approved):

webapp/src/Twig/TwigExtension.php
meisterT: I first had put floor, then changed it round, so clearly don't have a strong opinion.

thijskh (approved):

PR #1853, opened 2022-12-14
Test with latest PHP82 version

vmcj:
Codecoverage is still done on the PHP81 for now as it makes more sense to do this when something is merged to main.


vmcj:
Codecoverage is still done on the PHP81 for now as it makes more sense to do this when something is merged to main.

nickygerritsen (approved):

PR #1855, opened 2022-12-22
Don't use all PSR rules

vmcj:
Especially the one about the vertical space results in discussion. The side effects would result in splitting files without making them easier to understand (for now). The line length is something we should work on but would now clutter untouched codes in reviews.


vmcj:
@eldering and @meisterT I think you both have the most issues with the current annotations. With this PR those unhandled annotations should now not be displayed anymore.

meisterT (approved):
I am fine with starting with this. We can change it later in either direction incrementally.

eldering (approved):

PR #1857, opened 2022-12-27
fix potential errors in boolfind output validator

ir5:
I found the current implementation of output validator for boolfind example contains potential errors, which might cause segmentation fault and make judgehost halt.

+ Line 17: `new int(nbools);` is a typo of `new int[nbools];`. However, I believe it is safer to use a `vector` than to use a `new`.
+ Line 41: Using `%s` for int is inappropriate. This can cause out-of-range error.
+ Line 50 and 69: `.c_str()` is required to output a string object in vfprintf.

### DIFF
--- a/example_problems/boolfind/output_validators/boolfind_run/validate.cc
+++ b/example_problems/boolfind/output_validators/boolfind_run/validate.cc
@@ -5,6 +5,7 @@
 #include <cstring>
 #include <cmath>
 #include <unistd.h>
+#include <vector>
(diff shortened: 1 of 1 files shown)


vmcj:
Hi @ir5, thanks for your contribution.

I'll let an C++ expert look at it and will merge afterwards. Due to the holidays this might take a bit more time than normal.

edomora97 (approved):

Issue #1861, opened 2022-12-29
Document that `@EXPECTED_RESULTS@: ` tag is ignored for standard submission directories


mpsijm:
### Description of the problem
The `@EXPECTED_RESULTS@: ` tag is ignored when an imported submission is one of the standard submission directories (e.g., `accepted/`, `wrong_answer/`) since https://github.com/DOMjudge/domjudge/commit/3b9ec839379bef79a23633e1b51c22285265e683. This is not documented anywhere and led to confusion for me during the last few contests.

### Your environment
DOMjudge, since 7.0.0

### Current behaviour
Some examples, when importing a problem zip:
- For `submissions/accepted/solution.py`, the expected result is `["CORRECT"]`.
- For `submissions/accepted/solution.py` with any `@EXPECTED_RESULTS@: (...)` tag, the expected result is `["CORRECT"]`.
- For `submissions/mixed/solution.py` with `@EXPECTED_RESULTS@: ACCEPTED,TIME_LIMIT_EXCEEDED` tag, the expected result is `["CORRECT","TIMELIMIT"]`.
  - Also, after importing, the info alert at the top of the page says "Annotated result 'CORRECT, TIMELIMIT' does not match directory for submissions/mixed/solution.py".
- For `submissions/mixed/solution.py` without any `@EXPECTED_RESULTS@: ` tag, the expected result is `["MIXED"]`.
  - The info alert at the top of the page shows no warning for this, but perhaps that would be a good idea to add :stuck_out_tongue: 

### Any other information that you want to share?
I wouldn't mind if the commit linked above would be reverted (i.e., that the `@EXPECTED_RESULTS@: ` tag would take precedence over the submission's directory name again). But you probably have a good reason to keep the current behaviour :innocent:

vmcj:
I think the reason for this ordering is that stuff which is in the spec takes precedence over things which are not in it. Otherwise you would have different results between DOMjudge and another system following the CLICS spec when uploading the problemset.

eldering:
Indeed, there are two different methods to classify the expected outcome (directory and `@EXPECTED_RESULTS@` tag), and we somehow have to make a decision which one to prioritize. I think the current one is relatively simple, but might indeed not be expected, so would be good to document this.
FYI, there's an unresolved issue on how to add the tags to the problem package specification, see https://github.com/Kattis/problem-package-format/pull/4

mpsijm:
[...]

PR #1862, opened 2022-12-30
Mention EXPECTED_RESULTS in the docs

vmcj:
Closes: https://github.com/DOMjudge/domjudge/issues/1861

Currently the EXPECTED_RESULTS are not part of the ICPC problem package and I'm not sure if they will be soon. We mention the EXPECTED_RESULTS in the README of the example problems but not in the doc. This PR tries to fix this.

I'm not sure how much we want to mention here as it will become part of the spec?

~~Still need to link to the Judging verifier also but that is hard to do in a PR via GitHub itself. If we agree on the wording I'll also make that link so the functionality is bundled together with its primary usage in the docs.~~ We already link from the Judging Verifier to this page which is most likely the better route.

### DIFF
--- a/doc/manual/problem-format.rst
+++ b/doc/manual/problem-format.rst
@@ -12,13 +12,13 @@ On top, DOMjudge defines a few extensions:
    distributed to participants. The file extension determines any of
    three supported formats. If multiple files matching this pattern are
    available, any one of those will be used.
- * Annotation with ``@EXPECTED_RESULTS@: `` for jury solutions.
+ * Annotation with ``@EXPECTED_RESULTS@:`` for jury solutions.
    Use this to annotate the possible outcomes that a submission can have,
    that will be accepted by the Judging verifier.
    This extension is in consideration to be added to the ICPC specification,
    see https://github.com/Kattis/problem-package-format/pull/4.
    Note that, to match with the current ICPC specification,
-   the ``@EXPECTED_RESULTS@: `` tag is ignored in
+   the ``@EXPECTED_RESULTS@:`` tag is ignored in
    the four official submission subdirectories,
    but it will work if the tagged solution is in, e.g.,
    ``submissions/mixed/`` or ``submissions/rejected/``.
--- a/doc/manual/problem-format.rst
+++ b/doc/manual/problem-format.rst
@@ -12,6 +12,9 @@ On top, DOMjudge defines a few extensions:
    distributed to participants. The file extension determines any of
    three supported formats. If multiple files matching this pattern are
    available, any one of those will be used.
+ * Annotation with ``EXPECTED_RESULTS`` for jury solutions. The ICPC
+   problem package specification always has precedence. Use this to annotate
+   possible outcomes a submission can have to use together with the Judging verifier. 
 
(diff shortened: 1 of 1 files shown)


doc/manual/problem-format.rst
```diff
@@ -12,6 +12,9 @@ On top, DOMjudge defines a few extensions:
    distributed to participants. The file extension determines any of
    three supported formats. If multiple files matching this pattern are
    available, any one of those will be used.
+ * Annotation with ``EXPECTED_RESULTS`` for jury solutions. The ICPC
+   problem package specification always has precedence. Use this to annotate
+   possible outcomes a submission can have to use together with the Judging verifier.
```
eldering, line 17: It would be good to clarify how these tags should be used exactly, and maybe also refer to a sample submission.
BTW, if you're importing submissions from a problem package, then they have to be located inside a `submissions/*` directory, so will always be overridden, I guess.

mpsijm: Agreed that a clarification on the usage would be nice. Perhaps a link to https://github.com/Kattis/problem-package-format/pull/4 suffices, but I would at least write the tag in the exact required format (i.e., two `@`s and a space after the `:`).

Indeed, the `@EXPECTED_RESULTS@: ` tag is overridden when the submission is in one of the standard submission directories (which are `submissions/{accepted,wrong_answer,time_limit_exceeeded,run_time_error}/`. Perhaps we can explicitly state that this tag therefore only works in non-standard submission directories (like `mixed/` or `rejected/`).

mpsijm, line 17:
```suggestion
 * Annotation with ``@EXPECTED_RESULTS@: `` for jury solutions.
   Use this to annotate the possible outcomes that a submission can have,
   that will be accepted by the Judging verifier.
   This extension is in consideration to be added to the ICPC specification,
   see https://github.com/Kattis/problem-package-format/pull/4.
   Note that, to match with the current ICPC specification,
   the ``@EXPECTED_RESULTS@: `` tag is ignored in
   the four official submission subdirectories,
   but it will work if the tagged solution is in, e.g.,
   ``submissions/mixed/`` or ``submissions/rejected/``.
```

mpsijm: > maybe also refer to a sample submission

I also agree with this one, but the current example submissions in https://github.com/DOMjudge/domjudge/tree/main/example_problems do not exactly match the spec, so linking to them in their current state may be confusing. For many of these submissions, the `@EXPECTED_RESULTS@: ` tag is ignored because it lives in a "standard" submission directory. For BAPCtools, I've had to re-arrange many of these submissions to keep the tests passing (see https://github.com/RagnarGrootKoerkamp/BAPCtools/commit/8b12d05fa94c904d7ce008cbc4a47762132c8475) :stuck_out_tongue:

eldering (approved):

mpsijm (approved):

vmcj:
After https://github.com/DOMjudge/domjudge/pull/2006 is merged there will be a file we can link to as an example.

eldering (approved):

mpsijm (approved):

PR #1863, opened 2022-12-31
compare.cc (default validator): improve float WA message

mpsijm:
The WA message showed `-1` for undefined tolerance values, which was kind of confusing.
Additionally, it was not always clear whether it was the absolute and/or relative tolerance that was exceeded.
This PR adds some extra branches to show a more precise WA message.

The main reason for this change is that the output of this validator is not only visible to jury members, but also to participants for the sample cases, which caused some confusion. But also for jury members, the validator output should be more clear now :slightly_smiling_face: 

I couldn't find any tests for this default validator, but if there are any, I'll be happy to still add some more, to cover all the new branches.
For now, I've tested this validator manually, using some problems from [BAPC 2022](https://2022.bapc.eu/). Problem Failing Flagship has only an absolute tolerance, and problem Imperfect Imperial Units has only a relative tolerance.

@RagnarGrootKoerkamp you'll probably also like this :smile:

### DIFF
--- a/sql/files/defaultdata/compare/compare.cc
+++ b/sql/files/defaultdata/compare/compare.cc
@@ -115,7 +115,7 @@ void compare_float(std::string judge, std::string team, flt jval, flt tval, flt 
 const char *USAGE = "Usage: %s judge_in judge_ans feedback_dir [options] < team_out";
 
 int main(int argc, char **argv) {
-	if(argc < 4) {
+	if (argc < 4) {
 		judge_error(USAGE, argv[0]);
 	}
 	judgemessage = openfeedback(argv[3], "judgemessage.txt", argv[0]);
@@ -199,7 +199,7 @@ int main(int argc, char **argv) {
 				wrong_answer("String tokens mismatch\nJudge: \"%s\"\nTeam: \"%s\"", judge.c_str(), team.c_str());
 			}
 		} else {
-			if(strcasecmp(judge.c_str(), team.c_str()) != 0) {
+			if (strcasecmp(judge.c_str(), team.c_str()) != 0) {
 				wrong_answer("String tokens mismatch\nJudge: \"%s\"\nTeam: \"%s\"", judge.c_str(), team.c_str());
 			}
 		}
--- a/sql/files/defaultdata/compare/compare.cc
+++ b/sql/files/defaultdata/compare/compare.cc
@@ -72,27 +72,44 @@ FILE *openfeedback(const char *feedbackdir, const char *feedback, const char *wh
 	return res;
 }
 
-/* Test two numbers for equality, accounting for +/-INF, NaN and
- * precision. Float f2 is considered the reference value for relative
- * error.
+/* Test two floating-point numbers for equality, accounting for +/-INF, NaN, and precision.
+ * Float `jval` is considered the reference value for relative error.
  */
-int equal(flt f1, flt f2, flt float_abs_tol, flt float_rel_tol)
-{
-	flt absdiff, reldiff;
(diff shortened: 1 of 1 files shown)


sql/files/defaultdata/compare/compare.cc
```diff
@@ -176,16 +153,46 @@ int main(int argc, char **argv) {
 			if (!isfloat(team.c_str(), tval)) {
 				wrong_answer("Expected float, got: %s", team.c_str());
 			}
-			if (!equal(tval, jval, float_abs_tol, float_rel_tol)) {
-				wrong_answer("Too large difference.\n Judge: %s\n Team: %s\n Difference: %Lg\n (abs tol %Lg rel tol %Lg)",
-							 judge.c_str(), team.c_str(), fabsl(jval-tval), float_abs_tol, float_rel_tol);
+			/* Finite values are compared with some tolerance */
```
vmcj, line 156: Is there a way how this can still stay another method? Moving this into the main function is not the right place IMO.

I do like the improvement for the rest though.

sql/files/defaultdata/compare/compare.cc
```diff
@@ -176,16 +153,46 @@ int main(int argc, char **argv) {
 			if (!isfloat(team.c_str(), tval)) {
```
vmcj, line 153: I think you can also take this out as you have: https://github.com/DOMjudge/domjudge/pull/1863/files#diff-30594013bdc266f0b1b51ef87283423017bce078812a75f1c00a6a0716e1b7adR186

sql/files/defaultdata/compare/compare.cc
```diff
@@ -72,33 +72,10 @@ FILE *openfeedback(const char *feedbackdir, const char *feedback, const char *wh
...
-	}
-	/* Values in different classes are always different. */
-	return 0;
-}
-
 const char *USAGE = "Usage: %s judge_in judge_ans feedback_dir [options] < team_out";
 
 int main(int argc, char **argv) {
-	if(argc < 4) {
+	if (argc < 4) {
```
eldering, line 78: Nitpick: could you make these unrelated layout changes in a separate commit?

eldering (commented):
As a general comment: during the ICPC WFs we found a bug in this validator dealing with NULL bytes. I have a fix for that lying around that conflicts with this MR. Please don't merge this right away, so I can see what is the best way forward.

sql/files/defaultdata/compare/compare.cc
mpsijm: Fair point! I've rewritten the function to not return an `int` but to directly invoke `wrong_answer` with the correct message :slightly_smiling_face:

sql/files/defaultdata/compare/compare.cc
mpsijm: This `isfloat` is needed to populate `tval`, which is of type `flt`. Without it, the `isfinite`/`isinf`/`isnan` functions don't have a value for `tval` :slightly_smiling_face: (this should be clearer now that I've re-extracted the `compare_float` function)

sql/files/defaultdata/compare/compare.cc
mpsijm: Sure, done! :smile:

sql/files/defaultdata/compare/compare.cc
vmcj: Check!

vmcj:
@mpsijm I'll set this to a draft, not because I think its not ready yet but to make sure we don't merge by accident waiting for @eldering to take a look at it.

vmcj:
@mpsijm I'll close this one as #1917 is where @eldering improves further on this. If you disagree feel free to re-open.

Issue #405, opened 2018-09-03
Remove role restriction from event feed reading


meisterT:
In theory, we should everyone allow to read the event feed. However, before we can allow that we have to solve two issues:

a) make sure that we don't leak information that's not available through other channels to the teams or the public (e.g. number of test cases per problem), and
b) make sure that we cannot be DOSed easily by opening a handful of event-feed connections (which do not have a timeout).

For now, I'll restrict it to the jury again.

nickygerritsen:
In our current implementation of the new scoreboard we give out the event ID of the latest event, regardless of whether you have access to it. This should also be fixed.

vmcj:
Should we actually do this, in which cases would this be useful as I think the code will be harder to maintain even besides the issues @meisterT mentioned IMO.

meisterT:
We decided that if you actually want more people to read the event feed (with different permission/access levels), you should put a CDS in front of it, see https://tools.icpc.global/.

PR #1867, opened 2023-01-05
API reader has full access to all data

vmcj:
I think we either have to choose to limit on all endpoints (and we discussed this already in https://github.com/DOMjudge/domjudge/issues/405 for the eventfeed which is part of the API in our case) or should just keep it simple, API access is full access.

When there are good usecases where we want to give out some of the info we can re-add this code again (and actually add tests for it).


nickygerritsen (approved):

eldering:
Next time it would be better to add the rationale for the change to the commit message instead of just the PR. Currently the commit message isn't clearly linked to the code change, which is removing logic for team users.

PR #1868, opened 2023-01-05
Test POST of new groups via API

vmcj:
After discussion with @nickygerritsen we agreed that POST should not set the ID to stay in line with the wording in the CLICS spec.

This does cost duplication for the API doc as we now need to stop mentioning the id field in the doc.

### DIFF
--- a/webapp/config/packages/nelmio_api_doc.yaml
+++ b/webapp/config/packages/nelmio_api_doc.yaml
@@ -436,6 +436,28 @@ nelmio_api_doc:
                         lazy_eval_results:
                             type: boolean
                             description: Whether to use lazy evaluation for this problem. Defaults to the global setting
-
+                TeamCategoryPost:
+                    type: object
+                    required: [name]
+                    properties:
+                        hidden:
+                            type: boolean
+                            description: Show this group on the scoreboard?
+                            nullable: true
+                        icpc_id:
+                            type: string
+                            description: The ID in the ICPC CMS for this group
+                            nullable: true
+                        name:
+                            type: string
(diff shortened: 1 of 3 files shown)


webapp/config/packages/nelmio_api_doc.yaml
```diff
@@ -434,6 +434,28 @@ nelmio_api_doc:
                         lazy_eval_results:
                             type: boolean
                             description: Whether to use lazy evaluation for this problem. Defaults to the global setting
-
+                TeamCategoryPost:
```
nickygerritsen, line 437: Can't we, instead of this, use the normal TeamCategory object merged with a generic ID object? I think OpenApi can do that. That saves duplication.

nickygerritsen (approved):

webapp/config/packages/nelmio_api_doc.yaml
vmcj: It would be a negative merge though,

We need the Modal from TeamCategory minus the ID field. I'm not sure if OpenAPI can do that.

nickygerritsen: Sorry, then do it the other way around. Create a TeamCategoryWithoutID and then a merged one

nickygerritsen:
We decided that we can merge this as if, if the conflict is fixed. Will look into it.

vmcj:
> We decided that we can merge this as if, if the conflict is fixed. Will look into it.

I think I fixed the merge conflict,

nickygerritsen (approved):
If CI is green let's merge

PR #1877, opened 2023-01-15
chmod in busybox does not have --reference

vmcj:
This is needed to let domjudge configure/make on alpine linux (busybox like).


eldering (approved):
LGTM

PR #1878, opened 2023-01-18
Add csafe and cppsafe executables with sanitizers support

rmartinsanta:
We are using a modified version of the C executable in our introductory programming course. In contrast to other programming languages, C has a lot of behavior that is undefined but does not throw an immediate error. For example, it does not check if memory accesses are valid, as it is considered the responsibility of the user. Consider the following C code:

```c
#include<stdio.h>
#define max(a,b)	(a<b? b:a)
long long int a[100001]={0},n,i,c=0,d=0,t,x;
int main(){
	scanf("%I64d",&n);
	for(i=0;i<n;i++){
		scanf("%I64d",&x);
		a[x]+=x;
	}
	for(i=1;i<=100001;i++){
		t=c;
		c=max(c,d+a[i]),d=t;
	}
	printf("%I64d",c);
	return 0;
}
```

While it may be fit for purpose in a competitive programming context, it clearly contains an out of bounds access to the array, due to looping `i<=100001` instead of `i<100001`. The instructor may want to automatically detect and consider this kind of mistakes as not correct, and may want to force the user to fix the bug. Example in our Domjudge instance changing the C executable for the Csafe version, changes the veredict from correct to runtime error, and prints the line where the invalid access was produced.

image

Additional notes:
- We do not want to replace the current C and cpp executables, we would like to allow Domjudge administrators to decide which behavior is preferable for their needs.
`-static ` flag was removed from the script as it was incompatible with some sanitizer checks and I am not sure why it is necessary. 
- As we have manually configured it in our Domjudge instance as an executable, we do not know if the changes included in this PR are enough to make it work by default for new Domjudge instances, or even if the folder structure is correct. Do not hesitate to correct us
- An idea for further improvement may be creating custom error codes for different sanitizer failures: right now all fall under Runtime Error

Any suggestion or feedback is welcome. Thanks @isaaclo97 for the help testing and validating the idea.

### DIFF
--- a/sql/files/defaultdata/cppsafe/build
+++ b/sql/files/defaultdata/cppsafe/build
@@ -1,0 +1,2 @@
+#!/bin/sh
+# nothing to compile
--- a/sql/files/defaultdata/cppsafe/run
+++ b/sql/files/defaultdata/cppsafe/run
@@ -13,7 +13,7 @@ MEMLIMIT="$1" ; shift
 #           other languages autodetected by extension)
 # -Wall:    Report all warnings
 # -fsanitize: Enable the specified Sanitizers, see https://github.com/google/sanitizers for the full documentation
-# -fPIE -pie -fno-omit-frame-pointer -fno-optimize-sibling-calls: Make code position independant and disable optimizations that may obfuscate the info provided by the 
+# -fPIE -pie -fno-omit-frame-pointer -fno-optimize-sibling-calls: Make code position independent and disable optimizations that may obfuscate the info provided by the 
 sanitizers
 # -g: Include debugging/symbol information in the executable, improves sanitizers output if an error is detected
 # -O2:      Level 2 optimizations (default for speed)
--- a/sql/files/defaultdata/cppsafe/run
+++ b/sql/files/defaultdata/cppsafe/run
@@ -1,0 +1,22 @@
+#!/bin/sh
+
+# C++ compile wrapper-script for 'compile.sh'.
+# See that script for syntax and more info.
+
+DEST="$1" ; shift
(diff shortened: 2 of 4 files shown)


vmcj:
I can see the use case, but what is the reasoning to not add these comments to the main language. An administrator can change the executable and add these options already. In the current form this is duplication and the admin can already change these options via the UI whereas this PR still needs a change to the C/C++ language.

eldering:
I agree with @vmcj: I can see the use case for this in a teaching environment, but I'd rather not add this as an extra language to the main DOMjudge codebase. In the past we've had similar requests to e.g. add a version of the C or C++ compile scripts that link against specific libraries, such as `libgmp`. If we add these, then we'd slowly end up with a large collection of compile scripts for the same languages, all with slight variations, and we could even get into questions of creating possible combinations of these.
I think instead that it is better to leave it up to the administrator of an installation to add/modify versions of the compile scripts that they require. Various useful options to consider and how to do so could for example be added to https://github.com/DOMjudge/domjudge/wiki/Adding-a-new-language or a similar page, and we could clearly reference that from the admin manual.

vmcj:
Hi @rmartinsanta & @isaaclo97,

I've created https://github.com/DOMjudge/domjudge/wiki/DOMjudge-in-education which I'll link from the administrator manual. 

As explained by @eldering we don't see a proper solution yet to fully incorporate your change. 

I'll close this PR for now, but feel free to discuss what works best to promote these usecases to other educators in our chat (https://www.domjudge.org/chat). The current Slack has other users which also use DOMjudge in classes so maybe they might have ideas on what is the best way to share this knowledge.

PR #1879, opened 2023-01-22
Distinguish between API and web UI when determining last login time. Fixes #1561.

### LINKED ISSUE #1561: Separate API last login vs UI
We keep track of when a user last logged in. However, this is one value, both indicating API access as well as UI access.
During SWERC we noticed this is not ideal: SWERC used the [CCS autologin LightDM](https://github.com/GEHACK/lightdm-qt5-greeter-ccs-api) and IP autologin. This means machines are checking the API for contest start every 15 seconds and is auto logging in the users, which means the last login field is set every 15 seconds.
Although this was useful (to see if any machine didn't 'call home'), we would also like to know when a user logged in using the UI.

So the easiest approach would be to have two fields, one for API and one for UI.

Thoughts?

nickygerritsen:


nickygerritsen:
Ignore the spellcheck failing, this is intended.

webapp/src/Security/UserStateUpdater.php
```diff
@@ -31,7 +31,15 @@ public static function getSubscribedEvents(): array
...
-            $user->setLastLogin(Utils::now());
+            $firewallName = 'main';
+            if (method_exists($event->getAuthenticationToken(), 'getFirewallName')) {
+                $firewallName = $event->getAuthenticationToken()->getFirewallName();
+            }
+            if ($firewallName === 'main') {
+                $user->setLastLogin(Utils::now());
+            } elseif ($firewallName === 'api') {
+                $user->setLastApiLogin(Utils::now());
+            }
```
vmcj, line 42: Consider adding an `else` so we can detect if we add new firewalls later?

nickygerritsen: We *have* more firewalls, like `metrics`. What should we do with them? At first I had main => lastlogin else => lastapilogin, but does that make sense?

vmcj: metrics is also over the API so for that specific case it would work, but with another firewall this would not hold anymore. Maybe this is indeed the best solution.

nickygerritsen: Metrics has a different firewall though. Which I think we should add here, let me do it

webapp/templates/jury/user.html.twig
```diff
@@ -116,6 +116,16 @@
                         {% endif %}
                     </td>
                 </tr>
+                <tr>
+                    <th>Last API login</th>
+                    <td>
+                        {% if user.lastApiLogin %}
+                            {{ user.lastApiLogin | printtime('D d M Y H:i:s T') }}
```
vmcj, line 123: Can't we use a default (https://twig.symfony.com/doc/2.x/filters/default.html) here?

vmcj (approved):

webapp/templates/jury/user.html.twig
nickygerritsen: Can we? I don' think we can. This would become:

```
{{ user.lastApiLogin | default(whatgoesherethatmakessensebutwecanstillcallprinttimeon) | printtime('D d M Y H:i:s T') }}
```

If you know a good default tell me, but null isn't one, since that would print the current date/time

vmcj: I had hoped `default('-')` but that might need a change to the `printtime` function to handle that specific case.

meisterT (approved):

PR #1881, opened 2023-01-22
Allow only one external contest per contest. Fixes #1842.

### LINKED ISSUE #1842: Allow exactly one external contest source per contest

Currently, you can add multiple external contest sources per DOMjudge internal contest. This is not only conceptually not required but also leads to weird UI/UX issues:
image

Clicking on the warnings in the external contest sources menu currently brings you to the "external contest source" overview instead of the actual contest source that caused the warning.

nickygerritsen:
This also renames the pages, URL's, etc from 'external contest source' to just 'external contest'. The database table is still called external contest source though. Should we also rename that? Or we could move the fields to the contest table since it's basically a one-to-one, but I'm not sure about that. Input is very welcome.

It also makes it such that the event feed importer accepts the contest ID instead of external contest source ID, since that now uniquely identifies it and in that way we don't have to expose the ID's of these sources anymore.


meisterT (approved):
I like the simplification in the code and the web interface!

PR #1884, opened 2023-01-23
Don't break on URLencoded logging entries

vmcj:
Probably this should also be fixed in the calling function to check if there are not more % symbols being injected. Normally we would need to verify the actual tried fileaccess (Mayhem tries to read etc/passwd) but as this printf is only used for logging we seem to be safe.

Fixes: https://github.com/DOMjudge/domjudge/security/code-scanning/451

### DIFF
--- a/webapp/src/Logger/VarargsLogMessageProcessor.php
+++ b/webapp/src/Logger/VarargsLogMessageProcessor.php
@@ -7,6 +7,7 @@
 
 namespace App\Logger;
 
+use ValueError;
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

eldering:
In general it's not clear to me that this is the correct fix. Any content that may contain variable or user input should be part of `$record['context']` passed into `vsprintf` as arguments. Where are these URL encoded strings coming from and why do they cause an error?

PR #1886, opened 2023-01-24
setsid() failure is not an error

edomora97:
The setsid() syscall can fail only with EPERM, which means that the process is already the process group leader. This is not an error since we are trying to become the leader.

See `man 2 setsid`.

---

In particular `setsid()` fails when I try to run runpipe from the terminal. Probably bash sets the group leader before exec-ing. While running inside DJ the runpipe process is not the leader so it doesn't fail to setsid.

### DIFF
--- a/judge/runpipe.cc
+++ b/judge/runpipe.cc
@@ -887,7 +887,7 @@ int main(int argc, char **argv) {
 
   // Enter a new session since we are dealing with signals.
   if (setsid() < 0) {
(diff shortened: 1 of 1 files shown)


eldering (approved):

edomora97:
> This also turns any potential future other error codes into warnings, but given that I'd assume this to be a pretty stable interface, I think that's ok.

Yeah, I don't expect it to change anytime soon™.

PR #1889, opened 2023-01-26
Fix VarargsLogmessage error handling

eldering:


vmcj (approved):
I think this makes sense.

PR #1904, opened 2023-02-10
Limit routes if we only expose identifiers with digits

vmcj:
I'm not 100% sure what happens for some of those fields when we shadow as there the primary is in charge of setting the ID.

### DIFF
--- a/webapp/src/Controller/API/ContestController.php
+++ b/webapp/src/Controller/API/ContestController.php
@@ -175,7 +175,7 @@ class ContestController extends AbstractRestController
 
     /**
      * Get the banner for the given contest.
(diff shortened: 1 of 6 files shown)


webapp/src/Controller/API/ContestController.php
```diff
@@ -185,20 +185,20 @@ public function singleAction(Request $request, string $cid): Response
      * )
      * @OA\Parameter(ref="#/components/parameters/id")
```
nickygerritsen, line 186: Shouldn't we then drop this line?

nickygerritsen (approved):
All these are always internal ID's, only configuration data (teams, contests, etc.) can be external and thus be a string.

PR #1905, opened 2023-02-10
contest option to change tiebreaker from penalty to runtime

chistopher:
Hi. We have been using domjudge for 6 years now in our ICPC lectures.
However, we usually have weekly tasks for the student to solve and the penalty tiebreaker by domjudge doesn't fit well with this. That's why we show the runtime of correct submissions on the scoreboard and rank by summed time.
We found it a great tool to teach the students how practical runtime differs from asymptotics and how much of a performance impact different changes can have. 

We have been maintaining these changes on our own fork for some time but I decided to clean it up and submit a PR.
I would appreciate any feedback or help since this is the first PHP project I'm working on.

### DIFF
--- a/webapp/src/Entity/Contest.php
+++ b/webapp/src/Entity/Contest.php
@@ -295,12 +295,11 @@ class Contest extends BaseApiEntity implements AssetEntityInterface
     private bool $processBalloons = true;
 
     /**
-     * @ORM\Column(type="boolean", name="order_by_runtime",
+     * @ORM\Column(type="boolean", name="runtime_as_score_tiebreaker",
      *     options={"comment"="Is runtime used as tiebreaker instead of penalty?","default"=0},
      *     nullable=false)
-     * @Serializer\Exclude()
      */
-    private bool $order_by_runtime = false;
+    private bool $runtime_as_score_tiebreaker = false;
 
     /**
      * @ORM\Column(type="boolean", name="public",
@@ -730,15 +729,15 @@ class Contest extends BaseApiEntity implements AssetEntityInterface
         return $this->processBalloons;
     }
 
-    public function setOrderByRuntime(bool $orderByRuntime): Contest
+    public function setRuntimeAsScoreTiebreaker(bool $runtimeAsScoreTiebreaker): Contest
     {
-        $this->order_by_runtime = $orderByRuntime;
+        $this->runtime_as_score_tiebreaker = $runtimeAsScoreTiebreaker;
         return $this;
     }
 
-    public function getOrderByRuntime(): bool
+    public function getRuntimeAsScoreTiebreaker(): bool
     {
-        return $this->order_by_runtime;
+        return $this->runtime_as_score_tiebreaker;
     }
 
     public function setMedalsEnabled(bool $medalsEnabled): Contest
--- a/webapp/src/Entity/Contest.php
+++ b/webapp/src/Entity/Contest.php
@@ -295,6 +295,14 @@ class Contest extends BaseApiEntity implements AssetEntityInterface
     private bool $processBalloons = true;
 
     /**
+     * @ORM\Column(type="boolean", name="order_by_runtime",
+     *     options={"comment"="Is runtime used as tiebreaker instead of penalty?","default"=0},
+     *     nullable=false)
+     * @Serializer\Exclude()
+     */
+    private bool $order_by_runtime = false;
+
+    /**
      * @ORM\Column(type="boolean", name="public",
      *     options={"comment"="Is this contest visible for the public?",
      *              "default"=1},
@@ -720,6 +728,17 @@ class Contest extends BaseApiEntity implements AssetEntityInterface
     public function getProcessBalloons(): bool
     {
         return $this->processBalloons;
+    }
+
+    public function setOrderByRuntime(bool $orderByRuntime): Contest
+    {
+        $this->order_by_runtime = $orderByRuntime;
+        return $this;
+    }
+
+    public function getOrderByRuntime(): bool
+    {
+        return $this->order_by_runtime;
     }
 
     public function setMedalsEnabled(bool $medalsEnabled): Contest
--- a/webapp/src/Form/Type/ContestType.php
+++ b/webapp/src/Form/Type/ContestType.php
@@ -94,6 +94,14 @@ class ContestType extends AbstractExternalIdEntityType
             ],
             'help' => 'Disable this to stop recording balloons. Usually you can just leave this enabled.',
         ]);
+        $builder->add('orderByRuntime', ChoiceType::class, [
+            'expanded' => true,
+            'choices' => [
+                'Yes' => true,
+                'No' => false,
+            ],
+            'help' => 'Enable this to show runtimes in milliseconds on scoreboard and use them as tiebreaker instead of penalty.',
+        ]);
         $builder->add('medalsEnabled', ChoiceType::class, [
             'expanded' => true,
             'choices' => [
--- a/webapp/src/Form/Type/ContestType.php
+++ b/webapp/src/Form/Type/ContestType.php
@@ -100,7 +100,7 @@ class ContestType extends AbstractExternalIdEntityType
                 'Yes' => true,
                 'No' => false,
             ],
-            'help' => 'Enable this to show runtimes in milliseconds on scoreboard and use them as tiebreaker instead of penalty. The runtime of a submission is the maximum over all testcases.',
+            'help' => 'Enable this to show runtimes in seconds on scoreboard and use them as tiebreaker instead of penalty. The runtime of a submission is the maximum over all testcases.',
         ]);
         $builder->add('medalsEnabled', ChoiceType::class, [
             'expanded' => true,
--- a/webapp/src/Form/Type/ContestType.php
+++ b/webapp/src/Form/Type/ContestType.php
@@ -100,7 +100,7 @@ class ContestType extends AbstractExternalIdEntityType
                 'Yes' => true,
                 'No' => false,
             ],
-            'help' => 'Enable this to show runtimes in milliseconds on scoreboard and use them as tiebreaker instead of penalty.',
+            'help' => 'Enable this to show runtimes in milliseconds on scoreboard and use them as tiebreaker instead of penalty. The runtime of a submission is the maximum over all testcases.',
         ]);
         $builder->add('medalsEnabled', ChoiceType::class, [
             'expanded' => true,
--- a/webapp/src/Form/Type/ContestType.php
+++ b/webapp/src/Form/Type/ContestType.php
@@ -94,7 +94,7 @@ class ContestType extends AbstractExternalIdEntityType
             ],
             'help' => 'Disable this to stop recording balloons. Usually you can just leave this enabled.',
         ]);
-        $builder->add('orderByRuntime', ChoiceType::class, [
+        $builder->add('runtimeAsScoreTiebreaker', ChoiceType::class, [
             'expanded' => true,
             'choices' => [
                 'Yes' => true,
--- a/webapp/src/Service/ScoreboardService.php
+++ b/webapp/src/Service/ScoreboardService.php
@@ -359,7 +359,7 @@ class ScoreboardService
             // STEP 1:
             // runtime improvements should be possible for all correct submissions
             if (!is_null($judging) && $judging->getResult() == Judging::RESULT_CORRECT) {
-                $runtime = (int) round(1000*$judging->getMaxRuntime()); // round to milliseconds
+                $runtime = (int) floor(1000*$judging->getMaxRuntime()); // round to milliseconds
                 $runtimeJury = min($runtimeJury, $runtime);
                 if (!$submission->isAfterFreeze()) {
                     $runtimePubl = min($runtimePubl, $runtime);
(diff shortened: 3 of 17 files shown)


webapp/src/Entity/Contest.php
```diff
@@ -294,6 +294,14 @@ class Contest extends BaseApiEntity implements AssetEntityInterface
      */
     private bool $processBalloons = true;
 
+    /**
+     * @ORM\Column(type="boolean", name="order_by_runtime",
+     *     options={"comment"="Is runtime used as tiebreaker instead of penalty?","default"=0},
+     *     nullable=false)
+     * @Serializer\Exclude()
```
nickygerritsen, line 301: Perhaps we do want to expose this in the API?

webapp/src/Entity/Contest.php
```diff
@@ -294,6 +294,14 @@ class Contest extends BaseApiEntity implements AssetEntityInterface
      */
     private bool $processBalloons = true;
 
+    /**
+     * @ORM\Column(type="boolean", name="order_by_runtime",
+     *     options={"comment"="Is runtime used as tiebreaker instead of penalty?","default"=0},
```
nickygerritsen, line 299: I wonder if this should improve, since it's not only ordering by it but also showing it on the scoreboard. Not sure what's better though.

webapp/src/Service/ScoreboardService.php
```diff
@@ -148,7 +148,11 @@ public function calculateTeamRank(
         $restricted = ($jury || $freezeData->showFinal(false));
         $variant    = $restricted ? 'restricted' : 'public';
         $points     = $rankCache ? $rankCache->getPointsRestricted() : 0;
-        $totalTime  = $rankCache ? $rankCache->getTotaltimeRestricted() : 0;
+        $totalTime  = 0;
+        if($rankCache) {
```
nickygerritsen, line 152: phpcs already tells you, but can you add a space after the `if` before the `(`?

webapp/src/Utils/Scoreboard/Scoreboard.php
```diff
@@ -292,18 +298,24 @@ protected static function scoreboardCompare(TeamScore $a, TeamScore $b): int
...
+    protected function scoreCompare(TeamScore $a, TeamScore $b): int
     {
         // More correctness points than someone else means higher rank.
         if ($a->numPoints != $b->numPoints) {
             return $b->numPoints <=> $a->numPoints;
         }
         // Else, less time spent means higher rank.
-        if ($a->totalTime != $b->totalTime) {
-            return $a->totalTime <=> $b->totalTime;
+        if($this->getOrderByRuntime()) { // runtime ordering
```
nickygerritsen, line 311:
```suggestion
        if ($this->getOrderByRuntime()) { // runtime ordering
```

nickygerritsen (commented):
There are some small remarks and PHPcs has some opinions which you should fix, but I like the idea in general.

Two things:
* Could you upload a screenshot?
* We should probably add it to the `API\ScoreboardController` class as well for API users.

In general I think this feature is useful and is not too complex too maintain personally, but let's await the ohers.

chistopher:
Thanks. I'll look into it on Monday. The PR changes the following.
If a contest has the `order_by_runtime` option enabled then:
- the scoreboard cells show the runtime of the fastest submission in ms (this can be improved by submitting after the first accept)
- the score shows the summed runtime over all solved problems instead of penalty
- the fastest submission is marked in dark green instead of the first solved
- on the team site, the submission list includes a column for runtime (only if accept)
- and it's again shown in the submission modal

Warning: as for now, the runtime is always shown to teams (independent of contest option) which would change default domjudge behavior. This is just because I did not know how to disable it for contests without the option set :(
However, maybe you like the idea to show this data to the teams. After all, codeforces does it too.

Here are some pics :)
image
image
image

nickygerritsen:
> Warning: as for now, the runtime is always shown to teams (independent of contest option) which would change default domjudge behavior. This is just because I did not know how to disable it for contests without the option set :( However, maybe you like the idea to show this data to the teams. After all, codeforces does it too.

good question, we need to think about this. I’m not sure yet if it would “hurt”

eldering:
> > Warning: as for now, the runtime is always shown to teams (independent of contest option) which would change default domjudge behavior. This is just because I did not know how to disable it for contests without the option set :( However, maybe you like the idea to show this data to the teams. After all, codeforces does it too.
> 
> good question, we need to think about this. I’m not sure yet if it would “hurt”

I think this definitely changes things: teams can learn very relevant information from the runtime of another team's solution: for example make an estimate on the complexity of the solution. So I don't think we should expose this by default.

webapp/src/Entity/Contest.php
eldering: I'd name it something like `runtime_as_score_tiebreaker`

webapp/src/Service/ScoreboardService.php
```diff
@@ -335,6 +340,34 @@ public function calculateScoreRow(
         $submissionsPubl = $pendingPubl = $timePubl = 0;
         $correctJury     = false;
         $correctPubl     = false;
+        $runtimeJury     = PHP_INT_MAX;
+        $runtimePubl     = PHP_INT_MAX;
+
+        // do runtime separately to not interfere with break/continue logic in normal submission counting
+        // we want to allow runtime improvements even after first correct submission
```
eldering, line 347: I think it would be cleaner to not duplicate this loop, but instead make the `continue/break` logic depend on the scoring type. Creating code duplication is asking for bugs when we change something in one place and overlook the other.

nickygerritsen:
> > > Warning: as for now, the runtime is always shown to teams (independent of contest option) which would change default domjudge behavior. This is just because I did not know how to disable it for contests without the option set :( However, maybe you like the idea to show this data to the teams. After all, codeforces does it too.
> > 
> > 
> > good question, we need to think about this. I’m not sure yet if it would “hurt”
> 
> I think this definitely changes things: teams can learn very relevant information from the runtime of another team's solution: for example make an estimate on the complexity of the solution. So I don't think we should expose this by default.

Wait maybe I misunderstood. @chistopher it only shows this for your own team right? On the team page. The ones on the scoreboard only show if you enable the flag? If that’s not the case I agree with Jaap

chistopher:
> > > > Warning: as for now, the runtime is always shown to teams (independent of contest option) which would change default domjudge behavior. This is just because I did not know how to disable it for contests without the option set :( However, maybe you like the idea to show this data to the teams. After all, codeforces does it too.
> > > 
> > > 
> > > good question, we need to think about this. I’m not sure yet if it would “hurt”
> > 
> > 
> > I think this definitely changes things: teams can learn very relevant information from the runtime of another team's solution: for example make an estimate on the complexity of the solution. So I don't think we should expose this by default.
> 
> Wait maybe I misunderstood. @chistopher it only shows this for your own team right? On the team page. The ones on the scoreboard only show if you enable the flag? If that’s not the case I agree with Jaap

You are right. It only shows on the team page for the team itself. The scoreboard does not change at all if the option is disabled (which it is by default).

chistopher:
I made the changes you requested.
- renamed the option from `order_by_runtime` to `runtime_as_score_tiebreaker`
- the `/contests/<id>/scoreboard` API endpoint now provides either time/solved_first (as before) or runtime/fastest_implementation (if the contest tiebreaks by runtime).
- the `runtime_as_score_tiebreaker` contest option is visible in the API
- scoreboard computation logic does runtime computation during the same loop as submission counting

webapp/src/Service/ScoreboardService.php
```diff
@@ -344,6 +351,28 @@ public function calculateScoreRow(
...
 
+            // three things will happen in the loop in this order:
+            // 1. update fastest runtime
+            // 2. count submissions until correct submission
+            // 3. determine time of first correct submission
+
+            // STEP 1:
+            // runtime improvements should be possible for all correct submissions
+            if (!is_null($judging) && $judging->getResult() == Judging::RESULT_CORRECT) {
+                $runtime = (int) round(1000*$judging->getMaxRuntime()); // round to milliseconds
```
eldering, line 362: Nitpick: I think the correct thing to do here is to `floor`, not `round`. This is in general how time is dealt with. A clock showing hours and minutes doesn't display the next minute when it's past 30 seconds into that minute.

webapp/src/Service/ScoreboardService.php
```diff
@@ -386,21 +415,16 @@ public function calculateScoreRow(
...
             }
 
+            // STEP 3:
             $absSubmitTime = (float)$submission->getSubmittime();
             $submitTime    = $contest->getContestTime($absSubmitTime);
 
-            // if correct, don't look at any more submissions after this one.
             if ($judging->getResult() == Judging::RESULT_CORRECT) {
                 $correctJury = true;
                 $timeJury    = $submitTime;
```
eldering, line 424: This should not get updated anymore once set, as otherwise it would track the latest submission time, not the first.

webapp/src/Service/ScoreboardService.php
```diff
@@ -464,19 +488,21 @@ public function calculateScoreRow(
             'submissionsRestricted' => $submissionsJury,
             'pendingRestricted' => $pendingJury,
             'solvetimeRestricted' => (int)$timeJury,
+            'runtimeRestricted' => $runtimeJury === PHP_INT_MAX ? 0 : $runtimeJury,
```
eldering, line 491: Maybe better to use a sentinel value of `-1` or just `null`?

webapp/src/Utils/Scoreboard/Scoreboard.php
```diff
@@ -438,4 +450,28 @@ public function solvedFirst(Team $team, ContestProblem $problem): bool
     {
         return $this->matrix[$team->getTeamid()][$problem->getProbid()]->isFirst;
     }
+
+
+    /**
+     * Determine whether this team has the fastest correct implementation for this problem
+     */
+    public function isFastestImplementation(Team $team, ContestProblem $problem): bool
```
eldering, line 458: Nitpick: let's stick to existing terminology:
```suggestion
    public function isFastestSubmission(Team $team, ContestProblem $problem): bool
```

webapp/src/Service/ScoreboardService.php
chistopher: It will not be updated again because of the `continue` in line 415.

chistopher:
Thanks for the feedback. I'll look into it later.

webapp/src/Service/ScoreboardService.php
chistopher: I use PHP_INT_MAX because it is a neutral element for `min` and I can write 
```php
$runtimeJury = min($runtimeJury, $runtime);
...
$runtimePubl = min($runtimePubl, $runtime);
```
instead of
```php
if (is_null($runtimeJury) || $runtimeJury > $runtime) {
    $runtimeJury = $runtime;
}
...
if (is_null($runtimePubl) || $runtimePubl > $runtime) {
    $runtimePubl = $runtime;
}
```

eldering: To clarify: I wasn't referring to the `PHP_INT_MAX` (I agree with that), but the `0` after the `?`.

webapp/src/Form/Type/ContestType.php
```diff
@@ -94,6 +94,14 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             ],
             'help' => 'Disable this to stop recording balloons. Usually you can just leave this enabled.',
         ]);
+        $builder->add('runtimeAsScoreTiebreaker', ChoiceType::class, [
+            'expanded' => true,
+            'choices' => [
+                'Yes' => true,
+                'No' => false,
+            ],
+            'help' => 'Enable this to show runtimes in milliseconds on scoreboard and use them as tiebreaker instead of penalty.',
```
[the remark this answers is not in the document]
chistopher: Help is updated. :+1: 

Regarding the runtime in seconds, here is a picture how it would look like.
image
I think it looks more clean in milliseconds but if you prefer it in seconds I will commit the change.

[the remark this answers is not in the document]
eldering: Just discussed this with @meisterT and he agreed that like this in seconds looks clearer.

My reasoning is that these are standard units; also, it makes it easier later to change the display format for submissions where the runtime is tens or hundreds of seconds and we might truncate some decimals.

webapp/src/Service/ScoreboardService.php
chistopher: Ah, sorry. 
Idk if null would be good there. I used 0 to keep the `runtime_...` fields of ScoreCache consistent with the `solvetime_...` fields and those use 0 for unsolved problems and have `nullable=false` in the DB.

webapp/src/Form/Type/ContestType.php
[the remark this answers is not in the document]
chistopher: done :)

webapp/src/Service/ScoreboardService.php
eldering: Ah, ok, sorry: I wasn't aware that these are not nullable and already follow this behaviour. Then please ignore this.

eldering (approved):
LGTM, let's wait if @nickygerritsen has any comments left.

PR #1908, opened 2023-02-15
Fix maintainer-install on Ubuntu22.04

vmcj:
This was removed in 7db919726b527fbf4fae9c44b4fc26078d4361eb but I need this for cmap.sty on 22.04 after an upgrade from 20.04.


nickygerritsen (approved):
Yep I also need latex to build stuff, so I think this makes sense.

eldering (approved):

PR #1914, opened 2023-02-17
Replace Toggle buttons with toggle checkboxes.

nickygerritsen:


webapp/templates/jury/jury_macros.twig
```diff
@@ -18,6 +18,18 @@
...
 {% endmacro %}
 
+{% macro toggle_autosubmit_extrafooter() %}
+    <script>
+        $(function() {
+            $('[data-toggle]').on('change', function() {
+                var $form = $(this).closest('form');
+                setTimeout(function() {
+                    $form.submit();
+                }, 300);
```
meisterT, line 28: nit: add a comment here why we need that

meisterT (approved):

PR #1915, opened 2023-02-17
Add column/link to executable to the language table.

meisterT:
It is not very much visible now that this is a link.


nickygerritsen (approved):
I do agree with the annotations this time 😄

PR #1917, opened 2023-02-17
Improve compare script

eldering:
This combines #1863 with changes of mine that fix a bug comparing files containing null bytes.

### DIFF
--- a/sql/files/defaultdata/compare/compare.cc
+++ b/sql/files/defaultdata/compare/compare.cc
@@ -72,33 +72,50 @@ FILE *openfeedback(const char *feedbackdir, const char *feedback, const char *wh
 	return res;
 }
 
-/* Test two numbers for equality, accounting for +/-INF, NaN and
- * precision. Float f2 is considered the reference value for relative
- * error.
+/* Test two floating-point numbers for equality, accounting for +/-INF, NaN, and precision.
+ * Float `jval` is considered the reference value for relative error.
  */
-int equal(flt f1, flt f2, flt float_abs_tol, flt float_rel_tol)
-{
-	flt absdiff, reldiff;
+void compare_float(std::string judge, std::string team, flt jval, flt tval, flt float_abs_tol, flt float_rel_tol) {
 	/* Finite values are compared with some tolerance */
-	if ( std::isfinite(f1) && std::isfinite(f2) ) {
-		absdiff = fabsl(f1-f2);
-		reldiff = fabsl((f1-f2)/f2);
-		return !(absdiff > float_abs_tol && reldiff > float_rel_tol);
-	}
+	if (std::isfinite(tval) && std::isfinite(jval)) {
+		flt absdiff = fabsl(tval-jval);
+		flt reldiff = fabsl((tval-jval)/jval);
+		if (float_abs_tol >= 0 && float_rel_tol >= 0) {
+			if (absdiff > float_abs_tol || reldiff > float_rel_tol) {
+				wrong_answer("Too large difference.\n Judge: %s\n Team: %s\n Absolute difference: %Lg (tolerance: %Lg%s)\n Relative difference: %Lg (tolerance: %Lg%s)",
+							 judge.c_str(), team.c_str(),
+							 absdiff, float_abs_tol, absdiff > float_abs_tol ? ", exceeded" : "",
+							 reldiff, float_rel_tol, reldiff > float_rel_tol ? ", exceeded" : "");
+			}
+		} else if (float_abs_tol >= 0) {
+			if (absdiff > float_abs_tol) {
+				wrong_answer("Too large difference.\n Judge: %s\n Team: %s\n Absolute difference: %Lg (tolerance: %Lg, exceeded)",
+							 judge.c_str(), team.c_str(), absdiff, float_abs_tol);
+			}
+		} else if (float_rel_tol >= 0) {
+			if (reldiff > float_rel_tol) {
+				wrong_answer("Too large difference.\n Judge: %s\n Team: %s\n Relative difference: %Lg (tolerance: %Lg, exceeded)",
+							 judge.c_str(), team.c_str(), reldiff, float_rel_tol);
+			}
+		}
 	/* NaN is equal to NaN */
-	if ( std::isnan(f1) && std::isnan(f2) ) return 1;
+	} else if (std::isnan(jval) && std::isnan(tval)) {
+		return;
 	/* Infinite values are equal if their sign matches */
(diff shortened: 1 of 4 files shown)


nickygerritsen (approved):

sql/files/defaultdata/compare/compare.cc
```diff
@@ -171,22 +198,42 @@ int main(int argc, char **argv) {
 			wrong_answer("User EOF while judge had more output\n(Next judge token: %s)", judge.c_str());
 		}
 
+		std::string extra_msg = "";
+		bool nonprintable = false;
+		for(char c : judge) {
```
meisterT, line 203: nit: do you want to add a ` ` here for consistency reasons? Also below then.

meisterT:
The test failure in `testFetchExecutable` looks related.

sql/files/defaultdata/compare/compare.cc
```diff
@@ -72,30 +73,39 @@ FILE *openfeedback(const char *feedbackdir, const char *feedback, const char *wh
 	return res;
 }
 
+bool equal_case_insensitive(std::string a, std::string b)
+{
+	for(size_t i=1; i<a.length(); i++) a[i] = tolower(a[i]);
+	for(size_t i=1; i<b.length(); i++) b[i] = tolower(b[i]);
```
mzuenni, line 79: nit: notice that calling ```tolower(char)``` is actually undefined behaviour see [the reference](https://en.cppreference.com/w/cpp/string/byte/tolower). And you probably want to write ```std::tolower```.

eldering: Thanks, should be fixed now.

sql/files/defaultdata/compare/compare.cc
```diff
@@ -72,30 +73,45 @@ FILE *openfeedback(const char *feedbackdir, const char *feedback, const char *wh
...
+// The behavior of std::tolower on (signed) char is undefined.
+char tolower_char(char c)
+{
+    return static_cast<char>(std::tolower(static_cast<unsigned char>(c)));
+}
+
+bool equal_case_insensitive(std::string a, std::string b)
+{
+	for (size_t i=1; i<a.length(); i++) a[i] = tolower_char(a[i]);
+	for (size_t i=1; i<b.length(); i++) b[i] = tolower_char(b[i]);
```
mpsijm, line 85: Why do these loops start at `1`, and not at `0`? If I now print `Impossible` instead of `impossible`, this is not allowed according to this validator, even though it should be case-insensitive :sweat_smile: 

(big thanks to `NickyAndMark.java` at problem I of BAPC 2022 for triggering this :stuck_out_tongue:)

mpsijm (changes requested):
I've checked whether `validator.cc` still has the same behaviour on the problems that I mentioned in https://github.com/DOMjudge/domjudge/pull/1863#issue-1515145386. For float values, my fix still seems to work fine :smile: However, I did find a bug regarding the case-insensitive validation of strings:

sql/files/defaultdata/compare/compare.cc
nickygerritsen: These Nicky and Mark are great guys

mpsijm (approved):
Thanks for the fix! :smile:

sql/files/defaultdata/compare/compare.cc
```diff
@@ -72,33 +73,65 @@ FILE *openfeedback(const char *feedbackdir, const char *feedback, const char *wh
...
-	if ( std::isfinite(f1) && std::isfinite(f2) ) {
-		absdiff = fabsl(f1-f2);
-		reldiff = fabsl((f1-f2)/f2);
-		return !(absdiff > float_abs_tol && reldiff > float_rel_tol);
-	}
+	if (std::isfinite(tval) && std::isfinite(jval)) {
+		flt absdiff = fabsl(tval-jval);
+		flt reldiff = fabsl((tval-jval)/jval);
+		if (float_abs_tol >= 0 && float_rel_tol >= 0) {
+			if (absdiff > float_abs_tol || reldiff > float_rel_tol) {
```
nickygerritsen, line 99: After running this validator on BAPC and talking to @mpsijm over a Slack Huddle at the same time, we found out this should be:

```suggestion
			if (absdiff > float_abs_tol && reldiff > float_rel_tol) {
```

otherwise it fails if a submission is wrong EITHER relative or absolute instead or BOTH.

mpsijm: And just for completeness: this makes it consistent with [Kattis' default validator](https://github.com/Kattis/problemtools/blob/develop/support/default_validator/default_validator.cc#L150-L151). Note that they write `!( <= )` instead of simply `>`, but it should indeed be the same now. Thanks for finding this one! :smile:

nickygerritsen:
I rejudged all of last years BAPC and except the above float fix I only had some explainable differences, so looks good.

sql/files/defaultdata/compare/compare.cc
```diff
@@ -72,33 +73,65 @@ FILE *openfeedback(const char *feedbackdir, const char *feedback, const char *wh
...
-	}
+	if (std::isfinite(tval) && std::isfinite(jval)) {
+		flt absdiff = fabsl(tval-jval);
+		flt reldiff = fabsl((tval-jval)/jval);
+		if (float_abs_tol >= 0 && float_rel_tol >= 0) {
+			if (absdiff > float_abs_tol && reldiff > float_rel_tol) {
+				wrong_answer("Too large difference.\n Judge: %s\n Team: %s\n Absolute difference: %Lg (tolerance: %Lg%s)\n Relative difference: %Lg (tolerance: %Lg%s)%s",
+				             judge.c_str(), team.c_str(),
+				             absdiff, float_abs_tol, absdiff > float_abs_tol ? ", exceeded" : "",
+				             reldiff, float_rel_tol, reldiff > float_rel_tol ? ", exceeded" : "",
```
mpsijm, line 103: I'm gonna be very nit-picky and refactor these conditionals now. Looking at it again, the conditions are always `true` :grin: 
```suggestion
				wrong_answer("Too large difference.\n Judge: %s\n Team: %s\n Absolute difference: %Lg (tolerance: %Lg, exceeded)\n Relative difference: %Lg (tolerance: %Lg, exceeded)%s",
				             judge.c_str(), team.c_str(),
				             absdiff, float_abs_tol, reldiff, float_rel_tol,
```

I tested this on the BAPC 2022 problems again (locally), nothing breaks :slightly_smiling_face:

PR #1925, opened 2023-02-18
Add example Kotlin solution to team manual.

eldering:
Also replace raw PDF pagebreak by LaTeX \clearpage as the first doesn't work. Note that clearpage also flushes all floats (in this case code listings) to the current page.

### DIFF
--- a/doc/manual/team.rst
+++ b/doc/manual/team.rst
@@ -375,17 +375,21 @@ for different programming languages.
    :language: c
    :caption: *A solution in C*
 
+.. raw:: latex
+
+   \clearpage
+
 .. literalinclude:: ../examples/example.cc
    :language: cpp
    :caption: *A solution in C++*
 
-.. raw:: pdf
-
-   PageBreak
-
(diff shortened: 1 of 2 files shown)


nickygerritsen (approved):

meisterT (approved):

doc/manual/team.rst
```diff
@@ -375,18 +375,22 @@ for different programming languages.
...
-.. raw:: pdf
-
-   PageBreak
-
 .. literalinclude:: ../examples/example.java
    :language: java
    :caption: *A solution in Java*
 
+.. literalinclude:: ../examples/example.kt
+   :language: java
```
thijskh, line 391: This should be `kotlin` as the highlighting fails with the Java lexer.

thijskh (changes requested):

doc/manual/team.rst
eldering: Oops, copy&paste error. Fixed.

PR #1926, opened 2023-02-18
Make it obvious when you are in dev mode.

meisterT:
This is only done in the jury interface, so no change for the teams.

It will also help in cases where you have a local maintainer install (which is typically in dev mode) and you don't want to confuse it with a production instance.

Example:
image

### DIFF
--- a/webapp/src/Twig/TwigExtension.php
+++ b/webapp/src/Twig/TwigExtension.php
@@ -45,6 +45,7 @@ class TwigExtension extends AbstractExtension implements GlobalsInterface
     protected TokenStorageInterface $tokenStorage;
     protected AuthorizationCheckerInterface $authorizationChecker;
     protected string $projectDir;
+    protected bool $debug;
 
     public function __construct(
         DOMJudgeService               $dj,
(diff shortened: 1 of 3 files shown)


nickygerritsen (approved):

webapp/src/Twig/TwigExtension.php
```diff
@@ -68,6 +70,7 @@ public function __construct(
         $this->tokenStorage         = $tokenStorage;
         $this->authorizationChecker = $authorizationChecker;
         $this->projectDir           = $projectDir;
+        $this->debug                = $debug;
```
thijskh, line 73: Can just use `app.debug` instead in the template instead of adding a new global?

webapp/templates/jury/menu.html.twig
```diff
@@ -84,7 +84,16 @@
             {% endif %}
         </ul>
 
+        {% if debug %}
+            <ul class="navbar-nav ml-auto">
+                <li class="nav-item">
+                    <i class="fa-brands fa-dev fa-2x" style="color: white;"></i>
```
thijskh, line 90: Why an inline style here and a css file entry for the other aspect?

PR #1928, opened 2023-02-19
Add better error handling for problems with no test cases.

radwouters:
When installing DOMJudge in my local environment, I got an SQL error which was difficult to understand. Fortunately, there was Nicky G to help me out. The error is now a lot more understandable!

### DIFF
--- a/webapp/src/Service/DOMJudgeService.php
+++ b/webapp/src/Service/DOMJudgeService.php
@@ -53,6 +53,7 @@ use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInt
 use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
 use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 use ZipArchive;
+use function count;
 
 class DOMJudgeService
 {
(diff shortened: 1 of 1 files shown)


webapp/src/Service/DOMJudgeService.php
```diff
@@ -53,6 +53,7 @@
 use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
 use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 use ZipArchive;
+use function count;
```
nickygerritsen, line 56: Do we need this?

radwouters: No, it was a mixup in my phpstorm config. Fixed.

nickygerritsen (approved):

meisterT (approved):

thijskh:
Love this as it’s a common situation that’s now clearly handled.

Maybe a nitpick but is it a BadRequest when the serverside config is incomplete?

tuupke:
Thanks @radwouters! This definitely improves things!

A next step would be to better handle the case of having no testcases. A team should not be presented with a BadRequest  when submitting. They can't change anything to prevent this failure. A different way to handle this situation is to accept the submission but raise an internal error and disable the problem. This should result in the submission being queued until the problem gets enabled again.

We should consider to not allow judging on a problem with zero testcases, and only allow judging when the first testcase gets added. Maybe add a checkbox in the interface so you can decide whether to enable.

thijskh:
@tuupke is right and this is a long standing problem. However I would not get in the way of this concrete improvement over the status quo.

meisterT:
I think we all agree 👍

nickygerritsen:
I propose to merge this as is and then later do the fancy thing. Anyone opposed?

meisterT:
Do it!

PR #1929, opened 2023-02-19
In the scoreboard view, display `started` instead of `starts` after c…

meisterT:
…ontest start.

### DIFF
--- a/webapp/templates/partials/scoreboard.html.twig
+++ b/webapp/templates/partials/scoreboard.html.twig
@@ -35,7 +35,12 @@
                     {% set now = 'now'|date('U') %}
                     {{ now | printremainingminutes(current_contest.endtime) }}
                 {% else %}
-                    starts: {{ current_contest.starttime | printtime }} - ends: {{ current_contest.endtime | printtime }}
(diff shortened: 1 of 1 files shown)


webapp/templates/partials/scoreboard.html.twig
```diff
@@ -35,7 +35,12 @@
                     {% set now = 'now'|date('U') %}
                     {{ now | printremainingminutes(current_contest.endtime) }}
                 {% else %}
-                    starts: {{ current_contest.starttime | printtime }} - ends: {{ current_contest.endtime | printtime }}
+                    {% if current_contest.freezeData.started %}
+                        started:
+                    {% else %}
+                        starts:
+                    {% endif %}
+                    {{ current_contest.starttime | printtime }} - ends: {{ current_contest.endtime | printtime }}
```
thijskh, line 43: Should the same not be done for "ends"?

meisterT: Once the contest has ended, we match any of the above cases (contest over, or final standings)

nickygerritsen (approved):

PR #1933, opened 2023-02-21
Improve jury clarification list overview.

meisterT:
- stop repeating the word jury all over the page
- only have either from or to column, not both, in each sublist
- align team badges as we do for the submission list (and with that fix ugly line breaks in the table, an alternative would have been to fix the `&nbsp;`).

before:
before

after:
after


nickygerritsen (approved):
So much better

vmcj (approved):

PR #1936, opened 2023-02-23
Add roles needed to setup a contest like SWERC

vmcj:
Given that we use ADMIN in grafana and this was disabled for SWERC it makes sense to allow import for the API users. This way we can import all accounts in one go instead of having to add these extra users later.

If this gets approved I'll make some tests before merging (8.3+)

### DIFF
--- a/webapp/src/Controller/API/UserController.php
+++ b/webapp/src/Controller/API/UserController.php
@@ -358,6 +358,9 @@ class UserController extends AbstractRestController
         }
 
         $roles = $request->request->all('roles');
+        if (in_array('cds', $roles)) {
+            $roles = array_merge(['api_source_reader', 'api_writer', 'api_reader'], array_diff($roles, ['cds']));
+        }
         foreach ($roles as $djRole) {
             if ($djRole === '') {
                 continue;
--- a/webapp/src/Controller/API/UserController.php
+++ b/webapp/src/Controller/API/UserController.php
@@ -362,6 +362,9 @@ class UserController extends AbstractRestController
             if ($djRole === '') {
                 continue;
             }
+            if ($djRole === 'judge') {
+                $djRole = 'jury';
+            }
             $role = $this->em->getRepository(Role::class)->findOneBy(['dj_role' => $djRole]);
             if ($role === null) {
                 throw new BadRequestHttpException(sprintf("Role %s not found", $djRole));
--- a/webapp/src/Controller/API/UserController.php
+++ b/webapp/src/Controller/API/UserController.php
@@ -363,7 +363,7 @@ class UserController extends AbstractRestController
             $roles = ['cds'];
         }
         if (in_array('cds', $roles)) {
-            $roles = array_merge(['api_source_reader', 'api_writer', 'api_reader'], array_diff($roles, ['cds']));
+            $roles = ['api_source_reader', 'api_writer', 'api_reader', ...array_diff($roles, ['cds'])];
         }
         foreach ($roles as $djRole) {
             if ($djRole === '') {
--- a/webapp/src/Controller/API/UserController.php
+++ b/webapp/src/Controller/API/UserController.php
@@ -358,6 +358,10 @@ class UserController extends AbstractRestController
         }
 
         $roles = $request->request->all('roles');
+        // For the file import we change a CDS user to the roles needed for ICPC CDS.
+        if ($user->getUsername() === 'cds') {
+            $roles = ['cds'];
+        }
         if (in_array('cds', $roles)) {
             $roles = array_merge(['api_source_reader', 'api_writer', 'api_reader'], array_diff($roles, ['cds']));
         }
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -847,6 +847,9 @@ class ImportExportService
                 case 'api_writer':
                     $roles[] = $apiWriteRole;
                     break;
+                case 'cds':
+                    $roles += [$apiReadRole, $apiSourceRole, $apiWriteRole];
+                    break;
                 case 'analyst':
                 case 'staff':
                     // Ignore type analyst and staff for now. We don't have a useful mapping yet.
@@ -1180,6 +1183,9 @@ class ImportExportService
                     break;
                 case 'api_source_reader':
                     $roles[] = $apiSourceRole;
+                    break;
+                case 'cds':
+                    $roles += [$apiReadRole, $apiSourceRole, $apiWriteRole];
                     break;
                 case 'analyst':
                     // Ignore type analyst for now. We don't have a useful mapping yet.
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -790,14 +790,15 @@ class ImportExportService
      */
     public function importAccountsJson(array $data, ?string &$message = null, ?array &$saved = null): int
     {
-        $teamRole     = $this->em->getRepository(Role::class)->findOneBy(['dj_role' => 'team']);
-        $juryRole     = $this->em->getRepository(Role::class)->findOneBy(['dj_role' => 'jury']);
-        $adminRole    = $this->em->getRepository(Role::class)->findOneBy(['dj_role' => 'admin']);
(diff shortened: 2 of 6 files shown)


webapp/src/Service/ImportExportService.php
```diff
@@ -844,12 +847,21 @@ public function importAccountsJson(array $data, ?string &$message = null, ?array
                 case 'team':
                     $roles[] = $teamRole;
                     break;
+                case 'staff':
+                    $roles = $clarHandler;
+                    // Also add balloon role;
                 case 'balloon':
                     $roles[] = $balloonRole;
                     break;
```
nickygerritsen, line 855: Cds also needs source code reader.
Also if we do this, do we maybe want a special case for WF? Where the type is not cds (since that is not in spec) but the username is

vmcj: Can't we add the CDS as a role to the spec? But I'm fine with an extra if statement that if the type is either cds or the username is.

nickygerritsen: I don’t think we will het that role approved, especially since the permissions are very specific to us (we don’t want them to log in, so why should others care about a special role?$

webapp/src/Controller/API/UserController.php
```diff
@@ -361,10 +361,20 @@ public function addAction(Request $request): Response
         }
 
         $roles = $request->request->all('roles');
+        // For the file import we change a CDS user to the roles needed for ICPC CDS.
+        if ($user->getUsername() === 'cds') {
+            $roles = ['cds'];
+        }
+        if (in_array('cds', $roles)) {
+            $roles = array_merge(['api_source_reader', 'api_writer', 'api_reader'], array_diff($roles, ['cds']));
```
nickygerritsen, line 369:
```suggestion
            $roles = ['api_source_reader', 'api_writer', 'api_reader', ...array_diff($roles, ['cds'])];
```

webapp/src/Service/ImportExportService.php
```diff
@@ -847,6 +856,21 @@ public function importAccountsJson(array $data, ?string &$message = null, ?array
...
+                    $roles[] = $apiReadRole;
+                    break;
+                case 'api_source_reader':
+                    $roles[] = $apiSourceRole;
+                    break;
+                case 'api_writer':
+                    $roles[] = $apiWriteRole;
+                    break;
+                case 'cds':
+                    $roles = array_merge($roles, [$apiReadRole, $apiSourceRole, $apiWriteRole]);
```
nickygerritsen, line 872:
```suggestion
                    $roles += [$apiReadRole, $apiSourceRole, $apiWriteRole];
```

webapp/src/Service/ImportExportService.php
```diff
@@ -1165,6 +1196,21 @@ protected function importAccountsTsv(array $content, ?string &$message = null):
...
+                    $roles[] = $apiReadRole;
+                    break;
+                case 'api_writer':
+                    $roles[] = $apiWriteRole;
+                    break;
+                case 'api_source_reader':
+                    $roles[] = $apiSourceRole;
+                    break;
+                case 'cds':
+                    $roles = array_merge($roles, [$apiReadRole, $apiSourceRole, $apiWriteRole]);
```
nickygerritsen, line 1212:
```suggestion
                    $roles += [$apiReadRole, $apiSourceRole, $apiWriteRole];
```

webapp/tests/Unit/Controller/API/AccountBaseTestCase.php
```diff
@@ -0,0 +1,228 @@
...
+abstract class AccountBaseTestCase extends BaseTestCase
+{
+    protected ?string $apiUser = 'admin';
+
+    protected array $expectedAbsent = ['4242', 'nonexistent'];
+
+    public function helperVerifyApiUsers(string $myURL, array $objectsBeforeTest, array $newUserPostData, ?array $overwritten = null): void
+    {
+        $objectsAfterTest = $this->verifyApiJsonResponse('GET', $myURL, 200, $this->apiUser);
+        $newItems = array_map('unserialize', array_diff(array_map('serialize', $objectsAfterTest), array_map('serialize', $objectsBeforeTest)));
```
nickygerritsen, line 18:
```suggestion
        $newItems = array_map(unserialize(...), array_diff(array_map(serialize(...), $objectsAfterTest), array_map(serialize(...), $objectsBeforeTest)));
```

webapp/tests/Unit/Controller/API/AccountBaseTestCase.php
```diff
@@ -0,0 +1,228 @@
...
+        // Because we login again with the admin user we might see that one also
+        if (count($newItems)===2) {
+            self::assertContains('admin', array_column($newItems,'username'), "Found unexpected new/changed user");
```
nickygerritsen, line 21:
```suggestion
            self::assertContains('admin', array_column($newItems, 'username'), "Found unexpected new/changed user");
```

webapp/tests/Unit/Controller/API/AccountBaseTestCase.php
```diff
@@ -0,0 +1,228 @@
...
+    protected ?string $apiUser = 'admin';
+
+    protected array $expectedAbsent = ['4242', 'nonexistent'];
+
+    public function helperVerifyApiUsers(string $myURL, array $objectsBeforeTest, array $newUserPostData, ?array $overwritten = null): void
+    {
+        $objectsAfterTest = $this->verifyApiJsonResponse('GET', $myURL, 200, $this->apiUser);
+        $newItems = array_map('unserialize', array_diff(array_map('serialize', $objectsAfterTest), array_map('serialize', $objectsBeforeTest)));
+        // Because we login again with the admin user we might see that one also
+        if (count($newItems)===2) {
```
nickygerritsen, line 20:
```suggestion
        if (count($newItems) === 2) {
```

webapp/tests/Unit/Controller/API/AccountBaseTestCase.php
```diff
@@ -0,0 +1,228 @@
...
+                // be imported via the API file import.
+                continue;
+            }
+            $overwritten = $testUser[1] ?? null;
+            $testUser = $testUser[0];
+            $user = $testUser['username'];
+            $name = $testUser['name'];
+            $pass = $testUser['password'];
+            $role = $testUser['type'];
+            $tempData = ['id'=>$user, 'username'=> $user, 'name'=>$name, 'password'=>$pass, 'type'=>$role];
```
nickygerritsen, line 156:
```suggestion
            $tempData = ['id' => $user,  'username' => $user, 'name' => $name, 'password' => $pass, 'type' => $role];
```

webapp/tests/Unit/Controller/API/AccountBaseTestCase.php
```diff
@@ -0,0 +1,228 @@
...
+        foreach ($this->expectedObjects as $expectedObject) {
+            if (!isset($expectedObject['team_id'])) {
+                continue;
+            }
+            $url = $this->helperGetEndpointURL($this->apiEndpoint)."?team=".$expectedObject['team_id'];
+            $objects = $this->verifyApiJsonResponse('GET', $url, 200, $this->apiUser);
+            $found = false;
+            foreach ($objects as $possibleObject) {
+                if ($possibleObject['username'] == $expectedObject['username']) {
+                    $found = True;
```
nickygerritsen, line 218: See GHA phpcs comment here and below

webapp/tests/Unit/Controller/API/AccountBaseTestCase.php
```diff
@@ -0,0 +1,228 @@
...
```
vmcj, line 218:
```suggestion
                    $found = true;
```

webapp/tests/Unit/Controller/API/AccountBaseTestCase.php
```diff
@@ -0,0 +1,228 @@
...
+                    foreach ($expectedObject as $key => $value) {
+                        // Null values can also be absent.
+                        static::assertEquals($value, $possibleObject[$key] ?? null, $key . ' has correct value.');
+                    }
+                }
+            }
+            self::assertEquals(True,$found);
```
vmcj, line 225:
```suggestion
            self::assertEquals(true,$found);
```

PR #1940, opened 2023-02-26
Improve path checks

thijskh:


nickygerritsen (approved):

webapp/src/Controller/PublicController.php
```diff
@@ -133,7 +133,16 @@ public function scoreboardDataZipAction(RequestStack $requestStack, string $proj
 
         $publicPath = realpath(sprintf('%s/public/', $projectDir));
         foreach ($assetMatches[1] as $file) {
-            $zip->addFile($publicPath . '/' . $file, $file);
+            $filepath = realpath($publicPath . '/' . $file);
+            if (
+                substr($filepath, 0, strlen($publicPath)) !== $publicPath &&
+                substr($filepath, 0, strlen($vendorDir)) !== $vendorDir
+            ) {
+                // Path outside of known good dirs: path traversal
```
vmcj, line 141: Do we want to write something to a log or even throw an internal error?

thijskh: For now I'm just skipping it, not knowing whatever symlinks or other strange things people may be using... just nog exporting them in the zip

vmcj (approved):

webapp/src/Controller/PublicController.php
thijskh: I don't think we currently log things from the webinterface. Of course we could but at least currently there seems to be no expectation of logging from web UI interactions.

judge/judgedaemon.main.php
```diff
@@ -529,6 +534,13 @@ function fetch_executable_internal(
     $runuser .= '-' . $options['daemonid'];
 }
 
+if ($runuser === posix_getpwuid(posix_geteuid())['name'] ||
+    RUNGROUP === posix_getgrgid(posix_getegid())['name']
+) {
+    echo "Do not run the judgedaemon as the runser or rungroup.\n";
```
meisterT, line 540: Can we instead of what not to do say what to do, i.e. do we know at this point as which user this should be run with?

judge/judgedaemon.main.php
```diff
@@ -1264,11 +1276,11 @@ function judge(array $judgeTask): bool
         error("Could not create directory '$programdir'");
     }
 
-    // Cannot escape as we glob.
-    $cp_cmd = "cp -PRl '$workdir'/compile/* '$programdir'";
-    system($cp_cmd, $retval);
-    if ($retval!==0) {
-        error("Could not copy program to '$programdir'");
+    foreach (glob("'$workdir'/compile/*") as $compile_file) {
+        system('cp -PRl ' . dj_escapeshellarg($compile_file) . ' ' . djescapeshellarg($programdir), $retval);
```
meisterT, line 1280: There is an `_` missing in the second `dj_escapeshellarg`.

webapp/src/Controller/API/GeneralInfoController.php
```diff
@@ -327,6 +333,9 @@ public function countryFlagAction(Request $request, string $countryCode, string
         if (!Countries::alpha3CodeExists($alpha3code)) {
             throw new NotFoundHttpException("country $alpha3code does not exist");
         }
+        if (!preg_match('/^\d+x\d+$/', $size)) {
+            throw new BadRequestHttpException('invalid format for size parameter');
```
meisterT, line 337: nit: can we be more specific here and say that we would expect `<int>x<int>, e.g. 64x64`?

judge/judgedaemon.main.php
thijskh: I'm not sure we know that, basically it needs to be any user except the specials 'runuser' and 'root'.

webapp/src/Controller/API/GeneralInfoController.php
thijskh: I've gone with just making the two supported values explicit. We can do al kinds of dir reading in here to determine what flag aspects are there but that seems overkill to me, likely this will not change often.

PR #1941, opened 2023-02-26
Move unit tests to GHA

vmcj:
Not all uses are now implemented, the deprications is still missing for example.

Renamed the current job which checked if our codecov was out of date to fix tab completion.

CI fails but should be fixed after https://github.com/DOMjudge/domjudge/pull/1939

### DIFF
--- a/.github/jobs/base.sh
+++ b/.github/jobs/base.sh
@@ -59,9 +59,9 @@ echo -e "\033[0m"
 
 # configure, make and install (but skip documentation)
 make configure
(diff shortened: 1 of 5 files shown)


nickygerritsen (approved):

vmcj:
I've split out the commit in readable parts to see the difference between our current Gitlab config and the Github config.

vmcj:
We discussed to include the debs in the image, will do that later.

The only thing I'm not sure of is if we need to merge the unit test results but that can be done later if needed.

nickygerritsen (approved):
If it works I believe you

PR #1942, opened 2023-02-27
Balloon status is non-restricted for any staff member

vmcj:
Other alternatives is to add a check again in the template for the link or to add this specific role to the security firewall. This does add the balloons also to the menu but that makes sense for this user type.

### DIFF
--- a/webapp/src/Controller/Jury/BalloonController.php
+++ b/webapp/src/Controller/Jury/BalloonController.php
@@ -17,7 +17,7 @@ use Symfony\Component\Routing\Annotation\Route;
 
 /**
  * @Route("/jury/balloons")
- * @Security("is_granted('ROLE_JURY') or is_granted('ROLE_BALLOON')")
+ * @Security("is_granted('ROLE_ADMIN') or is_granted('ROLE_BALLOON')")
  */
 class BalloonController extends AbstractController
 {
--- a/webapp/src/Controller/RootController.php
+++ b/webapp/src/Controller/RootController.php
@@ -38,6 +38,9 @@ class RootController extends BaseController
             if ($this->dj->checkrole('balloon')) {
                 return $this->redirectToRoute('jury_balloons');
             }
+            if ($this->dj->checkrole('clarification_rw')) {
+                return $this->redirectToRoute('jury_clarifications');
+            }
         }
         return $this->redirectToRoute('public_index');
     }
--- a/webapp/templates/jury/index.html.twig
+++ b/webapp/templates/jury/index.html.twig
@@ -43,7 +43,9 @@
             </div>
             <div class="card-body">
                 <ul>
-                    <li><a href="{{ path('jury_balloons') }}">Balloon Status</a></li>
+                    {% if is_granted('ROLE_ADMIN') or is_granted('ROLE_BALLOON') %}
+                        <li><a href="{{ path('jury_balloons') }}">Balloon Status</a></li>
+                    {% endif %}
                     {% if is_granted('ROLE_CLARIFICATION_RW') %}
                         <li><a href="{{ path('jury_clarifications') }}">Clarifications</a></li>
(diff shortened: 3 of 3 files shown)


nickygerritsen:
Does it make sense for a clarification user? I'm not sure... Why would we have the balloon role otherwise?

vmcj:
> Does it make sense for a clarification user? I'm not sure... Why would we have the balloon role otherwise?

I would say the clarification user is a person who is trusted slightly more than the other balloon runners (lets say a "Rick, Kevin or Mylene") but the balloon runners might not know what to disclose and what not so the unfiltered clarification data is not something what you need to share (so by default you don't want to)

In the end I don't care much, so I'll implement the switch for the URL so another usecase is also possible.

nickygerritsen:
> > Does it make sense for a clarification user? I'm not sure... Why would we have the balloon role otherwise?
> 
> I would say the clarification user is a person who is trusted slightly more than the other balloon runners (lets say a "Rick, Kevin or Mylene") but the balloon runners might not know what to disclose and what not so the unfiltered clarification data is not something what you need to share (so by default you don't want to)

If this is the case we clearly named the role wrong, and it should be 'semi trusted user' or something. Currently the name strongly indicates the user is doing clarifications, not balloons.
> 
> In the end I don't care much, so I'll implement the switch for the URL so another usecase is also possible.

Yes, I think this is better.

thijskh:
The roles are roles, not levels, right? So there's no need to imply other roles from a role (except for admin maybe to keep it simple). So if someone wants ballons and clarifications, give them both roles. Same would go for a jury member that would want balloon access? Just keeps it simple. It's also not only about trust, also just about (a) not having many irrelevant menus and (b) not having the ability to accidentally change something which was not needed for your role in the first place.

nickygerritsen:
So your proposal is to ALSO drop balloon from jury? We can do that (but then need to change the annotation in the ballooncontroller to not allow jury).
I am fine with that approach as well.

thijskh:
What was the use case for adding it in the first place and why didn't it suffice to check the balloons box for this user?

nickygerritsen:
> What was the use case for adding it in the first place and why didn't it suffice to check the balloons box for this user?

🤷‍♂️

vmcj:
> What was the use case for adding it in the first place and why didn't it suffice to check the balloons box for this user?

Viewing the `git blame` it has been annotated as balloon+jury since symfony.

I think what Tobi did was add the link as jury already had access. His change did not add the balloons to the menu but only add a link in the overview page.

My change makes that you should have one of the roles otherwise the link is not shown.

I think the usecase Tobi had, was that during SWERC we did not easily see that balloon numbers were only going up. Given the procedure at the WF for balloons it is implied that getting your balloon on time, or even at all has some effect on the contest so I would say this is information the jury may view.

If we would now take away the jury annotation we need to think about if admin should have this information. And if this is actually nice for the admin user as when we add the balloon role to the admin we also get the menu item with it, most of us would consider that cluther.
If we don't we can not easily see that information without adding the role. I would say admin should have all access but I wouldn't want the interface to be cluttered with it when not needed. I don't have an opinion about removing jury, but than we should just change the annotation to have admin instead of jury?

nickygerritsen:
I would keep it simple:

* If you want the jury to view it, give them the role.
* The admin always has all roles. I know this is a bit 'messy' with balloon since we do not want the menu item to appear though.

vmcj:
> I would keep it simple:
> 
>     * If you want the jury to view it, give them the role.
> 
>     * The admin always has all roles. I know this is a bit 'messy' with balloon since we do not want the menu item to appear though.

Ok, let me see if I can make it like this.

vmcj:
@nickygerritsen see last commit for reasoning why it needs to be done like this.

nickygerritsen (approved):
Yes I expected this to work in this way.

PR #1943, opened 2023-02-28
Allow to call the API using CLI and use it to import the sample problems.

nickygerritsen:
This allows us to import these without having a webserver or a .netrc.

I think the Python code can probably be nicer, but I'm not sure how, so hoping @meisterT can assist.

I verified this works by stopping my webserver and importing.

When this is merged, we can re-enable running `install` instead of `bare-install` again in the contributor Docker image again as well 😄 .

### DIFF
--- a/misc-tools/dj_utils.py
+++ b/misc-tools/dj_utils.py
@@ -5,16 +5,19 @@ Part of the DOMjudge Programming Contest Jury System and licensed
 under the GNU GPL. See README and COPYING for details.
 '''
 
+import itertools
 import json
 import os
 import requests
 import requests.utils
+import shlex
+import subprocess
 import sys
 
 _myself = os.path.basename(sys.argv[0])
 _default_user_agent = requests.utils.default_user_agent()
 headers = {'user-agent': f'dj_utils/{_myself} ({_default_user_agent})'}
-api_url = 'unset'
+domjudge_webapp_folder_or_api_url = 'unset'
 ca_check = True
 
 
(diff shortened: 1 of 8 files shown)


misc-tools/dj_utils.py
```diff
@@ -91,19 +104,64 @@ def upload_file(name: str, apifilename: str, file: str, data: dict = {}):
...
+
+    Returns:
+        The parsed endpoint contents.
+
+    Raises:
+        RuntimeError when the command exit code is not 0.
+    '''
+
+    command = [
+        f'{domjudge_webapp_folder_or_api_url}/bin/console',
```
meisterT, line 148: this might need quoting as well?

misc-tools/dj_utils.py
```diff
@@ -91,19 +104,64 @@ def upload_file(name: str, apifilename: str, file: str, data: dict = {}):
...
+        'api:call',
+        '-m',
+        shlex.quote(method),
+    ] + list(itertools.chain(*[
```
meisterT, line 152: Instead of this I would just write something simpler like:

```
command = [
        f'{domjudge_webapp_folder_or_api_url}/bin/console',
...
        shlex.quote(method),
]
for item in data:
    command.extend(['-d', f'{shlex.quote(item)}={shlex.quote(data[item])}']
...
command.append(shlex.quote(name))
```

vmcj (approved):
Do we now test both options (symfony cli and python) in the CI?

misc-tools/dj_utils.py
nickygerritsen: Does that work for the actual command? If so then yes. I will try

misc-tools/dj_utils.py
nickygerritsen: So smart

nickygerritsen:
> Do we now test both options (symfony cli and python) in the CI?

We do in the integration since we import test data

meisterT (approved):

PR #1945, opened 2023-03-01
PHPUnit10 disallowes Test as suffix.

vmcj:
In the Symfony project they went with TestCase as suffix. Use that for consistency.

### DIFF
--- a/webapp/tests/E2E/Controller/ControllerRolesTraversalTest.php
+++ b/webapp/tests/E2E/Controller/ControllerRolesTraversalTest.php
@@ -2,10 +2,10 @@
 
 namespace App\Tests\E2E\Controller;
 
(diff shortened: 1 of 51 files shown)


nickygerritsen (approved):
I assume you want to update to PHPUnit 10 then? Is it compatible with PHP 7.4?

vmcj:
> I assume you want to update to PHPUnit 10 then?

At one point yes, but not today. It is one of de deprecations which we currently have, to me it makes sense to try to get that list to 0 most of the time and this was one which was easy to fix. By doing it now we can check if we "like" the different naming convention.

> Is it compatible with PHP 7.4?

No it is not, so this doesn't need to be merged soon.

PR #1951, opened 2023-03-05
Remove (after_)event_id from scoreboard API in strict mode.

eldering:
This was removed in the 2022-07 version of the Contest API, see https://ccs-specs.icpc.io/2022-07/


nickygerritsen (approved):

vmcj:
Do you want this one in 8.2?

eldering:
> Do you want this one in 8.2?

I guess if DOMjudge 8.2 is supposed to support Contest API specification 2022-07, then it should be included, but no rush yet, as I'm still migrating the json schema and validator to https://github.com/icpc/ccs-specs

PR #1953, opened 2023-03-05
Fixci import contest using cli

vmcj:


nickygerritsen:
This fix is now in #1943.

PR #1955, opened 2023-03-06
Bug and duplication fixes related to verdicts

eldering:
The logic bug was causing an error about an undefined index `aborted` at JudgementController.php line 176.

### DIFF
--- a/webapp/src/Controller/API/JudgementController.php
+++ b/webapp/src/Controller/API/JudgementController.php
@@ -149,9 +149,9 @@ class JudgementController extends AbstractRestController implements QueryObjectT
 
         $specificJudgingRequested = $request->attributes->has('id')
             || $request->query->has('ids');
-        // If we don't have correct permissions or didn't request a specific
-        // judging (necessary for the event log), then exclude some judgings:
-        if (!$roleAllowsVisibility && !$specificJudgingRequested) {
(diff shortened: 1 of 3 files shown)


nickygerritsen (approved):

webapp/src/Controller/API/JudgementController.php
```diff
@@ -151,7 +151,7 @@ protected function getQueryBuilder(Request $request): QueryBuilder
             || $request->query->has('ids');
         // If we don't have correct permissions or didn't request a specific
         // judging (necessary for the event log), then exclude some judgings:
-        if (!$roleAllowsVisibility && !$specificJudgingRequested) {
+        if (!($roleAllowsVisibility && $specificJudgingRequested)) {
```
vmcj, line 154: I prefer to read the  `!A || !B` over `!(A && B)` as it reads more natural. As this is the logic stated at L152, which given the naming of the variables can probably be removed as this is the same as at L154.

eldering: I've updated the comment to reflect the changed logic.
But I disagree that the original formulation was clearer: it had more negations, and now the comments can also be written in a positive way that "$roleAllowsVisibility && $specificJudgingRequested" implies you *can* see these judgements.

PR #1958, opened 2023-03-10
Fix showing problem text on jury problem and contest detail pages. Fixes #1782.

### LINKED ISSUE #1782: Add problemtext link to `jury/problems` page
In the problems overview (`jury/problems`), we should add a link to the problemtext, to make it easy to spot if there is any missing.

Probably we should do the same on the contestproblem overview (`jury/contests/<id>`).

nickygerritsen:


vmcj:
How does the page now look?

nickygerritsen:
Not at computer but there is one more icon next to the ones we already have

nickygerritsen:
Ignore this SL Scan failed job; I pushed to the upstream repo by accident and now it is confused.

vmcj (approved):
What would be the difference between the button you add (https://github.com/DOMjudge/domjudge/pull/1958/files#diff-16d4c63eb2228d878305402c06205062e0faf8e15022d0185c8588673a066aaeR390) and the bin icon in behaviour? The alt text looks the same.

nickygerritsen:
> What would be the difference between the button you add ([#1958 (files)](https://github.com/DOMjudge/domjudge/pull/1958/files#diff-16d4c63eb2228d878305402c06205062e0faf8e15022d0185c8588673a066aaeR390)) and the bin icon in behaviour? The alt text looks the same.

That I made a copy paste mistake that is now fixed.

PR #1959, opened 2023-03-10
Add direct button to public page for registering when enabled. Fixes #1780.

### LINKED ISSUE #1780: Make self-registration more prominent by adding a "Register" button
Currently, even when self-registration is enabled, we only display a login button here, not a "Register" button:
image

It is visible on the login page, so it should be easy to move to the main menu as well:
image

nickygerritsen:
Based on 2a69e59dd309beaa5328a249d7f208236223dca9.

Note that #1780 suggests to drop the `Don't have an account?` section because we don't test it often and we have no unit tests. Instead of doing that, I added some simple unit tests.

I also added a link to log in from the register page, of people land on that 'by accident'.

### DIFF
--- a/webapp/src/Twig/TwigExtension.php
+++ b/webapp/src/Twig/TwigExtension.php
@@ -12,6 +12,7 @@ use App\Entity\JudgingRun;
 use App\Entity\Language;
 use App\Entity\Submission;
 use App\Entity\SubmissionFile;
+use App\Entity\TeamCategory;
 use App\Entity\Testcase;
 use App\Service\AwardService;
 use App\Service\ConfigurationService;
@@ -119,6 +120,7 @@ class TwigExtension extends AbstractExtension implements GlobalsInterface
         $user = $this->dj->getUser();
         $team = $user?->getTeam();
 
+        $selfRegistrationCategoriesCount = $this->em->getRepository(TeamCategory::class)->count(['allow_self_registration' => 1]);
         // These variables mostly exist for the header template.
         return [
             'current_contest_id'            => $this->dj->getCurrentContestCookie(),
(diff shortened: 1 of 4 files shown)


webapp/src/Twig/TwigExtension.php
```diff
@@ -167,6 +169,7 @@ public function getGlobals(): array
                                                $this->authorizationChecker->isGranted('ROLE_ADMIN') &&
                                                $this->config->get('data_source') === DOMJudgeService::DATA_SOURCE_CONFIGURATION_AND_LIVE_EXTERNAL,
             'doc_links'                     => $this->dj->getDocLinks(),
+            'allow_registration'            => $selfRegistrationCategoriesCount !== 0,
```
nickygerritsen, line 172: Decided to put this in the Twig globals. This does mean one more query on one page, so do we want to do that? Or do we want to add this to all public pages separately? I'm not sure how/if we can do this query only on the public pages otherwise.

vmcj: So for every page under /public? Isn't it always needed for the menu for all pages?

nickygerritsen: If you are logged in not. So that’s only public since you need to be logged in everywhere else.
But the query is also not expensive… alternative is we write the value to the symfony cache and update it when needed, but that’s more complex

thijskh: I'm not seeing what you mean. I read from the code that this will be executed on all pages in all sections, not only public pages?

nickygerritsen: That's what I said what it currently does. I'm wondering if that's a problem and we should ONLY do it on public pages and if so, how.

vmcj (approved):
I don't like the query but when it becomes slow we can think about a proper solution.

nickygerritsen:
> I don't like the query but when it becomes slow we can think about a proper solution.

We could already add an index on the field. Not sure how many categories we have though. Other things like caching the result could be done, but all are way more work and I don't think it's worth it currently.

PR #1960, opened 2023-03-12
Let scripts use the CLI to internally access the API by default.

eldering:
This simplifies usage, and is typically what you want when you run these commands locally on the domserver.

### DIFF
--- a/doc/manual/import.rst
+++ b/doc/manual/import.rst
@@ -10,7 +10,7 @@ in your `.netrc`_ file. You need to install `httpie`_ and replace the
 ``<API_URL>`` in the examples below with the API URL of your local DOMjudge
 installation.
 
-For using the CLI, you need to replace ``<WEBAPP_DIR>`` with the path to
+To use the CLI, you need to replace ``<WEBAPP_DIR>`` with the path to
 the ``webapp`` directory of the DOMserver.
 
 Importing team categories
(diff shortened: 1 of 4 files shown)


vmcj:
There is a typo in the commit message.

nickygerritsen (approved):

doc/manual/import.rst
```diff
@@ -10,7 +10,7 @@ in your `.netrc`_ file. You need to install `httpie`_ and replace the
 ``<API_URL>`` in the examples below with the API URL of your local DOMjudge
 installation.
 
-For using the CLI, you need to replace ``<WEBAPP_DIR>`` with the path to
+To use the CLI, you need to replace ``<WEBAPP_DIR>`` with the path to
```
vmcj, line 13: I might be wrong, but I think we don't mention `<WEBAPP_DIR>` anymore after your commit.

So I propose something like:
```suggestion
To use the CLI, you need to run inside the ``webapp`` directory of the DOMserver.
```

If I understand how the current CLI is implemented.

nickygerritsen: Yea so we drop this whole line imho

nickygerritsen: Ah yes you are right. Sorry, this uses console except for in `Putting all API imports together`

PR #1963, opened 2023-03-18
Add API to unfreeze a contest.

nickygerritsen:
See https://github.com/icpc/ccs-specs/pull/134 for the spec PR.

I also added tests for this, as well as changing the contest start time, which we didn't have yet.

### DIFF
--- a/webapp/src/Controller/API/ContestController.php
+++ b/webapp/src/Controller/API/ContestController.php
@@ -358,20 +358,22 @@ class ContestController extends AbstractRestController
         $now      = (int)Utils::now();
         $changed  = false;
         if (!$request->request->has('id')) {
-            $response = new JsonResponse('Missing "id" in request.', Response::HTTP_BAD_REQUEST);
+            return new JsonResponse('Missing "id" in request.', Response::HTTP_BAD_REQUEST);
         } elseif (!$request->request->has('start_time') && !$request->request->has('scoreboard_thaw_time')) {
-            $response = new JsonResponse('Missing "start_time" or "scoreboard_thaw_time" in request.', Response::HTTP_BAD_REQUEST);
+            return new JsonResponse('Missing "start_time" or "scoreboard_thaw_time" in request.', Response::HTTP_BAD_REQUEST);
         } elseif ($request->request->get('id') != $contest->getApiId($this->eventLogService)) {
-            $response = new JsonResponse('Invalid "id" in request.', Response::HTTP_BAD_REQUEST);
+            return new JsonResponse('Invalid "id" in request.', Response::HTTP_BAD_REQUEST);
         } elseif ($request->request->has('start_time')) {
             // By default, it is not allowed to change the start time in the last 30 seconds before contest start.
             // We allow the "force" parameter to override this.
             if (!$request->request->getBoolean('force') &&
                 $contest->getStarttime() != null &&
                 $contest->getStarttime() < $now + 30) {
-                $response = new JsonResponse('Current contest already started or about to start.',
+                return new JsonResponse('Current contest already started or about to start.',
                     Response::HTTP_FORBIDDEN);
-            } elseif ($request->request->get('start_time') === null) {
+            }
+
+            if ($request->request->get('start_time') === null) {
                 $contest->setStarttimeEnabled(false);
                 $response = new Response('', Response::HTTP_NO_CONTENT);
                 $this->em->flush();
@@ -379,52 +381,51 @@ class ContestController extends AbstractRestController
             } else {
                 $date = date_create($request->request->get('start_time'));
                 if ($date === false) {
-                    $response = new JsonResponse('Invalid "start_time" in request.', Response::HTTP_BAD_REQUEST);
-                } else {
-                    $new_start_time = $date->getTimestamp();
-                    if (!$request->request->getBoolean('force') && $new_start_time < $now + 30) {
-                        $response = new JsonResponse('New start_time not far enough in the future.',
-                            Response::HTTP_FORBIDDEN);
-                    } else {
-                        $newStartTimeString = date('Y-m-d H:i:s e', $new_start_time);
-                        $contest->setStarttimeEnabled(true);
-                        $contest->setStarttime($new_start_time);
-                        $contest->setStarttimeString($newStartTimeString);
-                        $response = new Response('', Response::HTTP_NO_CONTENT);
-                        $this->em->flush();
-                        $changed = true;
-                    }
+                    return new JsonResponse('Invalid "start_time" in request.', Response::HTTP_BAD_REQUEST);
                 }
+
+                $new_start_time = $date->getTimestamp();
+                if (!$request->request->getBoolean('force') && $new_start_time < $now + 30) {
+                    return new JsonResponse('New start_time not far enough in the future.',
+                        Response::HTTP_FORBIDDEN);
+                }
+                $newStartTimeString = date('Y-m-d H:i:s e', $new_start_time);
+                $contest->setStarttimeEnabled(true);
+                $contest->setStarttime($new_start_time);
+                $contest->setStarttimeString($newStartTimeString);
+                $response = new Response('', Response::HTTP_NO_CONTENT);
+                $this->em->flush();
+                $changed = true;
             }
         } else { // $request->request->has('scoreboard_thaw_time')
             if (!$request->request->getBoolean('force') && $contest->getUnfreezetime() !== null) {
-                $response = new JsonResponse('Current contest already has a unfreeze time set.',
+                return new JsonResponse('Current contest already has a unfreeze time set.',
                     Response::HTTP_FORBIDDEN);
+            }
+
+            $date = date_create($request->request->get('scoreboard_thaw_time') ?? 'not a valid date');
+            if ($date === false) {
+                $response = new JsonResponse('Invalid "scoreboard_thaw_time" in request.', Response::HTTP_BAD_REQUEST);
             } else {
-                $date = date_create($request->request->get('scoreboard_thaw_time') ?? 'not a valid date');
-                if ($date === false) {
-                    $response = new JsonResponse('Invalid "scoreboard_thaw_time" in request.', Response::HTTP_BAD_REQUEST);
-                } else {
-                    $new_unfreeze_time = $date->getTimestamp();
-                    if (!$request->request->getBoolean('force') && $new_unfreeze_time < $now - 30) {
-                        $response = new JsonResponse('New scoreboard_thaw_time too far in the past.',
-                            Response::HTTP_FORBIDDEN);
-                    } else {
-                        $returnContest = false;
-                        if ($new_unfreeze_time < $now) {
-                            $new_unfreeze_time = $now;
-                            $returnContest = true;
-                        }
-                        $newUnfreezeTimeString = date('Y-m-d H:i:s e', $new_unfreeze_time);
-                        $contest->setUnfreezetime($new_unfreeze_time);
-                        $contest->setUnfreezetimeString($newUnfreezeTimeString);
-                        $this->em->flush();
-                        $changed = true;
-                        $response = new Response('', Response::HTTP_NO_CONTENT);
-                        if ($returnContest) {
-                            $response = $this->renderData($request, $contest);
-                        }
-                    }
+                $new_unfreeze_time = $date->getTimestamp();
+                if (!$request->request->getBoolean('force') && $new_unfreeze_time < $now - 30) {
+                    return new JsonResponse('New scoreboard_thaw_time too far in the past.',
+                        Response::HTTP_FORBIDDEN);
+                }
+
+                $returnContest = false;
+                if ($new_unfreeze_time < $now) {
+                    $new_unfreeze_time = $now;
+                    $returnContest = true;
+                }
+                $newUnfreezeTimeString = date('Y-m-d H:i:s e', $new_unfreeze_time);
+                $contest->setUnfreezetime($new_unfreeze_time);
+                $contest->setUnfreezetimeString($newUnfreezeTimeString);
+                $this->em->flush();
+                $changed = true;
+                $response = new Response('', Response::HTTP_NO_CONTENT);
+                if ($returnContest) {
+                    $response = $this->renderData($request, $contest);
                 }
             }
         }
--- a/webapp/src/Controller/API/ContestController.php
+++ b/webapp/src/Controller/API/ContestController.php
@@ -372,7 +372,6 @@ class ContestController extends AbstractRestController
                 $response = new JsonResponse('Current contest already started or about to start.',
                     Response::HTTP_FORBIDDEN);
             } elseif ($request->request->get('start_time') === null) {
-                $this->em->persist($contest);
                 $contest->setStarttimeEnabled(false);
                 $response = new Response('', Response::HTTP_NO_CONTENT);
                 $this->em->flush();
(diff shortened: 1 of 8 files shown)


webapp/src/Controller/API/ContestController.php
```diff
@@ -332,59 +332,106 @@ public function setBannerAction(Request $request, string $cid, ValidatorInterfac
...
-            $response = new JsonResponse('Contest paused :-/.', Response::HTTP_OK);
-            $this->em->flush();
-            $changed = true;
-        } else {
-            $date = date_create($request->request->get('start_time'));
-            if ($date === false) {
-                $response = new JsonResponse('Invalid "start_time" in request.', Response::HTTP_BAD_REQUEST);
+        } elseif ($request->request->has('start_time')) {
+            // By default, it is not allowed to change the start time in the last 30 seconds before contest start.
+            // We allow the "force" parameter to override this.
```
vmcj, line 373: I'm not sure if I like such a parameter. For the (un)locking of contests we also decided against such a parameter.

webapp/tests/Unit/Controller/API/ContestControllerAdminTest.php
```diff
@@ -128,4 +135,65 @@ public function testBannerManagement(): void
...
+        yield [['id' => 1], 400, "Missing \u0022start_time\u0022 or \u0022scoreboard_thaw_time\u0022 in request."];
+
+        // Tests for changing the start time
+        yield [['id' => 2, 'start_time' => null], 400, "Invalid \u0022id\u0022 in request."];
+        yield [['id' => 1, 'start_time' => null], 403, 'Current contest already started or about to start.', [DemoPreEndContestFixture::class]];
+        yield [['id' => 1, 'start_time' => null], 403, 'Current contest already started or about to start.', [DemoAboutToStartContestFixture::class]];
+        yield [['id' => 1, 'start_time' => date('Y-m-d\TH:i:s', strtotime('+15 seconds'))], 403, 'New start_time not far enough in the future.'];
+        yield [['id' => 1, 'start_time' => date('Y-m-d\TH:i:s', strtotime('+15 seconds')), 'force' => false], 403, 'New start_time not far enough in the future.'];
+        yield [['id' => 1, 'start_time' => 'some invalid start time'], 400, 'Invalid \u0022start_time\u0022 in request.'];
+        yield [['id' => 1, 'start_time' => date('Y-m-d\TH:i:s', strtotime('+15 seconds')), 'force' => true], 204];
```
vmcj, line 184:
```suggestion
        yield [['id' => 1, 'start_time' => date('Y-m-d\TH:i:s', strtotime('+15 seconds')), 'force' => true], 204];
        yield [['id' => 1, 'start_time' => 'some invalid start time'], 400, 'Invalid \u0022start_time\u0022 in request.'];
```

Makes it easier to read as those belong together by enumeration of options.

webapp/src/Controller/API/ContestController.php
nickygerritsen: We already had it for start time change. We should be consistent and drop both or keep both

webapp/tests/Unit/Controller/API/ContestControllerAdminTest.php
```diff
@@ -128,4 +135,65 @@ public function testBannerManagement(): void
...
+        yield [['id' => 1, 'start_time' => null, 'force' => true], 204, null, [DemoAboutToStartContestFixture::class]];
+        yield [['id' => 1, 'start_time' => null], 204];
+
+        // Tests for changing the unfreeze time
+        yield [['id' => 2, 'scoreboard_thaw_time' => date('Y-m-d\TH:i:s', strtotime('+15 seconds'))], 400, "Invalid \u0022id\u0022 in request."];
```
vmcj, line 189:
```suggestion
        yield [['id' => 4242, 'scoreboard_thaw_time' => date('Y-m-d\TH:i:s', strtotime('+15 seconds'))], 400, "Invalid \u0022id\u0022 in request."];
```

When you test with the "normal" db that contest does exist.

webapp/src/Controller/API/ContestController.php
```diff
@@ -332,59 +332,106 @@ public function setBannerAction(Request $request, string $cid, ValidatorInterfac
...
+            if (!$request->request->getBoolean('force') &&
+                $contest->getStarttime() != null &&
+                $contest->getStarttime() < $now + 30) {
+                $response = new JsonResponse('Current contest already started or about to start.',
+                    Response::HTTP_FORBIDDEN);
+            } elseif ($request->request->get('start_time') === null) {
+                $this->em->persist($contest);
+                $contest->setStarttimeEnabled(false);
```
eldering, line 381: Shouldn't these two statements swapped? First update starttime to false, then persist?

webapp/tests/Unit/Controller/API/ContestControllerAdminTest.php
```diff
@@ -128,4 +135,65 @@ public function testBannerManagement(): void
...
+        yield [['id' => 1, 'scoreboard_thaw_time' => null], 400, 'Invalid \u0022scoreboard_thaw_time\u0022 in request.'];
+        yield [['id' => 1, 'scoreboard_thaw_time' => 'some invalid start time'], 400, 'Invalid \u0022scoreboard_thaw_time\u0022 in request.'];
+        yield [['id' => 1, 'scoreboard_thaw_time' => date('Y-m-d\TH:i:s', strtotime('+15 seconds'))], 403, 'Current contest already has a unfreeze time set.', [DemoPostUnfreezeContestFixture::class]];
+        yield [['id' => 1, 'scoreboard_thaw_time' => date('Y-m-d\TH:i:s', strtotime('+15 seconds')), 'force' => false], 403, 'Current contest already has a unfreeze time set.', [DemoPostUnfreezeContestFixture::class]];
```
vmcj, line 193: Why can't we overwrite this without force in the API? This is what PATCH was designed for. If this would have been +15 minutes I would think it should work. Otherwise everyone will add the force all the time.

webapp/src/Controller/API/ContestController.php
```diff
@@ -332,59 +332,106 @@ public function setBannerAction(Request $request, string $cid, ValidatorInterfac
...
      */
-    public function changeStartTimeAction(Request $request, string $cid): Response
+    public function changeTimesAction(Request $request, string $cid): Response
     {
         $contest  = $this->getContestWithId($request, $cid);
-        $now      = Utils::now();
+        $now      = (int)Utils::now();
         $changed  = false;
         if (!$request->request->has('id')) {
             $response = new JsonResponse('Missing "id" in request.', Response::HTTP_BAD_REQUEST);
```
eldering, line 366: More a general comment, but can't we do an early return here and in similar places below, so that we don't have this long and deep nesting of if/elseif cases?

webapp/tests/Unit/Controller/API/ContestControllerAdminTest.php
nickygerritsen: Im fine with allow to overwrite existing unfreeze times. What do others think?

eldering: This PATCH method is very specifically for unfreezing a contest. I think it doesn't make sense to overwrite an unfreeze time by default, so a force parameter makes sense to me.

webapp/src/Controller/API/ContestController.php
nickygerritsen: The Persist can even be dropped. It only adds new entities but the contest already exists. 
(technically persist only makes the entity tracked so doesn’t matter where it goes if we would need it)

webapp/src/Controller/API/ContestController.php
nickygerritsen: Yes that’s better

webapp/src/Controller/API/ContestController.php
```diff
@@ -363,75 +363,74 @@ public function changeTimesAction(Request $request, string $cid): Response
         $now      = (int)Utils::now();
         $changed  = false;
         if (!$request->request->has('id')) {
-            $response = new JsonResponse('Missing "id" in request.', Response::HTTP_BAD_REQUEST);
+            return new JsonResponse('Missing "id" in request.', Response::HTTP_BAD_REQUEST);
         } elseif (!$request->request->has('start_time') && !$request->request->has('scoreboard_thaw_time')) {
```
eldering, line 367: Nitpick: I'd also change these `elseif`'s to `if`'s. Otherwise it seems at a glance that there's a more complicated control flow, than simply a list of "check for this error condition and throw an error".

webapp/src/Controller/API/ContestController.php
```diff
@@ -363,75 +363,74 @@ public function changeTimesAction(Request $request, string $cid): Response
...
                 }
+                $newStartTimeString = date('Y-m-d H:i:s e', $new_start_time);
+                $contest->setStarttimeEnabled(true);
+                $contest->setStarttime($new_start_time);
+                $contest->setStarttimeString($newStartTimeString);
+                $response = new Response('', Response::HTTP_NO_CONTENT);
+                $this->em->flush();
+                $changed = true;
             }
         } else { // $request->request->has('scoreboard_thaw_time')
```
eldering, line 405: Nitpick: I think this makes the control flow also easier to read. And due to the checks above this is equivalent.
```suggestion
        }
        if ($request->request->has('scoreboard_thaw_time')) {
```

webapp/src/Controller/API/ContestController.php
```diff
@@ -363,75 +363,74 @@ public function changeTimesAction(Request $request, string $cid): Response
...
             if (!$request->request->getBoolean('force') && $contest->getUnfreezetime() !== null) {
-                $response = new JsonResponse('Current contest already has a unfreeze time set.',
+                return new JsonResponse('Current contest already has a unfreeze time set.',
                     Response::HTTP_FORBIDDEN);
+            }
+
+            $date = date_create($request->request->get('scoreboard_thaw_time') ?? 'not a valid date');
+            if ($date === false) {
+                $response = new JsonResponse('Invalid "scoreboard_thaw_time" in request.', Response::HTTP_BAD_REQUEST);
```
eldering, line 413: This can also be a direct `return`, right?

nickygerritsen: Yea missed this one. Thanks

webapp/src/Controller/API/ContestController.php
```diff
@@ -332,59 +332,105 @@ public function setBannerAction(Request $request, string $cid, ValidatorInterfac
      *                 property="force",
      *                 description="Force overwriting the start_time even when in next 30s",
```
vmcj, line 333:
```suggestion
     *                 description="Force overwriting the start_time even when in next 30s or overwriting existing thaw_time",
```

webapp/tests/Unit/Controller/API/ContestControllerAdminTest.php
```diff
@@ -128,4 +135,65 @@ public function testBannerManagement(): void
...
+        yield [['id' => 1, 'start_time' => 'some invalid start time'], 400, 'Invalid \u0022start_time\u0022 in request.'];
+        yield [['id' => 1, 'start_time' => null, 'force' => true], 204, null, [DemoAboutToStartContestFixture::class]];
+        yield [['id' => 1, 'start_time' => null], 204];
+
+        // Tests for changing the unfreeze time
+        yield [['id' => 4242, 'scoreboard_thaw_time' => date('Y-m-d\TH:i:s', strtotime('+15 seconds'))], 400, "Invalid \u0022id\u0022 in request."];
+        yield [['id' => 1, 'scoreboard_thaw_time' => null], 400, 'Invalid \u0022scoreboard_thaw_time\u0022 in request.'];
+        yield [['id' => 1, 'scoreboard_thaw_time' => 'some invalid start time'], 400, 'Invalid \u0022scoreboard_thaw_time\u0022 in request.'];
+        yield [['id' => 1, 'scoreboard_thaw_time' => date('Y-m-d\TH:i:s', strtotime('+15 seconds'))], 403, 'Current contest already has a unfreeze time set.', [DemoPostUnfreezeContestFixture::class]];
+        yield [['id' => 1, 'scoreboard_thaw_time' => date('Y-m-d\TH:i:s', strtotime('+15 seconds')), 'force' => false], 403, 'Current contest already has a unfreeze time set.', [DemoPostUnfreezeContestFixture::class]];
```
vmcj, line 193:
```suggestion
        yield [['id' => 1, 'scoreboard_thaw_time' => date('Y-m-d\TH:i:s', strtotime('+15 seconds')), 'force' => false], 403, 'Current contest already has an unfreeze time set.', [DemoPostUnfreezeContestFixture::class]];
```

webapp/src/Controller/API/ContestController.php
```diff
@@ -330,62 +330,113 @@ public function setBannerAction(Request $request, string $cid, ValidatorInterfac
...
+                $contest->setStarttime($new_start_time);
+                $contest->setStarttimeString($newStartTimeString);
+                $response = new Response('', Response::HTTP_NO_CONTENT);
+                $this->em->flush();
+                $changed = true;
+            }
+        }
+        if ($request->request->has('scoreboard_thaw_time')) {
+            if (!$request->request->getBoolean('force') && $contest->getUnfreezetime() !== null) {
+                return new JsonResponse('Current contest already has a unfreeze time set.',
```
vmcj, line 412:
```suggestion
                return new JsonResponse('Current contest already has an unfreeze time set.',
```

webapp/tests/Unit/Controller/API/ContestControllerAdminTest.php
vmcj: also above.

webapp/src/Controller/API/ContestController.php
```diff
@@ -330,62 +330,113 @@ public function setBannerAction(Request $request, string $cid, ValidatorInterfac
...
-            $this->em->persist($contest);
-            $contest->setStarttimeEnabled(false);
-            $response = new JsonResponse('Contest paused :-/.', Response::HTTP_OK);
-            $this->em->flush();
-            $changed = true;
-        } else {
-            $date = date_create($request->request->get('start_time'));
-            if ($date === false) {
-                $response = new JsonResponse('Invalid "start_time" in request.', Response::HTTP_BAD_REQUEST);
+            return new JsonResponse('Missing "id" in request.', Response::HTTP_BAD_REQUEST);
```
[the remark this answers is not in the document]
nickygerritsen: There is probably a difference, but I am fine with raising an exception, seems we do that in other places.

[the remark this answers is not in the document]
vmcj: Apparantly it helps dev mode but also it will make the code coverage slightly more trustworthy. It doesn't punish for tests not covering such exceptions.

vmcj (approved):

webapp/src/Controller/API/ContestController.php
```diff
@@ -325,62 +325,109 @@ public function setBannerAction(Request $request, string $cid, ValidatorInterfac
...
-            $changed = true;
-        } else {
-            $date = date_create($request->request->get('start_time'));
-            if ($date === false) {
-                $response = new JsonResponse('Invalid "start_time" in request.', Response::HTTP_BAD_REQUEST);
+            throw new BadRequestHttpException('Missing "id" in request.');
+        }
+        if (!$request->request->has('start_time') && !$request->request->has('scoreboard_thaw_time')) {
+            throw new BadRequestHttpException('Missing "start_time" or "scoreboard_thaw_time" in request.');
+        }
```
eldering, line 365: I think the check that the request doesn't have both a `start_time` and `scoreboard_thaw_time` was removed.
This means that now if both are provided, the start time may get updated (and persisted) and then when handling the thaw time, the request may fail with a `BadRequestHttpException`. I think that should not be allowed. So probably the simplest is to not allow updating both.

nickygerritsen: Yep I think it's better to re-add that. I will do that, and add a test that checks for it as well.

eldering (approved):

PR #1968, opened 2023-03-19
Update to PHP 8.1.

meisterT:
Also upgrade prometheus_client_php to make it work with php 8.1


meisterT:
This is mostly to test CI. I will push more changes in individual commits after this one.

webapp/src/Controller/API/ProblemController.php
```diff
@@ -466,9 +466,8 @@ protected function getIdField(): string
     /**
      * Transform the given object before returning it from the API.
      * @param array $object
-     * @return ContestProblem|ContestProblemWrapper
      */
-    public function transformObject($object)
+    public function transformObject($object): \App\Entity\ContestProblem|\App\Helpers\ContestProblemWrapper
```
vmcj, line 470: Can we have the short form here?

lib/lib.misc.php
```diff
@@ -43,10 +44,10 @@ function overshoot_time(float $timelimit, string $overshoot_cfg) : float
 function overshoot_parse(float $timelimit, string $token) : float
 {
     $res = sscanf($token, '%d%c%n');
-    if (count($res)!=3) {
+    if (count((array) $res)!=3) {
```
meisterT, line 47: why is this cast here necessary?

meisterT: cc @nickygerritsen

rector.php
```diff
@@ -0,0 +1,88 @@
+<?php
```
meisterT, line 1: Did you intentionally  include this file in the commit, @nickygerritsen ?

lib/lib.misc.php
nickygerritsen: Beceause in theory sscanf can return false, which cant be used with count(). If we are sure it doesn't, we can add `/** @var array */`, but are we sure? The docs of sscanf are not super clear to me.

rector.php
nickygerritsen: nope

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -1121,10 +1106,7 @@ protected function determineFileChanged(array $files, array $oldFiles): array
         return $result;
     }
 
-    /**
-     * @param Judging|ExternalJudgement|null $judging
-     */
-    protected function processClaim($judging, Request $request, ?string &$claimWarning) : ?RedirectResponse
+    protected function processClaim(\App\Entity\Judging|\App\Entity\ExternalJudgement|null $judging, Request $request, ?string &$claimWarning) : ?RedirectResponse
```
vmcj, line 1109: Can we use the shorter form here (so with use?)

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -832,25 +821,21 @@ public function editSourceAction(Request $request, Submission $submission): Resp
...
+                'query_builder' => fn(EntityRepository $er) => $er->createQueryBuilder('p')
+                    ->join('p.contest_problems', 'cp')
+                    ->andWhere('cp.allowSubmit = 1')
+                    ->andWhere('cp.contest = :contest')
+                    ->setParameter('contest', $submission->getContest())
+                    ->orderBy('p.name'),
             ])
             ->add('language', EntityType::class, [
-                'class' => 'App\Entity\Language',
+                'class' => \App\Entity\Language::class,
```
vmcj, line 834: Can we have shorter form.

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -832,25 +821,21 @@ public function editSourceAction(Request $request, Submission $submission): Resp
 
         $formBuilder = $this->createFormBuilder($data)
             ->add('problem', EntityType::class, [
-                'class' => 'App\Entity\Problem',
+                'class' => \App\Entity\Problem::class,
```
vmcj, line 824: Shorter form.

webapp/src/Controller/Jury/ScoreboardController.php
```diff
@@ -16,15 +16,8 @@
...
-    protected DOMJudgeService $dj;
-    protected ScoreboardService $scoreboardService;
-
-    public function __construct(
-        DOMJudgeService $dj,
-        ScoreboardService $scoreboardService
-    ) {
-        $this->dj                = $dj;
-        $this->scoreboardService = $scoreboardService;
+    public function __construct(protected readonly DOMJudgeService $dj, protected readonly ScoreboardService $scoreboardService)
```
vmcj, line 19: In this case I think the {} can also be on the same line?

webapp/src/Controller/PublicController.php
```diff
@@ -288,24 +277,23 @@ public function problemTextAction(int $probId): StreamedResponse
...
     /**
      * @Route("/{probId<\d+>}/samples.zip", name="public_problem_sample_zip")
      */
     public function sampleZipAction(int $probId): StreamedResponse
     {
-        return $this->getBinaryFile($probId, function (int $probId, Contest $contest, ContestProblem $contestProblem) {
-            return $this->dj->getSamplesZipStreamedResponse($contestProblem);
-        });
+        return $this->getBinaryFile($probId, fn(
+            int $probId,
```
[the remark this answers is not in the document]
nickygerritsen: The line would be 174 characters long. I'm all for not limiting us to 80 or 100 characters but 174 is too much even for me.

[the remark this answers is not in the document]
nickygerritsen: Or do you mean the 3 arguments of the method? How would you make it look

```
return $this->getBinaryFile($probId, fn(int $probId, Contest $contest, ContestProblem $contestProblem
        ) => $this->dj->getSamplesZipStreamedResponse($contestProblem));
```

is super ugly IMHO

webapp/src/Controller/Jury/SubmissionController.php
nickygerritsen: I search for `\App` in the code now and replaced all

webapp/src/Entity/Contest.php
```diff
@@ -988,9 +976,13 @@ public function getAbsoluteTime(?string $time_string)
             // Take into account the removed intervals.
             /** @var RemovedInterval[] $removedIntervals */
             $removedIntervals = $this->getRemovedIntervals()->toArray();
-            usort($removedIntervals, function (RemovedInterval $a, RemovedInterval $b) {
-                return Utils::difftime((float)$a->getStarttime(), (float)$b->getStarttime());
-            });
+            usort(
```
vmcj, line 979: I think this is harder to read.

webapp/src/Helpers/OrdinalItem.php
```diff
@@ -13,16 +13,12 @@
...
-
-    /** @Serializer\Inline() */
-    protected object $item;
-
-    public function __construct(int $ordinal, object $item)
-    {
-        $this->ordinal = $ordinal;
-        $this->item    = $item;
+    public function __construct(
+        /** @Serializer\SerializedName("ordinal") */
```
vmcj, line 17: Do we check if moving this here keeps the stuff in the API doc correct?

webapp/src/Entity/Contest.php
nickygerritsen: Because it’s multi line? I’m not sure I like the long lines either

webapp/src/Helpers/OrdinalItem.php
nickygerritsen: I did and it does. It’s not only the doc, it’s the actual output

webapp/src/Service/CheckConfigService.php
```diff
@@ -761,9 +746,11 @@ public function checkSelfRegistration(): array
                     $selfRegistrationCategories[0]->getName());
             } else {
                 $desc .= sprintf("Team categories allowed for self-registered teams: %s.\n",
-                    implode(', ', array_map(function ($category) {
-                        return $category->getName();
-                    }, $selfRegistrationCategories)));
+                    implode(
+                        ', ',
```
vmcj, line 750: This would all fit on the same line IMO

webapp/src/Entity/Problem.php
```diff
@@ -390,7 +390,7 @@ public function removeContestProblem(ContestProblem $contestProblem): void
     /**
      * @return Collection|ContestProblem[]
      */
-    public function getContestProblems()
+    public function getContestProblems(): \Doctrine\Common\Collections\Collection|array
```
vmcj, line 393: Can we have the shortform here? Is there not an automation for this?

nickygerritsen: There is in php cs fixer

webapp/src/Service/DOMJudgeService.php
```diff
@@ -239,10 +221,7 @@ public function getUser(): ?User
         return $user;
     }
 
-    /**
-     * @return bool|float|int|string|InputBag|null
-     */
-    public function getCookie(string $cookieName)
+    public function getCookie(string $cookieName): bool|float|int|string|\Symfony\Component\HttpFoundation\InputBag|null
```
vmcj, line 224: Short form

webapp/src/Service/CheckConfigService.php
nickygerritsen: So
```
implode(', ', array_map(fn($category) => $category->getName(), $selfRegistrationCategories))
```
?

That is 113 characters long which is maybe a bit long, but I'm fine with doing that if others agree (but I think some won't 😛 )

webapp/src/Utils/Scoreboard/Filter.php
```diff
@@ -4,28 +4,19 @@
...
-        $this->countries    = $countries;
-        $this->categories   = $categories;
-        $this->teams        = $teams;
+        public array $affiliations = [],
+        public array $countries = [],
+        public array $categories = [],
+        public array $teams = []
+    )
+    {
     }
```
vmcj, line 20:
```suggestion
    ) {}
```

webapp/src/Entity/Contest.php
nickygerritsen: What about:
```
            usort(
                $removedIntervals,
                static fn(RemovedInterval $a, RemovedInterval $b) => Utils::difftime(
                    (float)$a->getStarttime(),
                    (float)$b->getStarttime()
                )
            );
```

Also not super readable though.

Alternative is
```
            usort(
                $removedIntervals, static fn(RemovedInterval $a, RemovedInterval $b) =>
                    Utils::difftime((float)$a->getStarttime(), (float)$b->getStarttime())
            );
```

but also a bit weird.

Or we leave it as the non-short option.

webapp/src/Utils/Scoreboard/Filter.php
nickygerritsen: We should make a decision on
```
    ) {
    }
```
(which is officially what PSR2/PSR12 want)

vs
```
    ) {}
```
(which saves a line and doesn't confuse too much)

Im fine with either (and PHPStorm has an explicit setting for this).

webapp/src/Entity/Contest.php
meisterT: I think we should just leave it as it was

webapp/src/Logger/VarargsLogMessageProcessor.php
```diff
@@ -18,14 +18,14 @@ class VarargsLogMessageProcessor implements ProcessorInterface
...
-        if (strpos($record['message'], '%') === false || empty($record['context'])) {
+        if (!str_contains($record['message'], '%') || empty($record['context'])) {
             return $record;
         }
 
         $res = false;
         try {
             $res = vsprintf($record['message'], $record['context']);
-        } catch (ValueError $e) {}
+        } catch (ValueError) {}
```
meisterT, line 28: phpcs (github actions) is complaining here, so we should probably change its rules

webapp/src/Service/CheckConfigService.php
meisterT: We could store the result of `array_map(...)` in a local var and then use that?

vmcj: Yes, but if the rest doesn't like it its fine by me. I'll resolve.

webapp/src/Helpers/OrdinalItem.php
vmcj: I didn't think of that, you're right.

webapp/src/Logger/VarargsLogMessageProcessor.php
vmcj: I think its to prevent against:
`} catch (ValueError) { command }`

Would we mind such code?

nickygerritsen: This seems to be the same 'discussion' as the empty constructor.
I don't like `{ command }` but would be fine with `{}`. Probably we can 'teach' phpcs that (and if we replace it by php-cs-fixer that one as well)

nickygerritsen: Mmm nope, php-cs-fixer at least doesn't allow it: https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/5475 
They currently link to PSR2 which doesn't allow it.

meisterT: Yeah, empty {} is fine, non-empty should be in separate lines

webapp/src/Entity/Contest.php
vmcj: I would be fine with the old situation (although its the only place where we use the non-short now?) or the last alternative (although it indeed looks a bit weird)

webapp/src/Controller/PublicController.php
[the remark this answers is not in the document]
vmcj: I think I agree.

webapp/src/Entity/Contest.php
nickygerritsen: I reverted this one and two other ones to the old format, since that's indeed beter.

webapp/src/Utils/Scoreboard/Filter.php
vmcj: I think we should for those situations go with {}, it makes it explicit that nothing happens whereas `{\n}` could be a forgotten statement.

nickygerritsen: I do dislike it if we can't make phpcs allow it (and only it). Could you investigate? I'm already bummed that php-cs-fixer just won't do it, but if phpcs can, that's at least something.

webapp/src/Logger/VarargsLogMessageProcessor.php
vmcj: I doubt we can teach the github action the difference. So we either have to take out the rule completely or accept that it will make annotations about it.

vmcj (approved):
Some things left where I commented but overall its an improvement IMO.

webapp/src/Logger/VarargsLogMessageProcessor.php
meisterT: I'm fine with taking the rule out.

webapp/src/Twig/TwigExtension.php
```diff
@@ -189,10 +167,9 @@ public function printremainingminutes(float $start, float $end): string
 
     /**
      * Print a time formatted as specified. The format is according to date().
-     * @param string|float $datetime
      * @param Contest|null $contest If given, print time relative to that contest start.
      */
-    public function printtime($datetime, ?string $format = null, ?Contest $contest = null): string
+    public function printtime(string|float|null $datetime, ?string $format = null, ?Contest $contest = null): string
```
vmcj, line 172:
```suggestion
    public function printtime(string|float ?$datetime, ?string $format = null, ?Contest $contest = null): string
```
@nickygerritsen  is this allowed?

nickygerritsen: No this was never. What you mean is `?string|?float` but that is also not allowed

webapp/src/Utils/Scoreboard/Filter.php
eldering: I completely agree that we should use `{}`. This is really clearer, and I think a lack of tooling support would be a bad reason to not use it.

PR #1970, opened 2023-03-19
Run PHP-CS-Fixer with a set of non intrusive rules

nickygerritsen:
This is my proposal for the PHP-CS-Fixer rules.

Some notes:

- You can view the rules in `.php-cs-fixer.php`.
- There is a separate commit for import ordering, since I'm not sure.
- The diff is big, but I think it's doable to read through it and see what the change is.
- If you disagree with a rule or want to discuss it, add a comment to a diff once, it doesn't make sense to add it everywhere. My ruleset is clearly not set in stone, but I think this is a good middle ground that changes things to make them more consistent without doing things that make no sense.

Run it yourself with:
```
lib/vendor/bin/php-cs-fixer fix --allow-risky=yes
```

(add `--dry-run -vvv` for fancy output (like what rule is used in a file).

Or set it up in PHPStorm: https://www.jetbrains.com/help/phpstorm/using-php-cs-fixer.html . It will then add scribbles below lines that are not OK.


webapp/src/Controller/Jury/RejudgingController.php
```diff
@@ -133,7 +133,7 @@ public function indexAction(): Response
                 $status = $rejudging->getValid() ? 'applied' : 'canceled';
                 $sort_order = 2;
             } elseif ($todo > 0) {
-                $perc   = (int)(100 * ((double)$done / (double)($done + $todo)));
```
vmcj, line 136: Why is this better?

webapp/src/Controller/Jury/UserController.php
```diff
@@ -130,7 +130,7 @@ public function indexAction(): Response
             }
 
             $userdata['user_roles'] = [
-                'value' => implode(', ', array_map(fn(Role $role) => $role->getDjRole(), $u->getUserRoles()))
+                'value' => implode(', ', array_map(fn (Role $role) => $role->getDjRole(), $u->getUserRoles())),
```
vmcj, line 133: I think we will forget the comma most of the time, so this will always be flagged.

nickygerritsen: My phpstorm format adds it and I do like it (for less git conflicts and changes) but if others don’t I’m fine with dropping it

webapp/src/Controller/Jury/RejudgingController.php
nickygerritsen: They are the same type so php cs fixer picks one. See https://cs.symfony.com/doc/rules/cast_notation/short_scalar_cast.html

vmcj (approved):
I suspect other people have a stronger opinion about this.

eldering:
Just for my understanding: for the rules that we add, what will happen with violations:
- Will the CI fail?
- Will each old/new violation show up in commit diffs in github PR?

nickygerritsen:
> Just for my understanding: for the rules that we add, what will happen with violations:
> 
> * Will the CI fail?
> * Will each old/new violation show up in commit diffs in github PR?

I’m think the consensus is to not do either. Only allow people to manually run it and maybe run it once every while

webapp/src/Controller/Jury/UserController.php
eldering: I'm fine with adding this (assuming here and in my other comments that this won't trigger CI failures or violation messages in the PRs).

judge/judgedaemon.main.php
```diff
@@ -543,14 +543,14 @@ function fetch_executable_internal(
...
 // Set static environment variables for passing path configuration
 // to called programs:
-putenv('DJ_BINDIR='      . BINDIR);
-putenv('DJ_ETCDIR='      . ETCDIR);
-putenv('DJ_JUDGEDIR='    . JUDGEDIR);
-putenv('DJ_LIBDIR='      . LIBDIR);
+putenv('DJ_BINDIR=' . BINDIR);
+putenv('DJ_ETCDIR=' . ETCDIR);
+putenv('DJ_JUDGEDIR=' . JUDGEDIR);
+putenv('DJ_LIBDIR=' . LIBDIR);
```
eldering, line 549: I don't like this removal of alignment spaces. This specific place is maybe not the best example of where it is useful, but there are better places where I feel strongly that alignment improves code readability a lot and should not be sacrificed.

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -952,7 +951,7 @@ private function addSingleJudgingRun(
 
         if (array_key_exists($runResult, $resultsRemap)) {
             $this->logger->info('JudgeTask %d remapping result %s -> %s',
-                                [ $judgeTaskId, $runResult, $resultsRemap[$runResult] ]);
+                                [$judgeTaskId, $runResult, $resultsRemap[$runResult]]);
```
eldering, line 954: I think that here the extra spaces improve readability, so I'm not sure I'm in favor of automatically removing them.

webapp/src/Controller/Jury/UserController.php
eldering: On second thought: I see there are some cases where there is a single element in an array (across multiple lines) and I think then, if that is really intentionally a single element and there's no more elements expected to be added to that declaration, then the extra comma is confusing. It seems that writing it in a single line statement, then the trailing comma is removed, but that seems inconsistent to me...

webapp/src/Controller/API/JudgehostController.php
nickygerritsen: I'm fine with leaving them.

judge/judgedaemon.main.php
nickygerritsen: Yeah let's disable this rule.

webapp/src/Controller/Jury/UserController.php
nickygerritsen: Then let's not do this rule either

PR #1974, opened 2023-03-31
Upgrade to Symfony 6.2

nickygerritsen:
This should be all to upgrade to 6.2 and remove all new deprecations. It has quite a lot of changed stuff; I suggest to check each commit individually to see what happened.

I'm happy how much of services.yaml is gone now. The only thing that is left and different from the default one Symfony ships has a comment and can be fixed when 6.3 is released in May.


webapp/src/Controller/Jury/ContestController.php
```diff
@@ -66,7 +66,7 @@ public function indexAction(Request $request): Response
     {
         $em = $this->em;
 
-        if ($doNow = (array)$request->request->get('donow')) {
+        if ($doNow = $request->request->all('donow')) {
```
vmcj, line 69: Just to be sure, the naming of the function does not reflect what actually happens? So we still only get 1 parameter?

nickygerritsen: It gets *all* donow array values

webapp/config/packages/doctrine.yaml
```diff
@@ -24,6 +24,7 @@ doctrine:
         #server_version: '15'
     orm:
         auto_generate_proxy_classes: true
+        enable_lazy_ghost_objects: false
```
vmcj, line 27: What does this do? According to google this would stop us from having the objects in memory but I assume that is what we want?

nickygerritsen: It’s a new feature which has a bug and breaks stuff. The lazy ghosts is a fancy thing that will auto load relations only when you need it (even more than normal), but it has an open bug we run into

webapp/src/Entity/Clarification.php
```diff
@@ -118,7 +119,7 @@ class Clarification extends BaseApiEntity implements ExternalRelationshipEntityI
         nullable: false,
         options: ['comment' => 'Has been answered by jury?', 'default' => 0]
     )]
-    #[Serializer\Groups(['RestrictedNonstrict'])]
+    #[Serializer\Groups([AbstractRestController::GROUP_RESTRICTED_NONSTRICT])]
```
vmcj, line 83: Can we short the AbstractRestController to ARC to make this a bit shorter? I agree with constants but now there is a lot more reading needed. I think we can short GROUP also (but I get if you want that in the name..)

webapp/src/Doctrine/DBAL/Types/JsonType.php
```diff
@@ -7,16 +7,12 @@
...
 
 /**
- * Class JsonType
- *
  * We overwrite this class because we want to preserve zero fractions when
  * storing JSON data in the database. Otherwise "5.0" will be converted to "5",
  * which gives some inconsistencies between the API and event feed.
  *
  * Also we always want a LONGTEXT field and not a JSON field, as that is only
  * supported by MySQL 5.7+.
```
vmcj, line 15: Not part of your PR but I think we shouldn't target MySQL <5.7 anymore.

vmcj (approved):
With my minimal symfony knowledge... LGTM.

webapp/src/Doctrine/DBAL/Types/JsonType.php
nickygerritsen: Mariadb doesn’t have json either

vmcj: Ok, thats a good reason to not change the code.

webapp/src/Entity/Clarification.php
nickygerritsen: I'm fine importing it as an `ARC` alias. I think we should keep the GROUP thing.

vmcj (approved):

webapp/.env.test
```diff
@@ -2,3 +2,5 @@
 KERNEL_CLASS='App\Kernel'
 APP_SECRET='$ecretf0rt3st'
 SYMFONY_DEPRECATIONS_HELPER=999999
+PANTHER_APP_ENV=panther
```
meisterT, line 5: out of curiosity: what is this?

webapp/config/packages/doctrine.yaml
```diff
@@ -19,14 +19,17 @@ doctrine:
         mapping_types:
             enum: string
         schema_filter: ~^(?!sessions)~
+        # IMPORTANT: You MUST configure your server version,
```
meisterT, line 22: s/server version/mysql major server version/ ?

webapp/src/Controller/Jury/LanguageController.php
```diff
@@ -140,10 +136,10 @@ public function indexAction(): Response
 
     // Note that the add action appears before the view action to make sure
     // /add is not seen as a language.
-    /**
-     * @Route("/add", name="jury_language_add")
-     * @IsGranted("ROLE_ADMIN")
-     */
+    #[IsGranted('ROLE_ADMIN')]
+    // Note that the add action appears before the view action to make sure
```
meisterT, line 140: This comment is now duplicated

webapp/src/FosRestBundle/FlattenExceptionHandler.php
```diff
@@ -7,19 +7,20 @@
...
- *
- * @package App\Controller\API
- */
 class FlattenExceptionHandler implements SubscribingHandlerInterface
 {
-    public function __construct(private readonly ExceptionValueMap $messagesMap, private readonly bool $debug)
-    {
-    }
+    public function __construct(
+        #[Autowire(service: 'fos_rest.exception.messages_map')]
```
meisterT, line 19: Why do we have to pass these Autowire annotations now? It seems that previously that was done by *magic*

webapp/tests/bootstrap.php
```diff
@@ -0,0 +1,16 @@
...
+require dirname(__DIR__) . '/config/load_db_secrets.php';
+
+if (file_exists(dirname(__DIR__) . '/config/bootstrap.php')) {
+    require dirname(__DIR__) . '/config/bootstrap.php';
+} elseif (method_exists(Dotenv::class, 'bootEnv')) {
+    (new Dotenv())->bootEnv(dirname(__DIR__) . '/.env');
+}
+
+if ($_SERVER['APP_DEBUG']) {
+    umask(0000);
```
meisterT, line 15: Why this?

webapp/.env.test
nickygerritsen: Panther is a web browser for testing. It came with one of the recipes but we can drop it

webapp/config/packages/doctrine.yaml
nickygerritsen: This is what symfony adds in a recipe. I’m reluctant to change it, since it will make future recipe updates conflict

webapp/src/Controller/Jury/LanguageController.php
nickygerritsen: Thanks for catching. Rector has a bug where this happens and I thought I got all of them

webapp/src/FosRestBundle/FlattenExceptionHandler.php
nickygerritsen: The magic was in services.yaml and you needed to know EXACTLY how to call your variable. I prefer to put the logic close to the code but if others disagree we can do the old thing again

webapp/tests/bootstrap.php
nickygerritsen: Ask symfony, it’s part of their current recipe and got auto added. Again, we can remove it but I’d like to stick to upstream as much as possible

meisterT: I would leave this out indeed. It seems weird that we would use a different umask in debug mode.

webapp/src/FosRestBundle/FlattenExceptionHandler.php
meisterT: SGTM

webapp/config/packages/doctrine.yaml
meisterT: hmmm, ok

webapp/.env.test
meisterT: I don't care, was mostly curious

webapp/tests/bootstrap.php
nickygerritsen: I will check if I can find what their reasoning is to see if it makes sense. Otherwise I will remove it

nickygerritsen: https://github.com/symfony/recipes/commit/57e4d9ee9628aba119a1166390e9c427c6dc00fa no reason given for it. I will drop it.

meisterT (approved):

PR #1981, opened 2023-04-10
Add label to teams. Fixes #1978.

### LINKED ISSUE #1978: Add support for team labels
https://github.com/icpc/ccs-specs/pull/143 adds a label to teams, which will be the new way to use team numbers (opposed to 'stealing' the ID and having strange inconsistencies if the number is not assigned yet).

We want to test this at NAC, so we should implement this. I was thinking:

- Add the field to the team DB table
- For the TSV import do not set the label.
- For the JSON import read the label. If it is not set, do not set it.
- For the API, expose it. if not set, use the API ID.
- For the UI: the team overview table should clearly have it. Maybe we hide ICPC there since the table will get full.
- We now have these team badges everywhere. Make them use the label if it is set (and show it in the popover).
- ~The legacy results.tsv should probably use the label if it is not empty to be backwards compatible.~ (Update: nope, it uses icpcid) New JSON-based exports are fine.

nickygerritsen:


doc/manual/import.rst
```diff
@@ -159,6 +159,7 @@ fields:
 
 - ``id``: the team ID. Must be unique
 - ``icpc_id`` (optional): an ID from an external system, e.g. from the ICPC CMS, may be empty
+- ``label``: (optional): the label, for example the saat number
```
vmcj, line 162:
```suggestion
- ``label``: (optional): the label, for example the seat number
```

webapp/src/Entity/Team.php
```diff
@@ -50,6 +51,13 @@ class Team extends BaseApiEntity implements ExternalRelationshipEntityInterface,
     #[Serializer\SerializedName('icpc_id')]
     protected ?string $icpcid = null;
 
+    #[ORM\Column(
+        nullable: true,
+        options: ['comment' => 'Team label, for example the saat number', 'collation' => 'utf8mb4_bin']
```
vmcj, line 56:
```suggestion
        options: ['comment' => 'Team label, for example the seat number', 'collation' => 'utf8mb4_bin']
```

webapp/src/Form/Type/TeamType.php
```diff
@@ -54,6 +54,11 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
                 )
             ]
         ]);
+        $builder->add('label', TextType::class, [
+            'label'       => 'Label',
+            'required'    => false,
+            'help'        => 'Optional label, for example the saat number.',
```
vmcj, line 60:
```suggestion
            'help'        => 'Optional label, for example the seat number.',
```

webapp/src/Service/ImportExportService.php
```diff
@@ -920,7 +921,8 @@ protected function importTeamData(array $teamData, ?string &$message, ?array &$s
                     $teamAffiliation = new TeamAffiliation();
                     $teamAffiliation
                         ->setExternalid($teamItem['team_affiliation']['externalid'])
-                        ->setName($teamItem['team_affiliation']['externalid'] . ' - auto-create during import');
+                        ->setName($teamItem['team_affiliation']['externalid'] . ' - auto-create during import')
+                        ->setShortname($teamItem['team_affiliation']['externalid'] . ' - auto-create during import');
```
vmcj, line 925: This is unrelated?

vmcj (approved):

webapp/src/Service/ImportExportService.php
nickygerritsen: It is but it breaks the test if I don’t add this. Short name is required

webapp/migrations/Version20230410144107.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    public function getDescription(): string
+    {
+        return 'Add label to teams.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE team ADD label VARCHAR(255) DEFAULT NULL COLLATE `utf8mb4_bin` COMMENT \'Team label, for example the saat number\' AFTER icpcid');
```
eldering, line 23:
```suggestion
        $this->addSql('ALTER TABLE team ADD label VARCHAR(255) DEFAULT NULL COLLATE `utf8mb4_bin` COMMENT \'Team label, for example the seat number\' AFTER icpcid');
```

webapp/templates/jury/entity_id_badge.html.twig
```diff
@@ -1,5 +1,7 @@
-<span class="badge badge-secondary id-badge" data-toggle="tooltip" data-placement="top" title="{% if externalId is not null %}External ID: {{ externalId }}, {% endif %}Internal ID: {{ idPrefix }}{{ id }}">
-    {% if externalId is not null %}
+<span class="badge badge-secondary id-badge" data-toggle="tooltip" data-placement="top" title="{% if label is defined and label %}Label: {{ label }}, {% endif %}{% if externalId is not null %}External ID: {{ externalId }}, {% endif %}Internal ID: {{ idPrefix }}{{ id }}">
```
eldering, line 1: What if `label = 0`?

eldering (changes requested):

eldering (approved):

PR #1983, opened 2023-04-12
Fix medal awards when skipping categories. Fixes #1982.

### LINKED ISSUE #1982: [BUG] The calculation of medals is not as expected
The calculation of medals directly skips the teams not in `Medal categories` resulting in a decrease in the total number of medals.

### Your environment

- DOMJudge 8.1.4
- Ubuntu 22.04 LTS
- nginx

### Steps to reproduce
image

- Step 1: Create a contest and set the number of medals to `28, 56, 84` (i.e. 10%, 20%, 30% of total participants) for all teams from category `Participants`, and  category `Observers` is the same sort order as `Participants`.
- Step 2: Run the contest.

### Expected behaviour
When the contest is running / After the contest is finished, there are 3 teams from `Observers` are ranked with 5, 13, 15 (in the gold medal area), gold medals should be awarded to 31st place, as these three teams do not give medals. 

### Actual behaviour
The gold medal was still handed out to 28th place and the 29th place was given a silver medal, which resulted in the wrong number of gold medals. This problem also exists in the Silver and Bronze areas, and scoreboard and award API's are both wrong.

It may be the same problem as https://github.com/DOMjudge/domjudge/pull/1742, and https://github.com/DOMjudge/domjudge/commit/09d25cbb1b27f60cb572d99799bd36e1103a606b should have fixed it (only in scoreboard), but the refactoring in https://github.com/DOMjudge/domjudge/commit/42daadc7b6fe8a773d78967854fc99dcff409226 introduces this issue again in 8.2.0.

nickygerritsen:


webapp/tests/Unit/Service/AwardServiceTest.php
```diff
@@ -81,16 +92,25 @@ protected function setUp(): void
...
-        // D    | x  x  x  x
+        // C    | x  2  3  x
+        // D    | x  x  x  12
+        // E    | x  x  x  13
+        // F    | x  x  x  14
+        // G    | x  x  x  15
+        // H    | x  x  x  x
+        // I    | x  x  x  x
         //
         // THis means A is the overall winner, will get a gold medal and is the winner
```
vmcj, line 103:
```suggestion
        // This means A is the overall winner, will get a gold medal and is the winner
```

vmcj (approved):

eldering (approved):

meisterT (approved):

PR #1989, opened 2023-04-28
Refactor streaming ZIP into utils class.

meisterT:


nickygerritsen (approved):

thijskh (approved):
Much neater indeed. The fact that the method as a side-effect deletes the passed file might warrent some kind of documentation?

PR #1992, opened 2023-04-28
Fix allowed types according to comment.

meisterT:


webapp/src/Entity/User.php
```diff
@@ -411,11 +411,11 @@ public function getRoleList(): array
...
     public function getType(): ?string
     {
-        // Types allowed by the CCS Specs Contest API in order of most permissions to least
+        // Types allowed by the CCS Specs Contest API in order of most permissions to least.
         // Either key=>value where key is the DOMjudge role and value is the API type or
-        // only value, where both the DOMjudge role and API type are the same
-        $allowedTypes = ['admin', 'api_writer' => 'admin', 'jury' => 'judge',
-                         'api_reader' => 'admin', 'api_writer' => 'admin', 'api_source_reader' => 'judge',
+        // only value, where both the DOMjudge role and API type are the same.
+        $allowedTypes = ['admin', 'api_writer' => 'admin', 'api_reader' => 'admin',
```
meisterT, line 417: Is the api reader necessarily admin?

vmcj: There is not a good translation from our roles to icpc admin, but I would say he's admin as he can see all config options. As this is only a cosmetic difference I would also be fine with jury.

vmcj (approved):
I had a reason to leave the ordering a bit like the original code but I think this makes more sense.

I've checked my discussions with @nickygerritsen and the summary was that there is not a clear ordering which works best. So your ordering reads easier.

webapp/src/Entity/User.php
nickygerritsen: In fact it's more admin than api_writer I think. api_writer can ONLY write, not read, right?

meisterT: I don't understand. In DOMjudge we typically distinguish between judge and admin in the amount of write operations they are allowed, right?

If the reader role cannot write, it is not admin, correct?

vmcj: https://github.com/DOMjudge/domjudge/blob/9a1627a9aff1a91cf17f46272a2e1b13ccff3e81/webapp/config/packages/security.yaml#L5

How I read this, both jury+admin have api_reader, so if you want to translate back (api_reader->DOMjudge role) we want the smallest role, given that admin is a superset of jury I think actually picking jury is better?

The question is what ICPC judge can do in other systems, can these judge roles do other things in CDS or PC2 or Kattis?

nickygerritsen: Nowhere is it defined what a role can actually do I believe, so it's hard to answer. Moreover, nowhere do we say roles need anything in the UI, it's only for the API, so that makes it a bit cumbersome.

webapp/src/Entity/User.php
```diff
@@ -411,11 +411,11 @@ public function getRoleList(): array
     #[Serializer\Type('string')]
     public function getType(): ?string
     {
-        // Types allowed by the CCS Specs Contest API in order of most permissions to least
+        // Types allowed by the CCS Specs Contest API in order of most permissions to least.
         // Either key=>value where key is the DOMjudge role and value is the API type or
```
nickygerritsen, line 415:
```suggestion
        // Either key=>value where key is the DOMjudge role and value is the API account type or
```

nickygerritsen (approved):

PR #1995, opened 2023-04-30
try to fix CI

meisterT:


webapp/src/Service/ImportExportService.php
```diff
@@ -1108,8 +1109,9 @@ protected function importAccountsTsv(array $content, ?string &$message = null):
             } else if (!in_array($type, $djRoles)) {
```
vmcj, line 1109: alignment seems wrong here.

Also not sure if you see the annotations but PSR12 wants `elseif` over `else if`

webapp/src/Service/ImportExportService.php
```diff
@@ -823,8 +823,9 @@ public function importAccountsJson(array $data, ?string &$message = null, ?array
...
```
vmcj, line 823: I think the problem is that we now want the keys instead of the values.
```suggestion
            } else if (!in_array($type, array_keys($djRoles))) {
```

meisterT: Right, I realized the same. Tests succeed now locally, let's see what CI finds...

PR #2003, opened 2023-05-02
CLICS uses _ instead of - in latest draft

vmcj:


webapp/src/Service/ImportExportService.php
```diff
@@ -48,22 +48,21 @@ public function getContestYamlData(Contest $contest): array
 
         $data = [
             'name' => $contest->getName(),
-            'short-name' => $contest->getShortname(),
-            'start-time' => Utils::absTime($contest->getStarttime(), true),
+            'short_name' => $contest->getShortname(),
```
nickygerritsen, line 51: If we want to use the (new) CLICS format this should be `name` and `name` should be `formal_name`.

webapp/src/Service/ImportExportService.php
```diff
@@ -74,7 +73,7 @@ public function getContestYamlData(Contest $contest): array
                 'label' => $contestProblem->getShortname(),
                 'letter' => $contestProblem->getShortname(),
                 'name' => $contestProblem->getProblem()->getName(),
-                'short-name' => $contestProblem->getProblem()->getExternalid(),
+                'short_name' => $contestProblem->getProblem()->getExternalid(),
```
nickygerritsen, line 76: This one doesn't *really* matter since it's not in the spec, but I'm fine changing it.

webapp/src/Controller/API/ContestController.php
```diff
@@ -428,15 +428,15 @@ public function getContestYamlAction(Request $request, string $cid): StreamedRes
         $penalty_time = $this->config->get('penalty_time');
         $response     = new StreamedResponse();
         $response->setCallback(function () use ($contest, $penalty_time) {
-            echo "name:                     " . $contest->getName() . "\n";
```
nickygerritsen, line 431: Should we use the getCOntestYamlData from the import export service here for less duplication?

nickygerritsen (approved):

webapp/src/Controller/API/ContestController.php
vmcj: We should yes, it will also remove this annoying whitespace issue.

webapp/src/Service/ImportExportService.php
vmcj: I prefer consistency so I hope we can rename them all somewhere in the future.

We could do those later when we develop DJ9 (not sure if this would be breaking BC), or I can leave both notations and we remove the `short-name` in another version.

vmcj: Changed for id

vmcj:
Further work would be to completely remove the contest-yaml endpoint and add an optional yaml parameter to the json endpoints or to have a flag to get the exports for older clics formats.

PR #2006, opened 2023-05-04
Extra python examples, more uniform code layout with tags, 

vmcj:
I've moved some solutions around when they had multiple tags and I've added one fail back which might not have been added in the rewrite. @meisterT touched those files last so knows best if this is actually an improvement.


example_problems/hello/submissions/accepted/extra-file-awk/test-hello.awk
```diff
@@ -1,4 +1,4 @@
-# This should give RUN-ERROR on the default problem 'hello', since it
+# This should give CORRECT on the default problem 'hello',
```
meisterT, line 1: super-nit: in the commit message of this file, you are fixing the comment, not the annotation.

example_problems/hello/submissions/accepted/test-hello.R
```diff
@@ -1,2 +1,5 @@
-cat("Hello World!\n")
+# This should give CORRECT on the default problem 'hello'.
```
meisterT, line 1: I am wondering whether we should remove these comments overall for correct submissions. They made sense when everything was in a single directory, but now not we are using the standard format.

I am also not sure that we should add the `@EXPECTED_RESULTS@` if there exactly one. It is a relict from the pre-problem package format times.

vmcj: I think we should put some in a non-standard folder just to make sure that this also does work with 1 result, and I changed this mostly to be consistent so I'm fine with removing it everywhere for the standard folders (but will than move some solutions to `mixed-verdicts` for the check)

Removing the comment is fine by me, I might have a slight preference for the comment as it gives more information than having to look in the file explorer if this is a failing submission without comment or that this is a correct submission so it does not have a comment. I don't care much for it though.

eldering: I think that these `@EXPECTED_RESULTS@` tags are useful, both when viewing the file as @vmcj already mentioned, but also because it means that this meta-data is now included in the source, so you can separately submit/import the file as jury and it will still pick up this information.

example_problems/hello/submissions/run_time_error/test-exitcode-2.py
```diff
@@ -0,0 +1,5 @@
+'''
+@EXPECTED_RESULTS@: RUN-ERROR
+'''
+
+exit(1)
```
eldering, line 5: Nitpick, but could we always end files with a newline? (Also below)

eldering (approved):

example_problems/hello/submissions/run_time_error/test-exitcode-2.py
vmcj: with a newline char or an actual empty line?

eldering: Only a newline char.

PR #2007, opened 2023-05-05
Upgrade bootstrap 5.3

thijskh:
This touches quite some things so no doubt a quirk here or there will be left or there are areas to improve even further.


webapp/src/Controller/Team/SubmissionController.php
```diff
@@ -95,6 +95,7 @@ public function createAction(Request $request, ?Problem $problem = null): Respon
         }
 
         $data = ['form' => $form->createView(), 'problem' => $problem];
+        $data['validFilenameRegex']    = SubmissionService::FILENAME_REGEX;
```
nickygerritsen, line 98: The whitespace around the = is a bit weird, isn't it?

webapp/templates/base.html.twig
```diff
@@ -45,16 +45,6 @@
         var domjudge_base_url = "{{ app.request.getBaseURL() }}";
     {% endif %}
     $(function () {
-        $('body').on('change', '.custom-file-input', function () {
```
nickygerritsen, line 48: Is the file widget now natively supported, or is it now different?

nickygerritsen (approved):
Looks good in general. I guess we just have to see where we missed stuff or what looks ugly.

nickygerritsen:
The toggles that are autosaved after clicking them (like on the language page) don't work anymore,

nickygerritsen:
Links are now underscored by default (without hovering), which to me looks more ugly.

thijskh:
> Links are now underscored by default (without hovering), which to me looks more ugly.

This is indeed a new default in BS5. I think it's some idea about accessibility. I also think it's not so beautiful so I'd indeed agree to remove this.

webapp/templates/base.html.twig
thijskh: The default view is i.m.o. beter and I've moved this functionality of listing all filenames to the place where it matters and solved it there in an even better way.

nickygerritsen: Yeah I just looked at it and it looks good.

thijskh:
Added the fixes in cba0fae

nickygerritsen (approved):

PR #2011, opened 2023-05-06
Add Adminer.php (for easy db fiddling)

nickygerritsen:
@ubergeek42 this is my attempt at making the Adminer code a bit better:
* It is now behind a config flag, which is disabled by default.
* It uses a global function that gets loaded in Composer.
* It doesn't use a separate bundle.

I tried loading  Adminer using Composer but that is horrible: they expect you to 'compile' the code and that seemed like not the way to go.

I don't like that I need to use $_GLOBALS for the etcDir, but I couldn't find any way to 'inject' it in the global function any other way.

This replaces #1775.


webapp/Makefile
```diff
@@ -23,7 +23,7 @@ install-domserver:
 # This must be done first to install with the rest.
 	$(MAKE) copy-bundle-assets
 	$(INSTALL_DIR) $(DESTDIR)$(domserver_webappdir);
-	for d in bin config migrations public src templates tests ; do \
+	for d in bin config migrations public Resources src templates tests ; do \
```
meisterT, line 26: nit: make this lowercase

webapp/Resources/functions.php
```diff
@@ -0,0 +1,10 @@
+<?php
```
meisterT, line 1: nit: add declare strict types as we do for other files?

meisterT (approved):

PR #2013, opened 2023-05-06
Show a message when viewing the scoreboard as balloon runner.

### LINKED ISSUE #1289: No scoreboard for balloon runners (non-public)

nickygerritsen:
Fixes #1289.

It seems @vmcj already fixed the logic itself, this just adds a message to not confuse balloon runners.


vmcj (approved):

vmcj:
This does leave the question about moving stuff to `/staff` which is a lot of work but removes the confusion.

nickygerritsen:
And outside the scope of this issue. I think for now this is fine, I would prefer not to move all kinds of stuff around.

PR #2015, opened 2023-05-06
Lock a lockfile to make sure the judgedaemon is only running once.

### LINKED ISSUE #1776: Add locking for judgedaemon to prevent duplicate instances
When you accidentally start a duplicate instance of the same judgedaemon (same host and same CPU core), then weird errors happen.

We should add some locking that detects and prevents this from happening. During the WFs in Dhaka this happened at least to me, @meisterT and @nickygerritsen, since there was already a judgedaemon running as a service.

nickygerritsen:
Fixes #1776

The logfile is unique per daemon, so it seemed like a good file to use?

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -563,6 +563,18 @@ if (empty($options['e'])) {
     if (!posix_getpwnam($runuser)) {
         error("runuser $runuser does not exist.");
     }
+
+    define('LOCKFILE', RUNDIR.'/judge.'.$myhost.'.lock');
+    if (($lockfile = fopen(LOCKFILE, 'c'))===false) {
+        error("cannot open lockfile '" . LOCKFILE . "' for writing");
+    }
+    if (!flock($lockfile, LOCK_EX | LOCK_NB)) {
+        error("cannot lock '" . LOCKFILE . "', is another judgedaemon already running?");
+    }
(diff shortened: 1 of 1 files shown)


judge/judgedaemon.main.php
```diff
@@ -538,6 +538,14 @@ function fetch_executable_internal(
     exit(1);
 }
 
+if (defined('STDLOG')) {
+    // Add a lock to the logfile; this makes sure we can only run the daemon once
+    if (!flock(STDLOG, LOCK_EX | LOCK_NB)) {
+        echo "Can lock logfile, is a daemon already running?\n";
```
vmcj, line 544: If you can, should you exit?

nickygerritsen: Can't*. But @eldering proposes to use a separate lockfile and he will implement that.

nickygerritsen: Also I'd rather not change this variable name now, its pretty old.

meisterT: Please also log the path to the lockfile

nickygerritsen: @eldering will use a separate file and I guess he can add the logmsg for the filename 😄

nickygerritsen (commented):
LGTM

eldering (approved):

PR #2016, opened 2023-05-06
Add feedback/teammessage.txt for automated teamfeedback

### LINKED ISSUE #991: Add feedback/teammessage.txt for automated teamfeedback
https://github.com/DOMjudge/domjudge/blob/3f8c8b1f8d69c142ea90d34653c55a2c07d195c7/sql/files/examples/boolfind_cmp/run#L9

The above file and the manual from October 2017 mentioned the existence of a `teammessage.txt` feedback file that, presumably, would be shown to the team as submission feedback like `judgemessage.txt` is shown to the judges.

I have recently had cause to desire this functionality. It would in general be especially useful for practise contests, where you don't want to give the students all the answers, but giving them automated feedback to help them figure it out could be very valuable.

Strangely, I can't even find where these mentions were ever removed. What happened to this feature? Was it removed deliberately, or did it just get forgotten half way through implemenation? Can we have this feature again in the future?

brinky:
Shows feedback/teammessage.txt. For the sample cases to teams. All to jury.


meisterT (approved):

nickygerritsen (approved):

webapp/templates/jury/submission.html.twig
```diff
@@ -757,6 +757,12 @@
                             {% else %}
                                 <pre class="output_text">{{ runsOutput[runIdx].output_system | parseRunDiff }}</pre>
                             {% endif %}
+
+                            {% if runsOutput[runIdx].team_message is not empty %}
+                                <h5>Judge message</h5>
```
thijskh, line 762: Since there is also a "judge message" concept in a run, might be less confusing to make explicit that this is the team message, not the thing called "judge message" in the output. Something like "(Judge) message to team"?

meisterT:
The integration test failure seems to be related: 

```
[May 06 19:43:45.837] judgedaemon[16726]: error: File does not exist: /opt/domjudge/judgehost/judgings/runner--azerasq-project-11698361-concurrent-0-0/endpoint-default/84/84/testcase00001/feedback/teammessage.txt
```

See https://gitlab.com/DOMjudge/domjudge/-/jobs/4238953312#L3414

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -626,6 +631,7 @@ public function addJudgingRunAction(
         $outputDiff   = $request->request->get('output_diff');
         $outputError  = $request->request->get('output_error');
         $outputSystem = $request->request->get('output_system');
+        $teamMessage  = $request->request->get('team_message');
```
nickygerritsen, line 634: If you do this, you can drop the null check
```suggestion
        $teamMessage  = $request->request->get('team_message', '');
```

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -900,6 +910,7 @@ private function addSingleJudgingRun(
         string $outputError,
         string $outputDiff,
         string $outputRun,
+        string $teamMessage,
```
nickygerritsen, line 913:
```suggestion
        ?string $teamMessage,
```

Then it also accepts null and you don't need any casting.

webapp/migrations/Version20230506141723.php
```diff
@@ -0,0 +1,36 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20230506141723 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
```
nickygerritsen, line 17: Can you set a description? Like "Add team message to judging run output"

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -941,6 +949,10 @@ private function addSingleJudgingRun(
                 ->setOutputSystem(base64_decode($outputSystem))
                 ->setMetadata(base64_decode($metadata));
 
+            if($teamMessage) {
```
nickygerritsen, line 952: Nit:
```suggestion
            if ($teamMessage) {
```

webapp/src/Entity/JudgingRunOutput.php
```diff
@@ -116,6 +123,17 @@ public function getOutputSystem(): string
         return $this->output_system;
     }
 
+    public function setTeamMessage(string $teamMessage) : JudgingRunOutput
+    {
+        $this->team_message = $teamMessage;
+        return $this;
+    }
+
+    public function getTeamMessage(): string
```
nickygerritsen, line 132: This might need
```suggestion
    public function getTeamMessage(): ?string
```
to work?

webapp/templates/jury/submission.html.twig
```diff
@@ -778,6 +778,12 @@
                             {% else %}
                                 <pre class="output_text">{{ runsOutput[runIdx].output_system }}</pre>
                             {% endif %}
+
+                            {% if runsOutput[runIdx].team_message is not empty %}
+                                <h5>Judge message</h5>
```
nickygerritsen, line 783: I would even add `for team` here, since for jury the judge message is more known than to teams. Maybe we don't need it for the team pages, only jury pages? cc @meisterT WDYT?

meisterT: agreed

nickygerritsen (approved):

PR #2018, opened 2023-05-06
Use enum for event feed format, also use it while reading.

### LINKED ISSUE #1669: Refactor shadow clics spec implementation
As mentioned in https://github.com/DOMjudge/domjudge/pull/1668, we should rewrite to use the same constant for the spec version.

In the future we might even want to read/detect the exact version and store it in a variable with the same constants https://github.com/DOMjudge/domjudge/pull/1668/files#diff-efa9e2dd8b5fccf408219fb062d6e943b7c49fa715163a4a600fbb23c941bae1R51 ...(?)

_Originally posted by @eldering in https://github.com/DOMjudge/domjudge/pull/1668#discussion_r962571586_

nickygerritsen:
Also add support for enums in the config service.
Fixes #1669.

### DIFF
--- a/webapp/migrations/Version20230506164639.php
+++ b/webapp/migrations/Version20230506164639.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20230506164639 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'change the value of the event_feed_format configuration key';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $eventFeedFormatConfig = $this->connection->fetchAssociative('SELECT * FROM configuration WHERE name = :name', ['name' => 'event_feed_format']);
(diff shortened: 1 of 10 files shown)


webapp/migrations/Version20230506164639.php
```diff
@@ -0,0 +1,47 @@
...
+    public function down(Schema $schema): void
+    {
+        $eventFeedFormatConfig = $this->connection->fetchAssociative('SELECT * FROM configuration WHERE name = :name', ['name' => 'event_feed_format']);
+        if ($eventFeedFormatConfig) {
+            $this->addSql('UPDATE configuration SET value = :value WHERE name = :name', [
+                'name' => 'event_feed_format',
+                'value' => $eventFeedFormatConfig['value'] === '"2022-07"' ? 1 : 0,
+            ]);
+        }
+
```
vmcj, line 40:
```suggestion

```

webapp/templates/jury/partials/external_contest_info.html.twig
```diff
@@ -54,6 +54,31 @@
                     {{ externalContestSource.lastEventId | default('-') }}
                 </td>
             </tr>
+            {% if sourceService.validContestSource and sourceService.apiVersion %}
+                <tr>
+                    <th>External provier information</th>
```
meisterT, line 59: typo in provider

meisterT (approved):

PR #2020, opened 2023-05-07
Add the problemset page to the jury interface.

### LINKED ISSUE #1781: Allow jury to see the problemset overview page before contest start
The jury should be able to see the problemset overview page that the teams will see even before contest start to see if all PDFs are there, colors are correct and attachments are available.

nickygerritsen:
Fixes #1781

Screenshots:
Screenshot 2023-05-07 at 10 41 36
Screenshot 2023-05-07 at 10 41 49

### DIFF
--- a/webapp/templates/jury/index.html.twig
+++ b/webapp/templates/jury/index.html.twig
@@ -27,7 +27,7 @@
                     <li><a href="{{ path('jury_judgehosts') }}">Judgehosts</a></li>
                     <li><a href="{{ path('jury_languages') }}">Languages</a></li>
                     <li><a href="{{ path('jury_problems') }}">Problems</a></li>
-                    <li><a href="{{ path('jury_problemset') }}">Problemset</a></li>
+                    <li><a href="{{ path('jury_problemset') }}">Team problemset page preview</a></li>
                     <li><a href="{{ path('jury_users') }}">Users</a></li>
                     <li><a href="{{ path('jury_teams') }}">Teams</a></li>
                     <li><a href="{{ path('jury_team_categories') }}">Team Categories</a></li>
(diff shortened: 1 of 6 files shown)


vmcj:
I would suggest a warning here that you only see this because the account is special.

nickygerritsen:
> I would suggest a warning here that you only see this because the account is special.

Special == jury? I mean should we add this warning to each page of the jury interface? I'm not sure if that makes sense. We can change the title to 'preview' or something?

vmcj:
> > I would suggest a warning here that you only see this because the account is special.
> 
> Special == jury? I mean should we add this warning to each page of the jury interface? I'm not sure if that makes sense. We can change the title to 'preview' or something?

Preview would be fine, but I think I would get jury users who think the contest is open to soon. If I'm the only one than leave it as is.

nickygerritsen:
I added a warning.

webapp/templates/jury/index.html.twig
```diff
@@ -27,6 +27,7 @@
                     <li><a href="{{ path('jury_judgehosts') }}">Judgehosts</a></li>
                     <li><a href="{{ path('jury_languages') }}">Languages</a></li>
                     <li><a href="{{ path('jury_problems') }}">Problems</a></li>
+                    <li><a href="{{ path('jury_problemset') }}">Problemset</a></li>
```
vmcj, line 30:
```suggestion
                    <li><a href="{{ path('jury_problems') }}">Problems (Global)</a></li>
                    <li><a href="{{ path('jury_problemset') }}">Problemset</a></li>
```

```suggestion
                    <li><a href="{{ path('jury_problems') }}">Problems</a></li>
                    <li><a href="{{ path('jury_problemset') }}">Current problemset</a></li>
```

```suggestion
                    <li><a href="{{ path('jury_problems') }}">Problems</a></li>
                    <li><a href="{{ path('jury_problemset') }}">Problemset (current contest)</a></li>
```

Without knowing the interface I would not know the difference between the problemset (specific to current contest) and the problems (all contests)

Neither of these suggestions really work, but its the best I could come up with.

vmcj (approved):
I like the feature.

webapp/templates/jury/index.html.twig
nickygerritsen: I pushed another suggestion

meisterT (approved):

Issue #1961, opened 2023-03-13
Output of Python code altered when judging


machitgarha:
### Description of the problem

In one of the contests here, a submission of Python code resulted in wrong answer, which was wrong. The problem was, for one test case, the generated output of the code seemed to be altered. We tested the exact same code with the exact same test case input, both locally and even using the JudgeHost's PyPy3 executable (located in `/chroot/domjudge/usr/bin/pypy3`), the output was different. This means the PyPy executable isn't the problem, it might be the compare script, the runguard, etc.

### Your environment

- Version 8.1.2 (via Docker)
- Fedora Linux 36 (Server Edition)
- Apache

### Steps to reproduce

- Create a contest with one problem including a test case with the following input and output:

    Input:
    ```txt
    4
    abbaac
    ac
    bccddaaf
    zzzzzzzzzz
    ```

    Output (i.e. answer):
    ```txt
    abac
    ac
    bcdaf
    zzzzzz
    ```

- Login as a team of the contest, and submit the following Python code for the problem:

    ```python
    for i in range(int(input())):
        x = input()

        print(x[::2] + x[-1])
    ```

    It should give you a wrong answer.

- Now go to the submission in the admin panel, and view the diff output.

### Expected behaviour

Output of the code ran manually: 

```txt
abac
ac
bcdaf
zzzzzz
```

### Actual behaviour

Output of the code ran in DOMJudge:

```txt
aba

a

bcda

zzzzz

```

I have no idea why it converted the last character of every line into a new line character.

### Any other information that you want to share?

*No response*

machitgarha:
@vmcj, upgraded Domserver and Judgehosts to 8.2.0, the problem is still present.

nickygerritsen:
Could you provide us with a dump of your database by any chance? I think that would help in this case.

machitgarha:
@nickygerritsen, I just created a fresh environment (i.e. new Docker containers) to reproduce the issue (and it's reproducable). And here is the `mysqldump` file: 

[mysqldump.txt](https://github.com/DOMjudge/domjudge/files/11072837/mysqldump.txt)

eldering:
I would not want to do this automatically. You mean with a flag on importing via the CLI?
Another option would be to warn on import when the input contains `\r\n` (windows) or `\r` (mac) style newlines.

nickygerritsen:
[...]

Issue #1777, opened 2022-11-08
On a problem with zero testcases, submitting causes a 500 error


eldering:
This is because at that time it fails to create a judge task according to @meisterT .

thijskh:
This is an improvement because earlier the judgedaemon would crash on such a problem.

thijskh:
Solution is to just not show problems with <= 0 testcases in the team view?

eldering:
> Solution is to just not show problems with <= 0 testcases in the team view?

Sounds ok, but we should then also clearly flag that problem as not used/submittable in the jury interface, so that there's no confusion about the state of that problem.

thijskh:
Alternative is to just accept stuff, but assume it's (implied) on allow_judge = 0 until it has testcases

meisterT:
Setting allow_judge=0 until it has test cases sounds reasonable to me

PR #2021, opened 2023-05-07
Add a few extra checks when importing problem zips

eldering:
This fixes a few bugs found along the way, and also #1777 and #1961. See individual commit messages for details.

It also addresses #1861 but doesn't completely solve it:

- Due to result remapping, submissions under `submissions/no_output` that have the `NO-OUTPUT` tag get a warning because the tag gets remapped to `WRONG-ANSWER` and then mismatches the directory.
- Directories like `submissions/mixed` are recognized as verdict and a mismatch is reported with the list of tagged verdicts.

### DIFF
--- a/webapp/src/Service/ImportProblemService.php
+++ b/webapp/src/Service/ImportProblemService.php
@@ -318,6 +318,14 @@ class ImportProblemService
             }
         }
 
+        $this->em->persist($problem);
+        if ($contestProblem) {
+            $contestProblem->setProblem($problem);
+            $contestProblem->setContest($contest);
+            $this->em->persist($contestProblem);
+            $this->em->flush();
+        }
+
         // Load the current testcases to see if we need to delete, update or insert testcases.
         $existingTestcases = [];
         if ($problem->getProbid()) {
@@ -584,7 +592,6 @@ class ImportProblemService
                 count($removedAttachments), join(',', $removedAttachments));
         }
 
-        $this->em->persist($problem);
         $this->em->wrapInTransaction(function () use ($testcases, $startRank) {
             $this->em->flush();
             // Set actual ranks if needed.
@@ -597,12 +604,6 @@ class ImportProblemService
             }
             $this->em->flush();
         });
-        if ($contestProblem) {
-            $contestProblem->setProblem($problem);
-            $contestProblem->setContest($contest);
-            $this->em->persist($contestProblem);
-            $this->em->flush();
-        }
 
         $cid = $contest?->getCid();
         $probid = $problem->getProbid();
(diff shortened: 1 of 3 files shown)


webapp/src/Service/ImportProblemService.php
```diff
@@ -716,11 +725,19 @@ public function importZippedProblem(
...
+                        $results = [$expectedResult];
                     } elseif (!in_array($expectedResult, $results)) {
-                        $messages['info'][] = sprintf("Annotated result '%s' does not match directory for %s",
-                                              implode(', ', $results), $path);
+                        $messages['danger'][] = sprintf(
+                            "Annotated result '%s' does not match directory for %s",
+                            implode(', ', $results), $path
+                        );
                     } elseif (!empty($expectedResult)) {
+                        if (count($results)>1) {
```
nickygerritsen, line 735: Nit: I would prefer to add spaces around `>`

webapp/src/Service/ImportProblemService.php
```diff
@@ -796,6 +813,20 @@ public function importZippedProblem(
 
         $messages['info'][] = sprintf('Saved problem %d', $problem->getProbid());
 
+        // Only here disable problem submit to make sure the jury submissions
+        // do get added above.
+        if ($contestProblem) {
+           $this->em->flush();
```
nickygerritsen, line 819: PHPCS is right here, this should use 4 spaces, not 2, to indent

nickygerritsen (approved):
2 small style nits, code looks good.

webapp/src/Service/ImportProblemService.php
```diff
@@ -318,6 +318,14 @@ public function importZippedProblem(
             }
         }
 
+        $this->em->persist($problem);
+        if ($contestProblem) {
+            $contestProblem->setProblem($problem);
+            $contestProblem->setContest($contest);
+            $this->em->persist($contestProblem);
+            $this->em->flush();
```
meisterT, line 326: should this be outside of the flush?

webapp/src/Service/ImportProblemService.php
```diff
@@ -796,6 +813,20 @@ public function importZippedProblem(
 
         $messages['info'][] = sprintf('Saved problem %d', $problem->getProbid());
 
+        // Only here disable problem submit to make sure the jury submissions
+        // do get added above.
+        if ($contestProblem) {
+            $this->em->flush();
+            $testcases = $problem->getTestcases()->toArray();
+            if (count(array_filter($testcases, function($t) { return !$t->getDeleted(); }))==0) {
+                $messages['warn'][] = 'No testcases present, disabling submitting for this problem';
```
meisterT, line 822: should this be danger?

PR #2024, opened 2023-05-07
Upgrade the component to development version

vmcj:
The current version does not understand enum in combination with $this. See: https://github.com/PHPCompatibility/PHPCompatibility/issues/1343

Normally we wouldn't like having to run a pre-release version but this is for a very minor component (check if code is acceptable in this PHP version) and the image is only used for this test so shouldn't have an effect on other tests.

### DIFF
--- a/.github/jobs/fix_pipelinecomponents_image.sh
+++ b/.github/jobs/fix_pipelinecomponents_image.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+set -eux
+
+echo "Set plugin config for version detection"
+phpcs --config-set installed_paths /app/vendor/phpcompatibility/php-compatibility
+      
+# Current released version does not know enums
+echo "Upgrade the compatibility for PHP versions"
+mydir=$(pwd)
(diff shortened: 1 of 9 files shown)


vmcj:
This PR is mostly to see if the issue is now fixed, should be draft.

.github/jobs/fix_pipelinecomponents_image.sh
```diff
@@ -0,0 +1,13 @@
...
+
+set -eux
+
+echo "Set plugin config for version detection"
+phpcs --config-set installed_paths /app/vendor/phpcompatibility/php-compatibility
+      
+# Current released version does not know enums
+echo "Upgrade the compatibility for PHP versions"
+mydir=${pwd}
+sed -i 's/"phpcompatibility\/php-compatibility": "9.3.5"/"phpcompatibility\/php-compatibility": "develop-dev"/g' /app/composer.json
```
nickygerritsen, line 11: shouldn't it be `dev-develop`?

nickygerritsen (approved):

PR #2026, opened 2023-05-07
Add support for importing submissions as shadow that we don't accept, add custom verdicts and update unverified badge

### LINKED ISSUE #272: Judge error in external diff
Kattis exposed a judge error to us, which was not shown in the external diff (probably because the submission was still judging on our side). We might want to extend the filter to show also judging submissions on our side to see how far we are behind Kattis.

### LINKED ISSUE #1848: "verified" badge on shadow page is confusing
For all submission lists, we display badges to show how many submissions we got, how many are queued, etc.

Now in shadow mode, when looking at differences, you see something like this:
image

Note that the differences are verified, but the submissions are not, making this confusing.

Ideas welcome how to make this less confusing.

### LINKED ISSUE #1849: In shadow mode, take in data even if DOMjudge fails to import submissions

In shadow mode, DOMjudge may reject submissions or fail to import them even though the primary system accepted it.

This could be because of additional (perhaps misconfigured) restrictions such as invalid filenames or because the primary system fails to produce ZIP files for some sources.

In DOMjudge, you still would want to ingest all the dependent data such as verdicts and test case results (as well as the source code) to look at it. Also scoring is based on this data in shadow mode.
 
There is multiple ways to solve this problem, one is to duplicate the submission entity and duplicate it for external submissions; the other one is to add a boolean to the existing entity to know if we want to create judge tasks for it. We will likely implement both to see which option is cleaner / has less maintenance overhead.

nickygerritsen:
Fixes #1849, fixes #272, fixes #1848

So this does a couple of things:
* It adds a flag to `submitSolution` to allow to import a submission that is not valid according to us. This will not create a judging or judgetasks, but will import the submission and show it as `import-error`. This allows us to import submissions (and their sourcecode if available) so we can still compare the shadow data.
* Add a config setting for custom judgement types and use this to import unknown judgement types as shadow. For this I refactored out the logic to get verdicts from verdicts.php, so it is less duplicated code. This now has a flag to also add external verdicts, which is used in a few places (shadow diff, the API).
* Replace the `x unverified` badge with a badge showing how many unverified shadow differences there are. Also only show the verify mark form when there is a difference.

### DIFF
--- a/webapp/src/Entity/Submission.php
+++ b/webapp/src/Entity/Submission.php
@@ -268,7 +268,7 @@ class Submission extends BaseApiEntity implements ExternalRelationshipEntityInte
         return $this;
     }
 
-    public function isImportError(): string
+    public function isImportError(): ?string
     {
         return $this->importError;
     }
--- a/webapp/src/Entity/Submission.php
+++ b/webapp/src/Entity/Submission.php
@@ -90,11 +90,13 @@ class Submission extends BaseApiEntity implements ExternalRelationshipEntityInte
     private ?string $entry_point = null;
 
     #[ORM\Column(
-        nullable: false,
-        options: ['comment' => 'Whether this submission was imported during shadowing but had an error while doing so.']
+        nullable: true,
+        options: ['comment' => 'If this submission was imported during shadowing but had an error while doing so, the error message.']
     )]
+    #[OA\Property(nullable: true)]
+    #[Serializer\Expose(if: "context.getAttribute('domjudge_service').checkrole('jury')")]
     #[Serializer\Groups([ARC::GROUP_NONSTRICT])]
-    private bool $importError = false;
+    private ?string $importError = null;
 
     #[ORM\ManyToOne(inversedBy: 'submissions')]
     #[ORM\JoinColumn(name: 'cid', referencedColumnName: 'cid', onDelete: 'CASCADE')]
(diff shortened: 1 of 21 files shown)


webapp/src/Entity/Submission.php
```diff
@@ -89,6 +89,13 @@ class Submission extends BaseApiEntity implements ExternalRelationshipEntityInte
     #[Serializer\Expose(if: "context.getAttribute('domjudge_service').checkrole('jury')")]
     private ?string $entry_point = null;
 
+    #[ORM\Column(
+        nullable: false,
+        options: ['comment' => 'Whether this submission was imported during shadowing but had an error while doing so.']
+    )]
+    #[Serializer\Groups([ARC::GROUP_NONSTRICT])]
+    private bool $importError = false;
```
meisterT, line 97: Can we make this a string, not a boolean? Or add an extra string? That contains the actual error?

meisterT (changes requested):

webapp/src/Entity/Submission.php
```diff
@@ -90,11 +90,13 @@ class Submission extends BaseApiEntity implements ExternalRelationshipEntityInte
     private ?string $entry_point = null;
 
     #[ORM\Column(
-        nullable: false,
-        options: ['comment' => 'Whether this submission was imported during shadowing but had an error while doing so.']
+        nullable: true,
+        options: ['comment' => 'If this submission was imported during shadowing but had an error while doing so, the error message.']
```
vmcj, line 94:
```suggestion
        options: ['comment' => 'The error message for submissions which got an error during shadow importing.']
```

I find the sentence hard to read otherwise.

meisterT (approved):

PR #2031, opened 2023-05-07
Display timelimits at time of running in runtime graph.

### LINKED ISSUE #1859: Display time limit per judging or judge run
In the graphs on the submission page, it would be nice to display the old time limit for earlier judgings. Possibly, the time limit could even change between running test cases.

As a jury member, I would like to see the results of rejudgings with different time limits, and how the time limits affected the verdict.

### Expected behaviour
On the page `/jury/submissions/{sid}`, it would be nice if the "Testcase Runtimes" and "Max Runtimes" graphs would show the different time limits for the different runs.

A (manually edited) screenshot of how I imagine this would look like. The first judging was ran with a time limit of 9 seconds, while the remaining two runs were run with a time limit of 8 seconds, which is also the current time limit of the problem.

image

nickygerritsen:
Fixes #1859

We decided to not show the current timelimit as a separate line, but just dasherize / not dasherize the line. Also no vertical bars, since it seems to look better.

Screenshot:
Screenshot 2023-05-07 at 20 45 52

cc @mpsijm

### DIFF
--- a/webapp/src/Controller/Jury/SubmissionController.php
+++ b/webapp/src/Controller/Jury/SubmissionController.php
@@ -209,9 +209,27 @@ class SubmissionController extends BaseController
             ->getQuery()
             ->getResult();
 
+        // These three arrays are indexed by judgingid
         /** @var Judging[] $judgings */
         $judgings    = array_map(fn($data) => $data[0], $judgingData);
         $maxRunTimes = array_map(fn($data) => $data['max_runtime'], $judgingData);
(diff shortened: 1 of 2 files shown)


nickygerritsen:
Also the line is now 2 pixels high, seems to be a bit better on the eyes.

eldering (approved):

mpsijm (approved):
Oooeeehhh, looking shiny! :star_struck:

Issue #1721, opened 2022-10-22
Filter away the jury submissions in the balloon interface


vmcj:
We should (default) filter away the balloons for jury submissions (done for the verifier), I think we should keep the ones done by a jury member to test that balloons functionality does work.

meisterT:
We do have a filter option on that page already:
![image](https://user-images.githubusercontent.com/418721/197345970-930359e8-b7de-420c-aaf9-e3661827245c.png)

So my proposal is to add a filter for team category and by default set it to all non-hidden categories. WDYT?

vmcj:
> We do have a filter option on that page already: ![image](https://user-images.githubusercontent.com/418721/197345970-930359e8-b7de-420c-aaf9-e3661827245c.png)
> 
> So my proposal is to add a filter for team category and by default set it to all non-hidden categories. WDYT?

I think that is a good fix, sounds like good first time issue.

nickygerritsen:
@brinky look at this beauty

nickygerritsen:
Fixed in #2034.

PR #2034, opened 2023-05-07
#1721 balloon category filter

brinky:
I think the default filter part should be generalized


webapp/templates/jury/balloons.html.twig
```diff
@@ -91,7 +107,16 @@
             });
 
             $('#clear-filters').on('click', function () {
-                $('select[data-filter-field]').val([]).trigger('change');
+                $('select[data-filter-field]').each(function () {
+                    var $filterField = $(this);
+                    if ($filterField.data('filter-field') == "category-id") {
+                        $filterField.val([
+                            {{- defaultCategories|join(',') -}}
```
nickygerritsen, line 114: I think there is a `| json` filter

nickygerritsen (commented):
Codestyle reports on some valid remarks as well.

nickygerritsen (approved):

Issue #2023, opened 2023-05-07
Deleting a submission doesn't cascade to judgetask


eldering:
I imported and then again deleted a problem with jury submissions. Then I start the judgedaemon and I get the error below.
It seems that deleting a submission does not cascade to the associated judgetask.
```
$ ./judgedaemon
[May 07 09:31:25.166] judgedaemon[65818]: Judge started on tiger [DOMjudge/8.3.0DEV/2cbd6060f]
[May 07 09:31:25.166] judgedaemon[65818]: 🔏 Executing chroot script: 'chroot-startstop.sh check'
[May 07 09:31:25.187] judgedaemon[65818]: Registering judgehost on endpoint default: http://localhost/domjudge/api
[May 07 09:31:25.328] judgedaemon[65818]: ⇝ Received 1 'judging_run' judge tasks (endpoint default)
[May 07 09:31:25.328] judgedaemon[65818]:   Working directory: /home/jaap/domjudge/git/domjudge/output/judgings/tiger/endpoint-default/2773/2796
[May 07 09:31:25.329] judgedaemon[65818]:   🔒 Executing chroot script: 'chroot-startstop.sh start'
[May 07 09:31:25.512] judgedaemon[65818]: warning: Error while executing curl GET to url http://localhost/domjudge/api/judgehosts/get_files/source/2773?: http status code: 404, request size = 0, response: array (
  'code' => 404,
  'message' => 'Source code for submission with ID \'2773\' not found',
  'class' => 'Symfony\\Component\\HttpKernel\\Exception\\NotFoundHttpException',
  'trace' => 
  array (
    0 => 
    array (
      'namespace' => '',
      'short_class' => '',
      'class' => '',
      'type' => '',
      'function' => '',
      'file' => '/home/jaap/domjudge/git/domjudge/webapp/src/Controller/API/JudgehostController.php',
      'line' => 1182,
      'args' => 
      array (
      ),
    ),
    1 => 
    array (
      'namespace' => 'App\\Controller\\API',
      'short_class' => 'JudgehostController',
      'class' => 'App\\Controller\\API\\JudgehostController',
      'type' => '->',
      'function' => 'getSourceFiles',
      'file' => '/home/jaap/domjudge/git/domjudge/webapp/src/Controller/API/JudgehostController.php',
      'line' => 1162,
      'args' => 
      array (
      ),
    ),
    2 => 
    array (
      'namespace' => 'App\\Controller\\API',
[...]

Issue #243, opened 2016-11-05
Problems Cannot be Deleted


jGleitz:
If I try to delete a problem, DOMJudge warns me:

> Warning, this will:
cascade to contestproblem
cascade to event
cascade to submission
cascade to testcase

which sounds reasonable to me. However, if I confirm the deletion, I get this error:

```
error: SQL error, Error#1451: Cannot delete or update a parent row: a foreign key constraint fails (`domjudge`.`judging_run`, CONSTRAINT `judging_run_ibfk_1` FOREIGN KEY (`testcaseid`) REFERENCES `testcase` (`testcaseid`)), query: 'DELETE FROM problem WHERE `probid` = "18" LIMIT 1'
```

If I understand this right, the problem cannot be deleted because it would result in database constraints of judging runs being violated. But as the deletion cascades to submissions anyway, this judging runs could be deleted, too.

eldering:
Indeed, see also #189. Clearly this has to be supported. It seems to me that the order of deleting referenced foreign keys should be fixed to allow this here. That is, if the problem first cascades to submissions, judgings, and judging_runs, then the testcases can be deleted afterwards.

eldering:
Note also that our check for foreign key dependencies in `www/jury/delete.php` is not complete because it doesn't search tables recursively.

eldering:
@jGleitz could you confirm if this fixes the issue for you?

jGleitz:
Sorry for the late reply

> @jGleitz could you confirm if this fixes the issue for you?

No, unfortunately not. Neither applying the changes of f7a63e6 nor using `delete.php` from `master` fixes the problem from me.

The error message is the same as in the initial report.

Do I have to upgrade DOMJudge completely to `master` to test the fix? I’m a bit reluctant to do that, as I don’t know an easy way to set up a DOMJudge test system, nor want to put my production system to an unstable version.

eldering:
Which version of DOMjudge are you currently using? If 5.1, then either applying f7a63e6 or using the version from `master` should have worked.
Could you do some manual testing on the database? (You can simply copy it to a different name or host.) In particular, if you execute directly in a mysql client or e.g. phpmyadmin the query
```
DELETE FROM problem WHERE probid = 'X' LIMIT 1
````
where `X` is an appropriate problem ID, does the query fail too?
Then, if you first execute
```
DELETE FROM submission WHERE probid = 'X'
```
and after that the perform the delete problem query, does it work?

[...]

PR #2040, opened 2023-05-08
Fix removal of dangling problem/submission related data.

meisterT:
The only way to delete a submission in the web interface is to delete a problem. The dependant data in other tables does not form a tree, so there is multiple paths which does not (necessarily) work in MySQL.

Handle this explicitly in the base controller when we are deleting the problem entity.

See https://github.com/DOMjudge/domjudge/issues/243 for some historical context.

Fixes https://github.com/DOMjudge/domjudge/issues/2023

### DIFF
--- a/webapp/migrations/Version20230508135106.php
+++ b/webapp/migrations/Version20230508135106.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20230508135106 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Replace jobid with foreign key constraint.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('DROP INDEX jobid ON queuetask');
+        $this->addSql('ALTER TABLE queuetask ADD judgingid INT UNSIGNED DEFAULT NULL COMMENT \'Judging ID\', DROP jobid');
+        $this->addSql('ALTER TABLE queuetask ADD CONSTRAINT FK_45E85FF85D5FEA72 FOREIGN KEY (judgingid) REFERENCES judging (judgingid) ON DELETE CASCADE');
+        $this->addSql('CREATE INDEX judgingid ON queuetask (judgingid)');
+    }
+
+    public function down(Schema $schema): void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE queuetask DROP FOREIGN KEY FK_45E85FF85D5FEA72');
+        $this->addSql('DROP INDEX judgingid ON queuetask');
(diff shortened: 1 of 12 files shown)


nickygerritsen (approved):

webapp/src/Controller/BaseController.php
```diff
@@ -219,21 +219,45 @@ protected function commitDeleteEntity(
...
-                // Deleting a problem is a special case: its dependent tables do not
-                // form a tree, and the deletion of judging_run can only cascade from
-                // judging, not from testcase. Since MySQL does not define the
-                // order of cascading deletes, we need to manually first cascade
-                // via submission -> judging -> judging_run.
+                // Deleting a problem is a special case:
+                // Its dependent tables do not form a tree (but something like a diamond shape),
+                // and there are multiple cascading removal paths from problem to its dependant
+                // tables.
+                // Since MySQL does not define the order of cascading deletes, we need to manually
```
eldering, line 226: Remove duplicate `manually`:
```suggestion
                // Since MySQL does not define the order of cascading deletes, we need to
```

webapp/src/Controller/BaseController.php
```diff
@@ -219,21 +219,45 @@ protected function commitDeleteEntity(
...
+                         INNER JOIN judging j ON jr.judgingid = j.judgingid
+                         INNER JOIN submission s ON j.submitid = s.submitid
+                         WHERE s.probid = :probid',
+                    ['probid' => $entity->getProbid()]
+                );
+
+                // Then delete all outstanding queueTasks.
+                $entityManager->getConnection()->executeQuery(
+                    'DELETE qt FROM queuetask qt
+                         INNER JOIN submission s ON qt.jobid = s.submitid
```
eldering, line 243: This should be submitid => judgingid.

webapp/migrations/Version20230508135106.php
```diff
@@ -0,0 +1,42 @@
...
+    public function getDescription(): string
+    {
+        return 'Replace jobid with foreign key constraint.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('DROP INDEX jobid ON queuetask');
+        $this->addSql('ALTER TABLE queuetask ADD judgingid INT UNSIGNED DEFAULT NULL COMMENT \'Judging ID\', DROP jobid');
```
eldering, line 24: It would be better to make this a `CHANGE COLUMN` statement (also in the down), so that data is preserved.

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -1339,7 +1339,7 @@ public function getJudgeTasksAction(Request $request): array
             // Note: we are joining on queue tasks here since if there is no more queue task, there is also no more
             // work to be done. If we would not do this join, the getJudgetasks would try to delete the queue task,
             // which is both slow and results in spamming the auditlog
-            ->innerJoin(QueueTask::class, 'qt', Join::WITH, 'qt.jobid = jt.jobid')
+            ->innerJoin(QueueTask::class, 'qt', Join::WITH, 'qt.judging = jt.jobid')
```
eldering, line 1342: So a `judgetask.jobid` could still refer to different types of tables?

meisterT: Yes, let's change that in a separate PR.

PR #2044, opened 2023-05-08
Make problem.externalid a unique key.

eldering:
This was only meant to be unique per contest, but this caused subtle issues with the POST /api/problems endpoint. In that case, there might be multiple problems with the given externalid and if this is used as external problem ID, then that would fail.


nickygerritsen (approved):

vmcj (approved):
LGTM

meisterT (approved):

Issue #862, opened 2020-08-31
Installation process not properly separated


thijskh:
If you follow the installation steps in the manual it does not currently work. This is because they are spilt over domserver and judgehost and it's not currently possible to install on a machine only a judgehost when installing only the dependencies of that judgehost.

The current (automake) system requires that you install the superset of dependencies for domserver, judgehost and submit client for it to pass. And you are required to pass configuration (baseurl) that is only relevant for one component (submit).

This is not a documentation issue but an issue with the installation system. It should be possible to install a judgehost on a system with no need for any dependencies of the domserver or for configuration options of the domserver.

thijskh:
One approach could be to give each component its own (automake or other) installation process in its own subdir in the tarball (odmjudge.tar.gz would contain `submit/`, `judge/` and `server/` and each has a configure. This keeps it very clean and there's virtually no shared code or options. (An even more radical variant of this is to split the judgehost and submit client off to separate repositories and released tarballs. Drawback is that it requires more fiddling if you _are_ installing everything on one machine.)

Another approach may be to keep the current integrated configuration but make it more dynamic by requiring that you --enable the parts that you are building on this machine.

mvr320:
> And you are required to pass configuration (baseurl) that is only relevant for one component (submit).
The docs also use it.

I think the --enable route is friendly enough and makes that people only have to stay in a single directory for everything to work. I think me as a user would prefer that and if I would be inexperienced with Linux this is the approach which can create the least _hard_ to debug situations.

Ofcourse my argument becomes invalid if people would use the debs/docker when configure is something they have problems with.

eldering:
Alternative/combination of already suggested options: I think that autoconf allows for sub-projects, so we could have independent subdirectories with their own configure/make code and then in the root directory a configure/make script that allows building all together.
I think that already matches pretty much how `webapp`, `submit`, `judge`, `doc` are split up, but the configuration might be the most tricky one to deal with.

vmcj:
Closed with: ef548499cdd1c727a974300cb2a03834fcc59d8f

PR #2051, opened 2023-05-11
Document apt packages manual

vmcj:
I'm not really sure about the 1st commit, the packages are not actually needed so this would be a "white lie", but I assume 99% of the people use the configure script and as thats broken (https://github.com/DOMjudge/domjudge/issues/862) this might make the route slightly better.

For the judgehosts I think we should actually either add `gcc` or `g++` as I think runguard is compiled outside of the chroot?

### DIFF
--- a/doc/manual/install-domserver.rst
+++ b/doc/manual/install-domserver.rst
@@ -37,14 +37,14 @@ GNU/Linux, or one of its derivative distributions like Ubuntu::
 
   sudo apt install acl zip unzip mariadb-server apache2 \
         php php-fpm php-gd php-cli php-intl php-mbstring php-mysql \
-        php-curl php-json php-xml php-zip composer ntp
+        php-curl php-json php-xml php-zip composer ntp python3-yaml
 
 The following command can be used on RedHat Enterprise Linux, and related
 distributions like CentOS (with EPEL) and Fedora::
 
   sudo yum install acl zip unzip mariadb-server httpd \
         php-gd php-cli php-intl php-mbstring php-mysqlnd \
-        php-xml php-zip composer ntp
+        php-xml php-zip composer ntp python3-pyyaml
 
 Due to an `issue <https://github.com/DOMjudge/domjudge/issues/862>`_ in our
 configure script you should also install the software for the judgehost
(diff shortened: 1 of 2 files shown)


doc/manual/install-judgehost.rst
```diff
@@ -31,7 +31,7 @@ Software requirements
 For Debian::
 
   sudo apt install make pkg-config sudo debootstrap libcgroup-dev \
-        php-cli php-curl php-json php-xml php-zip lsof procps
+        php-cli php-curl php-json php-xml php-zip lsof procps gcc g++
```
eldering, line 34: It is true that you need a C/C++ compiler, but that could also be `clang`. But maybe mention that as a comment below this.

eldering: The Red Hat packages should then also be updated below.

doc/manual/install-domserver.rst
```diff
@@ -37,7 +37,7 @@ GNU/Linux, or one of its derivative distributions like Ubuntu::
 
   sudo apt install acl zip unzip mariadb-server apache2 \
         php php-fpm php-gd php-cli php-intl php-mbstring php-mysql \
-        php-curl php-json php-xml php-zip composer ntp
+        php-curl php-json php-xml php-zip composer ntp gcc g++
```
eldering, line 40: I think it would be better to here also refer to the judgehost requirements due to #862 than to add the dependencies.

vmcj:
@eldering I've now also added a part about RHEL. It feels to small to do in another PR.

doc/manual/install-domserver.rst
```diff
@@ -74,6 +78,12 @@ DOMjudge uses a MySQL or MariaDB database server for information storage.
 Where this document talks about MySQL, it can be understood to also apply
 to MariaDB.
 
+For the install option (unless using bare-install) steps you need the
+python `yaml` module, for Debian GNU/Linux, or one of its derivative 
+distributions like Ubuntu::
+
+  sudo apt install python3-yaml
+
```
eldering, line 86: Let's just add this to the overall package install list. I don't think that's a big burden and it simplifies things.

doc/manual/install-domserver.rst
```diff
@@ -40,12 +40,16 @@ GNU/Linux, or one of its derivative distributions like Ubuntu::
...
 
 The following command can be used on RedHat Enterprise Linux, and related
-distributions like CentOS and Fedora::
+distributions like CentOS (with EPEL) and Fedora::
 
   sudo yum install acl zip unzip mariadb-server httpd \
         php-gd php-cli php-intl php-mbstring php-mysqlnd \
         php-xml php-zip composer ntp
 
+Due to a `bug <https://github.com/DOMjudge/domjudge/issues/862>`_ in our
```
eldering, line 49: Nitpick: I don't think this is really a bug, but the documentation is not in line with how installation works.
```suggestion
Due to an `issue <https://github.com/DOMjudge/domjudge/issues/862>`_ in our
```

eldering (approved):

PR #2052, opened 2023-05-11
Improve on default executable build scripts.

nickygerritsen:
* We need to pass `$@` to get the arguments.
* Java needs the correct classpath.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -387,6 +387,7 @@ function fetch_executable_internal(
                 $buildscript = "#!/bin/sh\n\n";
                 $execlang = false;
                 $source = "";
+                $unescapedSource = "";
                 foreach ($langexts as $lang => $langext) {
                     if (($handle = opendir($execbuilddir)) === false) {
                         error("Could not open $execbuilddir");
@@ -395,7 +396,8 @@ function fetch_executable_internal(
                         $ext = pathinfo($file, PATHINFO_EXTENSION);
                         if (in_array($ext, $langext)) {
                             $execlang = $lang;
-                            $source = $file;
+                            $unescapedSource = $file;
+                            $source = dj_escapeshellarg($unescapedSource);
                             break;
                         }
                     }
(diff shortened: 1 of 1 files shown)


judge/judgedaemon.main.php
```diff
@@ -419,12 +419,15 @@ function fetch_executable_internal(
                         $buildscript .= "javac -cp ./ -d ./ '$source'.java\n";
                         $buildscript .= "echo '#!/bin/sh' > run\n";
                         // no main class detection here
-                        $buildscript .= "echo 'java -cp ./ '$source >> run\n";
+                        $buildscript .= "echo COMPARE_DIR=\$(dirname \"\$0\") >> run\n";
+                        $buildscript .= "echo 'java -cp \"\$COMPARE_DIR\" '$source \"\$@\" >> run\n";
```
meisterT, line 423: Why was this not necessary before?

nickygerritsen: Did we ever have a Java compare or run script before?

eldering, line 423: The single quotes look strange here: they don't wrap around the full string to be echo-ed and it seems that this could cause issues if `$source` contains weird characters, e.g. spaces/tabs. I think the simplest solution would be go apply `dj_escapeshellarg` to `$source`. Then it can be used as is here, but in the other commands, the quotes around it should be removed.

eldering: Also, I think that `\"\$@\"` is not correctly quoted, as the run script must contain `"$@"`. Can you check that that is the case?

nickygerritsen: I will check. The single quote was already weird before right?

meisterT: I think but am not 100% sure. Perhaps it also broke after we moved this to the chroot.

eldering: Indeed, it was there already.

nickygerritsen: They were wrong indeed, now they seem to be correct. This is the contents of `build` and `run`:

```bash
domjudge@domjudge-contributor:/Users/nicky/Projects/domjudge/output/judgings/domjudge-contributor-1/endpoint-default/executable/compare/31/67bf2c1b54f229b56dce724588d6bedf/build$ cat build
#!/bin/sh

javac -cp ./ -d ./ 'Main.java'
echo '#!/bin/sh' > run
echo 'COMPARE_DIR=$(dirname "$0")' >> run
echo 'java -cp "$COMPARE_DIR" Main "$@"' >> run
chmod +x run


domjudge@domjudge-contributor:/Users/nicky/Projects/domjudge/output/judgings/domjudge-contributor-1/endpoint-default/executable/compare/31/67bf2c1b54f229b56dce724588d6bedf/build$ cat run
#!/bin/sh
COMPARE_DIR=$(dirname "$0")
java -cp "$COMPARE_DIR" Main "$@"
```

judge/judgedaemon.main.php
```diff
@@ -409,22 +411,25 @@ function fetch_executable_internal(
...
+                        $buildscript .= "gcc -Wall -O2 -std=gnu11 $source -o run -lm\n";
                         break;
                     case 'cpp':
-                        $buildscript .= "g++ -Wall -O2 -std=gnu++17 '$source' -o run\n";
+                        $buildscript .= "g++ -Wall -O2 -std=gnu++17 $source -o run\n";
                         break;
                     case 'java':
-                        $source = basename($source, ".java");
-                        $buildscript .= "javac -cp ./ -d ./ '$source'.java\n";
+                        $buildscript .= "javac -cp ./ -d ./ $source\n";
```
meisterT, line 420: nit: the `/`-es are unnecessary

nickygerritsen:
I'm merging this now since it seems I need the `chmod +x` change.

PR #2053, opened 2023-05-11
Reload contest problem before modification after call to eventLogServ…

meisterT:
…ice.

### DIFF
--- a/webapp/src/Service/ImportProblemService.php
+++ b/webapp/src/Service/ImportProblemService.php
@@ -609,7 +609,6 @@ class ImportProblemService
         $probid = $problem->getProbid();
         $this->eventLogService->log('problem', $problem->getProbid(), $problemIsNew ? 'create' : 'update', $cid);
 
-
         // Submit reference solutions.
         if ($contest === null) {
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

webapp/src/Service/ImportProblemService.php
```diff
@@ -609,6 +609,7 @@ public function importZippedProblem(
         $probid = $problem->getProbid();
         $this->eventLogService->log('problem', $problem->getProbid(), $problemIsNew ? 'create' : 'update', $cid);
 
+
```
vmcj, line 612:
```suggestion
```

webapp/src/Service/ImportProblemService.php
```diff
@@ -819,6 +820,11 @@ public function importZippedProblem(
             $this->em->flush();
             $testcases = $problem->getTestcases()->toArray();
             if (count(array_filter($testcases, fn($t) => !$t->getDeleted())) == 0) {
+                // We need to reload the contest problem after the call(s) to the eventLogService.
```
vmcj, line 822: It is unclear to my (on first glance) why we have to do this in the `if` and not above.

PR #2055, opened 2023-05-17
Suggested partial fix for splitting installation process

vmcj:
I tried the suggestion of @eldering (https://github.com/DOMjudge/domjudge/issues/862#issuecomment-684066056) as I think that is the correct solution. It gave some non trivial problems which given my experience with autoconf were difficult to solve.

This should already be a start to annotate what can be moved to other `autoconf` files and might even by a temporary fix to make everything work as expected. So even if we merge this I think we should leave the issue open as debt.


configure.ac
```diff
@@ -248,24 +272,29 @@ else
...
 # Check option to build documentation. This option is provided to
 # allow disabling it.
-AC_ARG_ENABLE(doc-build,AS_HELP_STRING([--enable-doc-build],
-[enable building documentation (default: yes).]))
-
-if test "x$enable_doc_build" = "xno"; then
-	AC_SUBST(DOC_BUILD_ENABLED,[no])
-else
-	AC_SUBST(DOC_BUILD_ENABLED,[yes])
+if test "x$INSTALL_DOMSERVER" != xno; then
```
eldering, line 289: I don't think docs should be a subset of the domserver configure target, but a separate target.

Also would be good to keep the flags and variables for docs and domserver/judgehost consistent.

configure.ac
```diff
@@ -11,6 +11,23 @@ AC_DEFUN([AX_WITH_COMMENT], [
 	AC_ARG_WITH([ax_comment_$1],[$2],[],[])
 	])
 
+AC_SUBST(INSTALL_DOMSERVER,no)
+AC_ARG_ENABLE([domserver],AS_HELP_STRING([--enable-domserver],
+ [configure the domserver (default: no).]),
```
eldering, line 16: I'd suggest:
```suggestion
 [configure and build the domserver (default: no).]),
```
Same below for judgehost.

configure.ac
```diff
@@ -31,6 +48,7 @@ AC_SUBST(DOMJUDGE_VERSION, $PACKAGE_VERSION)
 AC_DEFINE_UNQUOTED(DOMJUDGE_VERSION, "$PACKAGE_VERSION",
 [DOMjudge version number (alias for PACKAGE_VERSION)])
 
+if test "x$INSTALL_JUDGEHOST" != xno; then
```
eldering, line 51: I didn't check this in detail, but did you test that this works as expected with passing `--disable-judgehost` also?

eldering (commented):
In general the idea looks fine, but see some comments, mostly on consistency.

vmcj:
I'll work on this again in the future but we need some extra tests for this before I'm confident that we can work on this.

eldering (approved):
Please rebase/squash before merging.

PR #2067, opened 2023-05-30
Add separate ZIP's for public and unfrozen scoreboard.

### LINKED ISSUE #2060: Download of scoreboard fails when non-public contest
The jury UI has the option to download the static scoreboard. The FPC jury uses this to compare the different sites to make a set which is accessible for all sites.

nickygerritsen:
Fixes #2060.

### DIFF
--- a/webapp/src/Controller/Jury/ContestController.php
+++ b/webapp/src/Controller/Jury/ContestController.php
@@ -24,6 +24,7 @@ use App\Service\AssetUpdateService;
 use App\Service\ConfigurationService;
 use App\Service\DOMJudgeService;
 use App\Service\EventLogService;
+use App\Service\ScoreboardService;
 use App\Utils\Utils;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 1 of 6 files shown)


webapp/src/Service/DOMJudgeService.php
```diff
@@ -1499,4 +1509,74 @@ public function getVerdicts(bool $mergeExternal = false): array
...
+
+        // Also copy in the webfonts
+        $webfontsPath = $publicPath . '/webfonts/';
+        foreach (glob($webfontsPath . '*') as $fontFile) {
+            $fontName = basename($fontFile);
+            $zip->addFile($fontFile, 'webfonts/' . $fontName);
+        }
+        $zip->close();
+
+        return Utils::streamZipFile($tempFilename, 'contest.zip');
```
thijskh, line 1579: It returns a different filename than the above urls suggest it will.

webapp/src/Controller/Jury/ContestController.php
```diff
@@ -944,4 +946,34 @@ private function doLock(int $contestId, bool $locked): Response
...
+        Request $request,
+        ScoreboardService $scoreboardService
+    ): Response {
+        /** @var Contest $contest */
+        $contest = $this->em->getRepository(Contest::class)->find($contestId);
+        if (!$contest) {
+            throw new NotFoundHttpException(sprintf('Contest with ID %s not found', $contestId));
+        }
+        return $this->dj->getScoreboardZip($request, $requestStack, $contest, $scoreboardService, true);
+    }
```
thijskh, line 978: Should this just be one action with the type being a parameter of sorts? Now the difference between the two actions is literally 1 bit.

webapp/src/Service/DOMJudgeService.php
nickygerritsen: I’m fine with renaming the url yes

meisterT (approved):

PR #2068, opened 2023-05-30
Limit height of contest banner.

### LINKED ISSUE #2062: Make contest banner a max height
This image is taking up way too much space, we should probably limit the width or height:
image

nickygerritsen:
Fixes #2062.

Before:
Screenshot 2023-05-30 at 10 42 54

After:
Screenshot 2023-05-30 at 10 42 42

### DIFF
--- a/webapp/public/style_domjudge.css
+++ b/webapp/public/style_domjudge.css
@@ -391,10 +391,13 @@ tr.ignore td, td.ignore, span.ignore {
 }
 
 .banner {
-    width: 98%;
+    max-width: 98%;
+    max-height: 100px;
     height: auto;
     display: block;
     margin-bottom: 2rem;
+    margin-left: auto;
+    margin-right: auto;
 }
 
 #teampicture {
--- a/webapp/public/style_domjudge.css
+++ b/webapp/public/style_domjudge.css
@@ -392,7 +392,7 @@ tr.ignore td, td.ignore, span.ignore {
 
 .banner {
     max-width: 98%;
-    max-height: 100px;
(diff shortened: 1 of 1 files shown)


thijskh (approved):

webapp/public/style_domjudge.css
```diff
@@ -391,10 +391,13 @@ tr.ignore td, td.ignore, span.ignore {
 }
 
 .banner {
-    width: 98%;
+    max-width: 98%;
+    max-height: 100px;
```
eldering, line 395: How does this look on a retina display? Shouldn't the max height also be a percentage or given in `x`'s maybe?

nickygerritsen: The screenshot was from a retina screen. I don’t get what you mean with `x` but using a percentage won’t work: that’s a percentage of the page, not the image

eldering: For percentage: fair enough. I guess in CSS it's `ex`, the height of the letter x in the current font. Just so things scale relative to the size of the rest on the page.
But TBH this is somewhat nitpicky.

nickygerritsen: I will change it to some ex them. Or em, that’s also a thing

eldering: But `em` is the unit size of the horizontal width of an m. So for vertical sizes, `ex` makes more sense. Although this is nitpicking in practice. :-)

nickygerritsen: I used 12ex, which is about 100px on my machine.

PR #2076, opened 2023-06-06
Add support for exporting full results.

nickygerritsen:
As discussed on the NAC Slack channel, we currently only do the results.tsv where everybody outside the medals with the same number of solved problems had the same rank, but some contests (like NAC) want a unique rank per team, so add support for it.

### DIFF
--- a/webapp/src/Controller/Jury/ImportExportController.php
+++ b/webapp/src/Controller/Jury/ImportExportController.php
@@ -277,10 +277,11 @@ class ImportExportController extends BaseController
         ]);
     }
 
-    #[Route(path: '/export/{type<groups|teams|results>}.tsv', name: 'jury_tsv_export')]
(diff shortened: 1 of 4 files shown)


webapp/src/Controller/Jury/ImportExportController.php
```diff
@@ -334,7 +343,7 @@ public function exportHtmlAction(Request $request, string $type): Response
         }
     }
 
-    protected function getResultsHtml(Request $request): Response
+    protected function getResultsHtml(Request $request, bool $full = false): Response
```
meisterT, line 346: I would consider replacing `$full` (not only here) with another variable with the opposite meaning, e.g. `$onlyMedals = false`.

nickygerritsen: But it’s not only medals. It’s more “same rank for same number of problems”. Suggestions?

meisterT: meh, then leave it like it is now

PR #2077, opened 2023-06-10
Add tests for configure invocations

vmcj:
I've used bats for this as we already used that before. Due to how GHA works the test is only added when its in the domjudge/main.

Some simple tests on workings, in a future PR I want to split this to separate tests for the judgehost and domserver.

This also uncovers some errors:
- when running with the FHS option we don't output where to store the example_problems
- the systemd check seems to fail. `   ./configure: line 4466: --variable=systemdsystemunitdir: command not found`

I'm not sure if there are other things which need to be tested before I change the autoconf to split into relevant sections for domserver & judgehost?


.github/jobs/configure-checks/all.bats
```diff
@@ -0,0 +1,371 @@
...
+}
+repo-remove () {
+    apt-get remove $@ -y >/dev/null; apt-get autoremove -y 2>/dev/null
+}
+remove_gnuc () {
+    repo-remove gcc g++
+}
+remove_c () {
+    remove_gnuc; repo-remove clang
+}
```
eldering, line 36: Nitpick: I think there's a bit too much indirection here: why declare functions `remove_{gnuc,c}` that are used twice, but do directly call `repo-{remove,install}` below for other things? I'd just use the latter functions directly.

.github/jobs/configure-checks/all.bats
```diff
@@ -0,0 +1,371 @@
...
+
+setup() {
+    setup_user
+    for shared_file in config.log confdefs.h conftest.err; do
+        chmod a+rw $shared_file || true
+    done
+    repo-install gcc g++ libcgroup-dev
+}
+
+q_configure () {
```
eldering, line 20: What the reason for the `q_` prefix?

.github/jobs/configure-checks/all.bats
```diff
@@ -0,0 +1,371 @@
...
+    assert_line "checking for cl.exe... no"
+    assert_line "configure: error: in \`/domjudge':"
+    assert_line 'configure: error: no acceptable C compiler found in $PATH'
+    assert_line "See \`config.log' for more details"
+}
+
+compiler_assertions () {
+    run q_configure
+    # Depending on where we run this we might runas wrong user or lack libraries
+    # assert_success
```
eldering, line 57: This looks like a leftover commented line of code.

.github/jobs/configure-checks/all.bats
```diff
@@ -0,0 +1,371 @@
...
+    # compile_assertions_finished
+}
+
+@test "Install C/C++ compilers (Clang as alternative)" {
+    remove_gnuc
+    repo-install clang libcgroup-dev
+    compiler_assertions cc c++
+    assert_line "checking for gcc... no"
+    assert_line "checking for cc... cc"
+    assert_line "checking baseurl... checking for g++... no"
```
eldering, line 124: This looks wrong: there should be output coming from the baseurl check.

.github/jobs/configure-checks/all.bats
```diff
@@ -0,0 +1,371 @@
...
+   refute_line "    - example_problems.: /opt/domjudge/domserver/example_problems"
+   refute_line " * judgehost...........: /opt/domjudge/judgehost"
+   refute_line "    - bin..............: /opt/domjudge/judgehost/bin"
+   refute_line "    - etc..............: /opt/domjudge/judgehost/etc"
+   refute_line "    - lib..............: /opt/domjudge/judgehost/lib"
+   refute_line "    - libjudge.........: /opt/domjudge/judgehost/lib/judge"
+   refute_line "    - log..............: /opt/domjudge/judgehost/log"
+   refute_line "    - run..............: /opt/domjudge/judgehost/run"
+   refute_line "    - tmp..............: /opt/domjudge/judgehost/tmp"
+   refute_line "    - judge............: /opt/domjudge/judgehost/judgings"
```
eldering, line 214: I think it would be better to have less `refute_line`s. If we test a few that should be enough and have less code bloat.

.github/workflows/autoconf-check.yml
```diff
@@ -0,0 +1,30 @@
...
+        #version: [sid, testing, stable, oldstable]
+    runs-on: ubuntu-latest
+    container:
+      image: debian:${{ matrix.version }}
+    steps:
+      - name: Install everything for configure and testing
+        run: apt-get update; apt-get install make bats autoconf git -y; apt-get full-upgrade -y
+      - uses: actions/checkout@v3
+      #- name: Get bats tools
+      #  run: cp ../../../submit/assert.bats ./
```
eldering, line 25: More leftover commented code.

.github/jobs/configure-checks/all.bats
vmcj: Sure, when writing the tests I expected to need it more often, I agree at this moment it doesn't really help.

.github/jobs/configure-checks/all.bats
vmcj: I needed a prefix to signal that this is non standard. Any suggestion for a good name? Otherwise I'll go with `run_configure`.

.github/jobs/configure-checks/all.bats
vmcj: I agree it should, but when I run `make clean; make configure; ./configure | grep baseurl` this is the output I get in my testsetup.

.github/workflows/autoconf-check.yml
```diff
@@ -12,7 +12,8 @@ jobs:
   debian-family:
     strategy:
       matrix:
-        version: [sid, testing, stable]
+        version: [testing, stable]
+        #version: [sid, testing, stable]
```
eldering, line 16: If the messages have changed, then I think the tests should be written more relaxed to accomodate that instead of ignoring this. Tests that overly rely on specifics of the environment are brittle, and might cause more harm than do good.

vmcj: This was to test if that was the case, the annoying thing with GHA is that is stops testing if its aware that another test in the same file has failed. Because `sid` is the first one it stops the other ones.

So my plan was to properly fix this, but since you last reviewed I found out that the tests were never started so they work locally but not in GHA.

.github/jobs/configure-checks/all.bats
vmcj: I've added a fix for this.

.github/jobs/configure-checks/all.bats
vmcj: Rewritten to make it clear why this is there.

.github/jobs/configure-checks/all.bats
```diff
@@ -0,0 +1,349 @@
...
+   assert_line "    - run..............: /usr/local/var/run/domjudge"
+   assert_line "    - tmp..............: /tmp"
+   assert_line "    - judge............: /usr/local/var/lib/domjudge/judgings"
+   assert_line "    - chroot...........: /chroot/domjudge"
+   assert_line "    - cgroup...........: /sys/fs/cgroup"
+}
+
+@test "Alternative dirs together with FHS" {
+   setup
+   run run_configure "--enable-fhs --with-domserver_webappdir=/run/webapp --with-domserver_tmpdir=/tmp/domserver --with-judgehost_tmpdir=/srv/tmp --with-judgehost_judgedir=/srv/judgings --with-judgehost_chrootdir=/srv/chroot/domjudge --with-judgehost_cgroupdir=/sys/fs/altcgroup"
```
eldering, line 253: Nitpick: quoting all the configure options inside a single string works, but only because it is passed to `su -c` in `run_configure` which word expands it again.

PR #2078, opened 2023-06-11
Disable scoreboard refresh for static scoreboard after contest has be…

### LINKED ISSUE #2071: Thawed static scoreboard contains uncessary refreshes
as seen on nac2023 scoreboard,  post thawed, it's still refreshes.   Seems those refresh bits should be suppressed on a thawed scoreboard.  seen on v=8.3.0DEV/471813069

meisterT:
…en unfrozen.

Fixes #2071

### DIFF
--- a/webapp/src/Service/ScoreboardService.php
+++ b/webapp/src/Service/ScoreboardService.php
@@ -889,6 +889,9 @@ class ScoreboardService
              ],
              'static' => $static,
         ];
+        if ($static && $contest && $contest->getFreezeData()->showFinal()) {
+            $data['refresh']['after'] = null;
+        }
 
         if ($contest) {
             if ($request && $response) {


nickygerritsen (approved):
If people loaded the page before the unfreeze happened, does this work?

vmcj:
> If people loaded the page before the unfreeze happened, does this work?

@meisterT I think you forgot to respond to @nickygerritsen.

meisterT:
No, it does still refresh... :-/

meisterT:
With the most recent commit it does.

Why did we use AJAX for the static scoreboard anyway? Does that even work when copying it over to a different location like a GCP bucket?

nickygerritsen:
> With the most recent commit it does.
> 
> Why did we use AJAX for the static scoreboard anyway? Does that even work when copying it over to a different location like a GCP bucket?

It does, I spend quite some time on it. We do it because it doesn’t visually refresh the page

meisterT:
Ok, I tried to fix it differently now and it works for me. Please have a look and I'll fix the conflict if you agree that this approach is OK.

webapp/public/js/domjudge.js
```diff
@@ -429,9 +429,12 @@ function processAjaxResponse(jqXHR, data) {
...
         var dataCurrentContest = $('[data-current-contest]').data('current-contest');
         var currentContest = parseInt(dataCurrentContest);
+        var refreshStop = $('[data-ajax-refresh-stop]');
 
-        // If the contest ID changed from another tab, reload or whole page
-        if (dataCurrentContest !== undefined && newCurrentContest !== currentContest) {
+        // Reload the whole page if
+        // - we were signaled to stop refreshing, or
+        // - if the contest ID changed from another tab.
+        if ("0" in refreshStop || (dataCurrentContest !== undefined && newCurrentContest !== currentContest)) {
```
nickygerritsen, line 437: We set refreshStop to 1, so why are we checking for refreshStop? Also shouldn't the line above be:
```
var refreshStop = $('[data-ajax-refresh-stop]').data('ajax-refresh-stop');
```
?

nickygerritsen (commented):
A bit confused how this works currently

meisterT:
> A bit confused how this works currently

I can understand, there were a few left-over changes from previous attempts in there. Should be fixed and clearer now. Also rebased on top of main.

webapp/public/js/domjudge.js
```diff
@@ -435,10 +435,13 @@ function processAjaxResponse(jqXHR, data) {
...
-        var dataCurrentContest = $('[data-current-contest]').data('current-contest').toString();
+        var dataCurrentContest = $('[data-current-contest]').data('current-contest');
+        var refreshStop = $('[data-ajax-refresh-stop]');
 
-        // If the contest ID changed from another tab, reload or whole page
-        if (dataCurrentContest !== undefined && newCurrentContest !== dataCurrentContest) {
+        // Reload the whole page if
+        // - we were signaled to stop refreshing, or
+        // - if the contest ID changed from another tab.
+        if ("0" in refreshStop || (dataCurrentContest !== undefined && newCurrentContest !== dataCurrentContest.toString())) {
```
nickygerritsen, line 444: I still don't get this `0` here, when we are setting a `="1"`

webapp/public/js/domjudge.js
meisterT: I was basically testing whether this signal was there at all (had a key `0`), but yeah this is nicer. Fixed.

nickygerritsen (approved):

webapp/public/js/domjudge.js
meisterT: See above for explanation.

PR #2079, opened 2023-06-12
Add contest.scoreboard_type to API

eldering:
This is to comply with the specification at https://ccs-specs.icpc.io/draft/contest_api#contests

Also use consistent snake_case for serialized property names.


nickygerritsen (approved):

vmcj:
This fails because the contest-api script is not updated.

Issue #2088, opened 2023-06-26
Add popup notification for clarification messages


2018hahazhufeng:
### Description of the enhancement request
Currently when participant recieves a message from jury, it'll only appear quietly. But a browzer popup notification is neccesary, especially during contest participant doesn't notice the domjudge page frequently.

eldering:
Note that we already have browser notifications for the jury, although they were broken recently, see #481.

So it probably should not be too difficult to also implement this on the team side. Would you want to consider creating a pull request?

2018hahazhufeng:
I‘d like to make PRs but I have no relevant develop experience. Anyway I'll make a try

vmcj:
Alternative is to add the number of messages to the menu? We already do this for Balloons (when balloonrunner) and errors (when admin).

PR #2089, opened 2023-06-28
Add desktop notification for team receiving clarifications

2018hahazhufeng:
For team interface, add a button to enable notification. Implement sending desktop notification when receiving new clarification messages. Fix bugs of previous enable notification function. #2088

### DIFF
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -44,8 +44,6 @@ function enableNotifications()
             return true;
         });
     }
-
-    
 }
 
 function disableNotifications()
@@ -67,8 +65,8 @@ function disableNotifications()
 // client has already received to display each notification only once.
 function sendNotification(title, options = {})
 {
-    
     if ( getCookie('domjudge_notify')!=1 ) return;
+    
     // Check if we already sent this notification:
     var senttags = localStorage.getItem('domjudge_notifications_sent');
     if ( senttags===null || senttags==='' ) {
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -17,7 +17,14 @@ function enableNotifications()
               'Re-enable notification permission in the browser and retry.');
         return false;
     }
-
+    if ( Notification.permission==='granted' ) {
+        setCookie('domjudge_notify', 1);
+        sendNotification('DOMjudge notifications enabled.');
+        $("#notify_disable").removeClass('d-none');
+        $("#notify_disable").show();
+        $("#notify_enable").hide();
+        return true;
+    }
     if ( Notification.permission!=='granted' ) {
         Notification.requestPermission(function (permission) {
             // Safari and Chrome don't support the static 'permission'
@@ -29,15 +36,16 @@ function enableNotifications()
                 alert('Browser denied permission to send desktop notifications.');
                 return false;
             }
+            setCookie('domjudge_notify', 1);
+            sendNotification('DOMjudge notifications enabled.');
+            $("#notify_disable").removeClass('d-none');
+            $("#notify_disable").show();
+            $("#notify_enable").hide();
+            return true;
         });
     }
 
-    setCookie('domjudge_notify', 1);
-    sendNotification('DOMjudge notifications enabled.');
-    $("#notify_disable").removeClass('d-none');
-    $("#notify_disable").show();
-    $("#notify_enable").hide();
-    return true;
+    
 }
 
 function disableNotifications()
@@ -59,8 +67,8 @@ function disableNotifications()
 // client has already received to display each notification only once.
 function sendNotification(title, options = {})
 {
+    
     if ( getCookie('domjudge_notify')!=1 ) return;
-
     // Check if we already sent this notification:
     var senttags = localStorage.getItem('domjudge_notifications_sent');
     if ( senttags===null || senttags==='' ) {
@@ -80,7 +88,9 @@ function sendNotification(title, options = {})
     var not = new Notification(title, options);
 
     if ( link!==null ) {
-        not.onclick = function() { window.open(link); };
+        not.onclick = function() { 
+            window.location.href = link;
+        };
     }
 
     if ( options.tag ) {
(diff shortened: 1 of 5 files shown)


webapp/public/js/domjudge.js
```diff
@@ -80,7 +86,9 @@ function sendNotification(title, options = {})
     var not = new Notification(title, options);
 
     if ( link!==null ) {
-        not.onclick = function() { window.open(link); };
+        not.onclick = function() { 
+            window.location.href = link;
```
nickygerritsen, line 90: What is the reason this has been changed? I think it should both work but maybe you had a reason it didn't?

webapp/src/Controller/Team/MiscController.php
```diff
@@ -129,6 +130,12 @@ public function homeAction(Request $request): Response
         return $this->render('team/index.html.twig', $data);
     }
 
+    #[Route(path: '/updates', methods: ['GET'], name: 'team_ajax_updates')]
```
nickygerritsen, line 133: Given that the function contains 'unread notifications' in the name, maybe the route should be `/updates-unread-notifications`? and also not `team_ajax_updates`. Or we call the method `updates()` and we change the body to
```php
 return ['unread_clarifications' => $this->json($this->dj->getUnreadClarifications())];
```

webapp/templates/team/base.html.twig
```diff
@@ -14,6 +14,17 @@
...
+
+            /* Show the notification options if the browser supports it */
+            if ('Notification' in window) {
+                $('#notify_enable').click(enableNotifications);
+                if (getCookie('domjudge_notify') != 1) {
+                    $('#notify_enable').removeClass('d-none');
+                }
+            }
+            updateClarifications();
+            setInterval(updateClarifications, 20000);
```
nickygerritsen, line 26: Will this re-add all unread notifications every 20 seconds if teams dismiss them? That's maybe not what we want?

webapp/templates/team/menu.html.twig
```diff
@@ -53,8 +53,15 @@
                 </li>
             {% endif %}
         </ul>
-
-        <div class="ms-auto me-3">
+        
+        <div class="ml-auto me-3">
+            <a class="nav-link d-none" id="notify_enable">
```
nickygerritsen, line 58: We should also need the `notify_disable` link here right?

nickygerritsen (commented):
Some small comments 😄

webapp/public/js/domjudge.js
2018hahazhufeng: `window.open` will open the link in a new tab but won't switch to it, which might be confusing, so I replace that to open the link just in current tab.

nickygerritsen: Hmmm I think that does make more sense yes.

webapp/src/Controller/Team/MiscController.php
2018hahazhufeng: Sorry for the naming issue, I'll change it following your advice.

webapp/templates/team/base.html.twig
2018hahazhufeng: The sendNotification function use local storage to track which notification has been sent, so the same clar won't be popup twice.

nickygerritsen: Ah my bad, it has been a while since I looked at the method. Then it's fine.

webapp/templates/team/menu.html.twig
2018hahazhufeng: I was thinking of during the contest there'll be no disadvantage for a team to enable notifications, so no such button implemented. I'll add it if you think its necessary.

nickygerritsen: I would prefer teams to be able to enable/disable it. And if we show the enable button like here, we should also show the disable button.

2018hahazhufeng:
Yeah that's good. I agree

webapp/src/Controller/Team/MiscController.php
```diff
@@ -129,6 +130,12 @@ public function homeAction(Request $request): Response
...
+    public function updates(): JsonResponse
```
nickygerritsen, line 134: Sorry, maybe it should be `updatesAction` to reflect the other changes.

webapp/src/Service/DOMJudgeService.php
```diff
@@ -299,6 +299,21 @@ public function clearCookie(
...
     }
 
+    public function getUnreadClarifications(): array
+    {
+        $user           = $this->getUser();
+        $team           = $user->getTeam();
+        $clarifications = $team->getUnreadClarifications();
+        $unreadClarifications = [];
+        foreach ($clarifications as $clar) {
+            $unreadClarifications[] = [
```
nickygerritsen, line 309: Should we limit them to the currently selected contest?
I.e. use
```php
$contest = $this->getCurrentContest($team->getTeamId());
```

Then check if `$contest` is null (if so, do nothing) and if not only add clarifications that match the contest cid.

2018hahazhufeng: Sorry I didn't consider the case. I'll modify it.

webapp/src/Service/DOMJudgeService.php
```diff
@@ -299,6 +299,27 @@ public function clearCookie(
...
+        $user           = $this->getUser();
+        $team           = $user->getTeam();
+        $clarifications = $team->getUnreadClarifications();
+        $contest        = $this->getCurrentContest($team->getTeamId());
+        $unreadClarifications = [];
+        if ($contest === null) {
+            return $unreadClarifications;
+        }
+        foreach ($clarifications as $clar) {
+            if ($clar->getContest() === $contest) {
```
nickygerritsen, line 313: Did you check if this works? We might need
```suggestion
            if ($clar->getContest()->getCid() === $contest->getCid()) {
```

2018hahazhufeng: I've checked it really works. But I think it needs to be fixed

nickygerritsen (approved):
Looks good to me now. Let's see what others think.

thijskh (approved):

PR #2090, opened 2023-06-29
Restructure add/edit entity tests, add tests for expected failures

vmcj:
~~There are some comments on tests options which fail. I think leaving them as comment is smarter as they are todo to test but the current test does not simulate the proper behaviour.
I discussed with @nickygerritsen and removal of the webapp/var/cache does also not fix the assert/blank which is needed for this.~~

A small rewrite for the tests, this should make that we cover some unhappy paths.


webapp/src/Entity/Contest.php
```diff
@@ -389,9 +389,9 @@ public function getExternalid(): ?string
         return $this->externalid;
     }
 
-    public function setName(string $name): Contest
+    public function setName(?string $name): Contest
     {
-        $this->name = $name;
+        $this->name = strval($name);
```
nickygerritsen, line 394: I think I'd prefer `(string)$name` or even `$name ?? ''`

nickygerritsen: But maybe even better is the thing in the form types I suggested on Slack, with `empty_data`

nickygerritsen (approved):
Maybe rename the PR itself since it clearly isn't about contest timezones anymore (or at least about more)

webapp/src/Entity/Contest.php
vmcj: > But maybe even better is the thing in the form types I suggested on Slack, with `empty_data`

The reason I didn't went for that is that we now handle the case deeper in the code so if there is another way of passing the wrong data we still handle that in a reasonable way.

> I think I'd prefer (string)$name or even $name ?? ''

Fine with the first one,

vmcj: @nickygerritsen I've now also done what you mentioned in slack. I've left the work on the `set*Name` functions as I think the extra protection is useful.

nickygerritsen: I’m not really sure if I agree with passing wrong data (we also don’t have stuff like passing strings to functions that accept int nor should we) but I’m also not strongly against it

PR #2092, opened 2023-07-02
Remove unclear `ext` mention

vmcj:
It might have been from a type where languages can only have 1 extension, or been used as the externalid. Both can now be specified independant of the Entity ID.


nickygerritsen (approved):

eldering:
The "main extension" was the one that is used internally to store the filename. I guess this is not used anymore since we moved to supporting multiple files in a submission and kept the filenames as submitted.

PR #2093, opened 2023-07-03
Add an option to display the specific results of too-late submissions for teams

### LINKED ISSUE #1967: Consider adding an option to choose whether or not to display the results of TOO-LATE submissions on the Team page
DOMjudge is mainly used as a competition system.

However, we often use DOMjudge for training contest, and then after the training contest, the team members who take part in the training contest have the need to solve problems after the contest.

It would be useful to add an option to the configuration page to enable the display of TOO-LATE submissions on the team page.

2018hahazhufeng:
For team interface, show the specific judge results of too-late submissions. Meanwhile add a global configuration option to enable/disable this feature.
Closes #1967

### DIFF
--- a/etc/db-config.yaml
+++ b/etc/db-config.yaml
@@ -215,6 +215,11 @@
             default_value: false
             public: true
             description: Should teams be able to view a diff of their and the reference output on sample testcases?
+        -   name: show_too_late_result
+            type: bool
+            default_value: false
+            public: true
+            description: Show results of TOO-LATE submissions in team interface?
         -   name: show_balloons_postfreeze
             type: bool
             default_value: false
--- a/webapp/src/Controller/Team/SubmissionController.php
+++ b/webapp/src/Controller/Team/SubmissionController.php
@@ -111,12 +111,13 @@ class SubmissionController extends BaseController
     public function viewAction(Request $request, int $submitId): Response
     {
         $verificationRequired = (bool)$this->config->get('verification_required');
-        $showCompile      = $this->config->get('show_compile');
-        $showSampleOutput = $this->config->get('show_sample_output');
-        $allowDownload    = (bool)$this->config->get('allow_team_submission_download');
-        $user             = $this->dj->getUser();
-        $team             = $user->getTeam();
-        $contest          = $this->dj->getCurrentContest($team->getTeamid());
+        $showCompile          = $this->config->get('show_compile');
+        $showSampleOutput     = $this->config->get('show_sample_output');
+        $allowDownload        = (bool)$this->config->get('allow_team_submission_download');
+        $showTooLateResult    = $this->config->get('show_too_late_result');
+        $user                 = $this->dj->getUser();
+        $team                 = $user->getTeam();
+        $contest              = $this->dj->getCurrentContest($team->getTeamid());
         /** @var Judging|null $judging */
         $judging = $this->em->createQueryBuilder()
             ->from(Judging::class, 'j')
@@ -192,6 +193,7 @@ class SubmissionController extends BaseController
             'allowDownload' => $allowDownload,
             'showSampleOutput' => $showSampleOutput,
             'runs' => $runs,
+            'showTooLateResult' => $showTooLateResult,
         ];
         if ($actuallyShowCompile) {
             $data['size'] = 'xl';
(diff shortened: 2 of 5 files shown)


etc/db-config.yaml
```diff
@@ -215,6 +215,11 @@
             default_value: false
             public: true
             description: Should teams be able to view a diff of their and the reference output on sample testcases?
+        -   name: show_too-late_result
```
nickygerritsen, line 218: I would call it 
```suggestion
        -   name: show_too_late_result
```

to be consistent with other option naming.

etc/db-config.yaml
```diff
@@ -215,6 +215,11 @@
...
+            type: bool
+            default_value: false
+            public: true
+            description: Show specific results of TOO-LATE submissions in team interface?
```
nickygerritsen, line 222: `specific` doesn't add much here, so:
```suggestion
            description: Show results of TOO-LATE submissions in team interface?
```

webapp/src/Controller/Team/SubmissionController.php
```diff
@@ -114,6 +114,7 @@ public function viewAction(Request $request, int $submitId): Response
         $showCompile      = $this->config->get('show_compile');
         $showSampleOutput = $this->config->get('show_sample_output');
         $allowDownload    = (bool)$this->config->get('allow_team_submission_download');
+        $showToolateResult= $this->config->get('show_too-late_result');
```
nickygerritsen, line 117: Can we add a space before the `=`? (and realign the other stuff then)

nickygerritsen (approved):
That seemed easier than I would have expected 😄 .
Some minor comments but looks good.

webapp/src/Controller/Team/SubmissionController.php
2018hahazhufeng: Okay I'll change it.

nickygerritsen:
It's only some alignment for the whole table, right? In that case I wouldn't worry about it for now, we can fix it later.

2018hahazhufeng:
Yeah, the width of the result column is too big.

nickygerritsen:
> Yeah, the width of the result column is too big.

OK I'm fine with ignoring this for now.

nickygerritsen (approved):
LGTM

webapp/templates/team/partials/submission_list.html.twig
```diff
@@ -52,7 +52,16 @@
...
+                            {{ 'too-late' | printResult }} 
+                            {%- if showTooLateResult %}
+                                ({%- if submission.judgings.first is empty or submission.judgings.first.result is empty %}
+                                    {{- '' | printResult -}}
+                                {%- elseif verificationRequired and not submission.judgings.first.verified %}
+                                    {{- '' | printResult -}}
+                                {%- else %}
+                                    {{- submission.judgings.first.result | printResult -}}
+                                {%- endif %})
+                            {%- endif %}
```
vmcj, line 64: Is there a way to remove this duplication?

2018hahazhufeng: I've thought of this, but didn't find a better way to implement this.

The following code also deplicate some code but is shorter. Maybe it's better?
```
{%- if submission.submittime > submission.contest.endtime %}
    {{ 'too-late' | printResult }} 
{%- endif %}
{%- if submission.submittime <= submission.contest.endtime or showTooLateResult %}
    {%- if submission.submittime > submission.contest.endtime %}
        (
    {%- endif %}
    {%- if submission.judgings.first is empty or submission.judgings.first.result is empty %}
        {{- '' | printResult -}}
    {%- elseif verificationRequired and not submission.judgings.first.verified %}
        {{- '' | printResult -}}
    {%- else %}
        {{- submission.judgings.first.result | printResult -}}
    {%- endif %}
    {%- if submission.submittime > submission.contest.endtime -%}
        )
    {%- endif %}
{%- endif %}
```

vmcj: I like it a bit more. If you accept `too-late/<verdict>` you can also remove the last check.

2018hahazhufeng: Okay, I've made a commit of this.

vmcj (approved):

meisterT:
Thanks for your contribution! Can you please rebase this on top of head so we can merge it?

2018hahazhufeng:
Sorry I might have misunderstood and squashed those commits just now

PR #2094, opened 2023-07-03
Change compiler/runner version (commands) in API to conform with spec.

meisterT:
See
https://ccs-specs.icpc.io/draft/contest_api?highlight=languages#examples-5

### DIFF
--- a/webapp/src/Entity/Language.php
+++ b/webapp/src/Entity/Language.php
@@ -204,8 +204,6 @@ class Language extends BaseApiEntity
             if (!empty($this->getCompilerVersion())) {
                 $ret['version'] = $this->getCompilerVersion();
             }
-        } else {
-            return null;
         }
         return $ret;
     }
--- a/webapp/src/Entity/Language.php
+++ b/webapp/src/Entity/Language.php
@@ -119,15 +119,19 @@ class Language extends BaseApiEntity
     private Collection $versions;
 
     #[ORM\Column(type: 'blobtext', nullable: true, options: ['comment' => 'Compiler version'])]
+    #[Serializer\Exclude]
     private ?string $compilerVersion = null;
 
     #[ORM\Column(type: 'blobtext', nullable: true, options: ['comment' => 'Runner version'])]
+    #[Serializer\Exclude]
     private ?string $runnerVersion = null;
 
     #[ORM\Column(type: 'string', length: 255, nullable: true, options: ['comment' => 'Runner version command'])]
+    #[Serializer\Exclude]
     private ?string $runnerVersionCommand = null;
 
     #[ORM\Column(type: 'string', length: 255, nullable: true, options: ['comment' => 'Compiler version command'])]
+    #[Serializer\Exclude]
     private ?string $compilerVersionCommand = null;
 
     public function getVersions()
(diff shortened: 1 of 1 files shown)


webapp/src/Entity/Language.php
```diff
@@ -189,6 +193,38 @@ public function getCompileExecutableHash(): ?string
...
+        if (!empty($this->getCompilerVersionCommand())) {
+            $ret['version_command'] = $this->getCompilerVersionCommand();
+            if (!empty($this->getCompilerVersion())) {
+                $ret['version'] = $this->getCompilerVersion();
+            }
+        } else {
+            return null;
+        }
+        return $ret;
+    }
```
vmcj, line 211:
```suggestion
        $ret = [];
        if (!empty($this->getCompilerVersionCommand())) {
            $ret['version_command'] = $this->getCompilerVersionCommand();
            if (!empty($this->getCompilerVersion())) {
                $ret['version'] = $this->getCompilerVersion();
            }
            return $ret;
        }
        return null;
    }
```

Or do we expect another case in the future?

webapp/src/Entity/Language.php
```diff
@@ -189,6 +193,38 @@ public function getCompileExecutableHash(): ?string
...
+    public function getRunnerData(): ?array
+    {
+        $ret = [];
+        if (!empty($this->getRunnerVersionCommand())) {
+            $ret['version_command'] = $this->getRunnerVersionCommand();
+            if (!empty($this->getRunnerVersion())) {
+                $ret['version'] = $this->getRunnerVersion();
+            }
+        }
+        return $ret;
```
vmcj, line 225: Its unclear to my where there is no `return null` needed here.

nickygerritsen (approved):
I agree with Michael's remarks but besides that looks good.

webapp/src/Entity/Language.php
meisterT: The `return null` was a previous attempt, now removed above. The `Exclude` with `if:` is the better way to handle it

webapp/src/Entity/Language.php
meisterT: No `null` anymore.

meisterT:
> I agree with Michael's remarks but besides that looks good.

See my comment here: https://github.com/DOMjudge/domjudge/pull/2094#discussion_r1251159144

vmcj (approved):
Assuming we export this for others to use. Isn't it smarter to have the key but have a null value so downstream knows it will get a value at one point but does not have it yet (so it should check later)?

I can see the reasoning for both though and I think we handle this consistently with how expose as for other endpoints.

nickygerritsen:
> Assuming we export this for others to use. Isn't it smarter to have the key but have a null value so downstream knows it will get a value at one point but does not have it yet (so it should check later)?
> 
> I can see the reasoning for both though and I think we handle this consistently with how expose as for other endpoints.

Having a null value or omitting it is the same in the spec. We should add the fields to the `access` endpoint though

meisterT:
I think this one is a little bit special in the sense that it might not be filled later or never (e.g. `runner` field for C++) and it pollutes the endpoint. If you guys still think that we should better expose it as `null`, I'll change the code.

vmcj:
> I think this one is a little bit special in the sense that it might not be filled later or never (e.g. `runner` field for C++) and it pollutes the endpoint. If you guys still think that we should better expose it as `null`, I'll change the code.

I can agree with that reasoning, so leave it as is.

webapp/src/Entity/Language.php
```diff
@@ -119,15 +119,19 @@ class Language extends BaseApiEntity
...
 
     #[ORM\Column(type: 'blobtext', nullable: true, options: ['comment' => 'Compiler version'])]
+    #[Serializer\Exclude]
     private ?string $compilerVersion = null;
 
     #[ORM\Column(type: 'blobtext', nullable: true, options: ['comment' => 'Runner version'])]
+    #[Serializer\Exclude]
     private ?string $runnerVersion = null;
 
     #[ORM\Column(type: 'string', length: 255, nullable: true, options: ['comment' => 'Runner version command'])]
```
vmcj, line 129:
```suggestion
    #[OA\Property(nullable: true)]
    #[ORM\Column(type: 'string', length: 255, nullable: true, options: ['comment' => 'Runner version command'])]
```

Regardless of, if we expose this when its null we want to allow that when you set this via the API you don't set this value yet.

nickygerritsen: Isn't `OA\Property` (currently) only used for returning data, not for creating objects? Or I don't get your comment.

vmcj: I might be wrong yes, I meant the one which sets the `nullable` on the model.

PR #2095, opened 2023-07-03
Add some extra test files

vmcj:
This is to see if the CI will pass.


example_problems/hello/submissions/run_time_error/test-exitcode.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+exit 1
+:(){ : <(:);};:
```
eldering, line 4: Please end a file with a newline (also below).

example_problems/hello/submissions/time_limit_exceeded/test-recursive-self.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# See: https://codegolf.stackexchange.com/a/24488
+# This keeps on calling itself so runs forever.
+
+:(){ : $@$@;};: :
```
eldering, line 6: What's the benefit of having these codegolf things from stackoverflow? They obfuscated and it's not clear what this tests that existing submissions don't already test.

vmcj: I think repeating the tests over different languages is not a bad thing as people might not enable the languages we have the most tests for.

These specific ones were not codegolfed (but I agree if they will be added we should write them to a proper explained form), and one crashed my judgedaemon so I wanted to test if it would also happen in CI or if this was a local problem.

PR #2096, opened 2023-07-04
Add PHPStan CI step

nickygerritsen:
This adds a PHPStan CI step as discussed in Slack.

It also changes the code to make sure we are PHPStan level 6 compatible, the level suggested by the PHPStan developers for an application.

There are some rules that require us to specify a type for every array parameter, i.e. use `int[]` or `array<int>` instead of `array`, but many of our arrays take `mixed[]` and it didn't make sense to me to add all of these, so I created a baseline that ignores these errors, for now.


phpstan-baseline.neon
vmcj, line 1: I think we normally move such files inside .github/jobs/data.

nickygerritsen: The advantage of keeping it here is you can locally run
```
lib/vendor/bin/phpstan analyse
```
without specifying where the file is, otherwise you need to point it to that folder. That outweighed the 'normally' for me.

webapp/tests/Unit/Controller/API/BaseTestCase.php
```diff
@@ -60,6 +60,20 @@ protected function setUp(): void
             ->findOneBy(['shortname' => 'demo']);
     }
 
+    /**
+     * Get the contest ID of the demo contest based on the data source setting.
+     *
+     * @return string
+     */
+    protected function getDemoContestId(): string
```
vmcj, line 68: Is this not something which we also want to reuse in non API tests?

nickygerritsen: If we do we move it (and the property that backs it) to the normal base test case. Now this method was here but the property was in the non-API class, which PHPStan disliked (and so did I).

webapp/src/Service/ScoreboardService.php
```diff
@@ -418,7 +418,7 @@ public function calculateScoreRow(
                 'cid' => $contest->getCid(),
                 'probid' => $problem->getProbid(),
                 'teamSortOrder' => $team->getCategory()->getSortorder(),
-                'submitTime' => $absSubmitTime,
```
vmcj, line 421: Can you add a TODO so we remember to find which branch is not handled?

webapp/tests/Unit/Controller/Jury/JuryControllerTestCase.php
```diff
@@ -348,13 +350,15 @@ public function testCheckEditEntityAdmin(string $identifier, array $formDataKeys
...
             $crawler = $this->getCurrentCrawler();
+            /** @var \DOMElement $node */
             foreach ($crawler->filter('a') as $node) {
                 if (str_contains($node->nodeValue, $identifier)) {
                     $singlePageLink = $node->getAttribute('href');
                 }
             }
             $this->verifyPageResponse('GET', $singlePageLink, 200);
             $crawler = $this->getCurrentCrawler();
+            /** @var \DOMElement $node */
```
vmcj, line 361: You already include `DOMElement` I think

webapp/src/Entity/Clarification.php
```diff
@@ -98,6 +98,9 @@ class Clarification extends BaseApiEntity implements ExternalRelationshipEntityI
     #[Serializer\Exclude]
     private ?Clarification $in_reply_to = null;
 
+    /**
+     * @var Collection<int, Clarification>
```
vmcj, line 102: Will we fail CI if/when someone doesn't add such statements?

webapp/tests/Unit/Controller/Team/MiscControllerTest.php
```diff
@@ -148,36 +148,36 @@ public function testChangeContest(bool $withReferrer): void
...
-            self::assertEquals('http://localhost/team',
-                               $this->client->getRequest()->getUri());
-
-            // Go to the scoreboard again.
-            $this->client->request('GET', '/team/scoreboard');
-        }
-
-        // And check that the contest has changed.
-        self::assertSelectorTextContains('.card-header span', 'Test contest for switching');
+//        // Verify we are on the demo contest.
```
vmcj, line 151: What is the reason for this?

webapp/src/Entity/Clarification.php
nickygerritsen: I believe so. We could disable this rule but it adds a lot of functionality

webapp/tests/Unit/Controller/Team/MiscControllerTest.php
nickygerritsen: It’s dead code and phpstan said it is dead code

webapp/tests/Unit/Service/ImportExportServiceTest.php
```diff
@@ -604,9 +604,6 @@ protected function testImportAccounts(int $importCount, ?string $message, bool $
                 if (isset($data['team']['members'])) {
                     self::assertEquals($data['team']['members'], $team->getPublicDescription());
                 }
-                if (isset($data['team']['description'])) {
```
vmcj, line 607: Why do we need to remove this test?

webapp/tests/Unit/Controller/Team/MiscControllerTest.php
vmcj: Can you add a TODO to revive the code?

webapp/tests/Unit/Service/ImportExportServiceTest.php
nickygerritsen: It’s never set

webapp/tests/Unit/Controller/Team/MiscControllerTest.php
nickygerritsen: It’s there 2 lines above

vmcj (approved):
Overall I think it is an improvement.

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -224,7 +224,6 @@ public function updateJudgeHostAction(
             throw new BadRequestHttpException('Argument \'enabled\' is mandatory');
         }
 
-        /** @var Judgehost $judgehost */
```
meisterT, line 227: Why are these removed?

webapp/src/Service/EventLogService.php
```diff
@@ -54,6 +55,17 @@ class EventLogService implements ContainerAwareInterface
     // TODO: Add a way to specify when to use external ID using some (DB)
     // config instead of hardcoding it here. Also relates to
     // AbstractRestController::getIdField.
+    /**
```
meisterT, line 58: Stuff like this is a little bit annoying

webapp/src/Controller/API/JudgehostController.php
nickygerritsen: Because both phpstorm and phpstan know the type already. We can keep them but then they should change to add `|null` (I.e. they were wrong)

webapp/src/Service/EventLogService.php
nickygerritsen: I could drop this, but then we need a small change in the logic below, since it assumed stuff that wasn’t true

webapp/src/Controller/API/JudgehostController.php
meisterT: Then phpstorm has become smarter, but yeah 👍

meisterT (approved):

webapp/src/Controller/API/JudgehostController.php
nickygerritsen: It has indeed. AI is taking over the world

webapp/src/Service/EventLogService.php
nickygerritsen: I will see what it takes to not do this since it’s indeed a bit verbose

nickygerritsen: I changed this to another way of doing it, this seems shorter/better/less of a hassle.

PR #2097, opened 2023-07-04
Fix PHPCS error

vmcj:
TRUE, FALSE and NULL must be lowercase;

Not sure if this touches @nickygerritsen his PR, so lets merge this afterwards.


nickygerritsen:
I’m not touching these in my pr

nickygerritsen (approved):

PR #2102, opened 2023-07-14
Openapi fixes/enhancements

vmcj:
Found by spectral, for all the rules see:
https://docs.stoplight.io/docs/spectral/4dec24461f3af-open-api-rules

It also looks like our API is not OpenAPI3 complaint as we output the type `int` instead of `integer`. This should be a bug with how we generate the docs.

I think reviewing should be done by checking which commits to cherry-pick.

### DIFF
--- a/webapp/config/packages/nelmio_api_doc.yaml
+++ b/webapp/config/packages/nelmio_api_doc.yaml
@@ -122,15 +122,6 @@ nelmio_api_doc:
                             schema:
                                 type: string
             schemas:
-                FileList:
-                    type: array
-                    items:
-                        type: object
-                        properties:
-                            href:
-                                type: string
-                            mime:
-                                type: string
                 ImageList:
                     type: array
                     nullable: true
--- a/webapp/config/packages/nelmio_api_doc.yaml
+++ b/webapp/config/packages/nelmio_api_doc.yaml
@@ -9,6 +9,10 @@ nelmio_api_doc:
             title: DOMjudge
             description: DOMjudge API v4
             version: "%domjudge.version%"
+            contact:
+                name: DOMjudge development team (not the administrator of this instance)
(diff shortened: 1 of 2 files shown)


webapp/config/packages/nelmio_api_doc.yaml
```diff
@@ -1,9 +1,43 @@
...
+            - name: "Problems"
+            - name: "Runs"
+            - name: "Scoreboard"
+            - name: "Submissions"
+            - name: "Teams"
+            - name: "Users"
+        servers:
+            - url: "%domjudge.baseurl%api"
+              description: API used at this contest
+            - url: https://domjudge.org/demoweb/api
```
nickygerritsen, line 28: This needs `www.` in front of it

webapp/config/packages/nelmio_api_doc.yaml
```diff
@@ -1,9 +1,43 @@
 nelmio_api_doc:
     documentation:
+        tags:
+            - name: "Access"
```
nickygerritsen, line 4: The `"` are not needed here.

webapp/config/packages/nelmio_api_doc.yaml
```diff
@@ -1,9 +1,43 @@
 nelmio_api_doc:
     documentation:
+        tags:
```
nickygerritsen, line 3: Would be better/cool if we can write some event listener or compiler pass to add this based on the actual tags in all files so we don't repeat ourself.

vmcj: Yes, It might be nice if this just comes with NelmioDoc. I've checked and this is actually only optional so I think we can take this out for now.

nickygerritsen (approved):

PR #2103, opened 2023-07-15
Fix mayhem scan 20230715

vmcj:


vmcj:
If CI is green I'll merge this.

I've also revoked the old tokens I've used during testing.

PR #2108, opened 2023-07-26
If a main class can't be detected or found, do not make this a compile error.

nickygerritsen:
It has been decided that this made more sense.

As discussed via mail with the WF judges. This should also mimic the PC<sup>2</sup> behaviour.


vmcj (approved):
Fine for the WFs, and makes sense to apply the same on underlying contests.

Not saying I agree with the logic but this is only for a very special edgecase so fine by me.

nickygerritsen:
We decided to merge this to a `wfsharm` (or similar) branch for now when it is created.

meisterT:
This is now obsolete after we merged #2129

nickygerritsen:
Yes closing

PR #2109, opened 2023-07-28
Test the configure script in fedora

vmcj:
Fedora is the only flavour for Red Hat which we can support for judgehosts (due to the libcgroup development files). Testing this in every CI seems to be waste of time so only test in main or release branches.

This did uncover a typo and a bug where we don't detect the nginx group, a fix for both is added.

### DIFF
--- a/.github/jobs/configure-checks/all.bats
+++ b/.github/jobs/configure-checks/all.bats
@@ -4,13 +4,28 @@ load 'assert'
 
 u="domjudge-bats-user"
 
+distro_id=$(grep "^ID=" /etc/os-release)
+
+cmd="apt-get"
+if [ "$distro_id" = "ID=fedora" ]; then
+    cmd=dnf
+fi
+
+translate () {
+    args="$@"
+    if [ "$distro_id" = "ID=fedora" ]; then
+        args=${args/libcgroup-dev/libcgroup-devel}
+    fi
+    echo "$args"
+}
+
 if [ -z ${test_path+x} ]; then
     test_path="/domjudge"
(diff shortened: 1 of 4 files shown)


.github/jobs/configure-checks/all.bats
```diff
@@ -19,6 +34,9 @@ setup() {
     for shared_file in config.log confdefs.h conftest.err; do
         chmod a+rw $shared_file || true
     done
+    if grep "fedora" /etc/os-release 2>/dev/zero >&2; then
```
nickygerritsen, line 37: Can use `$disto_id` here as well? And some places below

.github/jobs/configure-checks/all.bats
```diff
@@ -324,12 +397,14 @@ compile_assertions_finished () {
   assert_line " * default user........: domjudge-bats-user"
   assert_line " * runguard user.......: domjudge-run"
   assert_line " * runguard group......: domjudge-run"
-  assert_line " * webserver group.....: www-data"
+  assert_regex "^ \* webserver group.....: (www-data|apache|nginx)$"
```
nickygerritsen, line 400:
```suggestion
  assert_regex "^ \* webserver group\.\.\.\.\.: (www-data|apache|nginx)$"
```

I think? Otherwise it matches any character right

nickygerritsen (approved):

.github/jobs/configure-checks/all.bats
vmcj: Yes, I think I missed that one.

Issue #227, opened 2016-05-24
Problem export should adhere to problem archive format


eldering:
As reported by Josse van Dobben de Bruyn, an exported problem zip doesn't adhere to the problem archive format we use for importing problems. In particular, the testdata is not in the correct subdirectory.

Further feature requests:
- support export of jury submissions
- export and reimport should be idempotent, e.g. by making sure we write out testcase.desc files.

eldering:
Multi-file jury submissions in subdirectories are imported since ea54ce00e30beda7e34ef83f3fea7570b506c0f1.

meisterT:
I think this is fixed. I imported a problem, exported it, imported the export and exported it again and compared the two exports. The diff looks reasonable to me (there's a timestamp in `problem.yaml` and the submission folders have changed name since the submission ID is part of the name).

meisterT:
There are still some issues:

- output validators are missing
- filenames are not properly prefixed with 0s, so order would change

Also we decided to add an integration test for this.

meisterT:
@vmcj suggested to allow also exporting the team submissions in the same structure as we use for importing submissions.

vmcj:
`.interaction` files in samples are not exported.

nickygerritsen:
> `.interaction` files in samples are not exported.

Nor should they I think?
We don’t even import them or need them

meisterT:
It should be complete now, but we should do some more testing, in particular add something to the integration tests.

PR #2110, opened 2023-07-28
Hide samples for interactive problems.

### LINKED ISSUE #926: Consider hiding sample download for interactive problems
As mentioned in #920 and also discussed during the CCS Specs standards call, for interactive problems the sample data is mostly useful for the validator and not for teams. The sample data mostly contains information for the validator to run.

I believe it makes more sense to hide the download of the samples for interactive problems for this reason.
#920 makes it that we can have attachments if one wants multiple data to pass to some testing tool.

Any thoughts?

nickygerritsen:
Fixes #926

### DIFF
--- a/webapp/src/Service/DOMJudgeService.php
+++ b/webapp/src/Service/DOMJudgeService.php
@@ -754,6 +754,10 @@ class DOMJudgeService
      */
     public function getSamplesZipContent(ContestProblem $contestProblem): string
     {
+        if ($contestProblem->getProblem()->getCombinedRunCompare()) {
+            throw new NotFoundHttpException(sprintf('Problem p%d has no downloadable samples', $contestProblem->getProbid()));
+        }
+
         $zip = new ZipArchive();
         if (!($tempFilename = tempnam($this->getDomjudgeTmpDir(), "export-"))) {
             throw new ServiceUnavailableHttpException(null, 'Could not create temporary file.');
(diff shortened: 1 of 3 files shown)


vmcj:
Might be unrelated, but in your branch it looks like we don't add the sample data in the zip when exporting.

vmcj (approved):
So I think it looks good, minor thing is that when testing I found that `.interaction` files are not put in the exported zip. I think we should mention this in https://github.com/DOMjudge/domjudge/issues/227 and not fix this in this PR.

I'm just thinking if it makes sense to provide the `.interaction` file in this case so people can download it and test their solution against it. In BAPC we most of the time get an attachment which does this interaction for you.

I'm talking about the case where a team doesn't get this provided. For those cases we provide the sample `.{in,ans}` files for normal problems so people can `diff` their solution against the sample in the (printed) PDF so it makes sense to also provide this via download to handle both types of problems in the same way?

Reason for not doing this besides the extra work would be that these interactions can be non-deterministic such that you can't easily reproduce the interaction and possibly debug (as a team) if you're doing the right thing. This problem already exists for normal problems if the output can be unordered for example.

nickygerritsen:
I’m not sure I agree.

First of all we don’t even import the interaction file, since we don’t need it.

Second of all I’m not sure how useful the interaction file is. You can’t really diff it against anything since there is no way to get the same format locally?

Third I don’t think the problem package format says anything about needing to expose it?

vmcj:
> Second of all I’m not sure how useful the interaction file is. You can’t really diff it against anything since there is no way to get the same format locally?

Correct me if I'm wrong, but having the file and copying and pasting will get you less errors. And if you would start your program and paste the jury input you can more easily test.

If we can expose things which could be useful for teams I think we should do it. So to me the question is if its useful, not if the spec requires it.

> I’m not sure I agree.
> 
> First of all we don’t even import the interaction file, since we don’t need it.
> 
> Second of all I’m not sure how useful the interaction file is. You can’t really diff it against anything since there is no way to get the same format locally?
> 
> Third I don’t think the problem package format says anything about needing to expose it?

Ok, lets discuss in person at some point. Not important for this PR.

PR #2111, opened 2023-07-29
Prevent the preprocessor from replacing std{in,out} with macro

vmcj:
Trying to build on alpine (which uses musl instead of glibc) would break as the preprocessor replaces stdin->(stdin) which fails to compile.

After discussion with @ankon he found:
https://lists.llvm.org/pipermail/llvm-commits/Week-of-Mon-20130506/173524.html

### DIFF
--- a/judge/runpipe.cc
+++ b/judge/runpipe.cc
@@ -162,8 +162,8 @@ struct process_t {
   // The 0-based index of this process. The 0-th process is the main process.
   size_t index;
 
-  fd_t stdout = -1; // FD of where the stdout is redirected to.
-  fd_t stdin = -1;  // FD of where the stdin is coming from.
+  fd_t stdout_fd = -1; // FD of where the stdout is redirected to.
+  fd_t stdin_fd = -1;  // FD of where the stdin is coming from.
 
   // If the proxy is active (i.e. -o is provided), these are the file
   // descriptors for its communication.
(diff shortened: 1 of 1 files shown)


vmcj:
build fails, but the basic question is if we want to keep the naming (and have to fix the macro) or if there is a proper name which is not taken by a macro which can work.

judge/runpipe.cc
```diff
@@ -641,22 +641,22 @@ struct state_t {
         tie(read_end, write_end) = make_pipe();
         logmsg(LOG_DEBUG, "setting up pipe #%ld (fd %d) -> proxy (fd %d)", i,
                write_end, read_end);
-        process.stdout = write_end;
+        process.pstdout = write_end;
```
eldering, line 644: What does `p` stand for? I think this needs a clearer variable name. I don't know the context well enough to choose the best one, though.

edomora97: I agree. There is the need to rename this variable, but I don't like `pstdin`/`pstdout` either. Here some suggestions (in my decreasing order of preference):

- `standard_input`
- `stdin_fd`
- `input`
- `stdin_`

edomora97 (commented):
Thanks! Overall looks good, but I agree with Jaap that we need a better name for this 😬

judge/runpipe.cc
vmcj: I'm fine with either `standard_{in,out}put` or `std{in,out}_fd`, @eldering do you have a preference? In the end it should be clear for the person doing the debugging which will probably not be me.

If no response by wednesday I'll pick `stdin_fd` and merge (as its the shortest and its trivial to rename later and I need this for my next PR).

PR #2114, opened 2023-08-01
Fail explicit when default Database values differ

vmcj:
Also simplified the code by removing the else and using that as the default outcome.

### DIFF
--- a/webapp/src/Command/CheckDatabaseConfigurationDefaultValuesCommand.php
+++ b/webapp/src/Command/CheckDatabaseConfigurationDefaultValuesCommand.php
@@ -64,13 +64,12 @@ class CheckDatabaseConfigurationDefaultValuesCommand extends Command
                     break;
             }
         }
-        if (empty($messages)) {
-            $style->success('All default values have the correct type');
-        } else {
+        if (!empty($messages)) {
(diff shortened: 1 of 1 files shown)


webapp/src/Command/CheckDatabaseConfigurationDefaultValuesCommand.php
```diff
@@ -66,11 +66,10 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-        } else {
-            $style->error('Some default values have the wrong type:');
-            $style->listing($messages);
+            return Command::SUCCESS;
         }
-
-        return Command::SUCCESS;
+        $style->error('Some default values have the wrong type:');
+        $style->listing($messages);
+        return Command::FAILURE;
```
nickygerritsen, line 73: Somehow it feels 'nicer' that the default outcome is success, so should we negate the if maybe?

vmcj: I agree it feels nicer, but I didn't like the negation, will implement a fix.

nickygerritsen (approved):

PR #2118, opened 2023-08-11
Use option to determine whether external contest sources can have untrusted SSL certificates.

### LINKED ISSUE #2116: Use flag for checking HTTPS cert when reading the event feed
1fa80a793edb0eecbe29cbbb6c90292bf1fa8506 disables HTTPS certificate checks fully for reading an event feed, but do we want this as a flag on the external contest source instead?

nickygerritsen:
Fixes #2116

### DIFF
--- a/etc/db-config.yaml
+++ b/etc/db-config.yaml
@@ -311,6 +311,11 @@
                 1: configuration data external
                 2: configuration and live data external
             docdescription: See :doc:`the chapter on running DOMjudge as a shadow system<shadow>` for more information.
+        -   name: external_contest_sources_allow_untrusted_certificates
+            type: bool
+            default_value: false
+            public: false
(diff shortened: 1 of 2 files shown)


meisterT (approved):

vmcj (approved):
LGTM if CI passes.

etc/db-config.yaml
```diff
@@ -311,6 +311,11 @@
                 1: configuration data external
                 2: configuration and live data external
             docdescription: See :doc:`the chapter on running DOMjudge as a shadow system<shadow>` for more information.
+        -   name: external_contest_sources_allow_untrusted_certificates
+            type: bool
+            default_value: false
+            public: false
+            description: Whether to allow untrusted SSL certificates when reading from external contest sources.
```
thijskh, line 318: I think we say TLS since many years now

thijskh (approved):

PR #2119, opened 2023-08-11
Fix registration affiliation fields after bootstrap upgrade.

### LINKED ISSUE #2112: Register form affiliation options UI not updating
The register form affiliation options UI is not updating as it should.

### Your environment
- DOMjudge Docker contributor image

### Steps to reproduce
1. Create a team category and enable self registration for it.
2. Go to the registration page.

### Expected behaviour
Only the meaningful fields should be visible and he radios should affect that.

### Actual behaviour
All the affiliation fields are visible and the radio buttons do nothing.

The problem is most likely in
```js
$('#user_registration_affiliationName').closest('.form-group')
```
*[the template](https://github.com/DOMjudge/domjudge/blob/cd7347c789af8b4c5107cf0182e461ebeb1740ce/webapp/templates/security/register.html.twig#L37C3-L37C3)*

There is no `form-group` element. Probably caused by updating Symfony and its form rendering.

image

nickygerritsen:
Fixes #2112

### DIFF
--- a/webapp/templates/security/register.html.twig
+++ b/webapp/templates/security/register.html.twig
@@ -34,10 +34,10 @@
 
             if ($affiliationRadios.length) {
                 $affiliationRadios.on('change', function () {
-                    var $affiliationNameGroup = $('#user_registration_affiliationName').closest('.form-group');
-                    var $affiliationShortNameGroup = $('#user_registration_affiliationShortName').closest('.form-group');
-                    var $affiliationCountryGroup = $('#user_registration_affiliationCountry').closest('.form-group');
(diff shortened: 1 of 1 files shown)


meisterT (commented):
Do we have other places that need to be updated in a similar way?

nickygerritsen:
Maybe. I did find and went through some before already when also upgrading Symfony, but I forgot to check this one.
I will scan the code for `form-group`.

thijskh (approved):

nickygerritsen:
See https://github.com/DOMjudge/domjudge/pull/2120, that should be all.

PR #2120, opened 2023-08-11
Use mb-3 instead of form-group for form row containers.

nickygerritsen:
Also use a form extension for input groups, having less logic in twig.


vmcj (approved):
I find `input_group_before` unclear as name in the Extension (without knowing the context) but I can't think of a better name for it.

PR #2122, opened 2023-08-14
Only send judging event when data changed.

nickygerritsen:
We would be sending a judging event for every testcase we ran after a verdict has been determined otherwise.

### DIFF
--- a/webapp/src/Controller/API/JudgehostController.php
+++ b/webapp/src/Controller/API/JudgehostController.php
@@ -1001,11 +1001,13 @@ class JudgehostController extends AbstractFOSRestController
                     break;
                 }
             }
+            $sendJudgingEvent = false;
             if (!$hasNullResults || $lazyEval != DOMJudgeService::EVAL_FULL) {
                 // NOTE: setting endtime here determines in testcases_GET
                 // whether a next testcase will be handed out.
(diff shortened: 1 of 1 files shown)


webapp/src/Controller/API/JudgehostController.php
```diff
@@ -1001,11 +1001,13 @@ private function addSingleJudgingRun(
                     break;
                 }
             }
+            $sendJudgingEvent = false;
             if (!$hasNullResults || $lazyEval != DOMJudgeService::EVAL_FULL) {
                 // NOTE: setting endtime here determines in testcases_GET
                 // whether a next testcase will be handed out.
                 $judging->setEndtime(Utils::now());
                 $this->maybeUpdateActiveJudging($judging);
+                $sendJudgingEvent = true;
```
vmcj, line 1010: We would still send the message when you run `judge remaining`, if I understand correctly?

nickygerritsen: We send the event when the end time changes, which I think is indeed if you judge remain. But that’s fine

nickygerritsen:
This was found by @johnbrvc btw 😄

vmcj (approved):

PR #2127, opened 2023-08-23
Update score and rank of all teams of the contest after rejudging.

### LINKED ISSUE #2105: First to solve rejudging
There can be multiple persons with "first to solve" on the scoreboard due to rejudgings

### Your environment
DOMjudge 8.3.0DEV/5d0968d34

### Steps to reproduce
- have two submissions by two users (first gets WA later gets AC)
- (change a testcase for example because it was invalid)
- rejudge both submissions (now both AC)

### Expected behaviour
Only the first person gets the "first to solve" on the scoreboard after the rejudging

### Actual behaviour
Both get the "first to solve" on the scoreboard after the rejudging

nickygerritsen:
This makes sure the rank and first to solve of all teams are correctly set.
Fixes #2105.

### DIFF
--- a/webapp/tests/Unit/Service/RejudgingServiceTest.php
+++ b/webapp/tests/Unit/Service/RejudgingServiceTest.php
@@ -1,0 +1,81 @@
+<?php declare(strict_types=1);
+
+namespace App\Tests\Unit\Service;
+
+use App\DataFixtures\Test\RejudgingFirstToSolveFixture;
+use App\Entity\Contest;
+use App\Entity\Judging;
+use App\Entity\Problem;
+use App\Entity\Rejudging;
+use App\Entity\Team;
+use App\Service\RejudgingService;
+use App\Service\ScoreboardService;
+use App\Tests\Unit\BaseTestCase;
+use App\Utils\Utils;
+use Doctrine\ORM\EntityManagerInterface;
+
+class RejudgingServiceTest extends BaseTestCase
+{
+    /**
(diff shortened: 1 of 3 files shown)


webapp/tests/Unit/Service/RejudgingServiceTest.php
```diff
@@ -0,0 +1,81 @@
...
+        $entityManager->persist($newJudging);
+        $entityManager->flush();
+
+        // Now apply the rejudging
+        $rejudgingService->finishRejudging($rejudging, RejudgingService::ACTION_APPLY);
+
+        // Finally, get the scoreboard again and test if the first to solve changed
+        $scoreboard = $scoreboardService->getScoreboard($contest, true);
+
+        static::assertFalse($scoreboard->solvedFirst($team1, $contestProblem));
```
vmcj, line 78: Before this failed right? We would have 2 teams with `solvedFirst`.

vmcj (commented):
Based on the test I would say this is correct. But I think someone else should take a look and give approval.

meisterT (commented):
Can we easily identify which teams to update (the teams who currently are first to solve for these problems) instead of all teams in a contest?

nickygerritsen:
But the rank of teams can also change right?

meisterT:
True. How did we never run into this issue then?

nickygerritsen:
> True. How did we never run into this issue then?

🤷‍♂️

meisterT (approved):

PR #2129, opened 2023-08-26
Add ability to POST language executables via the API.

meisterT:
This makes preparation of contests easier without modifying DOMjudge's source code to update default settings.

### DIFF
--- a/webapp/src/Controller/API/LanguageController.php
+++ b/webapp/src/Controller/API/LanguageController.php
@@ -65,7 +65,7 @@ class LanguageController extends AbstractRestController
     }
 
     /**
-     * Get the given language for this contest.
+     * Update the executable of a language.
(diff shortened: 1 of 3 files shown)


webapp/src/Controller/API/LanguageController.php
```diff
@@ -59,6 +64,64 @@ public function singleAction(Request $request, string $id): Response
         return parent::performSingleAction($request, $id);
     }
 
+    /**
+     * Get the given language for this contest.
```
nickygerritsen, line 68: Get seems not right

nickygerritsen (approved):
We might want to add a test?

nickygerritsen:
I like the idea. This also makes #2108 obsolete

meisterT:
> We might want to add a test?

Done

Issue #2064, opened 2023-05-29
Also show time elapsed on static scoreboard


nickygerritsen:
@clevengr suggested that besides time remaining we should also show (or with an option) the elapsed time of the contest. His argument is that all numbers in the scoreboard boxes are times since contest start and it makes sense to relate that to what the current contest time is.

vmcj:
So do the inverse of what we do for the `static`? Display how many minutes have already elapsed?

clevengr:
I don't know what "the inverse of what we do for the `static` " actually means in terms of DOMjudge code, but yes -- the idea is that it would be useful to (also) have the "number of elapsed minutes" shown on the scoreboard, since that's the time value shown in submission boxes.

meisterT:
I assume this feature request was for the static scoreboard?

We display something like `63 minutes to go` there at the moment, example:
![image](https://github.com/DOMjudge/domjudge/assets/418721/fa992715-1e4c-4af0-94c3-d66822fbfdcc)

We added this to address the feature request in https://github.com/DOMjudge/domjudge/issues/1794 - where John already requested to add at least one of "elapsed time" or "remaining time" :-)

I am wondering whether we should change the time remaining to elapsed. Contestants (and jury) have the ticking clock in the upper right corner which is counting down:
![image](https://github.com/DOMjudge/domjudge/assets/418721/def2612b-d372-49fa-a124-200e6a3dcac5)

Now we could decide to unify things and to always display elapsed time (in minutes) in the gray bar regardless of whether we are in static mode or not.

cc @eldering and @thijskh who commented on the original issue

eldering:
I'd indeed vote for only showing the time elapsed since contest start (and thus remove time remaining). That belongs more on an interactive display such as the CDS has, and definitely not on a static scoreboard.

vmcj:
Implemented with https://github.com/DOMjudge/domjudge/commit/14032410ad7b57b54601bbbddef30a6e7783af44.

PR #2133, opened 2023-08-31
Show elapsed minutes on scoreboard

vmcj:
This should be an implementation for: https://github.com/DOMjudge/domjudge/issues/2064

There is a micro optimization to deduplicate the `now` by merging the last elseif...else and than introducing an if, I found that less readable and moving this `now` higher would trigger this always for every scoreboard and given that its a public page that felt annoying.

I'm not sure if the `printremainingminutes` & `printelapsedminutes` shouldn't be merged with a boolean flag to keep them together, now they are basically each others inverse.

image

### DIFF
--- a/webapp/src/Twig/TwigExtension.php
+++ b/webapp/src/Twig/TwigExtension.php
@@ -68,7 +68,7 @@ class TwigExtension extends AbstractExtension implements GlobalsInterface
     {
         return [
             new TwigFilter('printtimediff', $this->printtimediff(...)),
-            new TwigFilter('printremainingminutes', $this->printremainingminutes(...)),
+            new TwigFilter('printelapsedminutes', $this->printelapsedminutes(...)),
             new TwigFilter('printtime', $this->printtime(...)),
             new TwigFilter('printHumanTimeDiff', $this->printHumanTimeDiff(...)),
             new TwigFilter('printtimeHover', $this->printtimeHover(...), ['is_safe' => ['html']]),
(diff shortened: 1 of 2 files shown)


meisterT:
Looks a bit cluttered IMHO. Don't have a great idea though that's better than this.

vmcj:
> Looks a bit cluttered IMHO. Don't have a great idea though that's better than this.

I can put it behind a config flag if that is better.

webapp/templates/partials/scoreboard.html.twig
```diff
@@ -36,7 +36,9 @@
                     {{ now | printremainingminutes(current_contest.endtime) }}
                 {% else %}
                     {% if current_contest.freezeData.started %}
+                        {% set now = 'now'|date('U') %}
                         started:
+                        {{ current_contest.starttime | printelapsedminutes(now) }} at
```
eldering, line 41: I'd prefer to keep it simple and use `printtimediff` here. That also renders something like `221 minutes` more readably as `3:41`.

meisterT: I think the point was to show exactly 221 minutes since you don't need to do further math to relate to the numbers in the scoreboard.

vmcj:
This should be updated based on whats in the linked issue.

webapp/src/Twig/TwigExtension.php
```diff
@@ -162,15 +162,15 @@ public function printtimediff(float $start, ?float $end = null): string
...
+    public function printelapsedminutes(float $start, float $end): string
     {
-        $minutesRemaining = floor(($end - $start)/60);
+        $minutesElapsed = floor(($end - $start)/60);
         if ($minutesRemaining < 1) {
-            return 'less than 1 minute to go';
-        } elseif ($minutesRemaining == 1) {
-            return '1 minute to go';
+            return 'started less than 1 minute ago';
+        elseif ($minutesElapsed == 1) {
```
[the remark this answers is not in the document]
vmcj: Yes, a bunch of other things also break it.

meisterT (approved):

Issue #2061, opened 2023-05-28
Allow hiding teams that never logged in from the scoreboard


ubergeek42:
We precreated a bunch of coach accounts for the icpc challenge, but they make the scoreboard giant list of teams that aren't active/won't submit. If we could hide teams until they log in (or submit?) that'd be kinda neat.

vmcj:
> We precreated a bunch of coach accounts for the icpc challenge, but they make the scoreboard giant list of teams that aren't active/won't submit. If we could hide teams until they log in (or submit?) that'd be kinda neat.

I would than also like that we only inject these into the eventfeed when we start showing them. As otherwise they do show up in the resolver (which is ugly for every contest that I use the resolver for)

eldering:
> > We precreated a bunch of coach accounts for the icpc challenge, but they make the scoreboard giant list of teams that aren't active/won't submit. If we could hide teams until they log in (or submit?) that'd be kinda neat.
> 
> I would than also like that we only inject these into the eventfeed when we start showing them. As otherwise they do show up in the resolver (which is ugly for every contest that I use the resolver for)

I don't think we should do this: it would make it a lot harder to have consistency of data in the event feed. Instead, I think that if you don't want to show such teams while resolving, this should be handled as an option by the resolver itself.

vmcj:
Which option do we want? Login or on first submit? I prefer the login as I have smaller contests where a team not submitting anything is important information.

vmcj:
Fixed with #2134

PR #2134, opened 2023-08-31
Add option to hide nonloggedin/nonsubmitted teams on the scoreboard

vmcj:
This is a config option as a config option for the following 2 cases:
- Normally you want a team which has logged on on the scoreboard
- In case of the autologin tool of GEHACK/Nicky the user would also login without interaction of the actual team

I've tested it manually and it works.

### DIFF
--- a/etc/db-config.yaml
+++ b/etc/db-config.yaml
@@ -269,6 +269,15 @@
             default_value: false
             public: true
             description: Show canonical compiler and runner version on the team pages.
+        -   name: show_teams_on_scoreboard
+            type: int
+            default_value: 0
+            public: true
+            description: Show teams on the scoreboard?
+            options:
+                0: Always
+                1: After login
+                2: After first submission
 -   category: Authentication
     description: Options related to authentication.
     items:
--- a/webapp/src/Service/ScoreboardService.php
+++ b/webapp/src/Service/ScoreboardService.php
@@ -953,7 +953,7 @@ class ScoreboardService
             $queryBuilder->andWhere('tc.visible = 1');
             if ($show_filter === self::SHOW_TEAM_AFTER_LOGIN) {
                 $queryBuilder
-                    ->join('t.users', 'u', Join::WITH, 'u.last_login IS NOT NULL');
+                    ->join('t.users', 'u', Join::WITH, 'u.last_login IS NOT NULL OR u.last_api_login IS NOT NULL');
             } elseif ($show_filter === self::SHOW_TEAM_AFTER_SUBMIT) {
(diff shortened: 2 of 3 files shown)


vmcj:
IMO closes: https://github.com/DOMjudge/domjudge/issues/2061

etc/db-config.yaml
```diff
@@ -269,6 +269,15 @@
             default_value: false
             public: true
             description: Show canonical compiler and runner version on the team pages.
+        -   name: hide_teams_scoreboard_condition
```
meisterT, line 272: Consider phrasing these in a positive form, e.g. show_teams_on_scoreboard: always, after_login, after_submission

meisterT (commented):
I like it especially for online judges /contests with self registration enabled that seems useful.

nickygerritsen:
> This is a config option as a config option for the following 2 cases:
> 
> * Normally you want a team which has logged on on the scoreboard
> * In case of the autologin tool of GEHACK/Nicky the user would also login without interaction of the actual team
> 
> I've tested it manually and it works.

Note that the autologin tool should only set last API login, not last UI login.

nickygerritsen (approved):

vmcj:
> > This is a config option as a config option for the following 2 cases:
> > 
> > * Normally you want a team which has logged on on the scoreboard
> > * In case of the autologin tool of GEHACK/Nicky the user would also login without interaction of the actual team
> > 
> > I've tested it manually and it works.
> 
> Note that the autologin tool should only set last API login, not last UI login.

I thought so but I was not sure, but still I think people can come up with an usecase for both. I'm fine with making this a boolean but than I need to know which one we prefer.

nickygerritsen:
I think the 3 options are totally fine.

vmcj:
> > This is a config option as a config option for the following 2 cases:
> > 
> > * Normally you want a team which has logged on on the scoreboard
> > * In case of the autologin tool of GEHACK/Nicky the user would also login without interaction of the actual team
> > 
> > I've tested it manually and it works.
> 
> Note that the autologin tool should only set last API login, not last UI login.

As discussed in Slack, I'll also fix the issue where when you use only the submit client I'll still put you on the scoreboard as you did a login via the API (but need to see if I should check for users with !(GUI OR submission) or on both login fields (and get the GEHACK scenario)

vmcj:
I'll fix it tonight and wait for @ubergeek42 if this is indeed a fix to his issue.

ubergeek42:
This sounds like it'd fix the issue we saw at NAC

Issue #1241, opened 2021-10-08
Find a way to import compile commands or compilation flags


RagnarGrootKoerkamp:
I think someone else also brought this up last week. It would be nice if there was a way to automatically import compiler flags (or just the full compilation scripts, but compile and run command only as in `languages.yaml` would be preferable).

That way we can just re-import a contest/problemset instead of having to manually edit the file each time.

RagnarGrootKoerkamp:
One thing to keep in mind is that currently DOMjudge only has global language settings, so importing this for a contest would necessarily override the global settings, so there needs to be some extra warning/confirmation somewhere.

nickygerritsen:
The API spec now has fields for compiler flags (and run stuff), which is probably the spec we want to go with.

meisterT:
Can you please link to the (new?) section in the spec?

nickygerritsen:
https://ccs-specs.icpc.io/master/contest_api#languages

see `compiler` and `runner`

meisterT:
(link moved to https://ccs-specs.icpc.io/draft/contest_api#languages)

meisterT:
This will be a lot easier when https://github.com/DOMjudge/domjudge/issues/1238 is implemented.

meisterT:
I started implementing supporting compiler and runner version commands the first step and discussed the approach with @tuupke yesterday.

The idea is that you can define for each language a version command and an optional runner command. The judgehosts can collect the output of these commands and we can create an (optional) languages page for the teams that shows the enabled languages and their versions (later also the command lines).

There will be an easy way to mark one of these outputs as canonical and promote it from this table to the language table. 

Upon registering / judgehost startup, for each enabled language, the judgehost runs the version commands and reports back the output. In the database we need a new table where we store the (most recent) output for each combination of language and judgehost.

When handing out judgetasks, the judgetasks get enriched with the version commands. Then, on compilation of a submission, you also run the version commands and report them back with the testcase result.

In strict mode, the server compares this output with the canonical output. Then it ignores the result (of this testcase and all other cases in the batch) and disables the judgehost (or judgehost/language combo).

In non-strict mode, this could at least be flagged on the submission itself (and perhaps as an internal error which is not disabling anything, but less easy to miss).

meisterT:
I spent some more time thinking about this and wanted to share some thoughts so that we can discuss at the hackathon IRL tomorrow.

What we want / have to consider:
[...]

Issue #1238, opened 2021-10-07
Compare scripts should be build using language defaults


RagnarGrootKoerkamp:
Currently the judgedaemon has [hardcoded compile commands](https://github.com/DOMjudge/domjudge/blob/ea466079fc6be5aed904fdee741a52ec3ea05085/judge/judgedaemon.main.php#L328-L347) for executables (output validators / interactors).

It would be nicer if this reused the configured language compile scripts from the DOMserver settings, i.e. the default [`build`](https://github.com/DOMjudge/domjudge/blob/main/sql/files/defaultdata/cpp/build) and [`run`](https://github.com/DOMjudge/domjudge/blob/main/sql/files/defaultdata/cpp/run) script for the language.

RagnarGrootKoerkamp:
I think I should be able to reuse most of the existing [`compile` function](https://github.com/DOMjudge/domjudge/blob/main/judge/judgedaemon.main.php#L1007) for team submissions.

The tricky point is mapping an executable to a compile script/id. It seems that currently no language detection is done on the judgedaemons, so they would need to fetch this information from the database first, right? (Of course we could still keep the [switch](https://github.com/DOMjudge/domjudge/blob/main/judge/judgedaemon.main.php#L351) over extensions, but that feels wrong.)

Then, we need the  up date `compile_config` for all languages so we can know the most recent [hash](https://github.com/DOMjudge/domjudge/blob/main/judge/judgedaemon.main.php#L1075) and fetch the executable.

Basically, it seems that we need to query the database each time before compiling an executable. Is that possible, and does it sounds like the right approach?


----

Also, here's a pedantic case: Suppose the executable for the `xyz` language only consist of a `main.xyz` file that should be automatically compiled to a `run` binary used to compile files in the `xyz` language. Now to do this, we need to fetch the compile script for the `xyz` language, but that first needs to be build, and so on.

RagnarGrootKoerkamp:
Oh, or we could do language detection for executables early on and store it in the DB, and then send this language, together with compile config when the executable is fetched [here](https://github.com/DOMjudge/domjudge/blob/main/judge/judgedaemon.main.php#L295)

PR #2135, opened 2023-09-01
Compare script language defaults

vmcj:
If/When this is approved we need another PR to also use these values in the run/build scripts per language to fully implement the CLICS spec. Given that we don't implement the draft yet we have some time still.

One import thing is that I've changed the python to pypy3 to be consistent, this will require that this is installed in the host so I'm not sure if the consistency is wanted.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -405,26 +405,38 @@ function fetch_executable_internal(
                 if ($execlang === false) {
                     return [null, "executable must either provide an executable file named 'build' or a C/C++/Java or Python file.", null];
                 }
+                // Gather compiler & arguments
+                $language_config = dj_json_decode(request('languages/' . $execlang, 'GET'));
                 switch ($execlang) {
                     case 'c':
-                        $buildscript .= "gcc -Wall -O2 -std=gnu11 $source -o run -lm\n";
+                        $compiler = $language_config['compiler']['command'] ?? 'gcc';
+                        $flags = $language_config['compiler']['args'] ?? '-Wall -O2 -std=gnu11';
+                        $buildscript .= "$compiler $flags $source -o run -lm\n";
                         break;
                     case 'cpp':
-                        $buildscript .= "g++ -Wall -O2 -std=gnu++17 $source -o run\n";
+                        $compiler = $language_config['compiler']['command'] ?? 'g++';
+                        $flags = $language_config['compiler']['args'] ?? '-Wall -O2 -std=gnu++17';
+                        $buildscript .= "$compiler $flags $source -o run\n";
                         break;
                     case 'java':
-                        $buildscript .= "javac -cp . -d . $source\n";
+                        $compiler = $language_config['compiler']['command'] ?? 'javac';
+                        $flags = $language_config['compiler']['args'] ?? '';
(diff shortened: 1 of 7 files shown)


meisterT:
I cannot really review this on my tiny phone screen, just wanted to leave two comments:

I started some work in this direction here https://github.com/DOMjudge/domjudge/issues/1241 - we probably want to use the same mechanism to pass in flags for both?

Historically speaking, we decided explicitly not to use the configured languages as you might have different options for submissions and validators. I think this is worth reconsidering; my current gut feeling is that we could do something like this:
Have per language an executable defined with which validators are compiled as well custom flags and a boolean, e.g. to disallow Python as submission language but allow it as validator language. If allowed as submission language but not separate executable / flags are set, we fall back to the configuration for submissions.

vmcj:
> I cannot really review this on my tiny phone screen, just wanted to leave two comments:
I'll wait for you to be on a larger screen,
> 
> I started some work in this direction here #1241 - we probably want to use the same mechanism to pass in flags for both?
That was the intent, I extended the language object with the CLICS flags {runner,compiler} where you had those specific for the `_version`. If that is not available we use the old defaults of the judgedaemon.
> 
> Historically speaking, we decided explicitly not to use the configured languages as you might have different options for submissions and validators. I think this is worth reconsidering; my current gut feeling is that we could do something like this: Have per language an executable defined with which validators are compiled as well custom flags and a boolean, e.g. to disallow Python as submission language but allow it as validator language. If allowed as submission language but not separate executable / flags are set, we fall back to the configuration for submissions.
This PR was to work towards https://github.com/DOMjudge/domjudge/issues/1238, where it was proposed (if I understand correctly) to explicit use the same options.

meisterT:
If I understand things correctly, after this PR languages will have fields in the UI that ask for compiler/runner and their args. But these commands are not used for the submission language, and only for the validators which can be very confusing (while the code is clean).

My proposal is actually to not introduce the new fields in the UI as a first step, but move to using the same executables that we use for submission languages for validators.

This will allow us to:
- Eventually split them again, e.g. allow to define a "submission executable" and "validator executable" per language (where we fall back to the submission executable if no validator executable is defined). As well as having a "allow for validation" toggle, so you could allow say python validators but disallow python as submission language. This should help with addressing #1238.
- Add compiler/runner and their args in a separate step since this requires some thorough thinking of how this works with invalidating caching in the judging backend.

vmcj:
> If I understand things correctly, after this PR languages will have fields in the UI that ask for compiler/runner and their args. But these commands are not used for the submission language, and only for the validators which can be very confusing (while the code is clean).
> 
> My proposal is actually to not introduce the new fields in the UI as a first step, but move to using the same executables that we use for submission languages for validators.
> 
> This will allow us to:
> 
>     * Eventually split them again, e.g. allow to define a "submission executable" and "validator executable" per language (where we fall back to the submission executable if no validator executable is defined). As well as having a "allow for validation" toggle, so you could allow say python validators but disallow python as submission language. This should help with addressing [Compare scripts should be build using language defaults #1238](https://github.com/DOMjudge/domjudge/issues/1238).
> 
>     * Add compiler/runner and their args in a separate step since this requires some thorough thinking of how this works with invalidating caching in the judging backend.

Makes sense, so I'll close this for now.

PR #2139, opened 2023-09-03
Construct ZIP files while uploading executables.

meisterT:
This is an improvement on top of #2129, since you do not need to check in ZIP files into a ccsconfig repository, but directories which are zipped up on the fly.
This makes changing and reviewing changes much easier.

### DIFF
--- a/misc-tools/configure-domjudge.in
+++ b/misc-tools/configure-domjudge.in
@@ -17,6 +17,7 @@ import json
 import os.path
 import requests
 import requests.utils
+import shutil
 import sys
 from typing import List, Set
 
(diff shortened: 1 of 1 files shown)


vmcj (approved):
I like this more, but if someone with the old FS layout would now use this, they would lose the original zips?

Why can't we have the check for `.zip` -> pick the old logic, elseif its a dir -> zip, upload, remove zip?

meisterT:
I thought about it, but the old thing has been around for a week, so the number of users is probably just me :-)

nickygerritsen (approved):

vmcj:
> I thought about it, but the old thing has been around for a week, so the number of users is probably just me :-)

Ok, I thought this was also in a released version. Than ripping it out is preferred.

Issue #2117, opened 2023-08-06
Display error in UI when reading event feed fails with a non-200


nickygerritsen:
Currently when reading from the event feed fails with a non-200 status code (for example when a 404 is produced), this is only written in the `event-feed.log` but there is no clear indication in the UI that this happened.

This was the case when you use the CDS as an upstream and had a `/` at the end of your CCS URL. That has been fixed now, but we should still handle the non-200 case better IMHO.

PR #2140, opened 2023-09-03
Expose when externalContestSource (URI) returns non200

vmcj:
To trigger this: create a contest with http://URL// (the double // is important)

~~Currently this does create the internal error but I'm not sure if we should do this, this is not really an internal error as its an external error. The InternalError class doesn't have support for this yet and we have nothing to disable.~~

~~Currently ignoring the error works, resolving keeps on hanging.~~

~~Opening the PR mostly for the discussion. Personally I think having an item under internalError view is nice as it alerts but its also a bit misusing the concept.~~ This will add 1 extra alert in the shadow tab of the menu and mention that we're down.

For the issue see: https://github.com/DOMjudge/domjudge/issues/2117

### DIFF
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -308,11 +308,14 @@ class ExternalContestSourceService
                 $fullUrl .= '&since_token=' . $this->getLastReadEventId();
             }
             $response = $this->httpClient->request('GET', $fullUrl, ['buffer' => false]);
-            if ($response->getStatusCode() !== 200) {
+            $statusCode = $response->getStatusCode();
+            $this->source->setLastHTTPCode($statusCode);
+            $this->em->flush();
+            if ($statusCode !== 200) {
                 $this->logger->warning(
                     'Received non-200 response code %d, waiting for five seconds ' .
                     'and trying again. Press ^C to quit.',
-                    [$response->getStatusCode()]
+                    [$statusCode]
                 );
                 sleep(5);
                 continue;
--- a/webapp/templates/jury/partials/external_contest_info.html.twig
+++ b/webapp/templates/jury/partials/external_contest_info.html.twig
@@ -43,6 +43,8 @@
                 <td>
                     {% if not externalContestSource.lastPollTime %}
                         Event feed reader never checked in.
+                    {% elseif externalContestSource.lastHTTPCode is not same as(200) %}
+                        Received {{ externalContestSource.lastHTTPCode }} HTTP code.
                     {% else %}
                         {{ status }}, last checked in {{ externalContestSource.lastPollTime | printtimediff }}s ago.
                     {% endif %}
--- a/webapp/migrations/Version20231006185028.php
+++ b/webapp/migrations/Version20231006185028.php
@@ -1,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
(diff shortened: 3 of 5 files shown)


nickygerritsen:
If we are adding fields to the contest source anyway, we could also add the actual error there and display it on that page?

vmcj:
> If we are adding fields to the contest source anyway, we could also add the actual error there and display it on that page?

We do show the HTTP error on the page, not on the internal error (yet) as I was not sure if we want the internal error.
If you had another way of doing this in mind I'm fine with that but this was in line with other code.

webapp/src/Controller/Jury/InternalErrorController.php
```diff
@@ -114,6 +114,8 @@ public function viewAction(int $errorId): Response
                 $affectedLink = $this->generateUrl('jury_executable', ['execId' => $disabled['execid']]);
                 $affectedText = $disabled['execid'];
                 break;
+            case 'externalContestSource':
```
nickygerritsen, line 117: This doesn't add anything if we don't do anything in the case

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -290,6 +303,34 @@ protected function setLastEvent(?string $eventId): void
             ->execute();
     }
 
+    protected function setLastHTTPCode(int $httpCode): void
+    {
+        // We use a direct query since we need to reload the source otherwise and we don't want that since that would
+        // take more queries.
+        $this->em
+            ->createQuery(
+                'UPDATE App\Entity\ExternalContestSource ecs
```
nickygerritsen, line 312: I would strongly prefer a sprintf with `ExternalContestSource::class` here so if we ever rename it IDE's will understand.

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -290,6 +303,34 @@ protected function setLastEvent(?string $eventId): void
             ->execute();
     }
 
+    protected function setLastHTTPCode(int $httpCode): void
+    {
+        // We use a direct query since we need to reload the source otherwise and we don't want that since that would
```
nickygerritsen, line 308: I don't get why / where we need to reload the entity to set (or get) the code. And even so, this one extra single query to set the HTTP code on reconnect doesn't seem too much compared to all the queries the event feed reader does?

webapp/templates/jury/partials/external_contest_info.html.twig
```diff
@@ -43,6 +43,8 @@
                 <td>
                     {% if not externalContestSource.lastPollTime %}
                         Event feed reader never checked in.
+                    {% elseif externalContestSource.lastHTTPCode is not same as(200) %}
+                        Received non HTTP200 ({{ externalContestSource.lastHTTPCode }}).
```
nickygerritsen, line 47:
```suggestion
                        Received non HTTP 200 ({{ externalContestSource.lastHTTPCode }}).
```

I don't think `HTTP200` is a single word. Or maybe even `non-200 HTTP code` or `Received {{ externalContestSource.lastHTTPCode }} HTTP code`?

webapp/src/Service/ExternalContestSourceService.php
vmcj: It was copied from above, so I assumed the same reasoning (and didn't fully understand why we didn't want to reload above).

nickygerritsen (approved):
I love it like this. Way simpler.

meisterT (approved):

PR #2143, opened 2023-09-05
Add option to filter submissions which are currently judging.

meisterT:


webapp/src/Service/SubmissionService.php
```diff
@@ -149,6 +149,14 @@ public function getSubmissionList(
                 $queryBuilder->andWhere('j.result IS NULL OR j.endtime IS NULL');
             }
         }
+        if (isset($restrictions['judging'])) {
+            dump($restrictions);
+            if ($restrictions['judging']) {
+                $queryBuilder->andWhere('j.starttime IS NOT NULL AND j.result IS NULL');
+            } else {
+                $queryBuilder->andWhere('j.starttime IS NULL OR j.result IS NOT NULL');
```
eldering, line 157: Does this logic still behave as expected when you view a rejudging? As then above the logic to join judgings is different. I didn't look in details TBH.

meisterT: Yes, it does, since it is based on a judging not submission.

eldering (approved):

PR #2144, opened 2023-09-17
Double hard wall time limit for interactive problems.

meisterT:
This accounts for wall time spent in the validator. We may likely want to make this configurable in the future. The current factor is under the assumption that the validator has to do approximately the same amount of work wall-time wise as the submission.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -1350,10 +1350,6 @@ function judge(array $judgeTask): bool
     }
 
     // do the actual test-run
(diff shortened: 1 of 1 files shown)


eldering (approved):
LGTM, but as already mentioned: in the long run would be good to have some configuration for this or otherwise make sure that admin users understand why and where this change is coming from.

meisterT:
Agreed, my course of action is to gain a little bit more experience with it and then decide at which level (global, contest, problem) we need what tuning option.

PR #2145, opened 2023-09-17
Fixes to executable table layout

vmcj:
Still WIP but some questions on layout

image
Do we think the different tables are improvement or cluther?

Should I re-sort the tables on enabled/disabled. I think given that we have the `search` option there it might make it cleaner.

The code tag looks like an improvement:
image

### DIFF
--- a/webapp/src/Controller/Jury/ExecutableController.php
+++ b/webapp/src/Controller/Jury/ExecutableController.php
@@ -58,7 +58,6 @@ class ExecutableController extends BaseController
             ->addOrderBy('e.type', 'ASC')
             ->addOrderBy('e.execid', 'ASC')
             ->getQuery()->getResult();
-        dump($executables);            
         $executables      = array_column($executables, 'executable', 'execid');
         $table_fields     = [
             'execid' => ['title' => 'ID', 'sort' => true,],
@@ -103,10 +102,6 @@ class ExecutableController extends BaseController
                 'link' => $this->generateUrl('jury_executable_download', ['execId' => $e->getExecid()])
             ];
 
-            //dump((bool)count($e->getLanguages()));
-            dump((bool)count($e->getProblemsCompare()));
-            dump((bool)count($e->getProblemsRun()));
-
             $executables_tables[$execType][] = [
                 'data' => $execdata,
                 'actions' => $execactions,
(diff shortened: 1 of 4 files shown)


webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -42,18 +44,21 @@ public function __construct(
...
         /** @var Executable[] $executables */
         $executables      = $em->createQueryBuilder()
-            ->select('e as executable, e.execid as execid')
+            ->select('e as executable, e.execid as execid, l.langid')
             ->from(Executable::class, 'e')
+            ->leftJoin(Language::class, 'l', Join::WITH, 'l.compile_executable = e.execid')
             ->addOrderBy('e.type', 'ASC')
             ->addOrderBy('e.execid', 'ASC')
             ->getQuery()->getResult();
+        dump($executables);
```
meisterT, line 61: left-over dump

webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -96,14 +103,29 @@ public function indexAction(Request $request): Response
...
+                'cssclass' => count($e->getLanguages()) || count($e->getProblemsCompare()) || count($e->getProblemsRun()) ? '' : 'disabled',
+            ];
+            $executables_table[] = [
                 'data' => $execdata,
                 'actions' => $execactions,
                 'link' => $this->generateUrl('jury_executable', ['execId' => $e->getExecid()]),
+                'cssclass' => count($e->getLanguages()) || count($e->getProblemsCompare()) || count($e->getProblemsRun()) ? '' : 'disabled',
             ];
         }
+        //dump($executables_tables);
```
meisterT, line 123: more dumps

webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -96,14 +103,29 @@ public function indexAction(Request $request): Response
                 'link' => $this->generateUrl('jury_executable_download', ['execId' => $e->getExecid()])
             ];
 
-            $executables_table[]            = [
+            //dump((bool)count($e->getLanguages()));
```
meisterT, line 106: remove dumps

meisterT:
I would try to have two tables, one with used and one unused executables and separating them by icon (and sorting) instead and see what looks better.

vmcj:
image

I've now also labeled executables as disabled when the language they belong to are not allowed to be submitted.

webapp/src/Controller/Jury/ExecutableController.php
vmcj: We have a CI which also flags this, as this was WIP I forgot to remove them (but will when merging).

PR #2148, opened 2023-09-21
Show unused executables as disabled

vmcj:
This splits over 2 tables, it does not mark the executable of languages (labeled as no submit) as disabled (yet).

### DIFF
--- a/webapp/src/Controller/Jury/ExecutableController.php
+++ b/webapp/src/Controller/Jury/ExecutableController.php
@@ -12,6 +12,7 @@ use App\Service\DOMJudgeService;
 use App\Service\EventLogService;
 use App\Utils\Utils;
 use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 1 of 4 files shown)


webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -39,9 +40,22 @@ public function __construct(
         protected readonly EventLogService $eventLogService
     ) {}
 
+    private function checkDisabled(Executable $e): bool
```
nickygerritsen, line 43: This should be `checkEnabled` if it returns true when enabled. Also maybe move this to the executable.php file?

vmcj:
This has suggestions from both @meisterT & @nickygerritsen as I had split an earlier PR over 2 different branches to find the actual issue.

nickygerritsen (approved):

PR #2149, opened 2023-09-21
Show type of executable with fontawesome icon

vmcj:
image

### DIFF
--- a/webapp/src/Controller/Jury/ExecutableController.php
+++ b/webapp/src/Controller/Jury/ExecutableController.php
@@ -56,6 +56,7 @@ class ExecutableController extends BaseController
             ->getQuery()->getResult();
         $executables      = array_column($executables, 'executable', 'execid');
         $table_fields     = [
+            'icon' => ['title' => 'type', 'sort' => false],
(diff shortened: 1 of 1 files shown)


webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -73,6 +74,23 @@ public function indexAction(Request $request): Response
                 }
             }
             $execdata['execid']['cssclass'] = 'execid';
+            $type = $execdata['type']['value'];
+            switch ($type) {
+                case 'compare':
+                    $execdata['icon']['icon'] = 'code-compare';
+                    break;
+                case 'compile':
+                    $execdata['icon']['icon'] = 'down-left-and-up-right-to-center';
```
meisterT, line 83: I prefer `language` over this but keep it if you like it.

meisterT (approved):

webapp/src/Controller/Jury/ExecutableController.php
vmcj: Language is fine by me.

PR #2150, opened 2023-09-21
Executables disenabled languages

vmcj:
Strictly speaking those are not disabled but not used in the instance. Because someone can still change them I prefer to show them as disabled to keep the interface clean.

image

### DIFF
--- a/webapp/src/Entity/Executable.php
+++ b/webapp/src/Entity/Executable.php
@@ -195,6 +195,11 @@ class Executable
         if (count($this->problems_compare) || count($this->problems_run)) {
             return true;
         }
+        foreach ($this->languages as $lang) {
+            if ($lang->getAllowSubmit()) {
+                return true;
+            }
+        }
         return false;
     }
 }


meisterT:
So, used and unused is the better wording?

webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -39,9 +40,27 @@ public function __construct(
         protected readonly EventLogService $eventLogService
     ) {}
 
+    private function checkDisabled(Executable $e): bool
+    {
+        if (in_array($e->getExecid(), ['compare', 'run', 'full_debug'])) {
```
meisterT, line 45: Read these from the config instead where they are selected as default scripts

webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -39,9 +40,29 @@ public function __construct(
         protected readonly EventLogService $eventLogService
     ) {}
 
+    private function checkUnused(Executable $e): bool
+    {
+        foreach(['compare', 'run', 'full_debug'] as $config_script) {
+            if ($e->getExecid() === (string)$this->config->get('default_' . $config_script)) {
```
meisterT, line 46: Do we need a null check here? In case someone uses an old database?

vmcj: I think a null check would not catch the exception for that case. So I'll wrap this in a `try..catch`

nickygerritsen (approved):

PR #2151, opened 2023-09-21
Added problem badges

vmcj:
image

### DIFF
--- a/webapp/src/Controller/Jury/ExecutableController.php
+++ b/webapp/src/Controller/Jury/ExecutableController.php
@@ -3,6 +3,7 @@
 namespace App\Controller\Jury;
 
 use App\Controller\BaseController;
+use App\Entity\ContestProblem;
 use App\Entity\Executable;
 use App\Entity\ExecutableFile;
 use App\Entity\ImmutableExecutable;
@@ -62,6 +63,7 @@ class ExecutableController extends BaseController
             'icon' => ['title' => 'type', 'sort' => false],
             'execid' => ['title' => 'ID', 'sort' => true,],
             'type' => ['title' => 'type', 'sort' => true,],
+            'badges' => ['title' => 'problems', 'sort' => false],
             'description' => ['title' => 'description', 'sort' => true,],
         ];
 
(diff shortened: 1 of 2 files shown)


nickygerritsen (approved):
Code style warnings but besides that LGTM

webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -59,12 +59,25 @@ public function indexAction(Request $request): Response
...
+            'badges' => ['title' => 'problems', 'sort' => false],
             'description' => ['title' => 'description', 'sort' => true,],
         ];
 
         $propertyAccessor  = PropertyAccess::createPropertyAccessor();
         $executables_table = [];
         foreach ($executables as $e) {
+            $badges = [];
+            if ($this->dj->getCurrentContest()) {
+                foreach([$e->getProblemsCompare(), $e->getProblemsRun()] as $problemList) {
```
meisterT, line 71: If you use `array_merge` here, you can reduce the two outer loops to a single one

webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -91,6 +104,7 @@ public function indexAction(Request $request): Response
                 default:
                     $execdata['icon']['icon'] = 'question';
             }
+            $execdata['badges']['value'] = $badges;
```
meisterT, line 107: Should we add an explicit sort here?

vmcj:
image

I tried to improve the speed of the queries when handling Tobi his remarks. @nickygerritsen this is non-trivial with Doctrine is there a better way?

webapp/src/Controller/Jury/ExecutableController.php
vmcj: Tried to fix this in the query to speed it up further.

webapp/templates/jury/jury_macros.twig
```diff
@@ -141,6 +141,10 @@
                             {{- (item.value|default(item.default|default('')))|affiliationLogo(item.title) -}}
                             {% elseif key == "warning_content" %}
                             {{- item.value|printWarningContent -}}
+                            {% elseif key == "badges" %}
+                                {% for badge in item.value %}
+                                    {{- badge|problemBadge -}}
```
nickygerritsen, line 146: Maybe we want a space between the badges?

nickygerritsen (approved):
> I tried to improve the speed of the queries when handling Tobi his remarks. @nickygerritsen this is non-trivial with Doctrine is there a better way?

Besides doing direct queries or maybe a DQL, I don't think so

webapp/templates/jury/jury_macros.twig
vmcj: image

Fine by me.

meisterT (approved):

PR #2154, opened 2023-09-23
Update ScoreboardMergeCommand

RagnarGrootKoerkamp:
Some changes I made:
1. Allow reading from local files directly. It turns out to be more reliable to first download scoreboards from the API and then merge them when servers go down after the contest. (This needs e.g. `- /path/to/scoreboards:/scoreboards` in the `docker-compose.yml` which needs to be added manually.)
2. Do not check that group IDs look like integers -- they can be strings.
3. Match problems by display name instead of ID, since sites differ in whether they use `A` or `problemshortname` as ID.

I also had to disable showing the banner image, but I changed that by commenting it out in the twig. This needs some better fix.

(More comments in the [script in BAPCtools](https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/master/bin/misc/merge_scoreboards.sh) that uses this.)

### DIFF
--- a/webapp/src/Command/ScoreboardMergeCommand.php
+++ b/webapp/src/Command/ScoreboardMergeCommand.php
@@ -89,8 +89,28 @@ class ScoreboardMergeCommand extends Command
                 'Alternating URL location of the scoreboard to merge and a comma separated list of group_ids to include.' . PHP_EOL .
                 'If an URL and it requires authentication, use username:password@ in the URL' . PHP_EOL .
                 'URL should have the form https://<domain>/api/v4/contests/<contestid>/ for DOMjudge or point to any ICPC Contest API compatible contest' . PHP_EOL .
-                'Only the /teams, /organizations, /problems and /scoreboard endpoint are used, so manually putting files in those locations can work as well.'
+                'Only the /teams, /organizations, /problems and /scoreboard endpoint are used, so manually putting files in those locations can work as well.' . PHP_EOL .
+                'Alternatively, you can mount local files directly in the container: add "- /path/to/scoreboards:/scoreboards" to "docker-compose.yml" and use "/scoreboards/eindhoven" as path.'
             );
+    }
+
+    /**
+     * url: "https://judge.gehack.nl/api/v4" or "/path/to/file"
+     * endpoint: "/teams"
+     * args: "?public=1" (ignored for files)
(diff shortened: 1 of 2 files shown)


webapp/src/Command/ScoreboardMergeCommand.php
```diff
@@ -59,7 +59,10 @@ public function __construct(
 
     protected function configure(): void
     {
-        $this
+        $this->setName('scoreboard:merge')
```
nickygerritsen, line 62: This is not needed anymore because of line 40

webapp/templates/public/scoreboard.html.twig
```diff
@@ -7,7 +7,7 @@
         {% if showExternalId(contest) %}
             {% set contestId = contest.externalid %}
         {% endif %}
-        {% set bannerImage = contestId | assetPath('contest') %}
+        {#% set bannerImage = contestId | assetPath('contest') %#}
```
nickygerritsen, line 10: What happened when this was enabled?

RagnarGrootKoerkamp:
```

In DOMJudgeService.php line 1363:
                                                                                                                     
  App\Service\DOMJudgeService::assetPath(): Argument #1 ($name) must be of type string, null given, called in /domj  
  udge/webapp/var/cache/dev/twig/05/0524521e8604d78b0578c15c32703153.php on line 108
  ```

webapp/src/Command/ScoreboardMergeCommand.php
RagnarGrootKoerkamp: right; removed

webapp/templates/public/scoreboard.html.twig
nickygerritsen: Lets add a check if contestId is not null then?

RagnarGrootKoerkamp: right yes done; that was easy in hindsight

nickygerritsen:
Can you look into the annotations reported by phpstan and the codestyle? Afterwards we can merge I tihnk.

RagnarGrootKoerkamp:
All should be/become green now

nickygerritsen:
Thanks!

PR #2157, opened 2023-09-24
Disarm timer in runguard after child has exited.

meisterT:
If we would not disarm the timer, there is a possibility that the timer sends us a SIGALRM while we are still busy with cleaning the sandbox up.

What you observe in these cases is a judging with wall time well below the time limit is judged as TLE, e.g.
```
Timelimit exceeded.
runtime: 0.288s cpu, 0.302s wall
memory used: 26066944 bytes
********** runguard stderr follows **********
/opt/domjudge/bin/runguard: warning: timelimit exceeded (hard wall time): aborting command
```

In practice, we saw the behavior happening when running many judgedaemons and domserver on a single machine while rejudging the whole contest (i.e.  under quite high load). In that case, the call `cgroup_delete_cgroup_ext` did sometimes hang for multiple seconds.

For easy reproducibility, you can also add an artificial delay in the clean up code, e.g. by adding something like:
```
const struct timespec artificial_delay = { 10, 0 };
nanosleep(&artificial_delay, NULL);
```

### DIFF
--- a/judge/runguard.c
+++ b/judge/runguard.c
@@ -1428,8 +1428,23 @@ int main(int argc, char **argv)
 		} else {
 			exitcode = WEXITSTATUS(status);
 		}
+		verbose("child exited with exit code %d", exitcode);
 
-        check_remaining_procs();
+		if ( use_walltime ) {
+			/* Disarm timer we set previously so if any of the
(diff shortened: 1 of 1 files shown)


edomora97 (approved):
Makes total sense! Thanks!

PR #2158, opened 2023-09-24
Use correct printf format specifier for filesize.

meisterT:
The default value for filesize for scripts is 2.5GB, so more than 2^31 bytes, previously this caused a message like

```
/home/sitowert/domjudge/bin/runguard [56839 @   0.001495]: verbose: setting filesize limit to -1610612736 bytes
```

After this commit, it is:
```
/home/sitowert/domjudge/bin/runguard [56839 @   0.001521]: verbose: setting filesize limit to 2684354560 bytes
```

### DIFF
--- a/judge/runguard.c
+++ b/judge/runguard.c
@@ -799,7 +799,7 @@ void setrestrictions()
 	setlim(STACK);
 
 	if ( filesize!=RLIM_INFINITY ) {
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

eldering:
Agreed, according to `man setrlimit` the `rlim_t` data type is an unsigned long, also double checked in the kernel source.

Issue #2160, opened 2023-09-26
Cannot get list of contests as a jury if there's no contest


Suckzoo:
### Description of the problem
When there is no contest at all, [this line](https://github.com/DOMjudge/domjudge/blob/8.2.1/webapp/src/Controller/Jury/ContestController.php#L382) of released code makes NPE.
As a result, a jury cannot enter contest page, and cannot make any contest.

### Your environment
- DOMjudge version 8.2.1 (latest stable)
- Operating system / Linux distribution and version Ubuntu 22.04
- Webserver Apache2

### Steps to reproduce
- Just bare-install and jump into contest page as an admin

### Expected behaviour
- The contests page should not return 500 response.

### Actual behaviour
- NPE occurs: calling isLocked on null.

### Any other information that you want to share?
- As of the main branch, the bug seems to be fixed.

vmcj:
> NOTE: If this is a _discussion starter_, you need any _installation help_ or have a _question_ on how to accomplish something, rather post at [our discussion channel](https://www.domjudge.org/chat) or send an email to our [DOMjudge-devel mailinglist](https://www.domjudge.org/mailman/postorius/lists/domjudge-devel.domjudge.org/) instead of filing an issue here.
> ### Description of the problem
> 
> When there is no contest at all, [this line](https://github.com/DOMjudge/domjudge/blob/8.2.1/webapp/src/Controller/Jury/ContestController.php#L382) of released code makes NPE. As a result, a jury cannot enter contest page, and cannot make any contest.
> ### Your environment
> 
>     * DOMjudge version 8.2.1 (latest stable)
> 
>     * Operating system / Linux distribution and version Ubuntu 22.04
> 
>     * Webserver Apache2
> 
> 
> ### Steps to reproduce
> 
>     * Just bare-install and jump into contest page as an admin
> 
> 
> ### Expected behaviour
> 
>     * The contests page should not return 500 response.
> 
> 
> ### Actual behaviour
> 
>     * NPE occurs: calling isLocked on null.
> 
> 
> ### Any other information that you want to share?
> 
>     * As of the main branch, the bug seems to be fixed.

Hi @Suckzoo, thank you for your issue. A PR would be welcome as I think you've already found the fix for the 8.2 branch.

Suckzoo:
[...]

PR #2161, opened 2023-09-27
Auto calculate num actions for jury tables.

Suckzoo:
ContestController is broken in 8.2.1. (https://github.com/DOMjudge/domjudge/issues/2160)

This PR cherry-picks [1294b6e](https://github.com/DOMjudge/domjudge/blob/1294b6e/webapp/src/Controller/Jury/ContestController.php).


nickygerritsen (changes requested):
This is not going to work as is, since the older PHP/Symfony we use doesn't understand the `#[Route]` etc. I will see if I can properly backport this.

nickygerritsen:
Thanks for the initial change, I updated it.

PR #2167, opened 2023-09-29
Allow reimport of contest & contest problems

vmcj:
This would allow having 2 contests with the same problems which is useful for the BAPC prelims where one DOMjudge instance hosts multiple contests for different locations (not to be confused with ICPC sites).

In the past the import-contest script would receive an 500 as we don't properly check if a problem with that name already exists.

This should make the import-contest idempotent.

### DIFF
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -322,18 +322,19 @@ class ImportExportService
         foreach ($problems as $problemData) {
             // Deal with obsolete attribute names. Also for name fall back to ID if it is not specified.
             $problemName  = $problemData['name'] ?? $problemData['short-name'] ?? $problemData['id'] ?? null;
+            $problemExternal = $problemData['id'] ?? $problemData['short-name'] ?? $problemLabel ?? null;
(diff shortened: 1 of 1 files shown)


webapp/src/Service/ImportExportService.php
```diff
@@ -299,7 +299,7 @@ public function importProblemsData(Contest $contest, array $problems, array &$id
             $this->em->persist($problem);
             $this->em->flush();
 
-            $contestProblem = new ContestProblem();
+            $contestProblem = $this->em->getRepository(ContestProblem::class)->findOneBy(['shortname' => $problemLabel, 'problem' => $problem]) ?: new ContestProblem();
```
vmcj, line 302:
```suggestion
            $contestProblem = $this->em->getRepository(ContestProblem::class)->findOneBy(['shortname' => $problemLabel, 'problem' => $problem, 'contest' => $contest]) ?: new ContestProblem();
```

vmcj:
I'll add a check for this case and add an optional confirmation that it should be overwritten.

vmcj:
The optional compare scripts for a problem are now recreated on every upload so that should be fixed first.

nickygerritsen (approved):

PR #2168, opened 2023-09-29
Check if username confirms to Entity regex

vmcj:
Explicit not done for the TSV to keep the old behaviour. We should check if this regex can be shared globally and used in the assertions on the Entity and through the different API endpoints for contstraints.

### DIFF
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -877,6 +877,15 @@ class ImportExportService
             $juryTeam = null;
             $roles    = [];
             $type     = $account['type'];
+            $username = $account['username'];
+
+            $icpcRegexChars = "[a-zA-Z0-9@._-]";
+            $icpcRegex = "/^" . $icpcRegexChars . "+$/";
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

webapp/src/Service/ImportExportService.php
```diff
@@ -877,6 +877,14 @@ public function importAccountsJson(array $data, ?string &$message = null, ?array
             $juryTeam = null;
             $roles    = [];
             $type     = $account['type'];
+            $username = $account['username'];
+
+            $icpcRegex = "/^[a-zA-Z0-9@._-]+$/";
+            if (!preg_match($icpcRegex, $username)) {
+                $message = sprintf('Username "%s" should follow regex pattern: %s', $username, $icpcRegex);
```
meisterT, line 884: While you are testing with a regex I think it is more user friendly to talk about the allowed set of characters here.

vmcj: That takes an extra variable but now done.

meisterT (approved):

PR #2170, opened 2023-09-30
Add option to upload languages configuration via API.

### LINKED ISSUE #2142: Allow uploading `languages.json` to the API
We currently can download the language configuration as part of the API.

It returns a JSON array similar to this:
```
[
    {
        "allow_judge": true,
        "compile_executable_hash": "3ac90f7921fe29186658a8806ac183ca",
        "compiler": {
            "version_command": "gcc --version"
        },
        "entry_point_name": null,
        "entry_point_required": false,
        "extensions": [
            "c"
        ],
        "filter_compiler_files": true,
        "id": "c",
        "name": "C",
        "time_factor": 1.0
    },
...
```

It would be great if you could also upload a `languages.json` to update the language configuration. I think we do care the most about `allow_judge` and `entry_point_required` but we could allow setting everything except `compile_executable_hash` for which we already have an endpoint to update executables.

meisterT:
Fixes #2142.

### DIFF
--- a/webapp/src/Controller/API/LanguageController.php
+++ b/webapp/src/Controller/API/LanguageController.php
@@ -10,6 +10,7 @@ use Doctrine\ORM\QueryBuilder;
 use FOS\RestBundle\Controller\Annotations as Rest;
 use Nelmio\ApiDocBundle\Annotation\Model;
 use OpenApi\Attributes as OA;
+use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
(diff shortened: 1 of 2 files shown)


webapp/src/Controller/API/LanguageController.php
```diff
@@ -122,6 +123,84 @@ public function updateExecutableActions(Request $request, string $id): void
         $this->dj->auditlog('executable', $language->getLangid(), 'updated');
     }
 
+    #[IsGranted('ROLE_ADMIN')]
```
vmcj, line 126: Do we want this to be API_WRITER? We do for other endpoints but as this is setup of the instance I can see why it can be ADMIN only.

webapp/src/Controller/API/LanguageController.php
```diff
@@ -122,6 +123,84 @@ public function updateExecutableActions(Request $request, string $id): void
...
+            /** @var Language $language */
+            $lang_id = $language['id'];
+            $lang = $this->em->getRepository(Language::class)->findOneBy(
+                [$idField => $lang_id]
+            );
+            if (!$lang) {
+                continue;
+            }
+            $lang->setAllowSubmit(true);
+            if (isset($language['name'])) {
```
nickygerritsen, line 159: We could use a PropertyAccessor loop here to get rid of most duplication. See also the problem import somewhere (I believe)

nickygerritsen (approved):

webapp/src/Controller/API/LanguageController.php
```diff
@@ -122,6 +123,84 @@ public function updateExecutableActions(Request $request, string $id): void
...
+
+        $idField = $this->eventLogService->externalIdFieldForEntity(Language::class) ?? 'langid';
+        foreach ($newLanguages as $language) {
+            /** @var Language $language */
+            $lang_id = $language['id'];
+            $lang = $this->em->getRepository(Language::class)->findOneBy(
+                [$idField => $lang_id]
+            );
+            if (!$lang) {
+                continue;
```
vmcj, line 156: Should we not indicate this to the one uploading that the language is missing?

webapp/src/Controller/API/LanguageController.php
meisterT: Looking at other places, I think this is fine as is, e.g.:
- https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Controller/API/ContestController.php#L68
- https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Controller/API/ProblemController.php#L57
- https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Controller/API/GeneralInfoController.php#L244

webapp/src/Controller/API/LanguageController.php
meisterT: I tried this and it was not really less code / more understandable (probably also because of the nested compiler/runner fields). Feel free to submit to this branch.

webapp/src/Controller/API/LanguageController.php
meisterT: Yes, and no:

- Logging is not really possible as we return the updated languages config, so that it can be compared in the script.
- We could error out, but that's quite brutal.
- We could try adding a new language, but of course stuff like executables are missing here.

What about this: I will add a TODO comment here for now; and when we decided on the best solution we can replace it?

webapp/src/Controller/API/LanguageController.php
nickygerritsen: Then I’m fine as is

webapp/src/Controller/API/LanguageController.php
vmcj: I'm fine with that.

PR #2172, opened 2023-10-01
Shadow create submission during freeze

vmcj:
Still create submission even if the URL to the file is not given. Useful if you use DOMjudge mainly as scoreboard after an eventfeed and not as a shadowing system.

In the past we would not create the submissions because the link was missing, as the earlier intent was to always create submssions even if we can't download the submission_file (yet) this is still in line with earlier code.


nickygerritsen (approved):

vmcj:
Nicky changed a lot after this one and this problem should have been fixed in main.

PR #2173, opened 2023-10-03
Minor fixes related to import.

nickygerritsen:
Found all these while working on the domlogo changes.

### DIFF
--- a/webapp/src/Service/DOMJudgeService.php
+++ b/webapp/src/Service/DOMJudgeService.php
@@ -654,7 +654,13 @@ class DOMJudgeService
             return null;
         }
 
-        return $this->jsonDecode($response->getContent());
+        $content = $response->getContent();
+
+        if ($content === '') {
+            return null;
+        }
+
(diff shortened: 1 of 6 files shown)


webapp/src/Service/DOMJudgeService.php
```diff
@@ -654,7 +654,13 @@ public function internalApiRequest(string $url, string $method = Request::METHOD
             return null;
         }
 
-        return $this->jsonDecode($response->getContent());
+        $content = $response->getContent();
+
+        if ($content === '') {
+            return null;
```
meisterT, line 660: Why is this necessary?

nickygerritsen: The message of 33db981027a0c58fce7e9b4445d6410d06ded8dd explains this, but the API returns an empty response for 204's, like when uploading an image. This can't be JSON decoded and apparently breaks the import script.

meisterT (approved):

PR #2175, opened 2023-10-07
Handle kotlinc symbolic link for determining KOTLIN_DIR

agcom:
The `KOTLIN_DIR` variable needs to point to the **compiler binary directory** containing the `kotlinc` binary; because, latter in the script, `KOTLIN_DIR` is used to point to the `kotlin-stdlib.jar`; sometimes `kotlinc` would be just a symbolic link residing in a directory included in the `PATH` environment variable.

It is possible to manually set the `KOTLIN_DIR` variable; but, with this change, unzipping the Kotlin installation and creating a symbolic link would suffice; see https://github.com/DOMjudge/domjudge-packaging/issues/156.

### DIFF
--- a/sql/files/defaultdata/kt/run
+++ b/sql/files/defaultdata/kt/run
@@ -19,7 +19,7 @@ COMPILESCRIPTDIR="$(dirname "$0")"
 # Note that you then also might want to fix this in the compiler and runner version commands.
 # For example
 # KOTLIN_DIR=/usr/lib/kotlinc/bin
-KOTLIN_DIR="$(dirname "$(command -v kotlinc)")"
+KOTLIN_DIR="$(dirname "$(realpath "$(command -v kotlinc)")")"
 
 # Stack size in the JVM in KB. Note that this will be deducted from
 # the total memory made available for the heap.


sql/files/defaultdata/kt/run
```diff
@@ -19,7 +19,7 @@ COMPILESCRIPTDIR="$(dirname "$0")"
 # Note that you then also might want to fix this in the compiler and runner version commands.
 # For example
 # KOTLIN_DIR=/usr/lib/kotlinc/bin
-KOTLIN_DIR="$(dirname "$(command -v kotlinc)")"
+KOTLIN_DIR="$(if [ -L "$(command -v kotlinc)" ]; then echo "$(dirname "$(readlink "$(command -v kotlinc)")")"; else echo "$(dirname "$(command -v kotlinc)")"; fi)"
```
nickygerritsen, line 22: I like the idea but maybe we should split this on multiple lines?
```suggestion
if [ -L "$(command -v kotlinc)" ]
then
    KOTLIN_DIR=$(dirname "$(readlink "$(command -v kotlinc)")")
else
    KOTLIN_DIR=$(dirname "$(command -v kotlinc)")
fi
```

Seems a bit more readable to me?

nickygerritsen: OK nvm your fix is even better.

nickygerritsen (approved):

nickygerritsen:
I'd like input from others as well here.

sql/files/defaultdata/kt/run
agcom: Checkout the new commit; I tried to keep it on a single line for easier replacement by the user (as stated in the example comments above).

eldering:
LGTM, but please squash the commits into a single one before merging.

vmcj:
@agcom can you squash the commits? Afterwards we'll merge it.

agcom:
@vmcj There you go. I thought there is an auto squash and merge option.

PR #2176, opened 2023-10-07
Upgrade chroot, upgrade debootstrap & install kotlin default

vmcj:
I assume we also upgrade the debootstrap deb when we upgrade the chroot although this is not strictly needed. I've also added Kotlin to the chroot as it is probably used as often as Java.

@thijskh, is it a problem that the new debootstrap debs are NMU (https://wiki.debian.org/NonMaintainerUpload)?


thijskh:
Not a problem

meisterT (approved):

PR #2182, opened 2023-10-20
Add `make` and `libcgroup` to DOMserver's software requirements

YoungMind1:


vmcj:
Hi @YoungMind1,

We already mention this here: https://github.com/DOMjudge/domjudge/pull/2182/files#diff-626c66d580fc2cc70b82c5f745a6206ae4414b2a2837be5dafd3a2f55eed0aa6R49 but I think it's fine to add this and remove it again when we fix that bug.

vmcj (approved):

eldering:
I added a commit on top to make it clear that these packages are here only because of judgehost requirements.

eldering (approved):

PR #2187, opened 2023-10-22
API: Allow setting external id for User entity

### LINKED ISSUE #2186: API: External ID for User entity cannot be set via API

When using `data_source=1` (i.e. configuration data external), User entity cannot be create via API endpoint `/api/v4/users` due to missing externalid attribute

### Your environment
DOMjudge 8.2.1

### Steps to reproduce
- set data_source to 1 (configuration data external)
- make API call
`curl -XPOST https://domserver/api/v4/users -Fusername=demo -Fname=demo -Fpassword= -F"roles[]=team"`

### Expected behaviour
External id can be set via API call or is automatically created. Like in the Teams entity, `id` attribute should be included in the API call:
`curl -XPOST https://domserver/api/v4/users -Fid=demo -Fusername=demo -Fname=demo -Fpassword= -F"roles[]=team"` and that should set the external id if using configuration data external mode.
```
{"last_login_time":null,"last_api_login_time":null,"first_login_time":null,"team":null,"team_id":null,"roles":["team"],"type":"team","id":demo,"username":"demo","name":"demo","email":null,"last_ip":null,"ip":null,"enabled":true}
```

### Actual behaviour
API call returns:
```
{"last_login_time":null,"last_api_login_time":null,"first_login_time":null,"team":null,"team_id":null,"roles":["team"],"type":"team","id":null,"username":"demo","name":"demo","email":null,"last_ip":null,"ip":null,"enabled":true}
```
No external id for the entity is created and the configuration checker shows an error "User XY does not have an external ID" which needs to be fixed by manually editing all User entities via GUI.

baierjan:
This will fix #2186 when using `data_source=1`. I am not sure how this will affect the API for other modes.

### DIFF
--- a/webapp/src/Controller/API/UserController.php
+++ b/webapp/src/Controller/API/UserController.php
@@ -329,6 +329,7 @@ class UserController extends AbstractRestController
             ->setName($request->request->get('name'))
             ->setEmail($request->request->get('email'))
             ->setIpAddress($request->request->get('ip'))
+            ->SetExternalid($request->request->get('id'))
             ->setPlainPassword($request->request->get('password'))
             ->setEnabled($request->request->getBoolean('enabled', true));
 


nickygerritsen (commented):
I think we need to add this field to the AddUser scheme. And maybe add a test? (But we could do that ourself )

vmcj:
@baierjan will you implement @nickygerritsen his feedback or should someone else look into this?

baierjan:
@vmcj I am not familiar with the code base, so it will be much faster if you can take look at it. Feel free to close this PR (it is just a hotfix anyway) and do it properly.

nickygerritsen:
I am working on a version that adds a test for this as well, PR incoming soon.

Issue #2188, opened 2023-10-27
Allow to set self registration flag when import groups.json


nickygerritsen:
To enable more automation during imports it would be nice if one can enable self registration for a team category by setting some flag in the JSON during group import.

PR #2200, opened 2023-10-30
Allow selfregistration property for groups via API

vmcj:
The intent is that only priviliged users can view the properties but not expose those to the generic users (Teams, Public).

Closes: https://github.com/DOMjudge/domjudge/issues/2188

### DIFF
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -595,7 +595,8 @@ class ImportExportService
             $line = Utils::parseTsvLine(trim($line));
             $groupData[] = [
                 'categoryid' => @$line[0],
-                'name' => @$line[1]
+                'name' => @$line[1],
+                'allow_self_registration' => false,
             ];
         }
 
@@ -618,6 +619,7 @@ class ImportExportService
                 'visible' => !($group['hidden'] ?? false),
                 'sortorder' => @$group['sortorder'],
                 'color' => @$group['color'],
+                'allow_self_registration' => $group['allow_self_registration'] ?? false,
             ];
         }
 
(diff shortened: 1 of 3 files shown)


webapp/src/Service/ImportExportService.php
```diff
@@ -618,6 +618,7 @@ public function importGroupsJson(array $data, ?string &$message = null, ?array &
                 'visible' => !($group['hidden'] ?? false),
                 'sortorder' => @$group['sortorder'],
                 'color' => @$group['color'],
+                'allow_self_registration' => @$group['allow_self_registration'],
```
nickygerritsen, line 621: Maybe
```suggestion
                'allow_self_registration' => $group['allow_self_registration'] ?? false,
```

is better here?

webapp/src/Service/ImportExportService.php
```diff
@@ -661,6 +662,9 @@ protected function importGroupData(array $groupData, ?array &$saved = null): int
                 ->setSortorder($groupItem['sortorder'] ?? 0)
                 ->setColor($groupItem['color'] ?? null)
                 ->setIcpcid($groupItem['icpc_id'] ?? null);
+            if (!empty($groupItem['allow_self_registration'])) {
```
nickygerritsen, line 665: If we do my above suggestion this if can be dropped.

webapp/src/Service/ImportExportService.php
```diff
@@ -661,6 +662,9 @@ protected function importGroupData(array $groupData, ?array &$saved = null): int
...
+                $teamCategory->setAllowSelfRegistration($groupItem['allow_self_registration']);
+            }
```
vmcj, line 667:
```suggestion
            $teamCategory->setAllowSelfRegistration($groupItem['allow_self_registration']);
```

webapp/src/Service/ImportExportService.php
```diff
@@ -595,7 +595,8 @@ protected function importGroupsTsv(array $content, ?string &$message = null): in
             $line = Utils::parseTsvLine(trim($line));
             $groupData[] = [
                 'categoryid' => @$line[0],
-                'name' => @$line[1]
+                'name' => @$line[1],
+                'allow_self_registration' => false,
```
vmcj, line 599: @nickygerritsen this is now also needed with your fix.

nickygerritsen: Sure. Seems fine

nickygerritsen (approved):

Issue #2199, opened 2023-10-30
Error while running `make dist` 


YoungMind1:
OS: Ubuntu 23.04
Packages: everything listed [here](https://www.domjudge.org/docs/manual/8.2/develop.html#bootstrapping-from-git-repository-sources)

I get the following output when running `make dist`
```
make -C doc distdocs
make[1]: Entering directory '/home/amir/Documents/domjudge/doc'
Recursing target `distdocs' into: manual examples
make[2]: Entering directory '/home/amir/Documents/domjudge/doc/manual'
make html team
make[3]: Entering directory '/home/amir/Documents/domjudge/doc/manual'
sphinx-build -M html . build 
Running Sphinx v5.3.0
WARNING: The config value `pdf_use_toc' has type `str', defaults to `bool'.
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 18 source files that are out of date
updating environment: [new config] 18 added, 0 changed, 0 removed
reading sources... [100%] upgrading                                                                                                                                                           
/home/amir/Documents/domjudge/doc/manual/install-domserver.rst:104: ERROR: Undefined substitution referenced: "phpversion".
/home/amir/Documents/domjudge/doc/manual/install-domserver.rst:128: ERROR: Undefined substitution referenced: "phpversion".
/home/amir/Documents/domjudge/doc/manual/install-domserver.rst:128: ERROR: Undefined substitution referenced: "phpversion".
/home/amir/Documents/domjudge/doc/manual/install-domserver.rst:128: ERROR: Undefined substitution referenced: "phpversion".
/home/amir/Documents/domjudge/doc/manual/install-domserver.rst:128: ERROR: Undefined substitution referenced: "phpversion".
/home/amir/Documents/domjudge/doc/manual/install-domserver.rst:143: ERROR: Undefined substitution referenced: "phpversion".
/home/amir/Documents/domjudge/doc/manual/install-domserver.rst:143: ERROR: Undefined substitution referenced: "phpversion".
/home/amir/Documents/domjudge/doc/manual/install-domserver.rst:143: ERROR: Undefined substitution referenced: "phpversion".
/home/amir/Documents/domjudge/doc/manual/quick-install.rst:40: ERROR: Undefined substitution referenced: "phpversion".
/home/amir/Documents/domjudge/doc/manual/quick-install.rst:40: ERROR: Undefined substitution referenced: "phpversion".
[...]

PR #2204, opened 2023-11-04
Address current failing sphinx packages for development

vmcj:
We do generate the docs already before so the virtualenv is not that problematic. I will investigate if we can do something else so we don't need this but for now this is the best we can do.

This does block releasing though (although this also exists in the older releases) as a dependency for us is broken.

Closes: https://github.com/DOMjudge/domjudge/issues/2199

### DIFF
--- a/doc/manual/develop.rst
+++ b/doc/manual/develop.rst
@@ -48,6 +48,17 @@ On Debian 11 (Bullseye) and above, instead install::
     python3-sphinx python3-sphinx-rtd-theme rst2pdf fontconfig python3-yaml \
     latexmk texlive-latex-recommended texlive-latex-extra tex-gyre
 
+On Debian 12 (Bullseye) and derived versions, install::
+
+  sudo apt install autoconf automake bats \
+    fontconfig python3-yaml latexmk texlive-latex-recommended \
+    texlive-latex-extra tex-gyre python3-venv
+
+and create a virtual environment to install these packages currently failing in the
+default repositories::
+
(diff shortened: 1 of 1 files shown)


eldering:
> We do generate the docs already before so the virtualenv is not that problematic. I will investigate if we can do something else so we don't need this but for now this is the best we can do.
> 
> This does block releasing though (although this also exists in the older releases) as a dependency for us is broken.
> 
> Closes: #2199

This explanation is confusing: we generate the doce before _what_? How does that make the virtualenv not that problematic?

Also, is #2205 an alternative for this PR?

vmcj:
> > We do generate the docs already before so the virtualenv is not that problematic. I will investigate if we can do something else so we don't need this but for now this is the best we can do.
> > This does block releasing though (although this also exists in the older releases) as a dependency for us is broken.
> > Closes: #2199
> 
> This explanation is confusing: we generate the doce before _what_? How does that make the virtualenv not that problematic?
> 
In a normal release I thought (I verified and we don't) we already package the team manual. In that case only the one running `make dist` would need the extra virtual environment. 

> Also, is #2205 an alternative for this PR?
Yes, I think we should merge the other one if we accept the difference in the manual, but documenting how people can run `make dist` when doing development is a cleaner fix for the original issue (it was also easier to make sure we do fix this, only later did I realize how to find the actual issue).

vmcj:
So by making the manual ugly in the other PR mentioned above, the issue is now fixed so I'll close this one.

PR #2205, opened 2023-11-04
This unbreaks our teammanual generation

vmcj:
Removing this seems to also remove the background color for the summary. We should investigate what was the goal of the LaTeX change and how we can reproduce it.

Also https://github.com/rst2pdf/rst2pdf/pull/846 is now merged and released so we can remove a warning about the bool/str.

Closes: https://github.com/DOMjudge/domjudge/issues/2199

### DIFF
--- a/doc/manual/conf.py
+++ b/doc/manual/conf.py
@@ -148,8 +148,6 @@ latex_elements = {
 \definecolor{noteBgColor}{rgb}{1,0,1}
 
 \definecolor{sphinxnoteBgColor}{RGB}{221,233,239}
-\renewenvironment{sphinxnote}[1]
-{\begin{sphinxheavybox}\sphinxstrong{#1} }{\end{sphinxheavybox}}
 
 \usepackage{fancyhdr}
 \pagestyle{fancy}


eldering:
This removes the background colour also of the code examples and a few other places. It clearly makes the layout a bit less nice, but if there's no other way to unbreak building the team manual without dirty hacks, then maybe this is worth it.
I've added @thijskh as reviewer, as I'm pretty sure he was the one who creates this layout.

vmcj:
I thought I had added the difference:
Original layout:
image

Now:
image

vmcj:
> This removes the background colour also of the code examples and a few other places. It clearly makes the layout a bit less nice, but if there's no other way to unbreak building the team manual without dirty hacks, then maybe this is worth it. I've added @thijskh as reviewer, as I'm pretty sure he was the one who creates this layout.

Yes, he had created the layout. One fix I found which is a dirty hack is altering the generated `.tex` file. That would keep the color for the code examples.

meisterT:
While I liked the previous layout more, I do think this is acceptable to fix the issue.

vmcj:
@thijskh is already working on a proper fix, for now I'll merge this one so this works again.

Issue #2179, opened 2023-10-14
[contest.yaml] Special fields for primary datasource when shadowing


vmcj:
It would be nice if we can add the settings for the external contest source in the contest.yaml as there is now only a one to one relation. 

This would make setting up the analyst instance a lot simpler.

PR #2208, opened 2023-11-05
Allow setting external contest source with contest.yaml

vmcj:
This makes it easier to setup the analyst instance during the finals.

I've done it so that we only overwrite the fields if explicit set so it is possible to remove the password from the contest and not remove it when someone updates the contest again to alter the time for example.

Closes: https://github.com/DOMjudge/domjudge/issues/2179

### DIFF
--- a/doc/manual/shadow.rst
+++ b/doc/manual/shadow.rst
@@ -54,6 +54,20 @@ In the DOMjudge admin interface, go to the *External Contest Sources* page and
 create an external contest source. Select the contest to import into and enter
 the source you want to import from.
 
+Another alternative is to use these DOMjudge only keys in the `contest.yaml`::
+
+  shadow:
+    type: contest-api
+    source: http(s)://url_to_external_system/api/contests/<contest_id>
+    username: user_in_external_system
+    password: password_for_user
+
+or::
+
+  shadow:
+    type: contest-archive
+    source: /path/on/domjudge/filestem/to/clics/compliant/archive
+
 Running the event feed import command
 -------------------------------------
 
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -5,6 +5,7 @@ namespace App\Service;
 use App\Entity\Configuration;
 use App\Entity\Contest;
 use App\Entity\ContestProblem;
+use App\Entity\ExternalContestSource;
 use App\Entity\Problem;
 use App\Entity\Role;
 use App\Entity\Team;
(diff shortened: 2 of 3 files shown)


webapp/src/Service/ImportExportService.php
```diff
@@ -251,6 +252,23 @@ public function importContestData(mixed $data, ?string &$errorMessage = null, st
         }
 
         $this->em->persist($contest);
+
+        $shadow_type = $data['shadow_type'] ?? null;
```
nickygerritsen, line 256: We should document that you can do this whole thing

vmcj: Makes sense, although the only users using this are us IMO.

nickygerritsen: Which I agree with but I don't want to check the code to see how the fields are called. Also, maybe we use a subkey?
```yaml
shadow:
  type: xxx
```
etc?

Also makes the code more readable.

doc/manual/shadow.rst
```diff
@@ -54,6 +54,20 @@ In the DOMjudge admin interface, go to the *External Contest Sources* page and
...
+  shadow:
+    type: contest-api
+    source: http(s)://url_to_external_system/api/contests/<contest_id>
+    username: user_in_external_system
+    password: password_for_user
+
+or::
+
+  shadow:
+    type: contest-api
```
nickygerritsen, line 68:
```suggestion
    type: contest-archive
```

right?

webapp/src/Service/ImportExportService.php
```diff
@@ -251,6 +252,22 @@ public function importContestData(mixed $data, ?string &$errorMessage = null, st
...
+            $externalSource = $this->em->getRepository(ExternalContestSource::class)->findOneBy(['contest' => $contest]) ?: new ExternalContestSource();
+            $externalSource->setContest($contest);
+            foreach (['source', 'username', 'password', 'type'] as $field) {
+                // Overwrite the existing value if the property is defined in the data: $externalSource-setSource($data['shadow']['source'])
+                $fieldFunc = 'set'.ucwords($field);
+                $fieldArgs = [$shadow[$field]];
+                if (isset($shadow[$field])) {
+                    $externalSource->$fieldFunc(...$fieldArgs);
+                }
+            }
```
nickygerritsen, line 267:
```suggestion
            foreach ($shadow as $field => $value) {
                $fieldFunc = 'set'.ucwords($field);
                $fieldArgs = [$value];
                if (method_exists($externalSource, $fieldFunc)) {
                    $externalSource->$fieldFunc(...$fieldArgs);
                }
            }
```

would now also work right? That way if we ever add fields it just works ™️

nickygerritsen (approved):

doc/manual/shadow.rst
vmcj: yes, wrong copy paste.

PR #2209, opened 2023-11-06
Check if a problem actually is used in contests when exporting

vmcj:
Probably there is a cleaner solution, but I think this does touch the underlying problem.


nickygerritsen:
Seems you added some commits from another PR

vmcj:
Yes, can you comment (or approve) on https://github.com/DOMjudge/domjudge/pull/2209/commits/c49b8914bfa3df80ff10916772b5c3b0b725798c and I'll remove the other ones tonight.

nickygerritsen (approved):

Issue #2059, opened 2023-05-27
Testcase in output/judgings/... directory is not easily matched with web UI


eldering:
In the work directory of a judging, we have a directory per testcase which is of the form `testcase00189`. However, the number in there does not match the testcase rank in the problem and there's no (easy?) way to correlate this number to the testcase in the jury UI.

We should change this so that debugging is easier.

PR #2211, opened 2023-11-08
Testcase judgehost naming

vmcj:
fix https://github.com/DOMjudge/domjudge/issues/2059

@eldering is this what you wanted in the issue? I'm not sure how/where to document the 2nd commit and if we want it.

### DIFF
--- a/webapp/src/Controller/Jury/ProblemController.php
+++ b/webapp/src/Controller/Jury/ProblemController.php
@@ -5,6 +5,7 @@ namespace App\Controller\Jury;
 use App\Controller\BaseController;
 use App\Entity\Contest;
 use App\Entity\ContestProblem;
+use App\Entity\Judgehost;
 use App\Entity\Judging;
 use App\Entity\Problem;
 use App\Entity\ProblemAttachment;
@@ -41,6 +42,7 @@ use Symfony\Component\Routing\Annotation\Route;
 use Symfony\Component\Security\Http\Attribute\IsGranted;
 use Symfony\Component\Yaml\Yaml;
 use ZipArchive;
+use Symfony\Component\DependencyInjection\Attribute\Autowire;
 
 #[IsGranted('ROLE_JURY')]
 #[Route(path: '/jury/problems')]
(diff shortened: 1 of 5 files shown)


eldering:
I think this solution is not so great for two reasons:
First, you now need to first go to the problem page to look up this

nickygerritsen (approved):

webapp/src/Controller/Jury/ProblemController.php
```diff
@@ -477,8 +478,18 @@ public function viewTextAction(int $probId): StreamedResponse
...
+    public function testcasesAction(
+        Request $request,
+        int $probId,
+        #[MapQueryParameter(name: 'testcase')]
+        ?int $testcaseFromRequest = null,
+    ): Response {
+        if ($testcaseFromRequest) {
+            $tc = $this->em->getRepository(Testcase::class)->find($testcaseFromRequest);
+            if ($tc) {
+                return $this->redirectToRoute('jury_problem_testcases', ['probId' => $tc->getProblem()->getProbId()]);
```
meisterT, line 490: I have the impression that this behavior can be very confusing. You are requesting one problem with a testcase from a different problem and are redirected?

Perhaps it becomes less confusing if we just add a separate route for it (in case it is useful)?

meisterT:
A screenshot would help for the second commit.

webapp/src/Controller/Jury/ProblemController.php
vmcj: So currently it is undocumented behaviour, and I think there are moments where you might want to know what testcase belongs with which problem (there is on the judgehost a system problem and you want to findout what kind of behaviour that problem has (loads of IO, loads of memory etc.)

I'm fine with just removing the commit, or just making a new route.

webapp/templates/jury/problem_testcases.html.twig
```diff
@@ -62,6 +63,7 @@
                                title="Move testcase down"><i class="fas fa-arrow-down"></i></a>
                         {% endif %}
                     </td>
+                    <td rowspan="2">{{ "%05d"|format(testcase.testcaseid) }}</td>
```
eldering, line 66: It would be nice to add a more complete path if it fits in the UI.

vmcj:
> A screenshot would help for the second commit.

image

And when you hover you get an educated guess for the absolute path. Educated as you can configure something else on the judgehost versus what you did on the domserver.

eldering:
I think #2324 is a better solution: both in the place/way the information is displayed and also by taking the actual directory from the judgehost instead of depending on configure/build time variables.

PR #2212, opened 2023-11-10
LGTM has stopped, is now incorporated in Github CodeQL

vmcj:
Should we also add badges for the supported versions, if people actually look at those badges they might prefer to know the version they run is reasonable error free.


nickygerritsen (approved):

eldering:
> Should we also add badges for the supported versions, if people actually look at those badges they might prefer to know the version they run is reasonable error free.

I think this is more for developers, so I don't think we need to change it there.

PR #2216, opened 2023-11-12
Make certain parts of the scoreboard sticky, relevant when having many problems

vmcj:
I encountered this when testing for something else. When there are many problems we don't fit in the viewport anymore (horizontal we move out of a browser window). If the viewer would scroll certain parts should stay:
- banner
- warnings (this is not a real contest), frozen status (contest frozen with 120 min remaining) etc.
- The legenda & filter button, the "generated on time"

I couldn't get the team names to become sticky as that would remove the border (https://stackoverflow.com/a/53559396).


vmcj:
The progress bar is not sticky yet.

example_problems/problems.yaml
```diff
@@ -1,15 +1,151 @@
-- id: hello
-  label: A
+- id: hello0-
```
meisterT, line 1: need to revert this file

gitlab/webstandard.sh
```diff
@@ -80,6 +80,10 @@ if [ "$TEST" = "w3cval" ]; then
     rm -f localhost/domjudge/css/select2-bootstrap*.css*
     rm -f localhost/domjudge/css/dataTables*.css*
     rm -f localhost/domjudge/jury/config/check/phpinfo*
+    # We need any value for bs-gutter-x to calculate with in CSS
```
meisterT, line 83: I don't understand what is happening here. Do you remove this from the CSS file to avoid it being flagged as non-standard? Is there a risk that this doesn't work in "all" browsers then?

vmcj: Given that there are at this point only 2 browsers and it works there I think we're fine, also bootstrap already uses CSS variables so we would have problems somewhere else also.

The main problem is that W3C has the CSS standard where this is a very strong candidate (not sure how this works for them to make this final) but the validator doesn't detect this properly. There are a lot of threads about this that this is more an issue with the W3C validator than with support for this feature.

meisterT:
What is the difference between this and https://github.com/DOMjudge/domjudge/pull/2132?

gitlab/webstandard.sh
meisterT: Did you test this also in the mobile view?

vmcj:
> What is the difference between this and #2132?

Everything, this does something completely else. In this one I make the overview (well tried, the team page seems to have been forgotten) and make that you can scroll when having loads of problems (so the scoreboard is quite wide)

The other one tries to keep the scoreboard summary (so based on the length) sticky when scrolling. So one works on horizontal scrolling (in a sort of edgecase) the other one when having multiple sortorders or enough teams to fill more than 1 viewport.

PR #2219, opened 2023-11-18
Disallow extensions with dots, Extension filter checks full extension

### LINKED ISSUE #2218: `doc` file can be submitted as a code

`doc` file can be submitted as code. It seems the file bypasses the file extension check.

### Your environment

- DOMjudge 8.2.2
- Ubuntu 20.04
- Nginx server

### Steps to reproduce

[A.zip](https://github.com/DOMjudge/domjudge/files/13400004/A.zip)

Unzip the archive first, and submit the `A.doc`  to any problem. Language setting sets as default.

setting

### Expected behaviour

expect

### Actual behaviour

actual

backend

ItsNiklas:
# Description

If the option `Filter files passed to compiler by extension list` is enabled for a language, only the filename ending with the extension will be checked. This may work for most languages, but can be annoying for e.g. the C language. Files like `submission.doc` can get through just because they end with a `c`. This is not the intended behavior of this option. Disabling this option allows the team to fully choose the language.

Honestly, I cannot think of any other case except maybe `pyc`. But I still think this is a small oversight in the code.

Fixes #2218.

Tested on my local development instance. The behaviour described in the issue is now as expected and I couldn't produce any other new bugs.

## Bug fix

Check the extension including the dot in the extension loop. This forces the file to end with exactly `.c`. The `ltrim` allows the check to complete even if the extensions are accidentally entered with a dot when creating a new language.

## Breaking

If `filterCompilerFiles` is on:
- We now assume that the filename contains a dot. I think this is acceptable since this is a contest control system, and this is the expected behavior when this option is enabled by an admin.
  - Maybe change the `FILENAME_REGEX` to capture this earlier...

### DIFF
--- a/webapp/src/Service/SubmissionService.php
+++ b/webapp/src/Service/SubmissionService.php
@@ -530,8 +530,9 @@ class SubmissionService
             if ($source !== 'shadowing' && $language->getFilterCompilerFiles()) {
                 $matchesExtension = false;
                 foreach ($language->getExtensions() as $extension) {
-                    $extensionLength = strlen($extension);
-                    if (substr($file->getClientOriginalName(), -$extensionLength) === $extension) {
+                    $extensionWithDot = '.' . ltrim($extension, '.');
+                    $extensionLength = strlen($extensionWithDot);
+                    if (substr($file->getClientOriginalName(), -$extensionLength) === $extensionWithDot) {
                         $matchesExtension = true;
                         break;
                     }
--- a/webapp/src/Service/SubmissionService.php
+++ b/webapp/src/Service/SubmissionService.php
@@ -530,9 +530,7 @@ class SubmissionService
             if ($source !== 'shadowing' && $language->getFilterCompilerFiles()) {
                 $matchesExtension = false;
                 foreach ($language->getExtensions() as $extension) {
-                    $extensionWithDot = '.' . $extension;
-                    $extensionLength = strlen($extensionWithDot);
-                    if (substr($file->getClientOriginalName(), -$extensionLength) === $extensionWithDot) {
+                    if (str_ends_with($file->getClientOriginalName(), '.' . $extension)) {
                         $matchesExtension = true;
                         break;
                     }
--- a/webapp/src/Service/SubmissionService.php
+++ b/webapp/src/Service/SubmissionService.php
@@ -530,7 +530,7 @@ class SubmissionService
             if ($source !== 'shadowing' && $language->getFilterCompilerFiles()) {
                 $matchesExtension = false;
                 foreach ($language->getExtensions() as $extension) {
-                    $extensionWithDot = '.' . ltrim($extension, '.');
+                    $extensionWithDot = '.' . $extension;
                     $extensionLength = strlen($extensionWithDot);
                     if (substr($file->getClientOriginalName(), -$extensionLength) === $extensionWithDot) {
                         $matchesExtension = true;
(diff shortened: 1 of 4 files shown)


vmcj (approved):

vmcj:
Hi @ItsNiklas, thank you for the fix. I'll merge this after one of the others also approves (which will most likely be at/after Friday).

webapp/src/Service/SubmissionService.php
```diff
@@ -530,8 +530,9 @@ public function submitSolution(
             if ($source !== 'shadowing' && $language->getFilterCompilerFiles()) {
                 $matchesExtension = false;
                 foreach ($language->getExtensions() as $extension) {
-                    $extensionLength = strlen($extension);
-                    if (substr($file->getClientOriginalName(), -$extensionLength) === $extension) {
+                    $extensionWithDot = '.' . ltrim($extension, '.');
```
thijskh, line 533: Why is the ltrim needed?

ItsNiklas: I was scared of this change backfiring.

Let's say an end user has added a new language, say `BQN`. He adds the extension `.bqn` to the extension list (accidentally with the dot) and turns on the filter. This works fine with the old code, and teams submitting BQN files get through.

The server gets updated and includes this commit. Now when users try to submit for this Language, they get the error
```None of the submitted files match... (allowed: .bqn)```.
Which does not help at all.

`extensionWithDot` would now contain two dots, which does not match any filename. The `ltrim` ensures that the variable always starts with only one dot and avoids weird errors.

What do you think? Should this be clarified/handled elsewhere?

Comment (thijskh) [reply webapp/src/Service/SubmissionService.php:533]: Since there is several pieces of code that work with the configured filename extensions, I think I would prefer to make the source format of our extensions simple/predictable, instead of handling it with or without `.` on all code that interprets it.

Although it does not explicitly forbid it, the [CLICS specification](https://ccs-specs.icpc.io/2023-06/contest_api#languages) only provides extensions bare, i.e. without `.`.

So it's probably best, in my opinion, to start forbidding adding `.` in the admin interface for a language, and assume in this code it will not already be present.

Comment (ItsNiklas) [reply webapp/src/Service/SubmissionService.php]: Good point.

**Comment by ItsNiklas** [comment]
I added the `Assert` to the Language entity / form.

image

Had to fix the code that created the delete button because it overode the error message. I added a Doctrine migration because the errors were pretty bad after the entity changed. Hopefully it is correct that way, I wasn't sure.

Thanks

File: webapp/src/Service/SubmissionService.php [hunk 2]
```diff
@@ -530,8 +530,9 @@ public function submitSolution(
             if ($source !== 'shadowing' && $language->getFilterCompilerFiles()) {
                 $matchesExtension = false;
                 foreach ($language->getExtensions() as $extension) {
-                    $extensionLength = strlen($extension);
-                    if (substr($file->getClientOriginalName(), -$extensionLength) === $extension) {
+                    $extensionWithDot = '.' . $extension;
+                    $extensionLength = strlen($extensionWithDot);
+                    if (substr($file->getClientOriginalName(), -$extensionLength) === $extensionWithDot) {
```
meisterT, line 535: nit: use `str_ends_with` here to make the intent more clear?

ItsNiklas: Makes sense, I guess it wasn't used before since the function is new in PHP 8.0.

Issue #2086, opened 2023-06-21
default `configure` does not install service files for judgedaemon, docs do mention those.


vmcj:
### Description of the problem
When you follow the manual we mention the service files so you can start a judgehost with a service. The problem is that we don't advice the `fhs` flag which does set this.

We can fix this in multiple ways:
- Make the FHS default
- Don't respect the prefix and just install those service files
- Mention this explicit in the docs
- Mention this in the steps after the `make install-judgehost` that someone should do this (and template the files in the `make judgehost` step)

In my opinion we should go for either option {3,4} or discuss {1}.

@eldering what are your thoughts?

eldering:
I think we can also (try to) install the service files when not in FHS mode, although that might fail if the user runs `make judgehost-install` without root.
I don't understand though why `systemd_unitdir` isn't set at https://github.com/DOMjudge/domjudge/blob/main/judge/Makefile#L36 from https://github.com/DOMjudge/domjudge/blob/main/configure.ac when FHS is not enabled.

vmcj:
> I think we can also (try to) install the service files when not in FHS mode, although that might fail if the user runs `make judgehost-install` without root. I don't understand though why `systemd_unitdir` isn't set at https://github.com/DOMjudge/domjudge/blob/main/judge/Makefile#L36 from https://github.com/DOMjudge/domjudge/blob/main/configure.ac when FHS is not enabled.

@eldering I think we discussed and agreed that if you **don't** do FHS mode we (should and ) will **not** install service files anymore.

So that requires a change in de documentation.

eldering:
Indeed, I'd propose to *never* install systemd unit files, in the same way that we don't install the apache or nginx config in `/etc/`, but just document how you can symlink/copy them. That will remove a good bit of complexity from our configure/makefile setup.

PR #2220, opened 2023-11-20
Fail when pkg-config can't find the systemd lib directory

vmcj:
I think there is a bit of refactoring which can be done, it feels like we test for this twice.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -225,8 +225,9 @@ AS_IF([test "x$with_systemdsystemunitdir" = "xyes" -o "x$with_systemdsystemunitd
      AS_IF([test "x$PKG_CONFIG" = "x"],AC_MSG_ERROR([systemd support requested but no pkg-config available to query systemd package]))
      def_systemdsystemunitdir=$($PKG_CONFIG --variable=systemdsystemunitdir systemd)
 
+     # TODO: This seems to fail when systemd is not installed so pkg-config can't find the variable, we don't error out there
      AS_IF([test "x$def_systemdsystemunitdir" = "x"],
-   [AS_IF([test "x$with_systemdsystemunitdir" = "xyes"],
+   [AS_IF([test "x$with_systemdsystemunitdir" = "xyes" -o "x$with_systemdsystemunitdir" = "xauto"],
     [AC_MSG_ERROR([systemd support requested but pkg-config unable to query systemd package])])
     with_systemdsystemunitdir=no],
    [with_systemdsystemunitdir="$def_systemdsystemunitdir"])])


configure.ac
```diff
@@ -225,8 +225,9 @@ AS_IF([test "x$with_systemdsystemunitdir" = "xyes" -o "x$with_systemdsystemunitd
      AS_IF([test "x$PKG_CONFIG" = "x"],AC_MSG_ERROR([systemd support requested but no pkg-config available to query systemd package]))
      def_systemdsystemunitdir=$($PKG_CONFIG --variable=systemdsystemunitdir systemd)
 
+     # TODO: This seems to fail when systemd is not installed so pkg-config can't find the variable, we don't error out there
```
eldering, line 228: This comment is not really clear to me: are we currently not erroring out or should we not be?

configure.ac
```diff
@@ -225,8 +225,9 @@ AS_IF([test "x$with_systemdsystemunitdir" = "xyes" -o "x$with_systemdsystemunitd
...
      AS_IF([test "x$def_systemdsystemunitdir" = "x"],
-   [AS_IF([test "x$with_systemdsystemunitdir" = "xyes"],
+   [AS_IF([test "x$with_systemdsystemunitdir" = "xyes" -o "x$with_systemdsystemunitdir" = "xauto"],
```
eldering, line 230: How is the value `yes` now different from `auto`?

configure.ac
vmcj: The comment can be removed as I tried to fix this with this PR.

In the case that `systemd` is not installed but `pkg-config` is:
- pkg-config will try to get the value here: https://github.com/DOMjudge/domjudge/pull/2220/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R226 but can't as systemd is not there, so not package specification can give the variable `systemdsystemunitdir`.
- We should fail, but we don't. I think because an empty string is stored and we don't fail on the non-zero exit code.

Next in the code:
`AS_IF([test "x$def_systemdsystemunitdir" = "x"]` will be true as the variable is empty. I copied the solution from here: https://github.com/DOMjudge/domjudge/pull/2220/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R224 but I'm not 100% if we can't rewrite this to make more sense, I think the auto is set here https://github.com/DOMjudge/domjudge/pull/2220/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R224 but I'm not sure where the case for `with_systemdsystemunitdir=yes` would be set.

configure.ac
vmcj: I have no idea, I copied this check from earlier and I think the `test "x$with_systemdsystemunitdir" = "xyes` might need to be removed?

eldering:
Closed in favor of fixing it along the lines of https://github.com/DOMjudge/domjudge/issues/2086

Issue #2072, opened 2023-06-02
Allow relative (de)activate_time when importing contest.yaml


mpsijm:
Follow-up of #1807.

### Description of the enhancement request
In `contest.yaml`, `activate_time` and `deactivate_time` now need to be full-blown ISO strings, but I would like to use relative times as well.

### Details
When creating a contest via the web interface, the "Activate time" and "Deactivate time" are allowed to be relative times. However, in `contest.yaml`, these fields are required to be in ISO format: https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Service/ImportExportService.php#L181-L195

And perhaps on a similar note, the `duration` field currently only allows relative times, whereas `scoreboard_freeze_duration` only allows relative times but has an alternative called `scoreboard-freeze`/`freeze` that only allows absolute timestamps. It would be nice if all of these are consistent, although perhaps it's a bad idea to deviate too much from https://ccs-specs.icpc.io/draft/contest_api#contests :stuck_out_tongue:

nickygerritsen:
The spec is indeed the reason these are only allowed to be the way they are. I would not be opposed to add support for this though. The fields without _ are legacy and I think we should not change them

PR #2221, opened 2023-11-22
Import relative contest times

vmcj:
This targets https://github.com/DOMjudge/domjudge/issues/2072

This has one slight flaw, the relative time is changed to an absolute time and stored as such. I'm not sure if there is a proper fix for this.

Also side-effect to test if the CI now works again.

### DIFF
--- a/webapp/src/Entity/Contest.php
+++ b/webapp/src/Entity/Contest.php
@@ -564,6 +564,7 @@ class Contest extends BaseApiEntity implements AssetEntityInterface
     {
         $this->starttimeString = $starttimeString;
 
+        $this->starttime = $this->getAbsoluteTime($starttimeString);
         $this->setActivatetimeString($this->getActivatetimeString());
         $this->setFreezetimeString($this->getFreezetimeString());
         $this->setEndtimeString($this->getEndtimeString());
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -124,6 +124,7 @@ class ImportExportService
             if ($contest) {
                 $timeValue = Utils::abstime($contest->getAbsoluteTime($timeValue));
             }
+            /** @var DateTime|bool $time */
             $time = new DateTime($timeValue);
         } else {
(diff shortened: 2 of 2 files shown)


webapp/src/Entity/Contest.php
```diff
@@ -564,6 +564,7 @@ public function setStarttimeString(string $starttimeString): Contest
     {
         $this->starttimeString = $starttimeString;
 
+        $this->starttime = $this->getAbsoluteTime($starttimeString);
```
eldering, line 567: If this wasn't here to begin with, I'm wondering if there was maybe a good reason for this.

eldering: I think in general if we want the same/similar logic for the UI and API, then we should rethink where and how parsing the input is handled in a common place. It feels that this is duplicating code in the import service that's already existing in other places.

vmcj: > If this wasn't here to begin with, I'm wondering if there was maybe a good reason for this.

I had the same question, so I was planning to not merge this PR before having discussed this at the hackathon.


> I think in general if we want the same/similar logic for the UI and API, then we should rethink where and how parsing the input is handled in a common place. It feels that this is duplicating code in the import service that's already existing in other places.

I agree, but than I prefer to move this one for after the release as mentioned in the commit message.

meisterT:
> This has one slight flaw, the relative time is changed to an absolute time and stored as such. I'm not sure if there is a proper fix for this.

That we store the time string as is (i.e. as relative time) and present it back to the user is intentional. If we move the contest start, and the user entered relative times, the respective other times need to move as well.

vmcj:
> > This has one slight flaw, the relative time is changed to an absolute time and stored as such. I'm not sure if there is a proper fix for this.
> 
> That we store the time string as is (i.e. as relative time) and present it back to the user is intentional. If we move the contest start, and the user entered relative times, the respective other times need to move as well.

Than this shouldn't be merged and requires some/loads of rework.

vmcj:
Not possible in current format.

Issue #1785, opened 2022-11-09
Inconsistency of clarification queue asssignment


eldering:
At the WF Dhaka dress rehearsal, this was reported by the judges:
![20221109_124110](https://user-images.githubusercontent.com/4692022/200758509-88da2a3f-0cfc-48d4-9393-6f84a3d74d20.jpg)
This showed as assigned to the jury queue on the clarifications overview page. After reassigning the original clarification request to the "Silly stuff" queue, it showed up there as (initially) expected.

We should probably only have a single queue for any clarification thread, or somehow make sure that the UX here is consistent.

thijskh:
More or less a duplicate/specific case of https://github.com/DOMjudge/domjudge/issues/453#issuecomment-447668438

nickygerritsen:
Proposal:
* Hide the edit button ever where except the top clarification
* Make that edit button edit all of them

vmcj:
> Proposal:
> 
>     * Hide the edit button ever where except the top clarification
> 
Most of the time a non-issue, but I prefer that for a long thread you can re-assign based on the last message, so I would prefer to have the button on the first and the last (or even only the last as most of the time that will also be the first...) it will save scrolling.
>     * Make that edit button edit all of them

Yes,

PR #2224, opened 2023-11-23
Change all clarifications in thread

vmcj:
By always finding the first in thread and changing the replies afterwards. Verified that this also works via the overview clarification page.

Partial fix for: https://github.com/DOMjudge/domjudge/issues/1785

### DIFF
--- a/webapp/src/Controller/Jury/ClarificationController.php
+++ b/webapp/src/Controller/Jury/ClarificationController.php
@@ -352,7 +352,13 @@ class ClarificationController extends AbstractController
         if ($queue === "") {
             $queue = null;
         }
-        $clarification->setQueue($queue);
+
+        // Either this is the first in a thread (of 1 or more) or this 2nd or more.
+        $curClarification = $clarification->getInReplyTo() ?? $clarification;
+        foreach ($curClarification->getReplies() as $reply) {
+            $reply->setQueue($queue);
+        }
+        $curClarification->setQueue($queue);
         $this->em->flush();
 
         if ($request->isXmlHttpRequest()) {


webapp/src/Controller/Jury/ClarificationController.php
```diff
@@ -352,7 +352,17 @@ public function changeQueueAction(Request $request, int $clarId): Response
...
             $queue = null;
         }
-        $clarification->setQueue($queue);
+
+        // Either this is the first in a thread (of 1 or more) or this 2nd or more.
+        $firstClarification = $clarification->getInReplyTo();
+        $curClarification = $clarification;
+        if ($firstClarification) {
+            $curClarification = $firstClarification;
+        }
```
nickygerritsen, line 361:
```suggestion
        $curClarification = $clarification->getInReplyTo() ?? $clarification;
```

Or maybe `?:` if it doesn't return `null` but `false`

vmcj: `public function getInReplyTo(): ?Clarification`

So `??` should be good?

thijskh (approved):

webapp/src/Controller/Jury/ClarificationController.php
```diff
@@ -352,7 +352,13 @@ public function changeQueueAction(Request $request, int $clarId): Response
         if ($queue === "") {
             $queue = null;
         }
-        $clarification->setQueue($queue);
+
+        // Either this is the first in a thread (of 1 or more) or this 2nd or more.
```
eldering, line 356: This comment is not really clear to me: `or this 2nd or more.`

vmcj: Removed the comment as Nicky his fix made this easy enough to read in code.

Issue #2191, opened 2023-10-28
Cannot view inactive judging on jury submission page


eldering:
### Description of the problem
When I browse to a submission on the jury interface that has multiple judgings, then I cannot view an inactive judging.

### Your environment
Current `main` branch, both at BAPC and my laptop.

### Actual behaviour
See below screenshot for URL http://localhost/domjudge/jury/submissions/106?jid=106, but it still shows the currently active judging 107.
![image](https://github.com/DOMjudge/domjudge/assets/4692022/0e03dca6-743a-4b6b-b4fb-2505b66fe172)

meisterT:
I cannot reproduce this behavior, neither in Chrome nor firefox. Did you type in the URL or did you click on the respective row in the judgings table?

meisterT:
While trying to provoke the issue, I found another one:
![image](https://github.com/DOMjudge/domjudge/assets/418721/64617e94-4abb-4742-bdf5-d77ada0ffcc5)

Note that the order in this table should be by judgingid, see https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Controller/Jury/SubmissionController.php#L221

I don't see where we would reorder things though.

meisterT:
I think this is fixed, feel free to re-open if you can still repro.

PR #2228, opened 2023-11-23
Improve y-axis of the submission graph format.

meisterT:
Found when looking at https://github.com/DOMjudge/domjudge/issues/2191

This changes the following things:
- Reduce amount of ticks for submissions with high runtimes
- Use integer ticks when higher precision only clutters the y-axis
- Add `s` to the labels to clearly indicate the unit

Examples:
image

image

### DIFF
--- a/webapp/templates/jury/partials/submission_graph.html.twig
+++ b/webapp/templates/jury/partials/submission_graph.html.twig
@@ -32,13 +32,14 @@
         'output-limit':   'black',
     } %}
     function create_chart(data, maxY) {
-        var tickStep = 1;
+        var tickStep = Math.floor(maxY/5);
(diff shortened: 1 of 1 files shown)


webapp/templates/jury/partials/submission_graph.html.twig
```diff
@@ -37,8 +37,11 @@
             tickStep = 0.2;
         } else if (maxY <= 5) {
             tickStep = 0.5;
+        } else {
+            tickStep = Math.floor(maxY/5);
```
nickygerritsen, line 41: This could go on line 35 or we remove the `= 1` there.

nickygerritsen (approved):

PR #2238, opened 2023-11-24
Store compiler/runner version history.

meisterT:
Previously, we only stored the most recent one per language/judgehost combination. This new format will allow us (in separate changes) to cross-reference in judgings and warn if they have not been using the canonical one.

Progress on https://github.com/DOMjudge/domjudge/issues/1241

### DIFF
--- a/webapp/migrations/Version20231124122006.php
+++ b/webapp/migrations/Version20231124122006.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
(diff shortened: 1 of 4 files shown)


webapp/migrations/Version20231124122006.php
```diff
@@ -0,0 +1,48 @@
...
+{
+    public function getDescription(): string
+    {
+        return 'Store a history of versions, instead of just the most recent one.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE queuetask DROP FOREIGN KEY FK_45E85FF85D5FEA72');
```
nickygerritsen, line 23: The migration added too much

nickygerritsen (approved):

PR #2240, opened 2023-11-24
Clarify non-threadedness of clarifications in code comment

eldering:


webapp/src/Controller/Jury/ClarificationController.php
```diff
@@ -353,7 +353,9 @@ public function changeQueueAction(Request $request, int $clarId): Response
             $queue = null;
         }
 
-        // Either this is the first in a thread (of 1 or more) or this 2nd or more.
+        // Find the original clarification if this is a reply, and then update
```
meisterT, line 356:
```suggestion
        // Find the original clarification: if this is a reply, and then update
```

Issue #2069, opened 2023-05-30
Download Contest Statements 


lperovskaya:
### Provide contestants with a single PDF file with contest statements in addition to existing ability to download each problem statement separately
Every contestant has to open each of problem statement pages at the beginning of the contest, or, if they have a printer access, send multiple print tasks to get a printed version of problem statements. Would be nice to have a single place to download a single pdf file with all problems in them.

### Expected behaviour
As soon as the contest starts the top of the list of problems page should have a way ( a separate button?) to download a concatenated PDF file with all problem statements

nickygerritsen:
I was thinking of having a contest.pdf you could upload on a contest, similar to the banner

nickygerritsen:
So after discussing this we have some questions / thoughts we first need to decide on:

* How to call the table / entity? `ContestStatement`?
* Do we allow more than PDF just like problem texts?
* Where do we put download button(s) for it? On the jury contest page and the team/public problemset page? But then where/how?
* Do we put this PDF in any of the ZIPs? Do we need a new ZIP?

Before we implement this, I think we need to answer these.

mpsijm:
Regarding the zip (the last question): BAPCtools allows exporting a `samples.zip` with the following structure, which we extract in the contestant PCs' home directories before the start of the contest:
```
- A
  - 1.in
  - 1.ans
  - ...
- B
  - ...
- ...
- contest.pdf
- contest-web.pdf
```
If you end up implementing a zip button that downloads all samples and the contest PDF, it would be nice if it would follow the same structure :slightly_smiling_face:

nickygerritsen:
Thanks that makes sense

PR #2241, opened 2023-11-24
Split the problem text content from the problem in the database.

nickygerritsen:
This is to prepare for the next Doctrine release, which [doesn't allow partial queries anymore](https://www.doctrine-project.org/projects/doctrine-orm/en/2.17/reference/partial-objects.html). This is also to make it consistent with other blobs in the database like submission files and problem attachments.

This is preparation for #2069 where I will do a similar thing.

Besides this there is one more partial query, which I will remove in a separate PR.

I know changing database stuff just for the ORM is not something to do lightly, but I don't think we have any better way. And this actually improves queries in the places where we used partial queries before IMHO.

### DIFF
--- a/webapp/migrations/Version20231124133426.php
+++ b/webapp/migrations/Version20231124133426.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20231124133426 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Split problem text content from problem';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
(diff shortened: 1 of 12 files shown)


webapp/migrations/Version20231124133426.php
```diff
@@ -0,0 +1,42 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20231124133426 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
```
vmcj, line 17: Can you describe what is happening?

webapp/src/Entity/Problem.php
```diff
@@ -259,39 +264,30 @@ public function getCombinedRunCompare(): bool
...
         $this->clearProblemtext = $clearProblemtext;
-        $this->problemtext = null;
+        $this->setProblemTextContent(null);
 
         return $this;
     }
 
     /**
-     * @return resource|string
+     * @return string|null
```
vmcj, line 286:
```suggestion
```

vmcj (changes requested):
I'm not the target audience for this one

nickygerritsen:
I propose to merge this after the current contest 'season', unless others think we should do it before so we can test it?

vmcj (approved):

PR #2242, opened 2023-11-24
Team manual back2tex

thijskh:


eldering:
Some general comments that not necessarily should be done in this PR:

- It would be nice if `doc/team/gentexconfig.in` would fetch some configuration defaults from `etc/db-config.yaml` or if the system is already installed from a running DOMjudge API.
- I'd make the only `make` targets under `doc/` to be `dist`, to pre-generate documentation to be shipped in the tarball, no targets that are meant to be run by the admin user installing DOMjudge. I think this makes things clearer and also simplifies the Makefiles.
- For the team manual, we should just provide a script (basically `doc/team/genteammanual`) that an admin user can run to regenerate the team manual with updated text and/or configuration. But this should be completely outside the Makefile structure.

vmcj:
@imylene13 is willing to look into this, do we have a specifc reason why we also include the teammanual in the admin manual?

PR #2256, opened 2023-11-25
Remove display of output diff to teams.

meisterT:
This was shown when the configuration option to display sample output was selected.
However, there is a risk that this leaks some information to the teams, since this information comes from `judgemessage.txt`.

We still display the content from `teammessage.txt`, in case a validator writes one.

### DIFF
--- a/webapp/templates/team/partials/submission.html.twig
+++ b/webapp/templates/team/partials/submission.html.twig
@@ -134,14 +134,6 @@
                                 <p class="nodata">There was no program output.</p>
                             {% endif %}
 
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

tuupke (approved):

eldering:
Note also that according to the [problem package specification](https://www.kattis.com/problem-package-format/spec/2023-07-draft.html#reporting-additional-feedback), this should not be shown to teams.

eldering (approved):

PR #2260, opened 2023-11-26
Set HTTPS fastcgi param correctly for http

### LINKED ISSUE #2177: Unwanted redirect from HTTP to HTTPS on login

Kevinjil:
Traefik also sets the 'X-Forwarded-Proto' header with regular HTTP traffic. As the 'HTTP' value is not empty, the 'HTTPS' fastcgi parameter was set and requests were incorrectly upgraded to 'HTTPS'.

Fixes #2177


vmcj (approved):
@tuupke @nickygerritsen can one of you take a look?

@Kevinjil can you also add a traefik config to the domjudge-packaging repo?

nickygerritsen (approved):

PR #2264, opened 2023-11-26
Use new PHP styleguide over PSR-2

vmcj:
The original intent was to pick a styleguide, we already follow it but this makes this explicit.


CODINGSTYLE.md
```diff
@@ -1,4 +1,4 @@
 DOMjudge project coding style guideline
 =======================================
 
-We follow the [PSR-2 coding style](https://www.php-fig.org/psr/psr-2/).
+We follow the [PSR-2 coding style](https://www.php-fig.org/psr/psr-12/).
```
vmcj, line 4:
```suggestion
We follow the [PSR-12 coding style](https://www.php-fig.org/psr/psr-12/).
```

nickygerritsen (approved):

PR #2267, opened 2023-11-30
Update installation documentation for Fedora.

mkfuron:
Update installation documentation to reflect experience installing DOMjudge on Fedora 38.  

Follow-up  on Issue 2222.

### DIFF
--- a/doc/manual/install-judgehost.rst
+++ b/doc/manual/install-judgehost.rst
@@ -39,8 +39,8 @@ For Debian::
 
 For RHEL 7/Fedora [*]_::
 
-  sudo yum install make pkgconfig sudo libcgroup-devel lsof \
-        php-cli php-mbstring php-xml php-process procps-ng gcc g++
+  sudo dnf install make pkgconfig sudo libcgroup-devel lsof \
+        php-cli php-mbstring php-xml php-process procps-ng gcc g++ \
         glibc-static libstdc++-static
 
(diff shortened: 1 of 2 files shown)


doc/manual/install-judgehost.rst
```diff
@@ -39,8 +39,8 @@ For Debian::
 
 For RHEL 7/Fedora [*]_::
 
-  sudo yum install make pkgconfig sudo libcgroup-devel lsof \
-        php-cli php-mbstring php-xml php-process procps-ng gcc g++
+  sudo dnf install make pkgconfig sudo libcgroup-devel lsof \
```
vmcj, line 42: Renaming to `dnf` here would break RHEL7 but the people still using that should know how to change this.

vmcj (approved):
Let's see if one of the others has an opinion and otherwise I'll merge it next week.

doc/manual/install-judgehost.rst
mkfuron: I am fine with removing the explicit reference to RHEL here, or noting that this would only be for 8.

vmcj: No, I think this is fine. Any university or company using RHEL7 should have gone to RHEL8 now (https://www.redhat.com/en/blog/end-maintenance-red-hat-enterprise-linux-7-almost-here)

PR #2268, opened 2023-11-30
Add OCaml language

vmcj:


sql/files/defaultdata/ocaml/run
```diff
@@ -0,0 +1,17 @@
...
+
+# Compilation needs to store intermediate files, `/tmp` is not writable
+export TMPDIR="."
+
+# Enable some libraries
+ocamlopt unix.cmxa str.cmxa bigarray.cmxa -o "$DEST" "$@" -ccopt -pipe
+
+# Consider also setting for faster executables:
+# -match-context-rows 100
+# -unsafe
```
eldering, line 16: Nitpick: would be better to add this comment above the compiler command.

webapp/src/DataFixtures/DefaultData/LanguageFixture.php
```diff
@@ -23,30 +23,31 @@ public function __construct(
...
+            ['kt',     'kotlin',     'Kotlin',      ['kt'],                      true,  'Main class', false,  true,   1,     'kt',                'kotlinc -version',   'kotlin -version'],
+            ['pas',    'pascal',     'Pascal',      ['pas', 'p'],                false, 'Main file',  false,  true,   1,     'pas',               '',                   ''],
+            ['pl',     'pl',         'Perl',        ['pl'],                      false, 'Main file',  false,  true,   1,     'pl',                '',                   ''],
+            ['plg',    'prolog',     'Prolog',      ['plg'],                     false, 'Main file',  false,  true,   1,     'plg',               '',                   ''],
+            ['py3',    'python3',    'Python 3',    ['py'],                      false, 'Main file',  true,   true,   1,     'py3',               'pypy3 --version',    'pypy3 --version'],
+            ['ocaml',  'ocaml',      'OCaml',       ['ml'],                      false, 'Main file',  false,  true,   1,     'ocaml',             'ocamlopt --version', ''],
```
eldering, line 44: From the compile command it seems that Ocaml doesn't have any specific entry point, so I think this should be:
```suggestion
            ['ocaml',  'ocaml',      'OCaml',       ['ml'],                      false, null       ,  false,  true,   1,     'ocaml',             'ocamlopt --version', ''],
```

sql/files/defaultdata/ocaml/run
```diff
@@ -8,10 +8,12 @@ MEMLIMIT="$1" ; shift
...
-# Enable some libraries
-ocamlopt unix.cmxa str.cmxa bigarray.cmxa -o "$DEST" "$@" -ccopt -pipe
-
+# Enable some libraries;
+# unix system calls, regular expressions and string processing & multi arrays
+# See: https://v2.ocaml.org/releases/5.1/api/index.html
 # Consider also setting for faster executables:
 # -match-context-rows 100
 # -unsafe
+ocamlopt unix.cmxa str.cmxa bigarray.cmxa -o "$DEST" "$@" -ccopt -pipe
```
eldering, line 17: Is there a good reason to add specifically these libraries? Why (not) add others as well? I think it's better to go with language defaults unless there's clearly a general reason to add/change something. We can just document some options and local admins can always customize things.

vmcj: This is what SWERC24 uses, so I assumed this is the reasonable default. Also I prefer to have an example of how to add the extra libraries.

But if you really prefer I can remove this and add it as a comment.

eldering (approved):

nickygerritsen (approved):

PR #2270, opened 2023-12-02
Use Jetbrains linters for QA

vmcj:
We can do this for all languages Jetbrains supports, for now this only gives what one already gets if/when using PhpStorm.

This will directly upload issues to the PR and upload findings in the Security tab. We can even fail if new issues are found (or accepted in the baseline).

I've checked with @nickygerritsen and the findings seem to be new compared to PHPStan and also most of the time seem to make sense.

Same as https://github.com/DOMjudge/domjudge/pull/2263 but the branch needs to be in the organisation for this to work as this needs a secret, we can decide to allow this on forks with some GA CI magic but that would make it possible to extract the secret.

### DIFF
--- a/.github/jobs/data/codespellignorefiles.txt
+++ b/.github/jobs/data/codespellignorefiles.txt
@@ -26,3 +26,4 @@ nv.d3.min*
 composer*
 ./doc/logos
 ./m4
+qodana.sarif.json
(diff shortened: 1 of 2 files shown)


vmcj:
I'll first try to get the PHPstan baseline smaller/empty.

PR #2275, opened 2023-12-05
Update dependencies to Symfony 6.4.

nickygerritsen:
Mostly for CI. If it passes we should be able to merge without issues.


webapp/config/packages/framework.yaml
```diff
@@ -4,6 +4,8 @@ framework:
     esi: false
     fragments: false
     http_method_override: true
+    #csrf_protection: true
+    annotations: true
```
vmcj, line 8: Isn't that the default already (https://symfony.com/doc/current/reference/configuration/framework.html#annotations)

nickygerritsen: Yes but it will change so I’d rather be explicit

vmcj (approved):
CI is green so fine by me.

PR #2282, opened 2023-12-29
Keep track of potential first to solves so we can update them when we get 'delayed' results.

### LINKED ISSUE #2281: The first solved is not calculated for some case

DOMjudge is unable to calculate first solved data correctly in a specific case described below.

Recalculation (Refresh scoreboard cache from jury UI) can recalculates correctly, but it'd be great if DOMjudge can handle this case.

### Your environment
Any of DOMjudge 8.2.1, 8.2.2, 2cd74e91c35122ad1eae325520e573c71b11a032 (using docker-contributor image on Ubuntu 22.04)

### Steps to reproduce
If the following happens in this order for a specific problem, the first solved is not calculated:
1. Team A submits a solution.
2. Team B submits a solution.
3. Team B's solution is returned as accepted. At this point, Team A's solution is still under judging.
4. Team A's solution is returned as incorrect.

At the point of returning the result of Team B (step 3), the first solved cannot be determined.
However, once the Team A's submission is returned (at step 4), because there's no pending submission that was submitted before the Team B's submission, the Team B's submission is expected to be marked as the first solved. 

For local testing, it's easy to reproduce this by extending timelimit of a problem longer (e.g. 60 seconds) and submitting a TLE solution as a team A's one.

### Expected behaviour
Once the Team B's solution can be determined as the first solved, then the scoreboard shows it.

### Actual behaviour
The first solved is still N/A even after the first solved can be determined.

Requesting recalculation of the scoreboard fixes it.

nickygerritsen:
Fixes #2281.

I don't know why we never found this before, but I think this scenario is real and I see no other easy way to fix this.

I have split up the logic of determining first to solve in two:

* If there are no correct submissions before the one being considered, this *can* be first to solve, but there are two options:
  * If there are no pending ones either, it is for sure a first to solve.
  * If there **are** pending ones, this means this can become a first to solve later, but we are not sure, so keep track of this

Then when we update a score cache entry with 0 pending items, we need to consider that we need to check submissions from other teams for the same problem, to see if they have any potential first to solve submissions and recalculate them.

This brings with it a performance penalty every time we update the score cache with 0 pending submissions, since we need to query for the same contest and problem all score caches with a potential first to solve. There is currently no index on this column, but there are indices on the contest, team and problem columns, so I think this should not be a big overhead. If it turns it this is an issue, we can add an index on `cid, probid, is_potential_first_to_solve`. The next extra overhead is when there are actually pending first to solves, since then we need to recalculate them. But I think this makes sense, since this is exactly what we try to solve.

I first wrote the testcase, which fails without my fix. If you add `$scoreboardService->calculateScoreRow($contest, $team2, $problem);` below the second `$scoreboardService->calculateScoreRow($contest, $team1, $problem);` it does succeed without my change, since then we do indeed update the data.

### DIFF
--- a/webapp/migrations/Version20231229094359.php
+++ b/webapp/migrations/Version20231229094359.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20231229094359 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add potential first to solve to scorecache.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE scorecache ADD is_potential_first_to_solve TINYINT(1) DEFAULT 0 NOT NULL COMMENT \'Is this potentially the first solution to this problem?\' AFTER is_first_to_solve');
+    }
+
+    public function down(Schema $schema): void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE scorecache DROP is_potential_first_to_solve');
+    }
+
+    public function isTransactional(): bool
+    {
+        return false;
+    }
+}
--- a/webapp/src/Service/ScoreboardService.php
+++ b/webapp/src/Service/ScoreboardService.php
@@ -254,7 +254,8 @@ class ScoreboardService
         Contest $contest,
         Team    $team,
         Problem $problem,
-        bool    $updateRankCache = true
+        bool    $updateRankCache = true,
+        bool    $updatePotentialFirstToSolves = true
     ): void {
         $this->logger->debug(
             "ScoreboardService::calculateScoreRow '%d' '%d' '%d'",
@@ -417,6 +418,7 @@ class ScoreboardService
         // See if this submission was the first to solve this problem.
         // Only relevant if it was correct in the first place.
         $firstToSolve = false;
+        $potentialFirstToSolve = false;
         if ($correctJury) {
             $params = [
                 'cid' => $contest->getCid(),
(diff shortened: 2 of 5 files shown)


webapp/src/Service/ScoreboardService.php
```diff
@@ -437,32 +439,49 @@ public function calculateScoreRow(
             //   - or already judged to be correct (if it is judged but not correct,
             //     it is not a first to solve)
             // - or the submission is still queued for judgement (judgehost is NULL).
-            $verificationRequiredExtra = $verificationRequired ? 'OR j.verified = 0' : '';
+            // If there are no valid correct submissions submitted earlier but there
+            // are submissions awaiting judgement, we are potentially the first to solve.
+            // We need to keep track of this since we later need to set the actual first to solve.
+            $verificationRequiredExtra = ($verificationRequired && !$useExternalJudgements) ? 'OR j.verified = 0' : '';
```
vmcj, line 445: Why is this change needed?

webapp/tests/Unit/Service/ScoreboardServiceTest.php
```diff
@@ -0,0 +1,123 @@
...
+                    ->setStarttime(Utils::now())
+                    ->setEndtime(Utils::now())
+                    ->setSubmission($submission)
+                    ->setResult($result)
+            );
+
+//        $judging->addRun($judgingRun = (new JudgingRun())
+//            ->setJudging($judging)
+//            ->setRuntime(0)
+//        );
```
vmcj, line 115: Leftover?

webapp/src/Service/ScoreboardService.php
nickygerritsen: External judgements don't have a verified status in this sense. This just never worked.

webapp/migrations/Version20231229094359.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    public function getDescription(): string
+    {
+        return 'Add potential first to solve to scorecache.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE scorecache ADD is_potential_first_to_solve TINYINT(1) DEFAULT 0 NOT NULL COMMENT \'Is this potentially the first solution to this problem?\' AFTER is_first_to_solve');
```
eldering, line 23: This seems to require a `_public` and `_jury` version: if you first solve a problem in the freeze and there's another one pending before yours, but also in the freeze, then the public/team should not know whether it was actually a first to solve or not.

In general, I'm a bit wary of adding extra complexity and columns. I think it would be better if we can figure out what scoreboard data depends what changes, and just fetch all that data and pass it to a single function that just recalculates the relevant scoreboard data.

nickygerritsen: But we also don’t have a jury and public version of the normal first to solve. So I’m wondering if we have some logic, where we don’t show it if there are blue boxes anyway?

I get what you mean but I’m not sure how to solve this easily. A new incorrect submission can affect a correct submission that came after it and currently we only handle the current submission, not any other

webapp/src/Service/ScoreboardService.php
```diff
@@ -499,6 +519,31 @@ public function calculateScoreRow(
...
+        // potential first to solves.
+        if ($updatePotentialFirstToSolves && $pendingJury === 0 && !$potentialFirstToSolve) {
+            /** @var ScoreCache[] $potentialFirstToSolves */
+            $potentialFirstToSolves = $this->em->createQueryBuilder()
+                ->from(ScoreCache::class, 's')
+                ->join('s.team', 't')
+                ->select('s', 't')
+                ->andWhere('s.contest = :contest')
+                ->andWhere('s.problem = :problem')
+                ->andWhere('s.is_potential_first_to_solve = 1')
```
meisterT, line 537: Is this restriction actually adding anything? We are hopefully only very rarely executing this codepath anyway, so it would not hurt to go here over all teams
- which have solved this problem (which is a restriction you are currently not adding)
- after us (right?)

nickygerritsen: We only set `s. is_potential_first_to_solve` if that team solved the problem and we have earlier pending submissions, so I'd say that we do filter on teams solving this problem.

My current code assumes that if we loop over all potentional first to solves for the same problem, we will find the one that we need to actually update.

Currently this code path is executed more than you think: it is executed when we have no pending jury submissions for this (team,problem) and we are not the first to solve ourselves.

If you have a better idea to implement this, I am open, but I don't follow your thought currently. Feel free to push it as a commit. I think the added test should catch if it works.

meisterT: I am thinking of something like https://github.com/DOMjudge/domjudge/commit/76f1640d4edc3d30adcbfe86911cd54acc4bb7d6

That should make sure we don't call it too often while not adding the complexity of the extra column. I still have to think through the potential race conditions...

nickygerritsen: Something like that could indeed work I think.

PR #2286, opened 2024-01-03
Number of deprecations has increased

vmcj:
See: https://gitlab.com/DOMjudge/domjudge/-/jobs/5851059819

We're currently a the latest LTS so have enough time to fix those in the future. Most are intruduced by the latest move to the LTS.

@nickygerritsen I thought you had a commit for this, did something change afterwards?


nickygerritsen:
I upgraded to 6.4.2, maybe that added some more.

nickygerritsen (approved):

PR #2287, opened 2024-01-03
Cleanup of PHPStan baseline

vmcj:
I'm planning to do this file by file, and created a commit per issue to make reviewing easier.

Not sure about 7cf609798c88693e0f55e11292571b97682912ce and if there is a more strict annotation.

### DIFF
--- a/webapp/src/Utils/Utils.php
+++ b/webapp/src/Utils/Utils.php
@@ -321,6 +321,8 @@ class Utils
 
     /**
      * Parse a hex color into it's three RGB values.
+     *
+     * @return array{int, int, int}
      */
     public static function parseHexColor(string $hex): array
     {
@@ -345,6 +347,8 @@ class Utils
 
     /**
      * Convert an RGB triple into a CSS hex color.
+     *
+     * @param array{int, int, int} $color
      */
     public static function rgbToHex(array $color): string
     {
@@ -506,6 +510,8 @@ class Utils
 
     /**
      * Compute the LCS diff of two lines.
+     *
+     * @return array{bool, string}
      */
     public static function computeLcsDiff(string $line1, string $line2): array
     {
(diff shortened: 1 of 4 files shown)


webapp/src/Utils/Utils.php
```diff
@@ -865,6 +874,9 @@ public static function parseTsvLine(string $line): array
 
     /**
      * Reindex the given array by applying the callback to each item.
+     *
+     * @param array<int|string, mixed> $array
+     * @return array<int|string, mixed>
```
nickygerritsen, line 879: This could become something like:
```suggestion
	 * @template TKey of array-key
     * @template UKey of array-key
     * @template V
     * @param array<TKey, V> $array
     * @param callable(V, TKey): UKey $callback
     * @return array<UKey, V>
```

webapp/src/Utils/Utils.php
```diff
@@ -345,6 +347,8 @@ public static function componentToHex(int $component): string
 
     /**
      * Convert an RGB triple into a CSS hex color.
+     *
+     * @param array<int, int> $color
```
nickygerritsen, line 351: Why not `int[]`?

webapp/src/Utils/Utils.php
```diff
@@ -506,6 +510,8 @@ public static function cutString(string $str, int $size): string
 
     /**
      * Compute the LCS diff of two lines.
+     *
+     * @return array<int, bool|string>
```
nickygerritsen, line 514:
```suggestion
     * @return array{bool, string}
```

See https://phpstan.org/writing-php-code/phpdoc-types#array-shapes

webapp/src/Utils/Utils.php
```diff
@@ -857,6 +864,8 @@ public static function toTsvField(string $field) : string
 
     /**
      * Split a line from a Tab Separated Values file into fields.
+     *
+     * @return array<int, string>
```
nickygerritsen, line 868:
```suggestion
     * @return string[]
```

nickygerritsen (approved):

PR #2288, opened 2024-01-03
PHPStan fixes #2

### LINKED ISSUE #2: Advance Configuration
All variable aspects of the CCS must be configurable prior to the start of a World Finals contest. That is, the CCS may not require that any changes be made to the CCS configuration once the contest has started.

vmcj:
What should we do with parts where we return an item from a dependency? Explicit have an annotation so we detect changes or leave it as is as those are overly broad?

### DIFF
--- a/webapp/src/FosRestBundle/FlattenExceptionHandler.php
+++ b/webapp/src/FosRestBundle/FlattenExceptionHandler.php
@@ -22,6 +22,9 @@ class FlattenExceptionHandler implements SubscribingHandlerInterface
         private readonly bool $debug
     ) {}
 
+    /**
+     * @return array<array{direction: int, format: string, type: string, method: string}>
+     */
     public static function getSubscribingMethods(): array
     {
         return [
@@ -44,6 +47,9 @@ class FlattenExceptionHandler implements SubscribingHandlerInterface
             $type);
     }
 
+    /**
+     * @return array{code: int, message: string}
+     */
     private function convertToArray(
         FlattenException $exception,
         Context $context
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -391,36 +391,6 @@ parameters:
 			path: webapp/src/Helpers/OrdinalArray.php
 
 		-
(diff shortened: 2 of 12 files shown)


webapp/src/FosRestBundle/FlattenExceptionHandler.php
```diff
@@ -44,6 +47,9 @@ public function serializeToJson(
             $type);
     }
 
+    /**
+     * @return array<string, string|int>
```
nickygerritsen, line 51: Another shape?

nickygerritsen:
> What should we do with parts where we return an item from a dependency? Explicit have an annotation so we detect changes or leave it as is as those are overly broad?

I don't know what you mean.

vmcj:
> > What should we do with parts where we return an item from a dependency? Explicit have an annotation so we detect changes or leave it as is as those are overly broad?
> 
> I don't know what you mean.

See: https://github.com/DOMjudge/domjudge/pull/2288/commits/9ecb21a832028e74c7ff84e804db761b75561a17#diff-6d7cf7ba24a3042c8310fec4a0ca3f644cd61c724beb660b4f5bfcf9e83f1291R40

Here we return the result of applying the function from one of our dependencies. In their code they don't have the annotation so I'm not sure what we want to do here?

nickygerritsen:
> > > What should we do with parts where we return an item from a dependency? Explicit have an annotation so we detect changes or leave it as is as those are overly broad?
> > 
> > 
> > I don't know what you mean.
> 
> See: [9ecb21a#diff-6d7cf7ba24a3042c8310fec4a0ca3f644cd61c724beb660b4f5bfcf9e83f1291R40](https://github.com/DOMjudge/domjudge/commit/9ecb21a832028e74c7ff84e804db761b75561a17#diff-6d7cf7ba24a3042c8310fec4a0ca3f644cd61c724beb660b4f5bfcf9e83f1291R40)
> 
> Here we return the result of applying the function from one of our dependencies. In their code they don't have the annotation so I'm not sure what we want to do here?

Yeah not sure, we can't do much I think. Or maybe use generics

vmcj:
So I've locally stashed the difficult parts and will look at those in the future. This should all be semi-trivial, The last commit is quite large but changing to another format also would lose information.

nickygerritsen (approved):

PR #2289, opened 2024-01-03
Stan fixes 20240103 #3

vmcj:
After this is approved I'll just merge rebase it with another approved one. Otherwise debugging if a unit test fails becomes really annoying on my side.

### DIFF
--- a/webapp/src/Controller/API/AbstractRestController.php
+++ b/webapp/src/Controller/API/AbstractRestController.php
@@ -230,6 +230,8 @@ abstract class AbstractRestController extends AbstractFOSRestController
 
     /**
      * @throws NonUniqueResultException
+     *
+     * @return Object[]
      */
     protected function listActionHelper(Request $request): array
     {
--- a/webapp/src/Controller/API/AbstractRestController.php
+++ b/webapp/src/Controller/API/AbstractRestController.php
@@ -267,7 +267,7 @@ abstract class AbstractRestController extends AbstractFOSRestController
             }
         }
 
-        /** @var array $objects */
+        /** @var Object[] $objects */
         $objects = $queryBuilder
             ->getQuery()
             ->getResult();
--- a/webapp/src/Controller/API/AccessController.php
+++ b/webapp/src/Controller/API/AccessController.php
@@ -24,6 +24,7 @@ class AccessController extends AbstractRestController
     /**
      * Get access information
      *
+     * @return array<string, array<int, array<string, array<int, string>|string>|string>>
      * @throws NoResultException
      * @throws NonUniqueResultException
      */
--- a/webapp/src/Controller/API/AwardsController.php
+++ b/webapp/src/Controller/API/AwardsController.php
@@ -60,6 +60,7 @@ class AwardsController extends AbstractRestController
     /**
      * Get the specific award for this contest.
      *
+     * @return array<array<string, array<int>|string>>
      * @throws Exception
      */
     #[Rest\Get('/{id}')]
--- a/webapp/src/Controller/API/AwardsController.php
+++ b/webapp/src/Controller/API/AwardsController.php
@@ -40,6 +40,7 @@ class AwardsController extends AbstractRestController
     /**
      * Get all the awards standings for this contest.
      *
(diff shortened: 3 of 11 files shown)


webapp/src/Controller/API/AbstractRestController.php
```diff
@@ -230,6 +230,8 @@ abstract protected function getIdField(): string;
 
     /**
      * @throws NonUniqueResultException
+     *
+     * @return Object[]
```
nickygerritsen, line 234: We could use an `@template` at the class level here again and then have this `@return T[]`.

webapp/src/Controller/API/AccessController.php
```diff
@@ -24,6 +24,7 @@ class AccessController extends AbstractRestController
     /**
      * Get access information
      *
+     * @return array<string, array<int, array<string, array<int, string>|string>|string>>
```
nickygerritsen, line 27: We should either:

* Use an array shape
* Change this method to return a DTO (data transfer object) instead of an array

I'm leaning towards the second option

webapp/src/Controller/API/AwardsController.php
```diff
@@ -40,6 +40,7 @@ public function __construct(
     /**
      * Get all the awards standings for this contest.
      *
+     * @return array<array<string,string|int[]>>|null
```
nickygerritsen, line 43: Should we create an `Award` DTO?

webapp/src/Controller/API/BalloonController.php
```diff
@@ -28,6 +28,7 @@ class BalloonController extends AbstractRestController
     /**
      * Get all the balloons for this contest.
      *
+     * @return array<int, int|string>
```
nickygerritsen, line 31: This does not return an array of strings? I think it returns an array of arrays. But maybe again we should use a DTO.

webapp/src/Controller/API/ContestController.php
```diff
@@ -437,6 +437,8 @@ public function getContestYamlAction(Request $request, string $cid): StreamedRes
 
     /**
      * Get the current contest state
+     *
+     * @return array<string, string|null>
```
nickygerritsen, line 441: Array shape or DTO?

webapp/src/Controller/API/GeneralInfoController.php
```diff
@@ -54,6 +54,8 @@ public function __construct(
 
     /**
      * Get the current API version
+     *
+     * @return array<string, int>
```
nickygerritsen, line 58: Array shape (or DTO but this one is small)

webapp/src/Controller/API/GeneralInfoController.php
```diff
@@ -71,6 +73,8 @@ public function getVersionAction(): array
 
     /**
      * Get information about the API and DOMjudge
+     *
+     * @return array<string, string|array<string, string>>
```
nickygerritsen, line 77: Array shape or DTO.

The advantage of the DTO btw is that we can also use it in the `OA\Response` I think.

webapp/src/Controller/API/ScoreboardController.php
```diff
@@ -42,6 +42,8 @@ public function __construct(
 
     /**
      * Get the scoreboard for this contest.
+     *
+     * @return array<string, string|array<string|int, string|array<string, int|string|array<string|int, array<string, int|string|bool>>>>>
```
nickygerritsen, line 46: Oof. 100% a DTO?

webapp/src/Controller/API/SubmissionController.php
```diff
@@ -499,6 +499,8 @@ public function getSubmissionFilesAction(Request $request, string $id): Response
 
     /**
      * Get the source code of all the files for the given submission.
+     *
+     * @return array<array<string, string>>
```
nickygerritsen, line 503: If we make SourceCodeList a DTO, we can use that

vmcj:
I'll close this one to not cluther the list. This will be future work as it requires significantly more work compared to the other PRs

PR #2293, opened 2024-01-06
Use const reference parameters for efficiency to not make copies

eldering:
Flagged by Coverity Scan.

### DIFF
--- a/sql/files/defaultdata/compare/compare.cc
+++ b/sql/files/defaultdata/compare/compare.cc
@@ -90,7 +90,7 @@ bool equal_case_insensitive(std::string a, std::string b)
 /* Test two floating-point numbers for equality, accounting for +/-INF, NaN, and precision.
  * Float `jval` is considered the reference value for relative error.
  */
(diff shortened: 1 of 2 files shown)


sql/files/defaultdata/compare/compare.cc
```diff
@@ -90,7 +90,7 @@ bool equal_case_insensitive(std::string a, std::string b)
 /* Test two floating-point numbers for equality, accounting for +/-INF, NaN, and precision.
  * Float `jval` is considered the reference value for relative error.
  */
-void compare_float(std::string judge, std::string team, flt jval, flt tval, flt float_abs_tol, flt float_rel_tol, std::string extra_msg) {
+void compare_float(const std::string &judge, std::string team, flt jval, flt tval, flt float_abs_tol, flt float_rel_tol, const std::string &extra_msg) {
```
edomora97, line 93: Probably also `team` can be `const&`.

edomora97 (approved):

sql/files/defaultdata/compare/compare.cc
eldering: Thanks, fixed.

meisterT (approved):

PR #2294, opened 2024-01-08
PHPStan fixes, mostly in Service files

vmcj:
Here are the most issues, to read the code it would be beneficial to indeed have the annotations here but it requires basically dumping all ways we call the code.


nickygerritsen (approved):

vmcj (commented):
So much better.

Issue #1327, opened 2021-11-08
Support more fields when importing teams


shuibinlong:
When importing a team `.tsv`/`.json`, I found that the import interface does not parse the field of the team location, but it is necessary for tasks in many competitions.

The same as team members(in `.tsv`), team contests, even team comments?

nickygerritsen:
Currently we advice to run a [CDS](tools.icpc.global/cds) if you need these. What is your use case to put them in DOMjudge?

shuibinlong:
> Currently we advice to run a [CDS](https://tools.icpc.global/cds/) if you need these. What is your use case to put them in DOMjudge?

Maybe currently, we hope to have members and location information when viewing team details on the scoreboard page(when clicking one team‘s row).

And I searched the relevant code, and it seems that it can be achieved by adding some fields to the `importTeamsTsv` and `importTeamsJson` functions in `webapp/src/Service/ImportExportService.php`

nickygerritsen:
> > Currently we advice to run a [CDS](https://tools.icpc.global/cds/) if you need these. What is your use case to put them in DOMjudge?
> 
> Maybe currently, we hope to have members and location information when viewing team details on the scoreboard page(when clicking one team‘s row).
> 
> And I searched the relevant code, and it seems that it can be achieved by adding some fields to the `importTeamsTsv` and `importTeamsJson` functions in `webapp/src/Service/ImportExportService.php`

I see the following issues with this:
* First, we (the whole ICPC world) is moving away from TSV to JSON, but that is fine. All these fields are also present in the JSON.
* We only have one field for location and one for members. How would you want to import these?
* For members this is in a different TSV/JSON, so you would need to import a separate file.
* Since both are only one field, it would be hard(er)/impossible to also expose this in the desired format in the API.

shuibinlong:
> > > Currently we advice to run a [CDS](https://tools.icpc.global/cds/) if you need these. What is your use case to put them in DOMjudge?
> > 
> > 
> > Maybe currently, we hope to have members and location information when viewing team details on the scoreboard page(when clicking one team‘s row).
> > And I searched the relevant code, and it seems that it can be achieved by adding some fields to the `importTeamsTsv` and `importTeamsJson` functions in `webapp/src/Service/ImportExportService.php`
> 
> I see the following issues with this:
> 
> * First, we (the whole ICPC world) is moving away from TSV to JSON, but that is fine. All these fields are also present in the JSON.
[...]

PR #2296, opened 2024-01-12
Add location to team API.

### LINKED ISSUE #2283: Expose room via team endpoint in API
Currently, `room` (`location` in DOMJudge internal) can be imported via JSON and exposed via balloon API and print command. Also, as a public content on the scoreboard, `room` should be accessible by the API via `team` endpoint like `public_description`.

[Slack Message](https://domjudge-org.slack.com/archives/CHHURFUM7/p1704004917597109?thread_ts=1703914015.010669&cid=CHHURFUM7)

nickygerritsen:
Fixes #2283.
Also rename room to location in database and UI to be more consistent.

### DIFF
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -854,7 +854,7 @@ class ImportExportService
                     'name' => @$team['name'],
                     'display_name' => @$team['display_name'],
                     'publicdescription' => $team['public_description'] ?? @$team['members'],
-                    'location' => @$team['location'],
+                    'location' => @$team['location']['description'],
                 ],
                 'team_affiliation' => [
                     'externalid' => $team['organization_id'] ?? null,
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -854,7 +854,7 @@ class ImportExportService
                     'name' => @$team['name'],
                     'display_name' => @$team['display_name'],
                     'publicdescription' => $team['public_description'] ?? @$team['members'],
-                    'room' => @$team['room'],
+                    'location' => @$team['location'],
                 ],
                 'team_affiliation' => [
                     'externalid' => $team['organization_id'] ?? null,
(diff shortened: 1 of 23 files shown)


vmcj (approved):
I think this is small enough that we don't need DJ9

cubercsl:
https://github.com/DOMjudge/domjudge/issues/1327#issuecomment-962911929

Maybe we should not change room to location?

> For `location` we could call it `room` (that's how it is called internally). Or name it something else, as long as it is not `location`.

nickygerritsen:
> [#1327 (comment)](https://github.com/DOMjudge/domjudge/issues/1327#issuecomment-962911929)
> 
> Maybe we should not change room to location?
> 
> > For `location` we could call it `room` (that's how it is called internally). Or name it something else, as long as it is not `location`.

You are right. I will think about what we should do

webapp/src/Service/ImportExportService.php
```diff
@@ -854,7 +854,7 @@ public function importTeamsJson(array $data, ?string &$message = null, ?array &$
                     'name' => @$team['name'],
                     'display_name' => @$team['display_name'],
                     'publicdescription' => $team['public_description'] ?? @$team['members'],
-                    'room' => @$team['room'],
+                    'location' => @$team['location']['description'],
```
vmcj, line 857: Is there a way how we can now just if unset pick the `x,y,rot` and put it as a string?

cubercsl: Usually we use the seat number (such as B05) for the `location` here, but x, y, rot in the ccs spec represent distance and rotation angle. I think they do not correspond well directly.

vmcj: You can already use the `description` field, so for your usecase you would put the seat number there (or would you still need more?)

I'm suggesting the situation where that field is not set and the spec is followed. During the ICPC WF we would get the information in DOMjudge without much extra effort so we know where a team would be located in case of problems.

nickygerritsen: But then if you import + export you get something else, while it would be nice if it's idempotent?

cubercsl: > You can already use the `description` field, so for your usecase you would put the seat number there (or would you still need more?)

AFAIK, the `room` is formally used in built-in printing and balloon.

PR #2306, opened 2024-01-19
Introduce data transfer objects.

nickygerritsen:
This seems like a big PR (and it probably is), but basically what it does is it changes many (but not all, it's only a start) of our 'magic' arrays into objects, mostly using PHP 8's new constructor property promotion. This has some advantages:

- Return types are typed, so your IDE, the PHP parser and PHPStan can (and will) complain if you do something strange.
- For example for the submissions filter it actually tells you in the DTO object what you can filter on, without you needing to check some docblock, or even worse the filter code.
- For API return types we can move a lot of the YAML configuration to actual PHP code, making it more typesafe again.

Most of these changes are for the API, but not all. The API part should be extensively tested by CI, and the few UI changes should be OK.

@vmcj and me were discussing that, if this gets approval soon ™️ , he might want to run it at SWERC to see if it works. It's easy to rollback IMHO and then we will find out soon enough if everything works. But I can also understand people might find this risky. But then, given the upcoming contest season for us, I'm not sure when to merge.

Note: I created separate commits so you can see what I did but in the end I will squash everything.


webapp/src/Service/AwardService.php
```diff
@@ -2,12 +2,16 @@
...
 
+use App\DataTransferObject\Award;
 use App\Entity\Contest;
 use App\Entity\Team;
 use App\Utils\Scoreboard\Scoreboard;
 
 class AwardService
 {
+    /**
+     * @var array<int, Award[]>
```
vmcj, line 13:
```suggestion
     * @var array<Award[]>
```

webapp/src/DataTransferObject/AccessEndpoint.php
```diff
@@ -0,0 +1,17 @@
...
+use JMS\Serializer\Annotation as Serializer;
+
+class AccessEndpoint
+{
+    /**
+     * @param string[] $properties
+     */
+    public function __construct(
+        public readonly string $type,
+        #[Serializer\Type('array<string>')]
```
vmcj, line 14:
```suggestion
        #[Serializer\Type('string[]')]
```

vmcj: Not sure if this is valid, if it is we have more places where this should/can be done.

webapp/src/Controller/API/AwardsController.php
```diff
@@ -80,9 +87,9 @@ public function singleAction(Request $request, string $id): array
     }
 
     /**
-     * Get the awards data for the given request and optional award ID.
+     * @return array{Contest, Scoreboard}
      */
-    protected function getAwardsData(Request $request, string $requestedType = null): ?array
+    protected function getContestAndScoreboard(Request $request): array
```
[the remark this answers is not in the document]
nickygerritsen: Git diff is horrible. I extracted a method since I changed the existing method that either returned one award or an array of awards in two methods

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -1406,16 +1419,18 @@ private function getTestcaseFiles(string $id): array
 
         $result = [];
         foreach (['input', 'output'] as $k) {
-            $result[] = [
-                'filename' => $k,
-                'content' => base64_encode($inout[$k]),
-            ];
+            $result[]   = new JudgehostFile(
```
vmcj, line 1422:
```suggestion
            $result[] = new JudgehostFile(
```

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -1380,15 +1390,18 @@ private function getExecutableFiles(string $id): array
 
         $result = [];
         foreach ($files as $file) {
-            $result[]   = [
-                'filename' => $file->getFilename(),
-                'content' => base64_encode($file->getFileContent()),
-                'is_executable' => $file->isExecutable(),
-            ];
+            $result[]   = new JudgehostFile(
```
vmcj, line 1393:
```suggestion
            $result[] = new JudgehostFile(
```

webapp/src/Service/CheckConfigService.php
```diff
@@ -316,55 +308,58 @@ public function checkScriptFilesizevsMemoryLimit(): array
...
-            ];
+        return new ConfigCheckItem(
+            caption: 'Compile file size vs. memory limit',
+            result: $result,
+            desc: 'If the script filesize limit is lower than the memory limit, then ' .
+            'compilation of sources that statically allocate memory may fail. We ' .
+            'recommend to include a margin to be on the safe side. The current ' .
+            '"script_filesize_limit" = ' . $this->config->get('script_filesize_limit') . ' ' .
+            'while "memory_limit" = ' . $this->config->get('memory_limit') . '.'
+        );
```
vmcj, line 319:
```suggestion
        return new ConfigCheckItem(
            caption: 'Compile file size vs. memory limit',
            result: $result,
            desc: 'If the script filesize limit is lower than the memory limit, then ' .
                'compilation of sources that statically allocate memory may fail. We ' .
                'recommend to include a margin to be on the safe side. The current ' .
                '"script_filesize_limit" = ' . $this->config->get('script_filesize_limit') . ' ' .
                'while "memory_limit" = ' . $this->config->get('memory_limit') . '.'
        );
```

vmcj (approved):
Overall I find this an improvement. Do we have a codecov report somewhere how much of the diff is covered by tests?

webapp/src/Service/AwardService.php
nickygerritsen: I did it this way since the array indices are not consecutive but the contest ID's

webapp/src/DataTransferObject/AccessEndpoint.php
nickygerritsen: Nope doesn't work.

webapp/src/Service/CheckConfigService.php
nickygerritsen: Does PSR12 say something about this? My .editorconfig, code style in PHPStorm as well as phpcs doesn't complain about this and PHPStorm even changes it back to be without indent.

nickygerritsen:
> Overall I find this an improvement. Do we have a codecov report somewhere how much of the diff is covered by tests?

First of all, seems codecov.io doesn't handle PR's from forks currently since we moved it to Github, but I ran it locally. I can only tell something about the API easily. For unit tests, we don't test:

- `contests/<x>/state`, but we test that during integration. Could still add it to a test.
- `contests/<x>/event-feed`, but again tested during integration. Could still add it to a test, at least with `stream=0`.
- `contests/<x>/status`, which we should maybe add.
- `contests/<x>/samples`, which we should maybe add a test for.
- Almost the whole `JudgehostController`, but we have integration tests for it.
- `contests/<x>/judgements/<x>`, but the list we do. Maybe add a test for the singular as well?
- POSTing to configure languages.
- The `getRuntimeAsScoreTiebreaker` mode for the scoreboard recently added.
- `contests/<x>/submissions/<x>/files` and `contests/<x>/submissions/<x>/source-code'`. We should probably add some tests.
- Adding teams, groups and organizations.

For the purpose of this PR I think we test almost everything, but this list might be a good start to improve our API coverage.

webapp/src/Service/CheckConfigService.php
vmcj: There MUST NOT be a hard limit on line length; the soft limit MUST be 120 characters; lines SHOULD be 80 characters or less. (PSR2), PSR12 doesn't have a new opinion.

I just find it ugly to look at but lets leave it.

PR #2317, opened 2024-02-05
Make cgroup mount not readonly, needed for newer Docker for Mac setups.

nickygerritsen:
Fixes DOMjudge/domjudge-packaging#170


vmcj (approved):
Still works for me.

PR #2321, opened 2024-02-09
Use DTO's for POST/PUT request bodies.

nickygerritsen:
Some notes:

- We need the Symfony built in serializer for request mapping, so that is why I'm adding it. We can't get rid of the JMS one, since we use virtualproperty and type annotations, which the Symfony one doesn't support.
- For adding submissions it's a bit ugly, since we have both the CLICS format and our own, so all fields are optional. Also our format uses file uploads, which the Symfony request mapper doesn't support.
- I haven't done anything to the judgehost controller, but should we? We can add way more typing there this way as well.
- All file uploads (logo's, etc) have also not been done.
- The contest patch call can't use the assert\callback to check for values, since there CLICS has a distinction between not passing a variable and setting it to `null`. A bit ugly.
- Some tests needed to be updated since the error messages are a bit different. Not a big issue I think.
- If you supply no body, the Symfony request mapper produces an empty 400 error. Also not a big issue I think.

After this we should discuss what more to change to DTO's @vmcj, you have gone through most code.


vmcj:
> After this we should discuss what more to change to DTO's @vmcj, you have gone through most code.

The first candidates would be the `*TwigData` arrays, `collect*Table` arrays and the `*Stats` arrays, the `DOMJudgeService` file also has a lot of ugly arrays 

The other individual functions:
- getConfigSpecification
- findExecutableOptions
- dataForEditor
- getStats

vmcj:
> Some notes:
> 
>     * We need the Symfony built in serializer for request mapping, so that is why I'm adding it. We can't get rid of the JMS one, since we use virtualproperty and type annotations, which the Symfony one doesn't support.

What is the new way of working with them? Is one the superset of the other or do they overlap?
> 
> 
>     * I haven't done anything to the judgehost controller, but should we? We can add way more typing there this way as well.

I would say so? It would make it much easier for people to create an IOI judge or if we plan to change the judgedaemon to another implementation?
> 
>     * All file uploads (logo's, etc) have also not been done.
> 
>     * The contest patch call can't use the assert\callback to check for values, since there CLICS has a distinction between not passing a variable and setting it to `null`. A bit ugly.

Can we change the CLICS spec or is CLICS correct here?
> 
>     * Some tests needed to be updated since the error messages are a bit different. Not a big issue I think.
Agree
> 
>     * If you supply no body, the Symfony request mapper produces an empty 400 error. Also not a big issue I think.
That makes sense right, don't repeat that request if we can't do anything with it.

nickygerritsen:
> > Some notes:
> > ```
> > * We need the Symfony built in serializer for request mapping, so that is why I'm adding it. We can't get rid of the JMS one, since we use virtualproperty and type annotations, which the Symfony one doesn't support.
> > ```
> 
> What is the new way of working with them? Is one the superset of the other or do they overlap?

Nope they have different skillsets... Just use the JMS stuff like we used to and for request params the Symfony one will be used. Hopefully in the future the Symfony one is more nature and we can replace the JMS one.
> 
> > ```
> > * I haven't done anything to the judgehost controller, but should we? We can add way more typing there this way as well.
> > ```
> 
> I would say so? It would make it much easier for people to create an IOI judge or if we plan to change the judgedaemon to another implementation?

Well the API itself won't change right, only the parsing logic. But I agree, it might be better.
> 
> > ```
> > * All file uploads (logo's, etc) have also not been done.
> > 
> > * The contest patch call can't use the assert\callback to check for values, since there CLICS has a distinction between not passing a variable and setting it to `null`. A bit ugly.
> > ```
> 
> Can we change the CLICS spec or is CLICS correct here?

I don't think we can easily change the CLICS spec. This is just a drawback of the serializer: there is no easy way to distinguish between `null` and that a value has not been set. What I *can* think of is to use an actual setter (and not a constructor argument) and then keep track there if it is actually called, but not sure if that's better.
> 
> > ```
> > * Some tests needed to be updated since the error messages are a bit different. Not a big issue I think.
> > ```
> 
> Agree
> 
> > ```
> > * If you supply no body, the Symfony request mapper produces an empty 400 error. Also not a big issue I think.
> > ```
> 
> That makes sense right, don't repeat that request if we can't do anything with it.

webapp/src/Controller/API/AbstractRestController.php
```diff
@@ -137,11 +137,15 @@ protected function renderCreateData(
...
+        $postfix = '';
         if ($routeType !== 'user') {
             $params['cid'] = $request->attributes->get('cid');
+            if ($params['cid'] === null) {
+                $postfix = '_1';
+            }
         }
         $headers = [
-            'Location' => $this->generateUrl("v4_app_api_{$routeType}_single", $params, UrlGeneratorInterface::ABSOLUTE_URL),
+            'Location' => $this->generateUrl("v4_app_api_{$routeType}_single$postfix", $params, UrlGeneratorInterface::ABSOLUTE_URL),
```
vmcj, line 148:
```suggestion
            'Location' => $this->generateUrl("v4_app_api_{$routeType}_single" . $postfix, $params, UrlGeneratorInterface::ABSOLUTE_URL),
```
I find it nicer to read, assuming it does work.

webapp/src/Controller/API/GroupController.php
```diff
@@ -89,14 +87,23 @@ public function singleAction(Request $request, string $id): Response
...
+        Request $request,
+        ImportExportService $importExport
+    ): Response {
         $saved = [];
-        $postedData = $request->request->all();
-        if (array_key_exists('id', $postedData)) {
-            throw new BadRequestHttpException("Cannot add group with ID");
-        }
-        $importExport->importGroupsJson([$postedData], $message, $saved);
+        $importExport->importGroupsJson([
```
vmcj, line 97: Should we add a `//TODO` to change the importGroupsJson to do the unpacking?

webapp/src/Controller/API/ProblemController.php
```diff
@@ -282,28 +284,19 @@ public function unlinkProblemAction(Request $request, string $id): Response
...
     )]
     #[OA\Parameter(ref: '#/components/parameters/id')]
-    public function linkProblemAction(Request $request, string $id): Response
+    public function linkProblemAction(
+        #[MapRequestPayload(validationFailedStatusCode: Response::HTTP_BAD_REQUEST)]
+        ContestProblemPut $contestProblemPut,
+        Request $request,
+        string $id
+    ): Response
     {
```
vmcj, line 299:
```suggestion
    ): Response {
```

phpcs complains

webapp/src/Controller/API/GroupController.php
nickygerritsen: Yeah was thinking the same

webapp/src/Controller/API/ProblemController.php
```diff
@@ -282,28 +284,19 @@ public function unlinkProblemAction(Request $request, string $id): Response
...
-        $required = ['label'];
```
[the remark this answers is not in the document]
nickygerritsen: Yup. Isn’t it great?

webapp/tests/Unit/Controller/API/GroupControllerTest.php
```diff
@@ -80,8 +80,9 @@ public function testNewAddedGroupPostWithId(): void
         $url = $this->helperGetEndpointURL($this->apiEndpoint);
         $postWithId = $this->newGroupsPostData[0];
         $postWithId['id'] = '1';
-        // CLICS does not allow POST to set the id value
-        $this->verifyApiJsonResponse('POST', $url, 400, 'admin', $postWithId);
+        // CLICS does not allow POST to set the id value. Our API will just ignore the property
+        $returnedObject = $this->verifyApiJsonResponse('POST', $url, 201, 'admin', $postWithId);
+        self::assertNotEquals($returnedObject['id'], $postWithId['id']);
```
vmcj, line 85: I'm not sure if I like/understand this,

Why does the behavior change?

webapp/tests/Unit/Controller/API/ContestControllerAdminTest.php
```diff
@@ -214,7 +214,7 @@ public function provideChangeTimes(): Generator
         // Note that if the first item contains "id", we replace it with the correct ID in the test
 
         // General tests
-        yield [[], 400, 'Missing \"id\" in request.'];
```
vmcj, line 217: Was it intentional to remove this one?

webapp/tests/Unit/Controller/API/GroupControllerTest.php
nickygerritsen: We used to “accept” and ID and then throw an error. Now we just ignore it.
We could still accept it and throw an error but why should we?

webapp/src/DataTransferObject/ContestProblemPut.php
```diff
@@ -0,0 +1,22 @@
...
+
+#[OA\Schema(required: ['label'])]
+class ContestProblemPut
+{
+    public function __construct(
+        #[OA\Property(description: 'The label of the problem to add to the contest')]
+        public readonly string $label,
+        #[OA\Property(description: 'Human readable color of the problem to add. Will be overwritten by `rgb` if supplied', nullable: true)]
+        public readonly ?string $color,
+        #[OA\Property(description: 'Hexadecimal RGB value of the color of the problem to add. Will be used if both `color` and `rgb` are supplied', nullable: true)]
```
vmcj, line 15:
```suggestion
        #[OA\Property(description: 'Hexadecimal RGB value of the color of the problem to add. Overrules `color` if supplied', nullable: true)]
```

Has nothing to do with your PR.

webapp/src/DataTransferObject/AddSubmission.php
```diff
@@ -0,0 +1,67 @@
+<?php declare(strict_types=1);
+
+namespace App\DataTransferObject;
+
```
vmcj, line 4:
```suggestion
```

phpcs

vmcj (approved):
LGTM

webapp/tests/Unit/Controller/API/GroupControllerTest.php
vmcj: Because now you would need to check the returned object if it is the same as what you expect.

Is there a case where you can set the id or is it just never allowed?

nickygerritsen: You can never set it. If you could, we should indeed add it to the DTO

webapp/src/Controller/API/AbstractRestController.php
vmcj: `"v4_app_api_{$routeType}_single{$postfix}"` would also be fine for me, I prefer to directly see that we do something with a variable there (in this case twice).

webapp/src/Controller/API/AbstractRestController.php
```diff
@@ -137,11 +137,15 @@ protected function renderCreateData(
         $params = [
             'id' => $id,
         ];
+        $postfix = '';
         if ($routeType !== 'user') {
             $params['cid'] = $request->attributes->get('cid');
+            if ($params['cid'] === null) {
+                $postfix = '_1';
```
meisterT, line 144: what's happening in this file?

webapp/src/Controller/API/GroupController.php
```diff
@@ -89,14 +87,23 @@ public function singleAction(Request $request, string $id): Response
         description: 'Returns the added group',
         content: new Model(type: TeamCategory::class)
     )]
-    public function addAction(Request $request, ImportExportService $importExport): Response
-    {
+    public function addAction(
+        #[MapRequestPayload(validationFailedStatusCode: Response::HTTP_BAD_REQUEST)]
```
meisterT, line 91: Do we have to add this everywhere?

webapp/tests/Unit/Controller/API/ClarificationControllerTest.php
```diff
@@ -132,12 +133,17 @@ public function testAddInvalidData(string $user, array $dataToSend, string $expe
...
+        if (str_starts_with($expectedMessage, '/')) {
+            static::assertMatchesRegularExpression($expectedMessage, $data['message']);
+        } else {
+            static::assertEquals($expectedMessage, $data['message']);
+        }
     }
 
     public function provideAddInvalidData(): Generator
     {
-        yield ['demo', [], "Argument 'text' is mandatory."];
```
meisterT, line 140: We don't display any text anymore?

webapp/src/Controller/API/AbstractRestController.php
nickygerritsen: This is broken currently already on main: If you create a team using `POST /api/teams` (and not `/api/contests/xxx/teams`, you will get a 500. This tries to fix this, but I can extract this in a separate commit if you want.

webapp/src/Controller/API/GroupController.php
nickygerritsen: The default is a 422. If you can live with that we can drop the `validationFailedStatusCode` part, but then we need to update the tests (since we sometimes return a 400 and sometimes a 422) or we always need to return a 422.

webapp/tests/Unit/Controller/API/ClarificationControllerTest.php
nickygerritsen: When you submit *nothing* there is an empty 400 (or 422 if we change the above thing) since Symfony doesn't know what to do. As soon as you supply *any* field (even an unknown one) we do display an error. It's a bit annoying but IMHO not a very big deal.

This test though should be added with an empty message probably.

PR #2322, opened 2024-02-10
Use a Symfony Form for jury clarifications.

### LINKED ISSUE #2311: Incomplete markdown rendering in clarification requests
Hi! SWERC 2023-24 was this weekend, here is a minor issue that we encountered with DOMjudge.

Some markdown elements such as quotes or inline code blocks are not rendered correctly, for example in clarification requests. 

### Your environment
DOMjudge 8.3.0DEV/5502c03cc

### Steps to reproduce
- Log into DOMJudge with jury capabilities
- Go to the "clarifications" tab
- Click "send clarification"
- Enter the following markdown:
~~~
*Italic*
**Bold**
`inline code`

 # Heading 1 
 ## Heading 2

 [Link](http://a.com) 

 ![Image](../../images/DOMjudgelogo.svg?v=8.3.0DEV/5502c03cc) 

- List e1
- List e2

1. list e1
2. list e2

---

```
code blocks
```
~~~
### Expected behaviour
Mardown elements are rendered correctly

### Actual behaviour
Blockquote, inline code and code blocks are note rendered properly

I suspect that the culprit is this  [this line](https://github.com/DOMjudge/domjudge/blob/5502c03cc880d8e5c6b0e570a7b875a07b1a2bf7/webapp/src/Controller/RootController.php#L54), where HTML escaping is done before rendering markdown, likely resulting in HTML special characters being replaced, erasing the markdown elements.

nickygerritsen:
Also sanitize only after converting to markdown. Fixes #2311

### DIFF
--- a/webapp/config/packages/html_sanitizer.yaml
+++ b/webapp/config/packages/html_sanitizer.yaml
@@ -1,0 +1,6 @@
+framework:
+    html_sanitizer:
+        sanitizers:
+            app.clarification_sanitizer:
+                allow_safe_elements: true
+                allow_relative_medias: true
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -420,10 +420,10 @@ function toggleExpand(event)
 function clarificationAppendAnswer() {
     if ( $('#clar_answers').val() == '_default' ) { return; }
     var selected = $("#clar_answers option:selected").text();
-    var textbox = $('#bodytext');
+    var textbox = $('#jury_clarification_message');
     textbox.val(textbox.val().replace(/\n$/, "") + '\n' + selected);
     textbox.scrollTop(textbox[0].scrollHeight);
-    previewClarification($('#bodytext') , $('#messagepreview'));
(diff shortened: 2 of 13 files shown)


vmcj (approved):
I've little experience with those forms.

meisterT:
Code looks fine, feel free to merge. But I do have a couple more suggestions/nits (will write separate comments for each):

code blocks for teams do not look right in preview:
image

Once sent, it looks ok:
image

Interestingly enough the same works on the jury side:
image

meisterT:
The reply button after starting a reply is useless and collapses the reply-form. I think it should be replaced by the "send" button: 
image

meisterT:
The reply form looks like a Matryoshka of boxes and clutter, remove at least these two:

image
 
Then "message" is unnecessary as there is only one text field and I don't see a use case to change the subject in a reply.

nickygerritsen:
Note that your comments are all about the team clarification page, which I didn't even touch. I do agree with the remakrs so we should fix them for sure.

meisterT (approved):

PR #2324, opened 2024-02-12
Record testcase directory in judging runs.

### LINKED ISSUE #2059: Testcase in output/judgings/... directory is not easily matched with web UI

meisterT:
This makes it much easier to find out where to go for debugging.

Fixes #2059.

Example:
image

Clicking the copy/paste icon copies the full path.

### DIFF
--- a/webapp/src/Controller/Jury/SubmissionController.php
+++ b/webapp/src/Controller/Jury/SubmissionController.php
@@ -394,6 +394,9 @@ class SubmissionController extends BaseController
                     '/\[output storage truncated after \d* B\]/',
                     (string)$runResult['output_run_last_bytes']
                 );
+                if ($firstJudgingRun) {
+                    $runResult['testcasedir'] = $firstJudgingRun->getTestcaseDir();
+                }
                 $runsOutput[] = $runResult;
             }
         }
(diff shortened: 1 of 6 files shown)


meisterT:
This is an alternative to https://github.com/DOMjudge/domjudge/pull/2211

nickygerritsen (approved):

vmcj (approved):
Fine by me, lets see which one @eldering wants.

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -394,6 +394,9 @@ public function viewAction(
                     '/\[output storage truncated after \d* B\]/',
                     (string)$runResult['output_run_last_bytes']
                 );
+                if ($firstJudgingRun) {
+                   $runResult['testcasedir'] = $firstJudgingRun->getTestcaseDir();
```
vmcj, line 398:
```suggestion
                    $runResult['testcasedir'] = $firstJudgingRun->getTestcaseDir();
```

nitpick: phpcs wants 4 spaces here.

eldering:
> Fine by me, lets see which one @eldering wants.

I think this solution is better than #2211 : both in the place/way the information is displayed and also by taking the actual directory from the judgehost instead of depending on configure/build time variables.

PR #2325, opened 2024-02-12
Last poll time is a decimal, so Doctrine uses strings by default when…

nickygerritsen:
… fetching from the database.

The Symfony debug bar complained about this.

### DIFF
--- a/webapp/src/Entity/ExternalContestSource.php
+++ b/webapp/src/Entity/ExternalContestSource.php
@@ -47,7 +47,7 @@ class ExternalContestSource
         nullable: true,
         options: ['comment' => 'Time of last poll by event feed reader', 'unsigned' => true]
     )]
-    private ?float $lastPollTime = null;
+    private string|float|null $lastPollTime = null;
 
     #[ORM\Column(
         type: 'smallint',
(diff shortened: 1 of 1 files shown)


vmcj (commented):
Why can't we steer doctrine to use a proper float? I find this ugly without the explanation

nickygerritsen:
Decimals aren't safe to put in a float since you loose precision, so they use strings. if you want floats, use `FloatType`.
Which we could do here but then the database column type will change as well to `DOUBLE PRECISISON` from `DECIMAL(X,Y)`

nickygerritsen:
> Decimals aren't safe to put in a float since you loose precision, so they use strings. if you want floats, use `FloatType`. Which we could do here but then the database column type will change as well to `DOUBLE PRECISISON` from `DECIMAL(X,Y)`

If we are going to change the column type I'd vote to do that in a different PR btw. But I'm not sure if we should do that at all.

meisterT (approved):
I think this is just making this one field align with all of our other `decimal` fields, so I believe it is Ok to merge.

We can discuss whether we want to change it in general separately, ISTR that @eldering looked into this at some point.

PR #2328, opened 2024-02-13
Export custom (interactive) compare scripts when downloading problem ZIP.

### LINKED ISSUE #2327: Exporting a problem with a custom validator produces invalid ZIP: custom output validator is missing

When downloading the ZIP of a problem with a custom validator, we write to the `problem.yaml`:

```yaml
validation: custom
```

But we do not add the custom validator files to the ZIP file.

We also write:

```ini
special_compare='circuitdesign_cmp'
```

to `domjudge-problem.ini`. We need to make sure this doesn't hurt (but looking at the code it seems to overwrite this with a custom validator if it comes in the ZIP)

### Your environment

DOMjudge version: f79b0c7d63dcdefcd438bac5de7e1757fbfa68a6

### Steps to reproduce

Download the ZIP for problem C from `nwerc18` from demoweb and try to import this again on a different instance.

### Expected behaviour

Problem gets imported correctly.

### Actual behaviour

You get an error: `Custom validator specified but not found.`

nickygerritsen:
Fixes #2327.

### DIFF
--- a/webapp/src/Controller/Jury/ProblemController.php
+++ b/webapp/src/Controller/Jury/ProblemController.php
@@ -270,7 +270,10 @@ class ProblemController extends BaseController
         $yaml = ['name' => $problem->getName()];
         if (!empty($problem->getCompareExecutable())) {
             $yaml['validation'] = 'custom';
+        } elseif ($problem->getCombinedRunCompare() && !empty($problem->getRunExecutable())) {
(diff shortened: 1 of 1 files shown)


nickygerritsen:
Tested by exporting, clearing my DB and importing again. Seems to produce something that makes sense.

meisterT (approved):

PR #2331, opened 2024-02-14
Shadowing: use DTO's

nickygerritsen:
This should have (almost) no functional changes, but it uses DTO's for reading from the event feed, making it that we can get rid of a lot of array shapes and that we get added type safety.

I did add the `provider` object to the API root endpoint, so we can reuse it here and added some small improvements, see the separate commits.


webapp/src/Serializer/Shadowing/EventDenormalizer.php
```diff
@@ -0,0 +1,115 @@
...
+            return false;
+        }
+        if ($type !== Event::class) {
+            return false;
+        }
+        return true;
+    }
+
+    /**
+     * @return array<string, bool|null>
```
vmcj, line 95:
```suggestion
     * @return array{0: bool, 1: null}
```
Looks shorter.

nickygerritsen: But also wrong, since the keys are strings, not integers.

webapp/src/Controller/API/GeneralInfoController.php
```diff
@@ -99,6 +100,10 @@ public function getInfoAction(
             version: self::CCS_SPEC_API_VERSION,
             versionUrl: self::CCS_SPEC_API_URL,
             name: 'DOMjudge',
+            provider: new ApiInfoProvider(
```
vmcj, line 103:
```suggestion
            //TODO: Add DOMjudge logo
            provider: new ApiInfoProvider(
```

webapp/src/DataTransferObject/Shadowing/GroupEvent.php
```diff
@@ -0,0 +1,17 @@
...
+    public function __construct(
+        public readonly string $id,
+        public readonly string $name,
+        public readonly ?string $icpcId,
+        public readonly ?string $type,
+        public readonly ?string $location,
+        public readonly ?bool $hidden,
+        public readonly ?int $sortorder,
+        public readonly ?string $color,
+    ) {}
```
vmcj, line 16: Probably unrelated but the fields here are not what we report in our api/doc and there we have some fields which we don't have here.

webapp/src/DataTransferObject/Shadowing/OrganizationEvent.php
```diff
@@ -0,0 +1,17 @@
...
+class OrganizationEvent implements EventData
+{
+    public function __construct(
+        public readonly string $id,
+        public readonly string $name,
+        public readonly ?string $icpcId,
+        public readonly ?string $formalName,
+        public readonly ?string $country,
+        public readonly ?string $url,
+        public readonly ?string $twitterHashtag,
```
vmcj, line 14: If we check this field, shouldn't we also verify the logo?

webapp/src/DataTransferObject/Shadowing/ProblemEvent.php
```diff
@@ -0,0 +1,14 @@
...
+namespace App\DataTransferObject\Shadowing;
+
+class ProblemEvent implements EventData
+{
+    public function __construct(
+        public readonly string $id,
+        public readonly string $name,
+        public readonly int $timeLimit,
+        public readonly ?string $label,
+        public readonly ?string $rgb,
```
vmcj, line 12: color is missing?

webapp/src/DataTransferObject/Shadowing/GroupEvent.php
nickygerritsen: Let's drop stuff we are not using here for now then?

webapp/src/DataTransferObject/Shadowing/OrganizationEvent.php
nickygerritsen: Let's drop it (and URL) since we don't use it

webapp/src/DataTransferObject/Shadowing/ProblemEvent.php
nickygerritsen: Correct, we are not checking it

webapp/src/DataTransferObject/Shadowing/ContestData.php
```diff
@@ -6,18 +6,10 @@ class ContestData
...
         public readonly string $name,
-        public readonly ?string $formalName,
-        public readonly ?string $shortname,
         public readonly string $duration,
         public readonly ?string $scoreboardFreezeDuration,
-        public readonly ?string $scoreboardThawTime,
-        public readonly ?string $scoreboardType,
         public readonly int $penaltyTime,
         public readonly ?string $startTime,
-        public readonly ?string $endTime,
```
vmcj: Should we add a `TODO` that we want to test the `{end,scoreboardThaw}Time`?

vmcj (approved):
Good enough to merge as is, there are some discussion points left for future PRs.

webapp/src/DataTransferObject/Shadowing/Award.php
```diff
@@ -0,0 +1,9 @@
+<?php declare(strict_types=1);
+
+namespace App\DataTransferObject\Shadowing;
+
+class Award implements EventData
+{
+    // Not used currently, so no fields
```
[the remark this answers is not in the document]
nickygerritsen: Otherwise we need special checks for unknown events. This makes the code a bit cleaner

webapp/src/DataTransferObject/Shadowing/ContestData.php
```diff
@@ -0,0 +1,16 @@
...
+class ContestData
+{
+    public function __construct(
+        public readonly string $id,
+        public readonly string $name,
+        public readonly string $duration,
+        public readonly ?string $scoreboardFreezeDuration,
+        public readonly int $penaltyTime,
+        public readonly ?string $startTime,
+        // TODO: check for end time and scoreboard thaw time
```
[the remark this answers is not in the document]
nickygerritsen: Nope. @vmcj wanted thus comment. It would be a new “feature”

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -735,30 +694,30 @@ protected function validateAndUpdateContest(string $entityType, ?string $eventId
...
-            $freezeSeconds        = floor(($freezeStartSeconds % 60) / 60);
-            $freezeMilliseconds   = $freezeStartSeconds - floor($freezeStartSeconds);
+            $durationInSeconds = $durationHourModifier * (int)$durationData[2] * 3600
+                + 60 * (int)$durationData[3]
+                + (double)sprintf('%d.%03d', $durationData[4], $durationData[5]);
+            $freezeStartSeconds = $durationInSeconds - $freezeInSeconds;
+            $freezeHour = floor($freezeStartSeconds / 3600);
+            $freezeMinutes = floor(($freezeStartSeconds % 3600) / 60);
+            $freezeSeconds = floor(($freezeStartSeconds % 60) / 60);
+            $freezeMilliseconds = $freezeStartSeconds - floor($freezeStartSeconds);
```
eldering, line 720: I disagree with this kind of removal of alignment spaces. It makes the code less readable.

webapp/src/Serializer/Shadowing/EventDenormalizer.php
```diff
@@ -0,0 +1,115 @@
...
+            throw new InvalidArgumentException('Unsupported data.');
+        }
+
+        if (!$this->serializer instanceof DenormalizerInterface) {
+            throw new LogicException('Cannot denormalize attribute "data" because the injected serializer is not a denormalizer.');
+        }
+
+        $eventType = EventType::fromString($data['type']);
+        if ($this->getEventFeedFormat($data, $context) === EventFeedFormat::Format_2022_07) {
+            $operation = !isset($data['data']) ? Operation::DELETE : Operation::CREATE;
```
vmcj, line 53:
```suggestion
            $operation = isset($data['data']) ? Operation::CREATE : Operation::DELETE;
```

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -258,15 +257,18 @@ public function getLastReadEventId(): ?string
...
     /**
      * @param string[] $eventsToSkip
      */
-    public function import(bool $fromStart, array $eventsToSkip, ?callable $progressReporter = null): bool
-    {
+    public function import(
+        bool $fromStart,
+        array $eventsToSkip,
+        ?callable $progressReporter = null
+    ): bool {
```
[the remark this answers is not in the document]
nickygerritsen: My bad. I am very used to (at work) reformat a file with a code style before saving to make it in sync.
I will try to undo all these reformat changes (even though they don't adhere to the code style).

webapp/src/DataTransferObject/Shadowing/Award.php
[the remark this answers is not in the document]
nickygerritsen: Would you prefer to add the special cases to the ExternalContestSourceService?

webapp/src/Serializer/Shadowing/EventDenormalizer.php
nickygerritsen: What is the reason for this?

webapp/src/Service/ExternalContestSourceService.php
nickygerritsen: My bad, this is what the code style currently says. I will undo these changes.

nickygerritsen:
@vmcj found an issue with legacy delete events. I will fix that but before doing that add tests for some events so we know they are parsed correctly.

webapp/src/Serializer/Shadowing/EventDenormalizer.php
meisterT: it's more readable?

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -2071,48 +2063,28 @@ protected function addOrUpdateWarning(
...
 
         $this->em->flush();
     }
 
-    protected function removeWarning(string $entityType, ?string $entityId, string $type): void
+    protected function removeWarning(EventType $eventType, ?string $entityId, string $type): void
     {
-        $hash       = ExternalSourceWarning::calculateHash($type, $entityType, $entityId);
-        $warning    = $this->em
+        $hash = ExternalSourceWarning::calculateHash($type, $eventType->value, $entityId);
```
vmcj, line 2081: Not sure if you touch this in your PR or if it was already broken but calculateHash expects a string for `$entityId` but that can be `null`

webapp/src/Service/ExternalContestSourceService.php
eldering: I don't know exactly what the current code style says, but I strongly think that automatic reformatting of alignment like this should be removed from it. Code formatting is a means to an end (better readability of code) and not a goal in itself.

nickygerritsen: I believe the problem is it wants some alignment and you can’t say “do whatever there was” but I hope I’m wrong since that would save me time

PR #2336, opened 2024-02-17
Move runpipe test to github actions

vmcj:
The logs can already be downloaded via the GitHub interface, so that part has been removed.


.github/jobs/chroot.bats
```diff
@@ -9,6 +9,10 @@ fi
 # Cleanup old dir
 rm -rf $CHROOT
 
+if [ "${ARCH}" = 'default_placeholder' ]; then
```
meisterT, line 12: Please add a comment here why this is necessary.

.github/workflows/runpipe.yml
```diff
@@ -0,0 +1,30 @@
+name: Run runpipe tests
+on:
+  push:
+    branches-ignore:
+      - main
```
meisterT, line 5: Why do we ignore this on main and other branches?

vmcj: We have the merge train, so we already had a PR where we ran this.

.github/jobs/chroot.bats
vmcj, line 12: // In GHA CI we use this option to trigger all non specific architectures.

vmcj, line 12:
```suggestion
// In GHA CI we use this option to trigger all non specific architectures.
if [ "${ARCH}" = 'default_placeholder' ]; then
```

meisterT:
LGTM once the failing test is fixed

vmcj:
I took the failing part out as that seems to be interesting on itself.

PR #2339, opened 2024-02-22
Always use cgroups in runguard.

meisterT:
There is no benefit in not using them, the library needs to be installed anyway to be able to compile runguard and it makes the flow in runguard more consistent.

Work towards #luigi (https://docs.google.com/document/d/1WZRwdvJUamsczYC7CpP3ZIBU8xG6wNqYqrNJf7osxYs/edit).


nickygerritsen (approved):

eldering (approved):
Makes sense: this is from the time we were switching from resource limits to cgroups.

PR #2343, opened 2024-02-23
Allow shadowing from PC2

vmcj:
1 out of 3 systems we shadow from needs an extra syntax, given that we only run this code ourselves I think this is safe.


nickygerritsen (approved):

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -735,30 +695,30 @@ protected function validateAndUpdateContest(string $entityType, ?string $eventId
...
-            $freezeSeconds        = floor(($freezeStartSeconds % 60) / 60);
-            $freezeMilliseconds   = $freezeStartSeconds - floor($freezeStartSeconds);
+            $durationInSeconds = $durationHourModifier * (int)$durationData[2] * 3600
+                + 60 * (int)$durationData[3]
+                + (double)sprintf('%d.%03d', $durationData[4], $durationData[5] ?? 0);
+            $freezeStartSeconds = $durationInSeconds - $freezeInSeconds;
+            $freezeHour = floor($freezeStartSeconds / 3600);
+            $freezeMinutes = floor(($freezeStartSeconds % 3600) / 60);
+            $freezeSeconds = floor(($freezeStartSeconds % 60) / 60);
+            $freezeMilliseconds = $freezeStartSeconds - floor($freezeStartSeconds);
```
eldering, line 721: Please don't just reformat code if you're not touching it, and if you do, then do it in separate commits. Also, I disagree with this kind of removal of alignment spaces. It makes the code less readable.

eldering (commented):
It seems this PR contains way more changes than what the description suggests, which makes it difficult to review.

webapp/src/Service/ExternalContestSourceService.php
vmcj: I think the reformatting was done in Nicky his PR.

I added a change to his commits but this branch is (currently) mostly to make the shadowing work for tonight.

eldering: Yes, I found out afterwards, so let's have that discussion there.

As a temporary I think this is fine, but if this planned to be merged into main, then please rebase it on top of Nicky's merged branch, so that only the actual changes are shown in the PR.

PR #2344, opened 2024-02-24
C++-ify `rungard`.

meisterT:


meisterT:
This is the next steps of merging runguard and runpipe (https://docs.google.com/document/d/1WZRwdvJUamsczYC7CpP3ZIBU8xG6wNqYqrNJf7osxYs/edit).

No functional changes intended. Small commits so it should be easy-ish to follow.

judge/runguard.cc
```diff
@@ -159,37 +150,35 @@ pid_t child_pid = -1;
...
+	{"stderr",     required_argument, nullptr,         'e'},
+	{"streamsize", required_argument, nullptr,         's'},
+	{"environment",no_argument,       nullptr,         'E'},
+	{"variable",   required_argument, nullptr,         'V'},
+	{"outmeta",    required_argument, nullptr,         'M'},
+	{"verbose",    no_argument,       nullptr,         'v'},
+	{"quiet",      no_argument,       nullptr,         'q'},
+	{"help",       no_argument,       &show_help,      1 },
+	{"version",    no_argument,       &show_version,   1 },
+	{ nullptr,     0,                 nullptr,          0 }
```
eldering, line 181: Nitpick: the 1 and 0 are not aligned.

judge/runguard.cc
```diff
@@ -836,6 +812,7 @@ void setrestrictions()
 	/* Set group-id (must be root for this, so before setting user). */
 	if ( use_group ) {
 		if ( setgid(rungid) ) error(errno,"cannot set group ID to `%d'",rungid);
+		gid_t aux_groups[10];
 		aux_groups[0] = rungid;
 		if ( setgroups(1, aux_groups) ) error(errno,"cannot clear auxiliary groups");
```
eldering, line 817: Nitpick: there's no need to declare this as an array if only a single group is set, let alone one of length 10. Maybe for clarity an array is still better though:
```suggestion
		gid_t aux_groups[1];
		aux_groups[0] = rungid;
		if ( setgroups(1, aux_groups) ) error(errno,"cannot clear auxiliary groups");
```
Also, I guess we can just clear all auxiliary groups, since `rungid` is already set as the main group, so
```
		if ( setgroups(0, NULL) ) error(errno,"cannot clear auxiliary groups");
```
would suffice? (Since this is a C system call, I'm not sure the `NULL` can be replaced by `nullptr`.

judge/Makefile
```diff
@@ -13,14 +13,12 @@ judgehost: $(TARGETS) $(SUBST_FILES)
 $(SUBST_FILES): %: %.in $(TOPDIR)/paths.mk
 	$(substconfigvars)
 
-runguard: LDFLAGS := $(filter-out -pie,$(LDFLAGS))
```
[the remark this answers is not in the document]
meisterT: It was introduced in https://github.com/DOMjudge/domjudge/commit/f61a86d999ffa3c1ba13867ad3c87ae9527028cb - my understanding is that it never was relevant for runguard but for checktestdata which still uses this option (https://github.com/DOMjudge/checktestdata/blob/main/Makefile#L73) and.

judge/runguard.cc
meisterT: I think you are right, but suggest to move that change to its own PR.

vmcj (approved):
LGTM, I assume you tested this with doing a replay of an old contest?

meisterT:
> LGTM, I assume you tested this with doing a replay of an old contest?

I did some limited testing on a few interactive and non-interactive problems so far, but I would like to do more testing before merging.

meisterT:
I did now a replay of a contest with ~1000 submissions. There was a single submission where the verdict changed and it was random (with `main` and this PR).

judge/Makefile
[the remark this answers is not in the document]
meisterT: I also checked and it doesn't use `-pie` for C++ anyway:
```
g++ -g -O2 -Wall -Wformat -Wformat-security -pedantic -fstack-protector -fPIE -D_FORTIFY_SOURCE=2 -std=c++11 -DREVISION="\"g11403e5a4\"" -I../lib -I../etc -o runguard runguard.cc ../lib/lib.error.c ../lib/lib.misc.c -lcgroup
```

eldering (approved):

Issue #2152, opened 2023-09-22
The admin user should default also get a team


vmcj:
We already do this in `dev` instances, but for production we always need the team as we import the jury submissions.

This issue is mostly to describe what would be the disadvantages.

meisterT:
The jury role is sufficient (and often personalized) for problem import. So I would not do this but encourage users to create a jury account before doing so.

We could check whether there's a team associated in the import script and ask for confirmation to proceed if not?

vmcj:
> The jury role is sufficient (and often personalized) for problem import. So I would not do this but encourage users to create a jury account before doing so.
> 
Sounds reasonable

> We could check whether there's a team associated in the import script and ask for confirmation to proceed if not?

That should be doable and is a better option.

vmcj:
Reopening as the intent doesn't match the code.

Currently a jury member can't upload problems via the GUI (and neither via API see: https://github.com/DOMjudge/domjudge/blob/0e3f272bafc63cf58bb6ce1f2a07b7fb04d15a34/webapp/src/Controller/API/ProblemController.php#L57)

@meisterT what should be changed? My original question or should we let jury members allow to upload problems? For the last part I would say using API_Writer is the better choice and maybe create a custom role for `bapctools` so a jury member can't break config easily and only via certain tools?

PR #2345, opened 2024-02-25
Default add DOMjudge team to admin user

vmcj:
We already did this in development for ourselves. Because jury users can't import problems and we don't want that, this is the easiest fix.

Closes: https://github.com/DOMjudge/domjudge/issues/2152

@meisterT I know you had a different opinion in the issue but I think the code has changed since that discussion.


vmcj:
Better option is to create another role which can do the problem upload, as this was also requested during EUC.

PR #2348, opened 2024-02-27
Do not escape $@ twice in validator run script generator.

nickygerritsen:
This would escape it in the output file, which is wrong.

Found while trying to shadow PC^2 for PacNW.


meisterT (commented):
How did we not notice this before?

nickygerritsen:
Nobody dares to write validators in Java or Python 🙈 .

meisterT:
We had plenty of validators in Python before. I want to do some code archeology before merging but not today

nickygerritsen:
3e65d980 might be interesting, that's when I changed it...

eldering:
We _should_ escape it in the output file: the output file is a build script that generates a run script. The run script should contain `"$@"` to pass its input arguments along to the Java/python interpreter as argument.

So I think this should indeed be investigated more closely before merging, and a more careful explanation is needed in the commit message.

nickygerritsen:
I agree it's weird, so I did some investigation.

With:

```php
                        // no main class detection here
                        $buildscript .= "echo 'COMPARE_DIR=\$(dirname \"\$0\")' >> run\n";
                        $mainClass = basename($unescapedSource, '.java');
                        $buildscript .= "echo 'java -cp \"\$COMPARE_DIR\" $mainClass \"\\\$@\"' >> run\n";
                        $buildscript .= "chmod +x run\n";
```

This is what `build` becomes:

```bash
#!/bin/sh

javac -cp . -d . 'Validator.java'
echo '#!/bin/sh' > run
echo 'COMPARE_DIR=$(dirname "$0")' >> run
echo 'java -cp "$COMPARE_DIR" Validator "\$@"' >> run
chmod +x run
```

and this is what ends up in `run` (which gets renamed to `runjury`):

```bash
#!/bin/sh
COMPARE_DIR=$(dirname "$0")
java -cp "$COMPARE_DIR" Validator "\$@"
```

And then with:

```php
                        // no main class detection here
                        $buildscript .= "echo 'COMPARE_DIR=\$(dirname \"\$0\")' >> run\n";
                        $mainClass = basename($unescapedSource, '.java');
                        $buildscript .= "echo 'java -cp \"\$COMPARE_DIR\" $mainClass \"\$@\"' >> run\n";
                        $buildscript .= "chmod +x run\n";
```

This is what `build` becomes:

```bash
#!/bin/sh

javac -cp . -d . 'Validator.java'
echo '#!/bin/sh' > run
echo 'COMPARE_DIR=$(dirname "$0")' >> run
echo 'java -cp "$COMPARE_DIR" Validator "$@"' >> run
chmod +x run
```

and this is what ends up in `run` (which gets renamed to `runjury`):

```bash
#!/bin/sh
COMPARE_DIR=$(dirname "$0")
java -cp "$COMPARE_DIR" Validator "$@"
```

Isn't this because the echo starts with a `'` which needs no escaping for `$`?

nickygerritsen:
> > Isn't this because the echo starts with a `'` which needs no escaping for `$`?
> 
> You're right, but the complicated escaping/quoting made that difficult to spot. Can you add this explanation to the commit message, so we don't get confused again?

Does it maybe make sense to even add that to the code as a comment? So we don't have to check the commit?

eldering (approved):

nickygerritsen:
@meisterT I assume you want to check/investigate as well a bit? I still do wonder why we made that change in 3e65d9805a6e14eb0ad947ede7ee57d0ef7f7984

judge/judgedaemon.main.php
```diff
@@ -418,13 +418,15 @@ function fetch_executable_internal(
...
+                        // Note: since the $@ is within single quotes, we do not need to double escape it.
+                        $buildscript .= "echo 'java -cp \"\$COMPARE_DIR\" $mainClass \"\$@\"' >> run\n";
                         $buildscript .= "chmod +x run\n";
                         break;
                     case 'py':
                         $buildscript .= "echo '#!/bin/sh' > run\n";
                         $buildscript .= "echo 'COMPARE_DIR=\$(dirname \"\$0\")' >> run\n";
-                        $buildscript .= "echo 'python3 \$COMPARE_DIR/$source' \"\\\$@\" >> run\n";
+                        // Note: since the $@ is within single quotes, we do not need to double escape it.
+                        $buildscript .= "echo 'python3 \$COMPARE_DIR/$source' \"\$@\" >> run\n";
```
meisterT, line 429: If I am not missing anything, this is still broken and actually python verifier might have been the reason why you did https://github.com/DOMjudge/domjudge/commit/3e65d9805a6e14eb0ad947ede7ee57d0ef7f7984 

Back then you probably realized that python is broken and then "fixed" Java as well. Now, I assume PC2 had Java validators and you "fixed" python as well?

So, if you look closely, then for Java the "whole" `echo` args are surrounded by `'` while for python the final `\"\$@\"` is not quoted.

Aligning both to the Java style is the best course of action IMHO.

nickygerritsen: Ah you are right. NAC had Python scripts and that broke. Let me make them in sync.

nickygerritsen: Done!

judge/judgedaemon.main.php
```diff
@@ -418,13 +418,15 @@ function fetch_executable_internal(
...
+                        // Note: since the $@ is within single quotes, we do not need to double escape it.
+                        $buildscript .= "echo 'java -cp \"\$COMPARE_DIR\" $mainClass \"\$@\"' >> run\n";
                         $buildscript .= "chmod +x run\n";
                         break;
                     case 'py':
                         $buildscript .= "echo '#!/bin/sh' > run\n";
                         $buildscript .= "echo 'COMPARE_DIR=\$(dirname \"\$0\")' >> run\n";
-                        $buildscript .= "echo 'python3 \$COMPARE_DIR/$source' \"\\\$@\" >> run\n";
+                        // Note: since the $@ is within single quotes, we do not need to double escape it.
+                        $buildscript .= "echo 'python3 \$COMPARE_DIR/$source \"\$@\"' >> run\n";
```
meisterT, line 429:
```suggestion
                        $buildscript .= "echo 'python3 \"\$COMPARE_DIR/$source\" \"\$@\"' >> run\n";
```

Without this change it won't work with spaces in `$COMPARE_DIR` or `$source`.

meisterT (changes requested):

judge/judgedaemon.main.php
nickygerritsen: Good catch

meisterT (approved):
Looks good to me now, but let's give @eldering another chance to look at it before merging

eldering (approved):
LGTM

PR #2349, opened 2024-02-27
Release 8.2.3

vmcj:
It has been a while since we released a BugFix release. I went through `git checkout upstream/82; git log upstream/main` and checked for any commit message mentioning a bug but I might have missed some (but I would say releasing anything is better than postponing forever).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,8 +1,9 @@
 DOMjudge Programming Contest Judging System
 
-Version 8.2.3DEV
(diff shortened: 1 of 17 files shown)


eldering:
I've pushed some commits that I think should be included or as documentation can safely be included.

vmcj:
I've now also fixed the tests,

nickygerritsen (approved):

PR #2351, opened 2024-03-02
Add some standalone tests for runguard.

meisterT:
This is not only useful in general for quicker testing of runguard, but will help to gain some confidence in runguard changes as we plan to do with
https://docs.google.com/document/d/1WZRwdvJUamsczYC7CpP3ZIBU8xG6wNqYqrNJf7osxYs/edit

### DIFF
--- a/gitlab/integration.sh
+++ b/gitlab/integration.sh
@@ -93,7 +93,7 @@ if [ ! -d ${DIR}/chroot/domjudge/ ]; then
 fi
 section_end judgehost
 
-section_start more_setup "Remaining setup (e.g. starting judgedaemon)"
+section_start more_setup "Remaining setup"
 
 # Download yajsv and ccs-specs for API check.
 cd $HOME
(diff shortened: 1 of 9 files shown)


gitlab/integration.sh
```diff
@@ -130,10 +127,21 @@ set -e
 if [ $PIN_JUDGEDAEMON -eq 1 ]; then
     PINNING="-n 0"
 fi
+section_end more_setup
+
+section_start runguard_tests "Running isolated runguard tests"
```
vmcj, line 132: You're extending the test which takes the longest, can we move this to a dedicated job? Or is the extra time neglectable?

judge/runguard_test/runguard_test.sh
```diff
@@ -0,0 +1,289 @@
+#!/bin/bash
```
vmcj, line 1: I assume you're aware that the other tests use `bats` but decided that this was better?

gitlab/integration.sh
meisterT: It takes ~32s on CI, out of ~13mins. I would say it's not worth adding the complication of a dedicated job (until we can enable it on github actions).

judge/runguard_test/runguard_test.sh
meisterT: Yeah, my test is allows for more flexibility that I think was necessary here.

vmcj (approved):
LGTM but I'm not an expert on runguard.

PR #2365, opened 2024-03-15
Allow warning_message during import. Fixes #2363.

### LINKED ISSUE #2363: Allow to import contest warning
In #1765 we added a contest warning, which replaces the thing we do every year with dd381242598462b911826cf98cda57d842b2385c.
It would be great if we can add this to the contest.yaml and import and this should be super easy to do. We can then add this to all test contests.

nickygerritsen:


eldering (approved):

webapp/src/Service/ImportExportService.php
```diff
@@ -209,7 +209,7 @@ public function importContestData(mixed $data, ?string &$errorMessage = null, st
                                $data['short_name'] ?? $data['shortname'] ?? $data['short-name'] ?? $data['id']
                            ))
             ->setExternalid($contest->getShortname())
-            ->setWarningMessage($data['warning-message'] ?? null)
+            ->setWarningMessage($data['warning_message'] ?? $data['warning-message'] ?? null)
```
meisterT, line 212: Why do we have both?

nickygerritsen: We also still have `short-name` and `shortname` for example.
I am fine only doing `_` one though if others agree.

meisterT: Where is this specified anyway, or is this our own extension of the spec?

vmcj: `warning message` is not in the spec, but I think it is good to have both ways of writing it, I for example tend to always use the `format-format` if I have to do this by hand while the logical thing if you export this would be that it has the `some_format` notation.

nickygerritsen: It indeed is not in the spec

nickygerritsen: So @meisterT what do you think? Allow both or only one?

meisterT: sure, support both

meisterT (approved):

PR #2368, opened 2024-03-16
Enable medals when setting medal categories.

nickygerritsen:


vmcj:
So if we add the medals field to the contest.yml and don't have this commit they would not show up on the scoreboard?

nickygerritsen:
We currently set medals to:

- disabled if you have no public categories
- still disabled but we add the public categories if we have any.

This fixes that the second case will set it to enabled.

vmcj (approved):

PR #2371, opened 2024-03-16
Analyst shadow differences

vmcj:
This also picks some commits from when we shadowed PC2

### DIFF
--- a/webapp/src/DataTransferObject/Shadowing/ProblemEvent.php
+++ b/webapp/src/DataTransferObject/Shadowing/ProblemEvent.php
@@ -7,7 +7,7 @@ class ProblemEvent implements EventData
     public function __construct(
         public readonly string $id,
         public readonly string $name,
-        public readonly int $timeLimit,
+        public readonly ?int $timeLimit,
         public readonly ?string $label,
         public readonly ?string $rgb,
     ) {}
--- a/webapp/src/DataTransferObject/Shadowing/SubmissionEvent.php
+++ b/webapp/src/DataTransferObject/Shadowing/SubmissionEvent.php
@@ -14,6 +14,6 @@ class SubmissionEvent implements EventData
         public readonly string $teamId,
         public readonly string $time,
         public readonly ?string $entryPoint,
-        public readonly array $files,
+        public readonly ?array $files,
     ) {}
 }
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -1448,6 +1448,10 @@ class ExternalContestSourceService
                 $zipUrl = $data->files[0]->href;
                 if (preg_match('/^https?:\/\//', $zipUrl) === 0) {
                     // Relative URL, prepend the base URL.
+                    // If both the base path ends with a / and the zip URL starts with one, drop one of them
+                    if (str_ends_with($this->basePath, '/') && str_starts_with($zipUrl, '/')) {
+                        $zipUrl = substr($zipUrl, 1);
+                    }
                     $zipUrl = ($this->basePath ?? '') . $zipUrl;
                 }
 
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -522,7 +522,8 @@ class ExternalContestSourceService
             case ExternalContestSource::TYPE_CCS_API:
                 try {
                     // The base URL is the URL of the CCS API root.
-                    if (preg_match('/^(.*\/)contests\/.*/',
+                    // Proper is '^(.*\/)contests\/.*/', but PC^2 doesn't expose this (yet).
+                    if (preg_match('/^(.*\/)contest(s\/.*)?/',
                                    $this->source->getSource(), $matches) === 0) {
                         $this->loadingError      = 'Cannot determine base URL. Did you pass a CCS API contest URL?';
                         $this->cachedContestData = null;
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -621,7 +621,7 @@ class ExternalContestSourceService
 
         // Note the @vars here are to make PHPStan understand the correct types.
         $method = match ($event->type) {
-            EventType::AWARDS, EventType::TEAM_MEMBERS, EventType::ACCOUNTS => $this->ignoreEvent(...),
+            EventType::AWARDS, EventType::TEAM_MEMBERS, EventType::ACCOUNTS, EventType::PERSONS => $this->ignoreEvent(...),
             EventType::STATE => $this->validateState(...),
             EventType::CONTESTS => $this->validateAndUpdateContest(...),
             EventType::JUDGEMENT_TYPES => $this->importJudgementType(...),
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -1436,7 +1436,7 @@ class ExternalContestSourceService
                     'message' => 'No source files in event',
                 ]);
                 $submissionDownloadSucceeded = false;
-            } elseif (($data->files[0]->mime ?? null) !== 'application/zip') {
+            } elseif ($data->files[0]->mime !== null && $data->files[0]->mime !== 'application/zip') {
                 $this->addOrUpdateWarning($event, $data->id, ExternalSourceWarning::TYPE_SUBMISSION_ERROR, [
                     'message' => 'Non-ZIP source files in event',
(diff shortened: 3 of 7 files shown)


webapp/src/DataTransferObject/Shadowing/ContestEvent.php
```diff
@@ -6,10 +6,10 @@ class ContestEvent implements EventData
 {
     public function __construct(
         public readonly string $id,
-        public readonly string $name,
+        public readonly ?string $name,
```
nickygerritsen, line 9: Although this is needed with the current WF repo, I'm not sure if we actually want to do this?

There will hopefully be a fix in the CDS and otherwise we can fix the config in the repo.

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -1023,17 +1028,22 @@ protected function validateAndUpdateProblem(Event $event, EventData $data): void
...
-            'name'      => $data->name,
-            'timelimit' => $data->timeLimit,
+            'name' => $data->name,
         ];
 
+        if ($data->timeLimit !== null) {
+            $toCheckProblem['timelimit'] = $data->timeLimit;
+        }
+
+        /* Disable as PC2 can have 2 problems with the same label
```
nickygerritsen, line 1038: This is still a pending discussion, and it won't happen at the finals so I'd rather not commit this for now.

webapp/src/DataTransferObject/Shadowing/ContestEvent.php
vmcj: Created another PR for this specific case.

webapp/src/Service/ExternalContestSourceService.php
vmcj: Taken the commit out.

nickygerritsen (approved):

webapp/src/DataTransferObject/Shadowing/ProblemEvent.php
```diff
@@ -7,7 +7,7 @@ class ProblemEvent implements EventData
     public function __construct(
         public readonly string $id,
         public readonly string $name,
-        public readonly int $timeLimit,
+        public readonly ?int $timeLimit,
```
eldering, line 10: I don't think we should merge into `main` things that are basically workarounds for bugs in PC^2's implementation. I'd rather we keep that in a separate branch and make sure PC^2 fixes this. Otherwise, we'll not notice anymore when such bugs occur in other systems we shadow and I think it's better to be strict here, also because the alternative leads to code complexity creep.

nickygerritsen: Is time limit a required property? If so I agre

webapp/src/DataTransferObject/Shadowing/SubmissionEvent.php
```diff
@@ -14,6 +14,6 @@ public function __construct(
         public readonly string $teamId,
         public readonly string $time,
         public readonly ?string $entryPoint,
-        public readonly array $files,
+        public readonly ?array $files,
```
eldering, line 17: Same as above: per spec this is required, so we shouldn't merge this in `main`.

webapp/src/DataTransferObject/Shadowing/ProblemEvent.php
vmcj: > Is time limit a required property? If so I agree

It seems to be, I'll revert.

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -1435,7 +1439,7 @@ protected function importSubmission(Event $event, EventData $data): void
                     'message' => 'No source files in event',
                 ]);
                 $submissionDownloadSucceeded = false;
-            } elseif (($data->files[0]->mime ?? null) !== 'application/zip') {
+            } elseif ($data->files[0]->mime !== null && $data->files[0]->mime !== 'application/zip') {
```
eldering, line 1442: FYI: I thought this was also invalid per spec, but indeed in this case the `mime` attribute may be omitted as it is assumed to be `application/zip`, see https://ccs-specs.icpc.io/2023-06/contest_api#submissions

webapp/src/DataTransferObject/Shadowing/ProblemEvent.php
eldering: According to https://ccs-specs.icpc.io/2023-06/contest_api#problems it is.

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -1444,6 +1448,10 @@ protected function importSubmission(Event $event, EventData $data): void
                 $zipUrl = $data->files[0]->href;
                 if (preg_match('/^https?:\/\//', $zipUrl) === 0) {
                     // Relative URL, prepend the base URL.
+                    // If both the base path ends with a / and the zip URL starts with one, drop one of them
+                    if (str_ends_with($this->basePath, '/') && str_starts_with($zipUrl, '/')) {
+                        $zipUrl = substr($zipUrl, 1);
+                    }
```
eldering, line 1454: I think this is wrong (and could maybe be specified a bit more clearly in the spec). If `$zipUrl` starts with a `/`, then it should be appended after the host part of `$this->basePath` per specification of how to resolve relative URIs.

I think it would be better here to use a library or so, since simply stripping a `/` and then concatenating is not correct. Also, I wonder whether the prefix `/` is intended behaviour by PC^2.

webapp/src/DataTransferObject/Shadowing/SubmissionEvent.php
vmcj: So when we are analyst and have the analysts access policy:
We start with access:
```
{
   "endpoints": [
     { "type": "submissions", "properties": ["id","files",...]}
   ]
}
```

But during the freeze we don't have that access anymore, what fields should the CDS give?

nickygerritsen: Indeed, files are omitted during the freeze for some users, so it is not required. The spec is a bit weird in that it a field doesn’t have a ? it can still be omitted if the user doesn’t have access

webapp/src/DataTransferObject/Shadowing/ProblemEvent.php
nickygerritsen: Also depends on whether all users have access to it. I will check the current access policy when I’m at my pc to see if there is any role that doesn’t have access

webapp/src/Service/ExternalContestSourceService.php
nickygerritsen: I think pc2 is wrong currently since they don’t support multi contest in the api. They are working on this, so maybe we should indeed not commit this until then

vmcj: So also revert https://github.com/DOMjudge/domjudge/pull/2371/commits/6680f111c01276d2bb9c7861b30a7eb792e26df4 ?

webapp/src/DataTransferObject/Shadowing/SubmissionEvent.php
eldering: Hmmm, yes, that's a fair point. This is a part of the spec that's a bit annoying where things differ in a bit ugly way depending on your role and time in the contest.

So this can be left as is, but it would be good to add a note to the code as to why it's optional.

webapp/src/Service/ExternalContestSourceService.php
eldering: I'd say so, yes.

Besides that, the existing implementation doesn't handle a relative URL that starts with a `/` correctly, but that's a separate issue.

webapp/src/DataTransferObject/Shadowing/SubmissionEvent.php
nickygerritsen: Yeah adding a comment makes sense.

webapp/src/DataTransferObject/Shadowing/ProblemEvent.php
nickygerritsen: The contest information access policy is not explicit on this: it says teams can't access 'other properties' from problems. But I added a comment there to discuss.

eldering: Are the time limits not visible in the problem set, or at least on the problems page view for the teams? I'd strongly push to make this public info, at least in WF/ICPC setting.

nickygerritsen: That is why I added the comment, I'd argue the same.

PR #2372, opened 2024-03-16
Do not send a judging update event for judgements that aren't valid (yet).

### LINKED ISSUE #2370: Rejudging judgements appear in the event feed while rejudging is cancelled.
I saw this happening at least for a repeated rejudging.

nickygerritsen:
Fixes #2370.


nickygerritsen:
I wonder why we never noticed this...

vmcj (approved):

PR #2373, opened 2024-03-16
CDS doesn't always send the name & penalty_time

vmcj:
It should according to the spec, but there is no good reason to fail on this.

Most likely this will be fixed, otherwise we should merge this after we created a `wfluxor` branch.


vmcj:
@nickygerritsen will the ICPC Tools team fix this or was this because we didn't set the `id` correctly so it would send 2 contests?

nickygerritsen:
Not sure, it is waiting on icpctools/icpctools#858

PR #2374, opened 2024-03-17
Fix new DB load script.

meisterT:
In particular:
- support spaces in filenames
- print actual list of choices
- only list files as choices, and only files with suffix `.sql.gz`
- start indexing with 1 as this script is meant to be run by humans

### DIFF
--- a/sql/dj_setup_database.in
+++ b/sql/dj_setup_database.in
@@ -398,29 +398,36 @@ dump)
 	;;
 
 load)
-	choice=""
 	DUMPNAME="$2"
+	FILE=""
 	if [ -z "$DUMPNAME" ]; then
-		set -- $(ls $DATABASEDUMPDIR)
-		ind=0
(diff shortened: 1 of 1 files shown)


vmcj (approved):
I assume you tested it, and it does look easier to read.

vmcj:
>     * print actual list of choices
> 
Out of interest, what failed?

meisterT:
> I assume you tested it, and it does look easier to read.

Yep, tried to load a dump with spaces and then run into this.


> > ```
> > * print actual list of choices
> > ```
> 
> Out of interest, what failed?

The previous code didn't list any choices with indices for me at all, should it have?

vmcj:
> > I assume you tested it, and it does look easier to read.
> 
> Yep, tried to load a dump with spaces and then run into this.
> 
> > > ```
> > > * print actual list of choices
> > > ```
> > 
> > 
> > Out of interest, what failed?
> 
> The previous code didn't list any choices with indices for me at all, should it have?

yes, pretty sur it did at one point for me.

PR #2376, opened 2024-03-17
Revert "PC^2 does not send timelimits for problems, so make them opti…

vmcj:
…onal."

We want to follow the spec and not be lenient for this import error. We have a separate branch for the customizations for PC2 so when we shadow them we can easily cherry-pick those.

This reverts commit a25d24bd1a7a9b15a277d0a28b14ec803f95c615.


eldering:
Also see my other comments in #2371

vmcj:
@eldering I thinks those are all the reverts needed and the extra comment. If you agree feel free to directly merge.

eldering (approved):
LGTM

PR #2379, opened 2024-03-17
Allow an `output_display_limit` of `-1`.

meisterT:
This means unlimited and is already supported by code and mentioned in the docs, but was disallowed by the regular expression.

While doing so, also move it from 'Judging' to 'Display' options.

### DIFF
--- a/etc/db-config.yaml
+++ b/etc/db-config.yaml
@@ -111,16 +111,9 @@
             type: int
             default_value: 50000
             public: false
-            description: Maximum size of error/system output stored in the database (in bytes); use `-1` to disable any limits.
-            regex: /^[1-9]\d*$/
-            error_message: A positive number is required.
-        -   name: output_display_limit
-            type: int
-            default_value: 2000
-            public: false
-            description: Maximum size of run/diff/error/system output shown in the jury interface (in bytes); use `-1` to disable any limits.
-            regex: /^[1-9]\d*$/
-            error_message: A positive number is required.
+            description: Maximum size of error/system output stored in the database (in bytes); use `-1` to disable any limits. See `Display` / `output_display_limit` for how to control the output *shown*.
+            regex: /^([1-9]\d*|-1)$/
+            error_message: A positive number or -1 is required.
         -   name: lazy_eval_results
             type: int
             default_value: 1
@@ -208,6 +201,13 @@
 -   category: Display
     description: Options related to the DOMjudge user interface.
     items:
+        -   name: output_display_limit
+            type: int
+            default_value: 2000
+            public: false
+            description: Maximum size of run/diff/error/system output shown in the jury interface (in bytes); use `-1` to disable any limits.
+            regex: /^([1-9]\d*|-1)$/
+            error_message: A positive number or -1 is required.
         -   name: show_pending
             type: bool
             default_value: true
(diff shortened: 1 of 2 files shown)


etc/db-config.yaml
```diff
@@ -208,6 +201,13 @@
 -   category: Display
     description: Options related to the DOMjudge user interface.
     items:
+        -   name: output_display_limit
+            type: int
+            default_value: 2000
+            public: false
+            description: Maximum size of run/diff/error/system output shown in the jury interface (in bytes); use `-1` to disable any limits.
+            regex: /^([1-9]\d*|-1)$/
+            error_message: A positive number is required.
```
nickygerritsen, line 210:
```suggestion
            error_message: A positive number or -1 is required.
```

nickygerritsen (approved):

eldering:
I think one reason to have this under judging, was that you'd also see the storage limit, because it might be confusing if the display limit doesn't have effect beyond the storage limit.

etc/db-config.yaml
```diff
@@ -114,13 +114,6 @@
             description: Maximum size of error/system output stored in the database (in bytes); use `-1` to disable any limits.
             regex: /^[1-9]\d*$/
```
eldering, line 115: You probably also want to change the regex for `output_storage_limit` then, and maybe refer to it from the description of `output_diisplay_limit`.

meisterT: Done, also fixed the submissions page.

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -390,7 +390,7 @@ public function viewAction(
                     $runResult['hostname'] = $firstJudgingRun->getJudgeTask()->getJudgehost()->getHostname();
                     $runResult['judgehostid'] = $firstJudgingRun->getJudgeTask()->getJudgehost()->getJudgehostid();
                 }
-                $runResult['is_output_run_truncated'] = preg_match(
+                $runResult['is_output_run_truncated'] = $outputDisplayLimit >= 0 && preg_match(
```
eldering, line 393: Is this correct? My understanding is that `$runResult['is_output_run_truncated']` indicates whether the final displayed output is truncated or not. This could have happened at storage and/or fetch time (above). If `$outputDisplayLimit == -1`, but output was truncated at storage time, then this seems wrong.
I think the simplest and clearest would be something like:
```php
                preg_match(
                    '/\[output storage truncated after \d* B\]/',
                    (string)$runResult['output_run_last_bytes']
                ) ||
                preg_match(
                    '/\[output display truncated after \d* B\]/',
                    (string)$runResult['output_run']
                );
```

As a side note: I can't find `TRUNCATE` from https://github.com/DOMjudge/domjudge/pull/2379/files#diff-9f6bb8e6b77ae7a86ba0f4456c3803399e916c0979d8c742dfa7253987e55423R340-R346 above as a MySQL string function, only https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_truncate. Does anyone know where this comes from?

meisterT: I sent a PR to fix this https://github.com/DOMjudge/domjudge/pull/2410

PR #2380, opened 2024-03-17
Add option for ondemand mode to do a full judging in 1 buttonpress

vmcj:
Normally we would need 2 button presses, 1 to start judging and the other after the `lazy` eval finished to go further. In case we shadow onDemand we are probably more interested in full submissions as you want to see the behaviour on all testcases.

In case of an analyst they would probably also want to know how far off a participant is.


webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -494,6 +494,15 @@ public function viewAction(
                 ->getSingleScalarResult();
         }
 
+        $evalOnDemand = false;
+        $problemLazyEvalResults = $submission->getContestProblem()->getLazyEvalResults();
+        if (((int)$problemLazyEvalResults === (int)DOMJudgeService::EVAL_DEFAULT
```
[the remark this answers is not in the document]
vmcj: > I don't see the benefit of casting both to int and then doing a strict comparison. Why not just `==`?

We did in another function, so I assumed we had a reason for that.

webapp/src/Controller/Jury/SubmissionController.php
meisterT, line 1: there are some phpcs checks that fail

meisterT (commented):
Can you please add a screenshot to the PR?

vmcj:
> Can you please add a screenshot to the PR?

image

webapp/src/Controller/Jury/SubmissionController.php
vmcj: Those are now fixed.

meisterT (approved):

PR #2382, opened 2024-03-19
Refactor evalOnDemand

vmcj:
This should make the code both easier to understand and even removes a minor bug.

No functional change intended (as we would never hit that bug in practice)

### DIFF
--- a/webapp/src/Controller/API/ProblemController.php
+++ b/webapp/src/Controller/API/ProblemController.php
@@ -332,18 +332,13 @@ class ProblemController extends AbstractRestController implements QueryObjectTra
 
         $contest = $this->em->getRepository(Contest::class)->find($this->getContestId($request));
 
-        $lazyEvalResults = 0;
-        if ($contestProblemPut->lazyEvalResults !== DOMJudgeService::LAZY_DEFAULT) {
-            $lazyEvalResults = (int)$contestProblemPut->lazyEvalResults;
-        }
-
         $contestProblem = (new ContestProblem())
             ->setContest($contest)
             ->setProblem($problem)
             ->setShortname($contestProblemPut->label)
             ->setColor($contestProblemPut->rgb ?? $contestProblemPut->color)
             ->setPoints($contestProblemPut->points)
-            ->setLazyEvalResults($lazyEvalResults);
+            ->setLazyEvalResults($contestProblemPut->lazyEvalResults);
 
         $this->em->persist($contestProblem);
         $this->em->flush();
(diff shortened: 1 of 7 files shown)


webapp/src/Entity/ContestProblem.php
```diff
@@ -75,7 +76,7 @@ class ContestProblem
         options: ['comment' => 'Whether to do lazy evaluation for this problem; if set this overrides the global configuration setting', 'unsigned' => true]
     )]
     #[Serializer\Exclude]
-    private ?int $lazyEvalResults = null;
+    private int $lazyEvalResults = 0;
```
nickygerritsen, line 79: We need a migration that:

* Updates existing `null` entries to `0`
* Makes the field non nullable

webapp/src/Controller/API/ProblemController.php
```diff
@@ -332,8 +332,8 @@ public function linkProblemAction(
 
         $contest = $this->em->getRepository(Contest::class)->find($this->getContestId($request));
 
-        $lazyEvalResults = null;
-        if ($contestProblemPut->lazyEvalResults !== null) {
+        $lazyEvalResults = 0;
+        if ($contestProblemPut->lazyEvalResults !== DOMJudgeService::LAZY_DEFAULT) {
             $lazyEvalResults = (int)$contestProblemPut->lazyEvalResults;
         }
```
nickygerritsen, line 338: Why not:
```suggestion
$lazyEvalResults = ($contestProblemPut->lazyEvalResults;
```

It's always an int now with a default of 0.

webapp/src/Entity/ContestProblem.php
```diff
@@ -193,17 +194,27 @@ public function getApiColor(): ?string
...
+    public function getLazyEvalResults(): int
     {
         return $this->lazyEvalResults;
     }
 
+    public function determineOnDemand(int $config_lazy_eval_results) {
+        if ($this->lazyEvalResults === DJS::EVAL_DEMAND) {
+            return true;
+        }
+        if ($this->lazyEvalResults === DJS::EVAL_DEFAULT && $config_lazy_eval_results) {
```
nickygerritsen, line 212: Shouldn't this be
```suggestion
        if ($this->lazyEvalResults === DJS::EVAL_DEFAULT && $config_lazy_eval_results === DJS:: EVAL_DEMAND) {
```
?

vmcj: Yes

webapp/src/Controller/API/ProblemController.php
vmcj: If only I actually committed what I discussed.

webapp/src/Entity/ContestProblem.php
vmcj: Done.

nickygerritsen (approved):

PR #2384, opened 2024-03-21
Get rid of final PHPStan baseline messages 🎉

nickygerritsen:


nickygerritsen:
We might want to raise the level to 7 though, since it seems to find some stuff. I will fix some stuff in a separate PR now.

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -113,6 +113,7 @@ public function getJudgehostsAction(
     /**
      * Add a new judgehost to the list of judgehosts.
      * Also restarts (and returns) unfinished judgings.
+     * @return array<array{jobid: int, submitid: int}>
```
vmcj, line 116:
```suggestion
     *
     * @return array<array{jobid: int, submitid: int}>
```

webapp/src/Controller/API/GeneralInfoController.php
```diff
@@ -163,6 +163,8 @@ public function getUserAction(): User
 
     /**
      * Get configuration variables.
+     *
+     * @return array<string, bool|int|array<string, string>|string>
```
vmcj, line 167:
```suggestion
     * @return array<string, bool|int|string|array<string, string>>
```

webapp/src/Controller/API/AbstractRestController.php
```diff
@@ -236,6 +150,7 @@ abstract protected function getIdField(): string;
 
     /**
      * @throws NonUniqueResultException
+     * @return array<U>
```
vmcj, line 153:
```suggestion
     * @return array<U>
     * @throws NonUniqueResultException
```

vmcj (approved):
CI happy... I'm happy

One file less to care about.

meisterT (approved):

PR #2386, opened 2024-03-22
Make TLE detection for interactive problems more robust.

meisterT:
While we plan to merge runguard and runpipe (see
https://docs.google.com/document/d/1WZRwdvJUamsczYC7CpP3ZIBU8xG6wNqYqrNJf7osxYs/edit#heading=h.i7kgdnmw8qd7), this is a "minimal" change to signal (in the literal meaning of the word) to runpipe that runguard killed the program because of a timelimit.

### DIFF
--- a/judge/runpipe.cc
+++ b/judge/runpipe.cc
@@ -919,10 +919,11 @@ struct state_t {
                 if (errno != EAGAIN && errno != EWOULDBLOCK) {
                     error(errno, "failed to read from tle pipe");
                 }
+            } else if (buffer[0] == 42) {
+                logmsg(LOG_WARNING, "child indicated TLE");
+                child_indicated_timelimit = true;
+                continue;
             }
-            logmsg(LOG_WARNING, "child indicated TLE");
-            child_indicated_timelimit = true;
-            continue;
         }
 
 
--- a/judge/runpipe.cc
+++ b/judge/runpipe.cc
@@ -39,6 +39,9 @@
 // emulate it by sending a message in the SIGCHLD signal handler using an extra
 // pipe.
 //
+// Additionally, we set up a signal handler for SIGUSR1 which runguard will use
+// to indicate when it killed its child process due to a time limit.
+//
 // If the proxy is not enabled (i.e. no traffic capturing), the pipes are setup
 // like this:
 //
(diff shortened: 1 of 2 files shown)


nickygerritsen (approved):

judge/runpipe.cc
```diff
@@ -611,6 +631,46 @@ struct state_t {
...
+     static fd_t write_end = -1;
+     if (write_end != -1) {
+         error(0, "install_sigchld_handler can be called only once");
+     }
+     write_end = fds[1];
+
+     logmsg(LOG_DEBUG, "exit handler will send event using %d -> %d", write_end,
+            read_end);
+
+     signal(SIGUSR1, [](int) {
```
eldering, line 657: I didn't dive into this in detail, but I think it would be better to use `sigaction` since it specifies behaviour more explicitly. In this case I guess it's not a big deal because `SIGUSR1` should only ever be delivered once.

judge/runpipe.cc
```diff
@@ -846,6 +913,18 @@ struct state_t {
           }
           continue;
         }
+        if (fd == child_timelimit_pipe) {
+            static char buffer[1];
+            if (read(child_timelimit_pipe, buffer, 1) != 1) {
```
eldering, line 918: Would be good to also check that `buffer` contains 42. Although I see it's also not done for the code you copy&pasted from.

judge/runpipe.cc
```diff
@@ -846,6 +913,18 @@ struct state_t {
...
+                if (errno != EAGAIN && errno != EWOULDBLOCK) {
+                    error(errno, "failed to read from tle pipe");
+                }
```
eldering, line 921: I'm not sure that the fall-through behaviour when `read` has an error (or returns zero) but the error is not `EAGAIN` or `EWOULDBLOCK` is correct here: in the original code for `child_exited_pipe` it would then check if a child exited and not do anything if not, but here `child_indicated_timelimit = true` is automatically set.

eldering:
I also pushed a small documentation fix commit.

judge/runpipe.cc
meisterT: After reading https://en.cppreference.com/w/cpp/utility/program/signal I think we are fine and this code snippet is in line with the one where we handle `SIGCHLD`.

judge/runpipe.cc
meisterT: Done

judge/runpipe.cc
meisterT: Done

eldering (approved):
I'm not a 100% sure there are no weird race conditions, but I guess the proof is in the pudding...

PR #2389, opened 2024-03-22
Add a flash to the config page showing what has changed.

meisterT:
Example:
image


nickygerritsen (approved):

niemela (approved):

vmcj (changes requested):
I'm not 100% in favor of this, I think fixing that we can have html/markdown in flash messages with an optional `rich=true` would be better.

I do like the idea of this, but I might even like it already before to have it client

PR #2391, opened 2024-03-22
Show more specific warnings on config changes.

meisterT:
- Only when scoring related options changed, one needs to refresh the scoreboard cache.
- When you change a judging related option, you should rejudge affected submissions.

As a follow-up improvement we should add a rejudging filter for all relevant submissions (we already display it on individual submissions) and allow rejudging all affected with one button click.

### DIFF
--- a/webapp/src/Controller/Jury/ConfigController.php
+++ b/webapp/src/Controller/Jury/ConfigController.php
@@ -85,8 +85,26 @@ class ConfigController extends AbstractController
             }
 
             if (empty($errors)) {
-                $this->addFlash('scoreboard_refresh', 'After changing specific ' .
-                    'settings, you might need to refresh the scoreboard.');
+                $needsRefresh = false;
+                $needsRejudging = false;
+                foreach ($diffs as $key => $diff) {
+                    $category = $this->config->getCategory($key);
+                    if ($category === 'Scoring') {
+                        $needsRefresh = true;
+                    }
+                    if ($category === 'Judging') {
+                        $needsRejudging = true;
+                    }
+                }
+
+                if ($needsRefresh) {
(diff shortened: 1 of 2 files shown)


webapp/src/Controller/Jury/ConfigController.php
```diff
@@ -85,8 +85,26 @@ public function indexAction(EventLogService $eventLogService, Request $request):
...
             if (empty($errors)) {
-                $this->addFlash('scoreboard_refresh', 'After changing specific ' .
-                    'settings, you might need to refresh the scoreboard.');
+                $needsRefresh = false;
+                $needsRejudging = false;
+                foreach ($diffs as $key => $diff) {
+                    $category = $this->config->getCategory($key);
+                    if ($category === 'Scoring') {
+                        $needsRefresh = true;
+                    }
```
vmcj, line 94: Can't you add an break here as you only need 1?

webapp/src/Service/ConfigurationService.php
```diff
@@ -70,6 +70,12 @@ public function get(string $name, bool $onlyIfPublic = false)
         return $value;
     }
 
+    public function getCategory(string $name): ?string
+    {
+        $spec = $this->getConfigSpecification()[$name] ?? null;
```
vmcj, line 75: Which configs don't have a category? And why should they not have a category?

nickygerritsen (approved):

vmcj (approved):

webapp/src/Controller/Jury/ConfigController.php
meisterT: No, there is two separate booleans here.

webapp/src/Service/ConfigurationService.php
meisterT: Done.

PR #2416, opened 2024-03-24
Unique constraint for balloons 

brinky:


nickygerritsen (approved):

webapp/src/Entity/Balloon.php
```diff
@@ -53,4 +70,38 @@ public function getSubmission(): Submission
...
+    {
+        return $this->contest;
+    }
+
+    public function setContest(Contest $contest): Balloon
+    {
+        $this->contest = $contest;
+        return $this;
+    }
+
```
vmcj, line 106:
```suggestion
```

vmcj (commented):
@nickygerritsen can we add a unit test for this case?

nickygerritsen (approved):

vmcj (approved):
I think we should maybe add a test with a fixture with this situation and make sure the migration passes. But let's do this outside of this PR.

PR #2421, opened 2024-03-24
Add toggles for booleans on contest page.

### LINKED ISSUE #2420: Add toggles for all boolean values on contest page
On the contest page, there is currently one boolean option that can be toggled directly, while the other's cannot:
image

It is currently a very specific route https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Controller/Jury/ContestController.php#L345
but it would be cool if we could add toggles in a generic way to all booleans, perhaps even on other pages as well.

nickygerritsen:
Fixes #2420

image

### DIFF
--- a/webapp/templates/jury/contest.html.twig
+++ b/webapp/templates/jury/contest.html.twig
@@ -107,10 +107,7 @@
                 <tr>
                     <th>Allow submit</th>
                     <td>
-                        <form action="{{ path('jury_contest_toggle', {'contestId': contest.cid, 'type': 'submit'}) }}" method="post" class="d-inline">
-                            <input type="checkbox" data-toggle="toggle" data-size="mini" data-on="Yes" data-off="No"
-                                   name="value" {% if contest.allowSubmit %}checked{% endif %}>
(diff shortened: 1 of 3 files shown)


webapp/templates/jury/contest.html.twig
```diff
@@ -126,15 +126,30 @@
                 {% endif %}
                 <tr>
                     <th>Process balloons</th>
-                    <td>{{ contest.processBalloons | printYesNo }}</td>
+                    <td>
+                        <form action="{{ path('jury_contest_toggle', {'contestId': contest.cid, 'type': 'balloons'}) }}" method="post" class="d-inline">
```
meisterT, line 130: let's create a partial for these?

meisterT (changes requested):

webapp/templates/jury/contest.html.twig
nickygerritsen: Done

meisterT:
We should also do the same here:
image


and here:
image

PR #2422, opened 2024-03-24
Sort teams on clarification for by identifier.

nickygerritsen:
As requested by @baierjan.

### DIFF
--- a/webapp/src/Form/Type/JuryClarificationType.php
+++ b/webapp/src/Form/Type/JuryClarificationType.php
@@ -39,6 +39,12 @@ class JuryClarificationType extends AbstractType
         } else {
             /** @var Team|null $limitToTeam */
             $teams = $this->em->getRepository(Team::class)->findAll();
+
+            // Sort teams by identifier
+            usort($teams, function (Team $a, Team $b) {
+                return strnatcmp($this->getTeamIdentifier($a), $this->getTeamIdentifier($b));
+            });
+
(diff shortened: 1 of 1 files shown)


nickygerritsen:
The question is if we want this at all. Jan likes it, but does everyone?

eldering:
> The question is if we want this at all. Jan likes it, but does everyone?

I guess it depends, but I think the request makes sense: if you mainly reference teams by their IDs then it is really annoying to go through a list sorted by names.

But I also think we should be consistent: don't sort one list by ID and another by name. So maybe create a configuration setting that controls the (default) sorting attribute, both in such dropdown lists but also on pages. Then the question is: do we only apply that to teams or also affiliations, categories, problems, ...

nickygerritsen:
> > The question is if we want this at all. Jan likes it, but does everyone?
> 
> I guess it depends, but I think the request makes sense: if you mainly reference teams by their IDs then it is really annoying to go through a list sorted by names.
> 
> But I also think we should be consistent: don't sort one list by ID and another by name. So maybe create a configuration setting that controls the (default) sorting attribute, both in such dropdown lists but also on pages. Then the question is: do we only apply that to teams or also affiliations, categories, problems, ...

I like this idea of having a global setting, but I'm not sure about sorting all entities.

nickygerritsen:
See #2817 for the proper way to do this.

Issue #2276, opened 2023-12-06
Improve visibility of "Scaling and load testing" information in DOMjudge documentation


taoky:
### Description of the problem
DOMjudge repo wiki [Scaling and load testing](https://github.com/DOMjudge/domjudge/wiki/Scaling-and-load-testing) contains important information for large contest configuration (including setting `max_children` value). However, this is not highlighted in [DOMjudge manual](https://www.domjudge.org/docs/manual/index.html). Though it gives a link to wiki in `doc/manual/overview.rst`, it could be neglected very easily -- As far as I know, some server admins for ICPC Asia Regional Contest have reported that they are unaware of this resource.

### Your environment

- `main` branch of DOMjudge/domjudge

### Steps to reproduce

1. View domserver installation and checklist pages: 
    - https://www.domjudge.org/docs/manual/main/install-domserver.html
    - https://www.domjudge.org/docs/manual/main/quick-install.html

### Expected behaviour
`max_children` setting (or "Scaling and load testing" wiki page) should be mentioned in relevant sections of the manual.

### Actual behaviour
There is no clear mention or guidance on these critical settings for configuring large-scale contests within the main DOMjudge manual.

### Any other information that you want to share?
N/A

PR #2424, opened 2024-03-24
Scaling wiki manual

vmcj:
Closes: https://github.com/DOMjudge/domjudge/issues/2276

### DIFF
--- a/doc/manual/install-domserver.rst
+++ b/doc/manual/install-domserver.rst
@@ -179,6 +179,20 @@ this approach is that the webserver is not aware of the actual client IP. This
 means that access logs for the webserver will still report the IP of the proxy
 or loadbalancer.
 
+Scaling the domserver
+---------------------
+
+Documenting all possible ways to properly size the domserver is out of scope for this
+manual. But properly think about the memory management for the PHP-FPM children with
+~40 per gigabyte of memory, 500 for a 16Gb system.
+
+* Consider the amount of submissions you expect at any given time, does the load stay uniform.
+* Should you expect all teams to solve the easy problems in the first hour.
+  as a rule of thumb you can expect: ``#testcases * #submissions/teams`` as extra load. Similar
+  for the harder problems (which look simple) where multiple wrong submissions can be expected.
+* For more info see the `wiki <https://github.com/DOMjudge/domjudge/wiki/Scaling-and-load-testing>`_
+
+
 Log in to DOMjudge
 ------------------
 The DOMserver should now be operational. You can access the web application
--- a/doc/manual/install-domserver.rst
+++ b/doc/manual/install-domserver.rst
@@ -186,7 +186,7 @@ Documenting all possible ways to properly size the domserver is out of scope for
 manual. But properly think about the memory management for the PHP-FPM children with
 ~40 per gigabyte of memory, 500 for a 16Gb system.
 
-* Consider the amount of submissions you expect at any given time, does the load stay uniform.
+* Consider the amount of submissions you expect at any given time, does the load stay uniform?
 * Do you expect all teams to solve the easy problems in the first hour?
   as a rule of thumb you can expect: ``#testcases * #submissions/teams`` as extra load. Similar
   for the harder problems (which look simple) where multiple wrong submissions can be expected.
--- a/doc/manual/install-domserver.rst
+++ b/doc/manual/install-domserver.rst
@@ -128,7 +128,7 @@ settings. Reload the web server for changes to take effect.
 
 An nginx webserver configuration snippet is also provided in
 ``etc/nginx-conf``.  You still need ``htpasswd`` from ``apache2-utils``
-though. To use this configuration, perform the following steps::
+though. To use this configuration, perform the following steps:
 
 .. parsed-literal::
 
@@ -139,7 +139,7 @@ though. To use this configuration, perform the following steps::
   service php\ |phpversion|-fpm reload
   service nginx reload
 
-On Fedora, use the following nginx configuration steps::
+On Fedora, use the following nginx configuration steps:
 
 .. parsed-literal::
 
--- a/doc/manual/install-domserver.rst
+++ b/doc/manual/install-domserver.rst
@@ -183,8 +183,8 @@ Scaling the domserver
 ---------------------
 
 Documenting all possible ways to properly size the domserver is out of scope for this
-manual. But properly think about the memory management for the PHP-FPM children with
-~40 per gigabyte of memory, 500 for a 16Gb system.
+manual. But properly think about the memory management for the PHP-FPM children, we advice
+40 FPM children per gigabyte of RAM memory so around 500 for a 16Gb system.
 
 * Consider the amount of submissions you expect at any given time, does the load stay uniform?
 * Do you expect all teams to solve the easy problems in the first hour?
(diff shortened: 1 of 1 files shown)


doc/manual/install-domserver.rst
```diff
@@ -179,6 +179,20 @@ this approach is that the webserver is not aware of the actual client IP. This
...
 or loadbalancer.
 
+Scaling the domserver
+---------------------
+
+Documenting all possible ways to properly size the domserver is out of scope for this
+manual. But properly think about the memory management for the PHP-FPM children with
+~40 per gigabyte of memory, 500 for a 16Gb system.
+
+* Consider the amount of submissions you expect at any given time, does the load stay uniform.
```
eldering, line 189:
```suggestion
* Consider the amount of submissions you expect at any given time, does the load stay uniform?
```

doc/manual/install-domserver.rst
```diff
@@ -179,6 +179,20 @@ this approach is that the webserver is not aware of the actual client IP. This
 means that access logs for the webserver will still report the IP of the proxy
 or loadbalancer.
 
+Scaling the domserver
+---------------------
+
+Documenting all possible ways to properly size the domserver is out of scope for this
+manual. But properly think about the memory management for the PHP-FPM children with
+~40 per gigabyte of memory, 500 for a 16Gb system.
```
eldering, line 187: This is not clearly phrased. Does it mean to say that 40 children consume 1 GB of memory? Is this a suggestion for how many to configure?

doc/manual/install-domserver.rst
```diff
@@ -179,6 +179,20 @@ this approach is that the webserver is not aware of the actual client IP. This
...
+
+* Consider the amount of submissions you expect at any given time, does the load stay uniform.
+* Should you expect all teams to solve the easy problems in the first hour.
```
eldering, line 190:
```suggestion
* Do you expect all teams to solve the easy problems in the first hour?
```

doc/manual/install-domserver.rst
```diff
@@ -179,6 +179,20 @@ this approach is that the webserver is not aware of the actual client IP. This
...
+  as a rule of thumb you can expect: ``#testcases * #submissions/teams`` as extra load. Similar
```
eldering, line 191:
```suggestion
  As a rule of thumb you can expect: ``#testcases * #submissions/teams`` as extra load. Similar
```
Not really clear what "extra load" means here.

doc/manual/install-domserver.rst
vmcj: Its copied from the php pool snippet, and it should say something about `total amount of memory in gigabytes divided by 40`, I fully agree its not clearly phrased but I can't think of a better fix which also reads natural.

doc/manual/install-domserver.rst
vmcj: What is intended is to address the POST per testcase in a very short amount of time.

doc/manual/install-domserver.rst
```diff
@@ -179,6 +179,22 @@ this approach is that the webserver is not aware of the actual client IP. This
 means that access logs for the webserver will still report the IP of the proxy
 or loadbalancer.
 
+Scaling the domserver
+---------------------
+
+Documenting all possible ways to properly size the domserver is out of scope for this
+manual. But properly think about the memory management for the PHP-FPM children, we advice
+40 FPM children per gigabyte of RAM memory so around 500 for a 16Gb system.
```
ubergeek42, line 187: I wonder if these numbers are still accurate? I think when I came up with this number I just ran htop on a running server (maybe make it busy by forcing some judgehosts to do some rejudging) and noted the rss for the fpm processes, and rounded to some nice number. But maybe it'd be good to double check that.

doc/manual/install-domserver.rst
```diff
@@ -179,6 +179,22 @@ this approach is that the webserver is not aware of the actual client IP. This
...
+
+* Consider the amount of submissions you expect at any given time, does the load stay uniform?
+* Do you expect all teams to solve the easy problems in the first hour?
+  As a rule of thumb you can expect: ``#testcases * #submissions/teams`` as extra load (HTTP traffic
+  from the judgehost reporting the results per testcase additionally on the team refreshing the page
+  to know the verdict). Similar for the harder problems (which look simple) where multiple wrong
```
ubergeek42, line 193: This is really hard to understand. I'm also not sure why dividing by the number of teams?
Maybe something like:
```suggestion
    Do not forget to account for "hidden load", i.e. requests that the judging system makes to the domserver. You can expect ``#testcases * #submissions/teams`` **requests/sec (or what unit is this?)** that you need to account for. Additionally, the scoreboard pages for spectators and teams auto-refresh every 30s (**not sure if this is still true, but I suspect it is**), so every 30 teams = 1 req/sec minimum (on top of teams refreshing manually to know the verdict from their submission).
```

I'm not sure how much to write here, but lazy evaluation is a big factor too.

ubergeek42 (approved):
left some comments, but anything in the docs is better than nothing in the docs, and at least there's the link to the scaling wiki page.

doc/manual/install-domserver.rst
vmcj: the `/`was not meant for dividing but more a `or` at this point the number of submissions is equal to the number of teams (1 per team for the simple one)

I fully agree that your way of reading it makes more sense as it mixes up 2 uses for the `/` symbol.

doc/manual/install-domserver.rst
vmcj: I think I can do that after the finals, sounds like a reasonable way for testing.

tuupke: Though I think 40/GiB is fine (we use a similar metric at $work, though the workload there is quite different) it might be better to change this to number of requests per second. The section title (which is great!) is regarding scaling domserver, but the body of the text is about how to fully utilize your/a server by running domserver. The wiki-link to 'scaling and load testing' does a better job, though even that link is inconsistent with the statement here (taking 70MiB as a 'more modern' number).

To me it makes more sense to base the requirements of the domserver on the (maximum) number of concurrent users and/or requests, because the number of fpm worker is the limiting factor of how many requests domserver can handle concurrently. Deriving a baseline for this is also much easier since we can simply calculate the number of concurrent (php) requests from the access log. (timestamp + request duration)

Having said that. I think this is fine for now. Lets use the finals to gather some data.

nit: isn't it `we advise` instead of `we advice`?

doc/manual/install-domserver.rst
tuupke: I agree with @nickygerritsen, dropping `/teams` makes it more legible here. To iterate on ubergeek's suggestion

```suggestion
Do not forget to account for "hidden load", i.e. requests that the judging system makes to the domserver. You can expect ``#testcases * #submissions`` requests that you need to account for. Not only are these requests spread out over the duration of the contest but they can also be spread out unevenly. The 'middle three' hours of a full five hour contest usually see fewer submissions than the first and last hours. 

This "hidden load" can be mitigated by keeping the "lazy evaluation" option enabled. Lazy evaluation stops whenever the judgehost decides a verdict can be reached. e.g. a submission who's verdict on the first sample testcase is "Wrong Answer" does not need to be checked against the other testcases. Lazy evaluation helps when it is expected that many submissions fail which is often the case.

Additionally, the scoreboard pages for spectators and teams auto-refresh every 30s, so every 30 teams = 1 req/sec minimum (on top of teams refreshing manually to know the verdict from their submission).
```

tuupke (approved):

PR #2428, opened 2024-03-24
Put `lib/vendor` close to the webapp

vmcj:
The files have nothing to do with binaries on the system but are only relevant for executable code from DOMjudge (which has no clearly defined place in FHS).

This was discussed in a private channel with the maintainers: https://domjudge-org.slack.com/archives/GJ2JX7B7X/p1709760201078489?thread_ts=1709540681.476949&cid=GJ2JX7B7X

The configure file is not cleaned up yet to clearly show this choice in the next minor release. If there is no new discussion this can be changed (as in removed) in the next major release.

I checked in a local container and it seems to work but I rather first discuss if this is the proper fix to handle this change in FHS.


thijskh:
I agree with the change as it's an improvement. I do however think we should just flatten it further: I perceive no useful split between "libdir" and "datadir" as it's used currently in the domserver and I would just install all those things at the same root (which might be libdir).

The whole distinction between `lib` and `share` in FHS is dated and quickly becoming irrelevant (if it was ever relevant in the first place..), imho.

I think the RH effect of the webapp ending up in `/run/webapp` and other parts of the app under `/usr` shows to me what we currently do is not what I'd expect.

vmcj:
> I agree with the change as it's an improvement. I do however think we should just flatten it further: I perceive no useful split between "libdir" and "datadir" as it's used currently in the domserver and I would just install all those things at the same root (which might be libdir).
> 

I'm fine with implementing but can you state where you want those things to endup (so quote me and just fill the right values)
```
#  * documentation.......: /usr/local/share/doc/domjudge
#  * domserver...........:
#     - bin..............: /usr/local/bin
#     - etc..............: /usr/local/etc/domjudge
#     - lib..............: /usr/local/lib/domjudge
#     - libvendor........: /usr/local/share/domjudge/lib/vendor
#     - log..............: /usr/local/var/log/domjudge
#     - run..............: /usr/local/var/run/domjudge
#     - sql..............: /usr/local/share/domjudge/sql
#     - tmp..............: /tmp
#     - webapp...........: /usr/local/share/domjudge/webapp
#     - example_problems.: /usr/local/share/domjudge/example_problems
#     - database_dumps...: /usr/local/var/lib/domjudge/db-dumps
```

eldering:
Either way: we must make sure that the system works with any randomly set path for paths which are configurable. So if someone configures `libvendordir` and `webappdir` like this (as is currently possible):

> ```
> #  * domserver...........:
> #     - lib..............: /usr/local/lib/domjudge
> #     - libvendor........: /my/weird/directory/lib/vendor
> #     - webapp...........: /random/other/dir/share/domjudge/webapp
> ```

then that must work, or it must not be possible to configure these paths.

If I have some time, I'm actually planning to write some build tests, that check this among other things.

nickygerritsen:
> Proposal: remove libvendordir setting and put it in `webapp/vendor`

This would also remove a customization we need to do after every Symfony upgrade, so I would be in favor.

vmcj:
> then that must work, or it must not be possible to configure these paths.
> 

I propose to go for that 2nd option, I think spending your developer time on this (as you're the expert on this together with @thijskh IMHO) is not worth it. I personally prefer that that time is spend on other code (as I don't really see the usecase to split here).

Probably this breaks the CI still as @nickygerritsen mentioned a file which also needs changes.

nickygerritsen:
If we do this, I would just move the thing alltogether, also in `composer.json` for example. Then libvendor can be completely removed from any configure scripts or makefiles, since it is already in the right place.

eldering:
> If we do this, I would just move the thing alltogether, also in `composer.json` for example. Then libvendor can be completely removed from any configure scripts or makefiles, since it is already in the right place.

Ok. I had the impression that in the past there were some other programs depending on this, but if that's not the case anymore, then sure.

vmcj:
> > If we do this, I would just move the thing alltogether, also in `composer.json` for example. Then libvendor can be completely removed from any configure scripts or makefiles, since it is already in the right place.
> 
> Ok. I had the impression that in the past there were some other programs depending on this, but if that's not the case anymore, then sure.

So you're also fine with just removing this completely from `configure.ac`? It would be much simpler in that case if we don't display it there at all but that might be `BC` for people who had the opinion that those `vendor` files should be somewhere else.

thijskh:
It seems highly unlikely to me that there's any deployer counting on this being configurable.

eldering:
> > > If we do this, I would just move the thing alltogether, also in `composer.json` for example. Then libvendor can be completely removed from any configure scripts or makefiles, since it is already in the right place.
> > 
> > Ok. I had the impression that in the past there were some other programs depending on this, but if that's not the case anymore, then sure.
> 
> So you're also fine with just removing this completely from `configure.ac`? It would be much simpler in that case if we don't display it there at all but that might be `BC` for people who had the opinion that those `vendor` files should be somewhere else.

Yes, I think we must then remove it from `configure.ac`: if it were still possible to modify there, then things would break.

vmcj:
> If we do this, I would just move the thing alltogether, also in `composer.json` for example. Then libvendor can be completely removed from any configure scripts or makefiles, since it is already in the right place.

@nickygerritsen it seems that it would put it next to webapp, not inside webapp. So we keep the problem for the copying for everything in webapp/public, did I do something wrong or are those indeed the proper locations (`/vendor` & `/webapp`)?

nickygerritsen:
Hmmm maybe we should move the composer.json to webapp as well? Then it should be standard again

vmcj:
> Hmmm maybe we should move the composer.json to webapp as well? Then it should be standard again

That would fix 1 issue yes, not sure if it introduces other ones.

webapp/Makefile
```diff
@@ -9,37 +9,83 @@ include $(TOPDIR)/Makefile.global
...
 	-rm -rf public/bundles/nelmioapidoc
 	mkdir -p public/bundles/nelmioapidoc
-	cp -R ../lib/vendor/nelmio/api-doc-bundle/Resources/public/* public/bundles/nelmioapidoc/
+	cp -R ../webapp/vendor/nelmio/api-doc-bundle/Resources/public/* public/bundles/nelmioapidoc/
 
 clean-l:
 	-rm -rf public/bundles/nelmioapidoc
 
-install-domserver:
+# Change baseDir in composer autogenerated files
```
thijskh, line 40: Is this still needed?

webapp/Makefile
```diff
@@ -9,37 +9,83 @@ include $(TOPDIR)/Makefile.global
...
 	$(INSTALL_DIR) $(DESTDIR)$(domserver_webappdir);
-	for d in bin config migrations public resources src templates tests ; do \
+	for d in bin config migrations public resources src templates tests vendor; do \
 		$(call install_tree,$(DESTDIR)$(domserver_webappdir),$$d) ; \
 	done
+	$(call fix_composer_paths,$(DESTDIR)$(domserver_webappdir)/vendor)
 # Change webapp/public/doc symlink
 	rm $(DESTDIR)$(domserver_webappdir)/public/doc
 	ln -s $(domjudge_docdir) $(DESTDIR)$(domserver_webappdir)/public/doc
 # Now change all relative symlinks in webapp/public to their correct paths
```
thijskh, line 59: Is this stil needed?

vmcj: I tested and it seems to work without, lets see if CI finds further issues.

gitlab/unit-tests.sh
```diff
@@ -20,7 +20,7 @@ echo "UPDATE user SET teamid = 1 WHERE userid = 1;" | mysql domjudge_test
 cp webapp/.env.test /opt/domjudge/domserver/webapp/
 
 # We also need the composer.json for PHPunit to detect the correct directory.
-cp composer.json /opt/domjudge/domserver/
+cp webapp/composer.json /opt/domjudge/domserver/
```
nickygerritsen, line 23: Shouldn't this then move to `/opt/domjudge/domserver/webapp`?

phpstan.dist.neon
```diff
@@ -22,4 +22,4 @@ parameters:
             message: "#Method .* return type has no value type specified in iterable type array#"
             path: webapp/src/DataFixtures/Test
 includes:
-    - lib/vendor/phpstan/phpstan-doctrine/extension.neon
+    - webapp/vendor/phpstan/phpstan-doctrine/extension.neon
```
nickygerritsen, line 25: We might (later) want to move this file and the phpunit file to webapp as well? Since they all test webapp only anyway.

gitlab/unit-tests.sh
vmcj: Yes, not sure why tests did work.
```suggestion
cp webapp/composer.json /opt/domjudge/domserver/webapp/
```

phpstan.dist.neon
vmcj: It would be nice to also run PHPStan on the judgedaemon but I suspect we'll never reach that step so fine by me.

nickygerritsen (approved):

PR #2429, opened 2024-03-25
Revert "Make the summary footer sticky on desktop"

vmcj:
This broke `main`

This reverts commit 1da1ff5cba55d05d7292f2df539609779a1056d3.

### DIFF
--- a/webapp/public/style_domjudge.css
+++ b/webapp/public/style_domjudge.css
@@ -612,18 +612,6 @@ tr.ignore td, td.ignore, span.ignore {
     background-color: #05264c;
 }
 
-.scoreboard .problem-summary {
-    text-align: left;
-}
-
-.scoreboard .summaryline {
-    border-top: 2px solid black;
-}
-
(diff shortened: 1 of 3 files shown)


nickygerritsen (approved):

gitlab/ci/webstandard.yml
```diff
@@ -1,8 +1,5 @@
 webstandard_check_role:
   image: domjudge/gitlabci:2.1
-  only:
```
meisterT, line 3: is this part intentional?

vmcj: No, the commit is marked as `[delete] ...` which means that before merging I would remove that commit. It makes that on gitlab we would also run the extra tests for admin as that's why I broke `main`, we only test a subset of the pages in the `webstandard` job to speed up the CI, when I set it up that way I accepted the small risk that we would break `main` as what happened now.

When we move to `GHA` we can test everything as they make it possible to stop matrix jobs when one of them fails.

PR #2432, opened 2024-03-28
Fix reuploading problemset.

nickygerritsen:
Also fix uploading using PUT and some small bugs in the contest import script wrt problemset upload.

### DIFF
--- a/webapp/src/EventListener/ParseFilesForPutListener.php
+++ b/webapp/src/EventListener/ParseFilesForPutListener.php
@@ -1,0 +1,44 @@
+<?php declare(strict_types=1);
+
+namespace App\EventListener;
+
+use Riverline\MultiPartParser\Converters\HttpFoundation;
+use Symfony\Component\EventDispatcher\Attribute\AsEventListener;
+use Symfony\Component\HttpFoundation\File\UploadedFile;
+use Symfony\Component\HttpKernel\Event\ControllerEvent;
+
+#[AsEventListener(priority: 1)]
+class ParseFilesForPutListener
+{
+    public function __invoke(ControllerEvent $event): void
+    {
+        // When we have a PUT request, the files on the request will not be filled, since PHP doesn't
+        // fill the $_FILES array. We can manually parse the request to fill it using a library.
+        // This code is based on https://github.com/symfony/symfony/issues/36409#issuecomment-612442260.
+        $request = $event->getRequest();
+        if ($request->isMethod('PUT')) {
+            $document = HttpFoundation::convert($request);
+            // Not all PUT requests are multipart, for example JSON PUT requests aren't. So check
+            // to see if this is a multipart request and otherwise do nothing.
+            if (!$document->isMultiPart()) {
+                return;
+            }
+
+            foreach ($document->getParts() as $part) {
+                if (!$part->isFile()) {
+                    continue;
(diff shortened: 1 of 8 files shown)


webapp/src/EventListener/ParseFilesForPutListener.php
```diff
@@ -0,0 +1,42 @@
...
+{
+    public function __invoke(ControllerEvent $event): void
+    {
+        // When we have a PUT request, the files on the request will not be filled, since PHP doesn't
+        // fill the $_FILES array. We can manually parse the request to fill it using a library.
+        // This code is based on https://github.com/symfony/symfony/issues/36409#issuecomment-612442260.
+        $request = $event->getRequest();
+        if ($request->isMethod('PUT')) {
+            $document = HttpFoundation::convert($request);
+            if (!$document->isMultiPart()) {
```
meisterT, line 21: what is happening here?

nickygerritsen: On that line or the whole file? The whole file I tried to explain in the comment and that line... checks if this request was actually a multipart upload, which is how browsers / HTTP do file uploads.

meisterT: This line. When can this be a non-multipart request but end up here? And should this silently return in this case?

nickygerritsen: If you do a JSON PUT for example? I will add a comment.

webapp/tests/Unit/Controller/API/ContestControllerAdminTest.php
```diff
@@ -200,6 +200,23 @@ public function testProblemsetManagement(): void
         $callbackData = ob_get_clean();
         self::assertEquals(file_get_contents($problemsetFile), $callbackData);
 
+        // Upload the problemset again, this time using PUT to also test that
+        $problemsetFile = __DIR__ . '/../../../../../webapp/public/doc/logos/DOMjudgelogo.pdf';
+        $problemset = new UploadedFile($problemsetFile, 'DOMjudgelogo.pdf');
+        $this->verifyApiJsonResponse('PUT', $url . '/problemset', 204, $this->apiUser, null, ['problemset' => $problemset]);
+
+        // Verify we still have a problemset
+        $object = $this->verifyApiJsonResponse('GET', $url, 200, $this->apiUser);;
```
vmcj, line 209:
```suggestion
        $object = $this->verifyApiJsonResponse('GET', $url, 200, $this->apiUser);
```

webapp/src/EventListener/ParseFilesForPutListener.php
nickygerritsen: Does this added comment make it more clear?

mpsijm (approved):
Can confirm that (re-)uploading the PDF with `POST` (or `PUT`) works on my machine with https://github.com/RagnarGrootKoerkamp/BAPCtools/commit/b94d81efe7a8a0a5b396e8be080ced367a153d2c :smile: 

(I now have trouble with reuploading problem ZIPs though, but I'll ~~#blamenicky~~ ask Nicky about that separately :stuck_out_tongue:)

meisterT (approved):

PR #2435, opened 2024-03-29
Bats ci failure

vmcj:


.github/jobs/configure-checks/all.bats
```diff
@@ -131,10 +131,11 @@ compile_assertions_finished () {
 
 @test "Install GNU C/C++ compilers" {
     # The test above already does this
-    # repo-remove clang
-    # repo-install gcc g++ libcgroup-dev
-    # compiler_assertions
-    # compile_assertions_finished
+    skip
```
meisterT, line 134: if we skip the test why not remove it? or does skip mean something different?

vmcj: It's mostly there to annotate why we don't currently have to test with it. I'm fine with removing it.

meisterT (approved):

PR #2436, opened 2024-03-29
Show toggles for contest booleans at problems page.

nickygerritsen:


meisterT (approved):
Looks good, I think we can re-use the same partial on jury/contest.html.twig

nickygerritsen:
Added that as well.

meisterT (approved):

meisterT:
I hopefully fixed the webstandard complaints and force pushed to your branch,

PR #2447, opened 2024-04-06
Initial code for a mobile friendly scoreboard.

nickygerritsen:
Missing:

* ❤️.
* Probably some logic based on whether things are enabled (like problems). Affiliation display logic should be mostly there.
* Extensive testing for non-finals settings.
* Maybe some cleanup of the code/logic.
* Maybe not do this on non public scoreboards?

But we want to use this for the public scoreboard in Luxor, so hence this initial PR.

### DIFF
--- a/webapp/templates/partials/scoreboard_table.html.twig
+++ b/webapp/templates/partials/scoreboard_table.html.twig
@@ -479,8 +479,12 @@
         {% endif %}
     </tr>
         <tr style="height: 32px;">
-            <td class="no-border"/>
-            <td colspan="2">
+            {% if showAffiliationLogos %}
+                {% set problemSpan = 3 %}
+            {% else %}
+            {% set problemSpan = 2 %}
+            {% endif %}
+            <td colspan="{{ problemSpan }}">
                 {% for problem in problems %}
                     {% set matrixItem = scoreboard.matrix[score.team.teamid][problem.probid] %}
                     {{ problem | problemBadgeMaybe(matrixItem) }}
--- a/webapp/templates/partials/scoreboard_table.html.twig
+++ b/webapp/templates/partials/scoreboard_table.html.twig
@@ -28,7 +28,7 @@
     </style>
 {% endif %}
 
-<table class="d-none d-md-block scoreboard center {% if jury %}scoreboard_jury{% endif %}">
+<table class="d-none d-md-table scoreboard center {% if jury %}scoreboard_jury{% endif %}">
 
     {# output table column groups (for the styles) #}
     <colgroup>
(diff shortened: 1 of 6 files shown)


webapp/templates/partials/scoreboard_table.html.twig
```diff
@@ -419,9 +588,46 @@
     {% endfor %}
 </style>
 <script>
-    document.querySelectorAll(".forceWidth:not(.toolong)").forEach(el => {
+    document.querySelectorAll(".desktop-scoreboard .forceWidth:not(.toolong)").forEach(el => {
         if (el instanceof Element && el.scrollWidth > el.offsetWidth) {
             el.classList.add("toolong");
         }
     });
+    window.resizeMobileTeamNames = function() {
```
meisterT, line 596: should we move this to our JS file so it is not loaded every time?

nickygerritsen: Good idea, should be done now.

meisterT:
> Maybe not do this on non public scoreboards?

Agreed. 

What is strange: I looked now on my phone (and not with chrome devtools) and it is way too wide... but only until the first refresh, then it's fine. Do we fix the width too late?

nickygerritsen:
Might be that I broke something when moving this to the js file. Will check

meisterT:
I quickly looked into the hearts. It is non-trivial to re-use the same code, so we might have to duplicate it.

nickygerritsen:
> Might be that I broke something when moving this to the js file. Will check

Should be fixed now, it indeed executed the javascript before the page was loaded, which broke when I moved the code to the .js file.

It now also only shows the mobile scoreboard on the public page.

vmcj:
@nickygerritsen you want to retarget this to `wfastana`?

nickygerritsen:
I rebased it on Astana then rebased Astana on this so it's already there. Changing the target will tell you hopefully there are no changes

meisterT:
For the record: Pending submissions didn't work in the end, but we may have broken them by the triangle thingy in the jury interface.

mpsijm:
Looking quite fancy already 😄 Some feedback: there's some empty space below the country flag, that can be used for the list of problems. In the past [BAPC](https://2024.bapc.eu/scoreboard/), there were 13 problems, of which the 13th was on a new line, but it would've fit if this empty space had been used:

Screenshot

Of course, there will always be a number of problems where the last one will have to be on a new line, but on mobile, every little bit of space helps 😄

nickygerritsen:
@mpsijm you seem to have frontend skills. Do you want to try this?
I can rebase the branch on main and target it to main

nickygerritsen:
> Looking quite fancy already 😄 Some feedback: there's some empty space below the country flag, that can be used for the list of problems. In the past [BAPC](https://2024.bapc.eu/scoreboard/), there were 13 problems, of which the 13th was on a new line, but it would've fit if this empty space had been used:
> 
> Screenshot
> 
> Of course, there will always be a number of problems where the last one will have to be on a new line, but on mobile, every little bit of space helps 😄
> 
> P.S. note that the target branch is still `wfluxor` 🙂

I fixed this, it now goes below the flags.

meisterT (approved):

PR #2449, opened 2024-04-06
Mask the teams in the ICPC online mirror mode API

vmcj:


webapp/src/Controller/API/TeamController.php
```diff
@@ -49,6 +49,14 @@ public function __construct(
         parent::__construct($entityManager, $dj, $config, $eventLogService);
     }
 
+    private function mask_ranking(?Response $normalResponse, string $entity="team"): ?Response
```
nickygerritsen, line 52: Formatting as reported by phpcs, but I would also call the method `maskRanking`

nickygerritsen (approved):

webapp/src/Controller/API/TeamController.php
meisterT, line 52: Instead of using a function let's inline the check. It's basically the same amount of code and makes it more clear what is happening.

vmcj: So duplicate the code in this case?

meisterT: It's basically one statement that needs to be duplicated, so I prefer the clarity on what's actually happening over reading another function

vmcj: I've removed the function now.

vmcj: This is now changed.

nickygerritsen (approved):

meisterT (approved):

PR #2452, opened 2024-04-07
Simplify and clean up submission page.

meisterT:
This organizes the flow of information to change the focus on the important things.

Before:
image

After:
image


eldering:
I think the header with team, problem, timelimit etc is a bit too compact. Some more space in between like before would be better.

nickygerritsen:
> I think the header with team, problem, timelimit etc is a bit too compact. Some more space in between like before would be better.

You mean horizontal spacing right? I think I agree, it's less clear now if the icons belong to the item before or after.

PR #2459, opened 2024-04-14
Check for minimum password length upon self registration.

### LINKED ISSUE #2458: Self-registering allows weak password

When I self-register, I can use a password that consists of only 8 lowercase characters, while if I change/set a password from the jury interface (for the very same user I just created) it tells me that the password requires at least 10 characters.

### Your environment

DOMjudge `wfluxor-online` branch.

### Expected behaviour

I'd expect the same password policy to be applied.

nickygerritsen:
Fixes #2458.

### DIFF
--- a/webapp/src/Controller/SecurityController.php
+++ b/webapp/src/Controller/SecurityController.php
@@ -2,6 +2,7 @@
 
 namespace App\Controller;
 
+use App\Controller\Jury\UserController;
 use App\Entity\Team;
 use App\Entity\TeamAffiliation;
 use App\Entity\TeamCategory;
@@ -12,6 +13,8 @@ use App\Service\DOMJudgeService;
 use Doctrine\ORM\EntityManagerInterface;
 use Ramsey\Uuid\Uuid;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\DependencyInjection\Attribute\Autowire;
+use Symfony\Component\Form\FormInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\HttpException;
@@ -25,7 +28,9 @@ class SecurityController extends AbstractController
     public function __construct(
         private readonly DOMJudgeService $dj,
         private readonly ConfigurationService $config,
-        private readonly EntityManagerInterface $em
+        private readonly EntityManagerInterface $em,
+        #[Autowire(param: 'min_password_length')]
+        private readonly int $minimumPasswordLength,
     ) {}
 
     #[Route(path: '/login', name: 'login')]
@@ -103,7 +108,12 @@ class SecurityController extends AbstractController
         $registration_form->handleRequest($request);
         if ($registration_form->isSubmitted() && $registration_form->isValid()) {
             $plainPass = $registration_form->get('plainPassword')->getData();
-            $password  = $passwordHasher->hashPassword($user, $plainPass);
+            if (strlen($plainPass) < $this->minimumPasswordLength) {
+                $this->addFlash('danger', "Password should be " . $this->minimumPasswordLength . "+ chars.");
+                return $this->redirectToRoute('register');
+            }
+
+            $password = $passwordHasher->hashPassword($user, $plainPass);
             $user->setPassword($password);
             if ((string)$user->getName() === '') {
                 $user->setName($user->getUsername());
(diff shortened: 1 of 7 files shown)


webapp/src/Controller/SecurityController.php
```diff
@@ -103,58 +105,62 @@ public function registerAction(
         $registration_form->handleRequest($request);
         if ($registration_form->isSubmitted() && $registration_form->isValid()) {
             $plainPass = $registration_form->get('plainPassword')->getData();
-            $password  = $passwordHasher->hashPassword($user, $plainPass);
-            $user->setPassword($password);
-            if ($user->getName() === null) {
-                $user->setName($user->getUsername());
-            }
+            if (strlen($plainPass) < UserController::MIN_PASSWORD_LENGTH) {
+                $this->addFlash('danger', "Password should be " . UserController::MIN_PASSWORD_LENGTH . "+ chars.");
```
vmcj, line 109:
```suggestion
                $this->addFlash('danger', "Password should be " . UserController::MIN_PASSWORD_LENGTH . "+ chars.");
                return $this->redirectToRoute('login');
```

Would make the PR so much easier to read I think.

All other changes are just indentations right?

webapp/src/Controller/SecurityController.php
```diff
@@ -103,58 +105,62 @@ public function registerAction(
         $registration_form->handleRequest($request);
         if ($registration_form->isSubmitted() && $registration_form->isValid()) {
             $plainPass = $registration_form->get('plainPassword')->getData();
-            $password  = $passwordHasher->hashPassword($user, $plainPass);
-            $user->setPassword($password);
-            if ($user->getName() === null) {
-                $user->setName($user->getUsername());
-            }
+            if (strlen($plainPass) < UserController::MIN_PASSWORD_LENGTH) {
```
eldering, line 108: Wouldn't it make more sense to store this constant here in the security controller, or possibly in some other global place?

webapp/src/Controller/SecurityController.php
eldering: All remaining changes are indentation indeed, and I agree that here it makes a lot of sense to return early, not just to make the diff simpler, but more importantly to reduce code nesting complexity.

nickygerritsen: We shouldnt return to login though, but register. will fix.

webapp/src/Controller/SecurityController.php
nickygerritsen: Do you have a good suggestion? Since we also use it in the jury controller.

eldering: Maybe in `webapp/config/static.yaml.in`? Not sure whether we should also make it configurable from `configure` or so...

nickygerritsen: I do like that idea, but then maybe we do that in a separate issue / PR?

meisterT (approved):

thijskh:
Probably we need to tell the user what length we expect when we present the form? Not reject it after submission with a surprise limit that was never communicated before?

nickygerritsen:
> Probably we need to tell the user what length we expect when we present the form? Not reject it after submission with a surprise limit that was never communicated before?

The `minlength` html attribute kinda does this, right?

webapp/tests/Unit/Controller/PublicControllerTest.php
```diff
@@ -210,17 +210,17 @@ public function selfRegisterProvider(): Generator
                     continue;
                 }
                 yield[['username'=>'minimaluser', 'teamName'=>'NewTeam','affiliation'=>'none'],'shirt-recognize-bar-together', $fixtures, $category];
-                yield[['username'=>'bruteforce', 'teamName'=>'Fib(4)','affiliation'=>'none'],'0112', $fixtures, $category];
-                yield[['username'=>'fullUser', 'name'=>'Full User', 'email'=>'email@domain.com','teamName'=>'Trial','affiliation'=>'none'],'.', $fixtures, $category];
+                yield[['username'=>'bruteforce', 'teamName'=>'Fib(4)','affiliation'=>'none'],'0112345678', $fixtures, $category];
```
vmcj, line 213:
```suggestion
                yield[['username'=>'bruteforce', 'teamName'=>'Fib(9)','affiliation'=>'none'],'01123581321', $fixtures, $category];
```

webapp/src/Controller/SecurityController.php
nickygerritsen: I moved it to the same place as the enabling of removed intervals. Made sense to me.

nickygerritsen:
> > Probably we need to tell the user what length we expect when we present the form? Not reject it after submission with a surprise limit that was never communicated before?
> 
> The `minlength` html attribute kinda does this, right?

I added an explicit help text for it.

PR #2489, opened 2024-04-20
CDS does not provide timelimit

meisterT:
We disable requiring timelimit in the event feed for the analyst instance.


vmcj:
I think we should t merge this as it's required by the spec, I propose to either hide those changes behind sn analyst spec or just check on asystest before.

meisterT:
I left a comment in https://github.com/icpctools/icpctools/issues/933#issuecomment-2070455774

PR #2491, opened 2024-04-20
Ignore the `map-info` event in the event feed

meisterT:
Similarly to the other ignored event types, we need to drop this event to successfully parse the event feed of external sources.

### DIFF
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -620,7 +620,7 @@ class ExternalContestSourceService
 
         // Note the @vars here are to make PHPStan understand the correct types.
         $method = match ($event->type) {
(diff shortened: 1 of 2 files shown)


nickygerritsen (approved):

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -620,7 +620,7 @@ public function importEvent(Event $event, array $eventsToSkip): void
 
         // Note the @vars here are to make PHPStan understand the correct types.
         $method = match ($event->type) {
-            EventType::AWARDS, EventType::TEAM_MEMBERS, EventType::ACCOUNTS, EventType::PERSONS => $this->ignoreEvent(...),
+            EventType::ACCOUNTS, EventType::AWARDS, EventType::MAP_INFO, EventType::PERSONS, EventType::TEAM_MEMBERS => $this->ignoreEvent(...),
```
eldering, line 623: It would be better to have a catch-all for unknown events. According to the standard a system may always add its own event types and we can't foresee all of them.

PR #2492, opened 2024-04-20
Make judgedaemon service shut down gracefully

meisterT:
Also let it listen to SIGQUIT for a harder shutdown aborting any ongoing judging, besides of course SIGKILL.


meisterT:
@vmcj @eldering I don't know which of you wrote this and whether it was tested

vmcj:
I discussed it with @eldering but didn't test it (yet)

nickygerritsen (approved):

eldering:
I _think_ I tested this, but would also say it's pretty safe, since it doesn't fundamentally change anything.

PR #2496, opened 2024-04-20
Ignore TLS protocol error.

meisterT:
The logs gets spammed with roughtly 1GiB/s on rare occasions.


vmcj (approved):
In the future we should find a way to detect such repeating entries and log them in a different way.

nickygerritsen (approved):

PR #2497, opened 2024-04-21
Ignore additional bronze medals in public scoreboard until unfreeze

meisterT:


nickygerritsen (approved):

nickygerritsen (approved):

webapp/src/Utils/Scoreboard/Scoreboard.php
```diff
@@ -48,6 +48,14 @@ public function __construct(
         $this->calculateScoreboard();
     }
 
+    /**
+     * @return bool Whether this Scoreboard has restricted access (either a jury member can see, or after unfreeze).
+     */
+    public function isRestricted(): bool
```
eldering, line 54: I think this function name is ambiguous. It can easily be read to mean that this is a restricted scoreboard that only contains frozen data. However, it means that this is the full scoreboard that has restricted access.

I'd propose to rename it to something like `hasRestrictedAccess`.

PR #2498, opened 2024-04-21
Fix missing events in the feed due to out of order commits

meisterT:
Events may become available to read by the event-feed API controller out of event_id order because they are (sometimes) created inside a transaction. The old code was only looking for events with ID larger than the last one emitted, which could lead to skipping events. This happened at the WFs in Luxor especially with judging events for a compiler error, as inside a transaction two events are logged and then the scoreboard is recalculated, leaving plenty of time for another event to get logged in the mean time.

Fix this by querying for *all* events and making sure that we're not missing any sequential event_ids. If we do wait and retry for one second, and only then skip these. This means that event filtering must now be done in code rather than in the SQL query.


webapp/src/Controller/API/ContestController.php
```diff
@@ -714,39 +716,77 @@ public function getEventFeedAction(
...
+                        $lastFoundId = $event->getEventid();
+                        break;
                     }
-                    $qb = $qb
-                        ->andWhere('e.endpointtype NOT IN (:restricted_types)')
-                        ->setParameter('restricted_types', $restricted_types);
+                    $expectedId++;
                 }
+                if ($missingEvents) {
+                    if ($missingEventRetries==0) {
```
nickygerritsen, line 751: Nit: formatting

nickygerritsen (approved):
PHP CS has some warnings as well

webapp/src/Controller/API/ContestController.php
meisterT: Done, also the PHP CS warnings

PR #2503, opened 2024-04-22
Set errno to zero before strtol

wx-csy:
According to the manpage of [strtol (Notes)](https://linux.die.net/man/3/strtol), we have to zero errno before calling `strtol` if we want  to use errno to determine if an error occurred.

I've corrected (hopefully) all such bugs in runguard, where some prevented me successfully running judgehost on our server. Not sure whether such kind of bugs exist elsewhere in this project.

### DIFF
--- a/judge/runguard.c
+++ b/judge/runguard.c
@@ -682,6 +682,7 @@ long read_optarg_int(const char *desc, long minval, long maxval)
 	long arg;
 	char *ptr;
 
+	errno = 0;
 	arg = strtol(optarg,&ptr,10);
 	if ( errno || *ptr!='\0' || arg<minval || arg>maxval ) {
 		error(errno,"invalid %s specified: `%s'",desc,optarg);
@@ -699,6 +700,7 @@ void read_optarg_time(const char *desc, double *times)
 	/* Check for soft:hard limit separator and cut string. */
 	if ( (sep=strchr(optcopy,':'))!=NULL ) *sep = 0;
 
+	errno = 0;
 	times[0] = strtod(optcopy,&ptr);
 	if ( errno || *ptr!='\0' || !finite(times[0]) || times[0]<=0 ) {
 		error(errno,"invalid %s specified: `%s'",desc,optarg);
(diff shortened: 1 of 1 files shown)


judge/runguard.c
```diff
@@ -999,8 +1002,9 @@ int main(int argc, char **argv)
 			break;
 		case 'u': /* user option: uid or string */
 			use_user = 1;
-			runuid = strtol(optarg,&ptr,10);
 			runuser = strdup(optarg);
```
meisterT, line 1005: I guess we also would want some error checking on strdup returning null and its errno?

judge/runguard.c
```diff
@@ -1014,8 +1018,9 @@ int main(int argc, char **argv)
 			break;
 		case 'g': /* group option: gid or string */
 			use_group = 1;
-			rungid = strtol(optarg,&ptr,10);
 			rungroup = strdup(optarg);
```
meisterT, line 1021: same here?

judge/runguard.c
```diff
@@ -699,15 +700,17 @@ void read_optarg_time(const char *desc, double *times)
...
 	if ( errno || *ptr!='\0' || !finite(times[0]) || times[0]<=0 ) {
 		error(errno,"invalid %s specified: `%s'",desc,optarg);
 	}
 
 	/* And repeat for hard limit if we found the ':' separator. */
 	if ( sep!=NULL ) {
+		errno = 0;
 		times[1] = strtod(sep+1,&ptr);
-		if ( errno || *ptr!='\0' || !finite(times[1]) || times[1]<=0 ) {
+		if ( errno || *(sep+1)=='\0' || *ptr!='\0' || !finite(times[1]) || times[1]<=0 ) {
```
meisterT, line 713: what are you checking here?

judge/runguard.c
```diff
@@ -1014,8 +1018,9 @@ int main(int argc, char **argv)
 			break;
 		case 'g': /* group option: gid or string */
 			use_group = 1;
-			rungid = strtol(optarg,&ptr,10);
 			rungroup = strdup(optarg);
+			errno = 0;
+			rungid = strtol(optarg,&ptr,10);
```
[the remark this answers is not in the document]
meisterT: I think they wanted to not have any other function call that could set `errno` in between `strtol` and the check. See also my comment.

judge/runguard.c
wx-csy: this is checking if the number after ':' is empty

seems not necessary here because `times[1]<=0` also catches this case; but explicit check improves readability?

judge/runguard.c
wx-csy: added

judge/runguard.c
wx-csy: also added

meisterT (approved):
LGTM, please squash to a single commit.

wx-csy:
> LGTM, please squash to a single commit.

Done!

vmcj (approved):
I'll wait for @eldering his review and merge afterwards.

eldering (approved):

PR #2506, opened 2024-04-23
Add rank to first to solve and region winners in results HTML.

### LINKED ISSUE #2480: Show rank of region winner in wf_results
At the finals they would like to know in which order to hand out plaques to region winners. 

We should add the rank of the team in the corresponding table, and perhaps do the same for the first to solve.

nickygerritsen:
Fixes #2480.

Note that the rank can be the same for two teams (if they are outside of the medals) or even be `-` if they are in honorable mentions. See attached screenshot for the WF 47 output.

image

### DIFF
--- a/webapp/src/Controller/Jury/ImportExportController.php
+++ b/webapp/src/Controller/Jury/ImportExportController.php
@@ -377,16 +377,19 @@ class ImportExportController extends BaseController
         $ranked        = [];
         $honorable     = [];
         $regionWinners = [];
+        $rankPerTeam   = [];
 
         $sortOrder = $request->query->getInt('sort_order');
 
         foreach ($this->importExportService->getResultsData($sortOrder, full: $full) as $row) {
             $team = $teamNames[$row[0]];
+            $rankPerTeam[$row[0]] = $row[1];
 
             if ($row[6] !== '') {
                 $regionWinners[] = [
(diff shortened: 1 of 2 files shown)


meisterT:
Did you intentionally add the rank once in the end and once in the middle?

nickygerritsen:
I did, it didn't look great adding it after the team for region or before the team for FTS. But if you think it does, I can change it.

vmcj:
> I did, it didn't look great adding it after the team for region or before the team for FTS. But if you think it does, I can change it.

Can you make it:
Region winners:
Region     > Team > Rank
Problem   > Team > Rank > Time

Where the Team will shift a bit to the left as a column has moved? Or is this what you didn't look nice?

nickygerritsen:
It will look like this, which is good enough I think:
image

I just pushed it.

vmcj (approved):
> It will look like this, which is good enough I think: image
> 
> I just pushed it.

No easy way to align the rank for both (and the same for team)? 

Otherwise fine by me.

nickygerritsen:
> > It will look like this, which is good enough I think: image
> > I just pushed it.
> 
> No easy way to align the rank for both (and the same for team)?
> 
> Otherwise fine by me.

Nope and the tables where not aligned before either

PR #2508, opened 2024-04-23
Catch errors when adding problems YAML.

### LINKED ISSUE #2482: Handle invalid RGB value in problemset
When importing a problemset with invalid RGB values (such as `#tobiii`), it crashes the import, and provokes an internal server error that can only be fixed by manual database changes.

nickygerritsen:
Fixes #2482.

### DIFF
--- a/webapp/src/Controller/API/ProblemController.php
+++ b/webapp/src/Controller/API/ProblemController.php
@@ -102,8 +102,11 @@ class ProblemController extends AbstractRestController implements QueryObjectTra
         // Note: we read the JSON as YAML, since any JSON is also YAML and this allows us
         // to import files with YAML inside them that match the JSON format
         $data = Yaml::parseFile($file->getRealPath(), Yaml::PARSE_DATETIME);
-        if ($this->importExportService->importProblemsData($contest, $data, $ids)) {
+        if ($this->importExportService->importProblemsData($contest, $data, $ids, $messages)) {
             return $ids;
+        }
+        if (!empty($messages)) {
+            throw new BadRequestHttpException($this->dj->jsonEncode($messages));
         }
         throw new BadRequestHttpException("Error while adding problems");
     }
--- a/webapp/src/Controller/API/ProblemController.php
+++ b/webapp/src/Controller/API/ProblemController.php
@@ -105,10 +105,11 @@ class ProblemController extends AbstractRestController implements QueryObjectTra
         if ($this->importExportService->importProblemsData($contest, $data, $ids, $messages)) {
             return $ids;
         }
+        $message = "Error while adding problems";
         if (!empty($messages)) {
-            throw new BadRequestHttpException($this->dj->jsonEncode($messages));
+            $message .= ': ' . $this->dj->jsonEncode($messages);
         }
-        throw new BadRequestHttpException("Error while adding problems");
+        throw new BadRequestHttpException($message);
     }
 
     /**
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -309,8 +309,9 @@ class ImportExportService
      *              problems?: array{name?: string, short-name?: string, id?: string, label?: string,
      *                              letter?: string, label?: string, letter?: string}} $problems
      * @param string[]|null $ids
+     * @param array<string, string[]> $messages
      */
-    public function importProblemsData(Contest $contest, array $problems, array &$ids = null): bool
+    public function importProblemsData(Contest $contest, array $problems, array &$ids = null, ?array &$messages = []): bool
     {
         // For problemset.yaml the root key is called `problems`, so handle that case
         // TODO: Move this check away to make the $problems array shape easier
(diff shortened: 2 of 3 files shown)


webapp/src/Service/ImportExportService.php
```diff
@@ -339,14 +352,35 @@ public function importProblemsData(Contest $contest, array $problems, array &$id
...
+                        $problem->getExternalid(),
+                        $error->getPropertyPath(),
+                        $error->getMessage()
+                    );
+                }
+            }
+
+            if ($hasErrors) {
+                return false;
+
```
vmcj, line 372:
```suggestion

```

webapp/src/Service/ImportExportService.php
```diff
@@ -309,8 +309,9 @@ public function importContestData(mixed $data, ?string &$errorMessage = null, st
      *              problems?: array{name?: string, short-name?: string, id?: string, label?: string,
      *                              letter?: string, label?: string, letter?: string}} $problems
      * @param string[]|null $ids
+     * @param array<string, string[]> $messages
      */
-    public function importProblemsData(Contest $contest, array $problems, array &$ids = null): bool
+    public function importProblemsData(Contest $contest, array $problems, array &$ids = null, ?array &$messages = []): bool
```
vmcj, line 314:
```suggestion
    public function importProblemsData(Contest $contest, array $problems, array &$ids = null, ?array &$messages = []): true
```

webapp/src/Service/ImportExportService.php
```diff
@@ -329,8 +330,20 @@ public function importProblemsData(Contest $contest, array $problems, array &$id
                 ->setTimelimit($problemData['time_limit'] ?? 10)
                 ->setExternalid($problemData['id'] ?? $problemData['short-name'] ?? $problemLabel ?? null);
 
-            $this->em->persist($problem);
-            $this->em->flush();
+            $hasErrors = false;
+            $errors    = $this->validator->validate($problem);
+            if ($errors->count()) {
+                $hasErrors = true;
```
vmcj, line 336:
```suggestion
            if ($hasErrors = $errors->count()) {
```

Not sure if we like this,

webapp/src/Service/ImportExportService.php
nickygerritsen: Well no, since it doesn't return true all the time anymore now?

webapp/src/Service/ImportExportService.php
nickygerritsen: I don't really like it, especially if `problem` HAS errors and contestproblem does not.

webapp/src/Controller/API/ProblemController.php
```diff
@@ -102,9 +102,12 @@ public function addProblemsAction(Request $request): array
...
         // to import files with YAML inside them that match the JSON format
         $data = Yaml::parseFile($file->getRealPath(), Yaml::PARSE_DATETIME);
-        if ($this->importExportService->importProblemsData($contest, $data, $ids)) {
+        if ($this->importExportService->importProblemsData($contest, $data, $ids, $messages)) {
             return $ids;
         }
+        if (!empty($messages)) {
+            throw new BadRequestHttpException($this->dj->jsonEncode($messages));
+        }
         throw new BadRequestHttpException("Error while adding problems");
```
eldering, line 111: Wouldn't it be better to merge the contents of these exceptions into a single one? So that it is both clear where the error happens and what the underlying cause is (if any messages are there). I think this would also make the logic of the code here less weird: first conditionally throw an exception, and then finally still throw an exception.

webapp/src/Service/ImportExportService.php
eldering: I'd do it like:
```
           $errors    = $this->validator->validate($problem);
           $hasErrors = $errors->count();
           if ($hasErrors) {
```
although I see that `$hasErrors` is again set further down. But then maybe from linebreaks or name make it clear that this variable has more global scope, while `$errors` is just reused.
Will leave it up to you.

webapp/src/Controller/API/ProblemController.php
nickygerritsen: So you suggest:

```php
$message = "Error while adding problems";
if (!empty($messages)) {
	$message .= ': ' . $this->dj->jsonEncode($messages);
}
throw new BadRequestHttpException($message);
```
?

nickygerritsen:
I somehow can't comment on your second remark, but yeah hasErrors is a more global scope. I will add a linebreak and maybe even a comment.

webapp/src/Service/ImportExportService.php
vmcj: Missed that line indeed.

webapp/src/Controller/API/ProblemController.php
eldering: Yes, that looks good to me!

vmcj:
> I somehow can't comment on your second remark, but yeah hasErrors is a more global scope. I will add a linebreak and maybe even a comment.

You can here: https://github.com/DOMjudge/domjudge/pull/2508#discussion_r1579041421 it's a response to your response on my message. GitHub displays that in a really ugly way.

webapp/src/Service/ImportExportService.php
nickygerritsen: I used two separate variables, what about that idea? It makes it a bit more readable.

eldering (approved):
LGTM

PR #2516, opened 2024-04-30
Use new images for the team manual.

nickygerritsen:


vmcj (approved):
The originals were a bit smaller but I don't think this will explode our release tarballs.

meisterT (approved):

PR #2523, opened 2024-05-01
Send less event updates for judgements in parallel mode.

nickygerritsen:
We used to send event updates for all testcases that came in after we already determined the final verdict for a problem. Now we check if we already have a previous end time for the judgement, and then we don't do this anymore.

Note that race conditions can still make it such that we send multiple update events, where two judgehosts know the result at the same time. However, this happens way less often.

Do we still want to update the end time? We currently do but then never send an update event anymore, even though the end time changed. So is this actually valid? Or do we not want to merge this and accept that we send updates (in Luxor we had submissions with 47 updates, while only 1 is really needed)? Or maybe in lazy mode we do not take into account the run time of any testcase that came after the first wrong one?

Found by @johnbrvc.

### DIFF
--- a/webapp/migrations/Version20241122150726.php
+++ b/webapp/migrations/Version20241122150726.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20241122150726 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add max runtime for verdict to judging';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE judging ADD max_runtime_for_verdict NUMERIC(32, 9) UNSIGNED DEFAULT NULL COMMENT \'The maximum run time for all runs that resulted in the verdict\'');
+    }
+
+    public function down(Schema $schema): void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE judging DROP max_runtime_for_verdict');
+    }
+
+    public function isTransactional(): bool
+    {
+        return false;
+    }
+}
--- a/webapp/src/Controller/API/JudgehostController.php
+++ b/webapp/src/Controller/API/JudgehostController.php
@@ -1023,9 +1023,11 @@ class JudgehostController extends AbstractFOSRestController
             if (!$hasNullResults || $lazyEval !== DOMJudgeService::EVAL_FULL) {
                 // NOTE: setting endtime here determines in testcases_GET
                 // whether a next testcase will be handed out.
+                $this->logger->error('Judging %d is done, setting endtime, it had %d', [ $judging->getJudgingid(), $judging->getEndtime() ]);
+                $this->logger->error('Verdict is %s', [$result]);
+                $sendJudgingEvent = !$judging->getEndtime();
                 $judging->setEndtime(Utils::now());
                 $this->maybeUpdateActiveJudging($judging);
-                $sendJudgingEvent = true;
             }
             $this->em->flush();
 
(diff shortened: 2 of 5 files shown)


meisterT:
Does it hurt?

nickygerritsen:
@johnbrvc does it hurt for you, more than that you run a codepath to check if the judgement changed?

I wonder if this is a clics discussion as well, can a judgement change after an `end_time` has been set?

meisterT:
IIRC, we did this intentionally to make the event feed and endpoints match. 

Cc @eldering

nickygerritsen:
> IIRC, we did this intentionally to make the event feed and endpoints match.
> 
> Cc @eldering

Yeah this is what I'm thinking as well, hence my last sentence.

johnbrvc:
It doesn't hurt per se, but it is quite confusing, and pollutes the logs
(in addition, it bloats the event feed).  Which judgement for a specific
judgement_id  are we supposed to believe?  the first?  the 20th?  the
last?
It seems that once a judgement_type_id is set (WA/AC, etc), then that's
it, unless the judge_type_id changes (which does not happen very often, if
at all - manual intervention).
I'm not sure what use anything after that is?  The "runs" entries have each
test case's disposition and execution time, so why does the final judgement
record update?

meisterT:
> It doesn't hurt per se, but it is quite confusing, and pollutes the logs

You could decide to only log if there is any contradicting information in the update.


> Which judgement for a specific judgement_id  are we supposed to believe?  the first?  the 20th?  the
last?

The last as we are sending updated information.

I think this is in line with https://ccs-specs.icpc.io/draft/contest_api#general-requirements which specifically allows duplicate events and "when an object changes one or more times a notification will be sent" as well "the latest notification sent for any object is the correct and current state of that object."


> I'm not sure what use anything after that is?  The "runs" entries have each test case's disposition and execution time, so why does the final judgement record update?

Technically something was judged after the previously recorded endtime, so we update it to reflect that. With that it is exposed in the specific API endpoint. To not have conflicting information in the feed and the API, we also send it in the feed.

johnbrvc:
> > I'm not sure what use anything after that is?  The "runs" entries have each test case's disposition and execution time, so why does the final judgement record update?
> 
> Technically something was judged after the previously recorded endtime, so we update it to reflect that. With that it is exposed in the specific API endpoint. To not have conflicting information in the feed and the API, we also send it in the feed.

I guess this is my question.  How can you have something that was already judged (and given a judgment (WA/AC/TLE, etc/), be judged again, the same way?   It seems (IMHO) that the judgment record should be sent when the judging is complete, or, don't include the `judgement_type_id`.  This could be my philosophical or intuitive understanding of what a "judgement" is, that is, is it "right" or "wrong"; I don't need to now it is "wrong" 43 times, even though technically, wrong is wrong, even if it is wrong 43 times.

meisterT:
> I guess this is my question. How can you have something that was already judged (and given a judgment (WA/AC/TLE, etc.), be judged again, the same way? 

That's just the nature of the first incorrect test case (in order) determines the final verdict. When that is known, we let teams and downstream clients know although there is still judging going on. But these remaining test cases cannot influence the final verdict.

eldering:
> > I guess this is my question. How can you have something that was already judged (and given a judgment (WA/AC/TLE, etc.), be judged again, the same way?
> 
> That's just the nature of the first incorrect test case (in order) determines the final verdict. When that is known, we let teams and downstream clients know although there is still judging going on. But these remaining test cases cannot influence the final verdict.

I agree with @johnbrvc that it doesn't really make sense to update the judgement after it has a verdict.  Even though it is still running on our side, that really should/does not not the judgement anymore.

I think what we should do, is simply still send the runs, not the judgements anymore. Since these don't affect the judgement, there's no need to update our judgement end time (either internally or externally), so we don't have to send out an event for that. Indeed, otherwise we should have, since the event feed should match the contents of the API endpoints.

nickygerritsen:
Note that we also update the max run time to be the max of all runs. This then also needs to change to only take the max of runs that added to the verdict, which might be hard. If we run in parallel, we might already have run some later test cases which would then be in the max run time during running but not anymore after it is done.

eldering:
After discussion IRL, we agreed to set the judging endtime only once, together with the verdict.

webapp/migrations/Version20241122150726.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    public function getDescription(): string
+    {
+        return 'Add max run time for verdict to judging';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE judging ADD max_run_time_for_verdict NUMERIC(32, 9) UNSIGNED DEFAULT NULL COMMENT \'The maximum run time for all runs that resulted in the verdict\'');
```
eldering, line 23: For consistency with other DB fields:
```suggestion
        $this->addSql('ALTER TABLE judging ADD max_runtime_for_verdict NUMERIC(32, 9) UNSIGNED DEFAULT NULL COMMENT \'The maximum run time for all runs that resulted in the verdict\'');
```

PR #2527, opened 2024-05-04
Improve error message on inconsistent ZIP files.

meisterT:
When using the graphical ZIP compress options on Mac OS or Ubuntu, the resulting ZIP archive will be inconsistent, tripping up `libzip`.

With `ziptool` you could provoke the error to learn more about the inconsistency but that is unfortunately not available in PHP:
```
can't open zip archive 'a.zip': Zip archive inconsistent: entry 2: local and central headers do not match
```

`libzip` worked around this in
https://github.com/nih-at/libzip/commit/b3e3b19e2e4b7d4651d01148f123129429ecc69b which was released in
https://github.com/nih-at/libzip/releases/tag/v1.10.1 but that version didn't make it into even unstable Debian/Ubuntu distributions yet.

### DIFF
--- a/webapp/src/Service/DOMJudgeService.php
+++ b/webapp/src/Service/DOMJudgeService.php
@@ -710,14 +710,18 @@ class DOMJudgeService
     public function openZipFile(string $filename): ZipArchive
     {
         $zip = new ZipArchive();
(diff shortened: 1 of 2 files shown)


vmcj (approved):

vmcj:
We should also backport this.

PR #2529, opened 2024-05-04
Hide 1px rectangles for nvd3 graphs.

### LINKED ISSUE #2528: Statistics graphs: do not display anything for 0 attempts.
See:
image

It looks like C/G are attempted, but they are not. We should check if we can disable this behaviour in the graph.

meisterT:
Unfortunately nvd3 does display a 1px bar for the value 0: https://github.com/novus/nvd3/blob/master/src/models/discreteBar.js#L198

This works around it but would also make very small values disappear.

Fixes #2528.

### DIFF
--- a/webapp/public/style_domjudge.css
+++ b/webapp/public/style_domjudge.css
@@ -655,3 +655,7 @@ blockquote {
 	color: DimGray;
 	margin-left: auto;
 }
+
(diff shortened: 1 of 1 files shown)


meisterT:
I will close this PR right away it is just for demonstration purposes. It would have the problem mentioned in the commit description, also losing information...

PR #2530, opened 2024-05-08
Remove ONLINE_JUDGE & DOMJUDGE check in example submissions

vmcj:
We might have in the past but in the default installation we don't set that environment variable anymore.

I'm not 100% if I'm correct here and I don't think this is important enough to merge before the weekend as this is only content.

### DIFF
--- a/example_problems/hello/submissions/accepted/extra-file-cs/test-hello.cs
+++ b/example_problems/hello/submissions/accepted/extra-file-cs/test-hello.cs
@@ -11,10 +11,6 @@ public class Hello
 {
 	public static void Main(string[] args)
 	{
-#if ONLINE_JUDGE
 		Console.Write("Hello world!\n");
-#else
-		Console.Write("ONLINE_JUDGE not defined\n");
-#endif
 	}
 }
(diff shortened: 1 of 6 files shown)


example_problems/hello/submissions/accepted/extra-file-cs/test-hello.cs
```diff
@@ -12,9 +12,9 @@ public class Hello
 	public static void Main(string[] args)
 	{
 #if ONLINE_JUDGE
-		Console.Write("Hello world!\n");
-#else
 		Console.Write("ONLINE_JUDGE not defined\n");
```
nickygerritsen, line 15: This message is now wrong

vmcj, line 15:
```suggestion
		Console.Write("ONLINE_JUDGE defined\n");
```

example_problems/hello/submissions/accepted/extra-file-cs/test-hello.cs
```diff
@@ -12,7 +12,7 @@ public class Hello
 	public static void Main(string[] args)
 	{
 #if ONLINE_JUDGE
-		Console.Write("ONLINE_JUDGE not defined\n");
+		Console.Write("ONLINE_JUDGE defined\n");
```
nickygerritsen, line 15: Actually why not remove it alltogether?

nickygerritsen (approved):

PR #2532, opened 2024-05-09
Add compiler version commands for C#, Rust, and Scala to DefaultData/LanguageFixture.php

mpsijm:
Tested at https://fpcs2024.bapc.eu/jury/versions :smile:


nickygerritsen (approved):
Nice! This doesn’t work if you upgrade but then again the others don’t either and people can set it themselves if they ever update so let’s ship it 🚢

vmcj (approved):
We can do the migration when we have more of them.

PR #2538, opened 2024-05-10
Don't set the empty string as we display it on the scoreboard

vmcj:
We state we fallback but would pick the empty string now,


meisterT:
I seem to remember that we also had an issue with `null`s but cannot find it quickly in the code.

vmcj:
> I seem to remember that we also had an issue with `null`s but cannot find it quickly in the code.

I do remember the same, but PHPStan seems to be happy so it should be allowed. I think I do prefer `null` for this specific case or to make the choice explicit here, so if `display_name` is not set start to guess on something else, the earlier code did not do that so I didn't add it but this feels a bit like having too many fields for the same thing.

meisterT (approved):

nickygerritsen (approved):

PR #2540, opened 2024-05-11
Skip judging parts of the integration tests.

meisterT:
This is unfortunate but gitlab seems to have enabled cgroups v2 on their runners and we still use cgroups v1, see https://github.com/DOMjudge/domjudge/issues/1072

### DIFF
--- a/gitlab/integration.sh
+++ b/gitlab/integration.sh
@@ -81,11 +81,25 @@ mount
 mount -o remount,exec,dev /builds
 section_end mount
 
+section_start check_cgroup_v1 "Checking for cgroup v1 availability"
+grep cgroup$ /proc/filesystems
+if [ $? -eq 0 ]; then
+    cgroupv1=1
+else
+    echo "Skipping tests that rely on cgroup v1"
+    cgroupv1=0
+fi
+section_end check_cgroup_v1
+
 section_start judgehost "Configure judgehost"
 cd /opt/domjudge/judgehost/
 sudo cp /opt/domjudge/judgehost/etc/sudoers-domjudge /etc/sudoers.d/
 sudo chmod 400 /etc/sudoers.d/sudoers-domjudge
-sudo bin/create_cgroups
+if [ $cgroupv1 -ne 0 ]; then
+    # We allow this to go wrong as some gitlab runners do not have the
+    # swapaccount kernel option set.
+    sudo bin/create_cgroups || cgroupv1=0
+fi
 
 if [ ! -d ${DIR}/chroot/domjudge/ ]; then
     cd ${DIR}/misc-tools
@@ -129,19 +143,23 @@ if [ $PIN_JUDGEDAEMON -eq 1 ]; then
 fi
 section_end more_setup
 
-section_start runguard_tests "Running isolated runguard tests"
-sudo addgroup domjudge-run-0
-sudo usermod -g domjudge-run-0 domjudge-run-0
-cd ${DIR}/judge/runguard_test
-make test
-section_end runguard_tests
+if [ $cgroupv1 -ne 0 ]; then
+    section_start runguard_tests "Running isolated runguard tests"
+    sudo addgroup domjudge-run-0
+    sudo usermod -g domjudge-run-0 domjudge-run-0
+    cd ${DIR}/judge/runguard_test
+    make test
+    section_end runguard_tests
+fi
 
-section_start start_judging "Start judging"
-cd /opt/domjudge/judgehost/
-
-sudo -u domjudge bin/judgedaemon $PINNING |& tee /tmp/judgedaemon.log &
-sleep 5
-section_end start_judging
+if [ $cgroupv1 -ne 0 ]; then
+    section_start start_judging "Start judging"
+    cd /opt/domjudge/judgehost/
+    
+    sudo -u domjudge bin/judgedaemon $PINNING |& tee /tmp/judgedaemon.log &
+    sleep 5
+    section_end start_judging
+fi
 
 section_start submitting "Importing Kattis examples"
 export SUBMITBASEURL='http://localhost/domjudge/'
@@ -164,9 +182,7 @@ for i in hello_kattis different guess; do
 done
 section_end submitting
 
-section_start judging "Waiting until all submissions are judged"
-# wait for and check results
-NUMSUBS=$(curl --fail http://admin:$ADMINPASS@localhost/domjudge/api/contests/1/submissions | python3 -mjson.tool | grep -c '"id":')
+section_start curlcookie "Preparing cookie jar for curl"
 export COOKIEJAR
 COOKIEJAR=$(mktemp --tmpdir)
 export CURLOPTS="--fail -sq -m 30 -b $COOKIEJAR"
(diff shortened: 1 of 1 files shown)


meisterT:
Trying this as an alternative to https://github.com/DOMjudge/domjudge/pull/2531 which skips the integration tests all together.

vmcj (approved):
I think any fix will do so lets merge first and afterwards try to get more back,
- We could check for cgroupv1 and if found still go on? I suspect they still have a fleet of older machines.
- We can see if we can get cgroupv1 back with `docker:dind` as that _might_ spin up a small vm. It does on macos/windows but I suspect it's smart enough to detect linux can already do cgroup on the host.
- I think we can skip L92 & L136-148 and check the API there and even get a bit further, it would require also checking if we're cgroupv2, because if we're we will never get everything verified.

meisterT:
Good ideas, added the check and continued with everything that we could.

gitlab/integration.sh
```diff
@@ -81,11 +81,26 @@ mount
...
+    cgroupv1=0
+fi
+section_end check_cgroup_v1
+
 section_start judgehost "Configure judgehost"
 cd /opt/domjudge/judgehost/
 sudo cp /opt/domjudge/judgehost/etc/sudoers-domjudge /etc/sudoers.d/
 sudo chmod 400 /etc/sudoers.d/sudoers-domjudge
-sudo bin/create_cgroups
+if [ $cgroupv1 -ne 0 ]; then
```
vmcj, line 99:
```suggestion
if [ $cgroupv -eq 1 ]; then
```

gitlab/integration.sh
```diff
@@ -81,11 +81,26 @@ mount
...
+    # We allow this to go wrong as some gitlab runners do not have the
+    # swapaccount kernel option set.
+    sudo bin/create_cgroups || cgroupv1=0
```
vmcj, line 102:
```suggestion
    sudo bin/create_cgroups || cgroupv=0
```
The choice for cgroupv=0 is explicit here but I would argue that we should fail for this case? If the gitlab runners are inconsistent I would prefer that we fail get a retry on another runner. 

Why can this fail on a machine with cgroupv1?

gitlab/integration.sh
```diff
@@ -129,19 +144,23 @@ if [ $PIN_JUDGEDAEMON -eq 1 ]; then
...
-section_start runguard_tests "Running isolated runguard tests"
-sudo addgroup domjudge-run-0
-sudo usermod -g domjudge-run-0 domjudge-run-0
-cd ${DIR}/judge/runguard_test
-make test
-section_end runguard_tests
-
-section_start start_judging "Start judging"
-cd /opt/domjudge/judgehost/
+if [ $cgroupv1 -ne 0 ]; then
```
vmcj, line 147:
```suggestion
if [ $cgroupv -eq 1 ]; then
```

gitlab/integration.sh
```diff
@@ -129,19 +144,23 @@ if [ $PIN_JUDGEDAEMON -eq 1 ]; then
...
+    sudo usermod -g domjudge-run-0 domjudge-run-0
+    cd ${DIR}/judge/runguard_test
+    make test
+    section_end runguard_tests
+fi
 
-sudo -u domjudge bin/judgedaemon $PINNING |& tee /tmp/judgedaemon.log &
-sleep 5
-section_end start_judging
+if [ $cgroupv1 -ne 0 ]; then
```
vmcj, line 156:
```suggestion
if [ $cgroupv1 -eq 1 ]; then
```

gitlab/integration.sh
```diff
@@ -175,61 +192,68 @@ export CURLOPTS="--fail -sq -m 30 -b $COOKIEJAR"
 CSRFTOKEN=$(curl $CURLOPTS -c $COOKIEJAR "http://localhost/domjudge/login" 2>/dev/null | sed -n 's/.*_csrf_token.*value="\(.*\)".*/\1/p')
 # Make a second request with our session + csrf token to actually log in
 curl $CURLOPTS -c $COOKIEJAR -F "_csrf_token=$CSRFTOKEN" -F "_username=admin" -F "_password=$ADMINPASS" "http://localhost/domjudge/login"
-
-# Send a general clarification to later test if we see the event.
-curl $CURLOPTS -F "sendto=" -F "problem=1-" -F "bodytext=Testing" -F "submit=Send" \
```
vmcj, line 180: I think we can still send the clarification here as the domserver is online.

gitlab/integration.sh
```diff
@@ -175,61 +192,68 @@ export CURLOPTS="--fail -sq -m 30 -b $COOKIEJAR"
...
-            head $dir/testcase001/program.meta
+section_end curlcookie
+
+if [ $cgroupv1 -ne 0 ]; then
+    section_start judging "Waiting until all submissions are judged"
+    # wait for and check results
+    NUMSUBS=$(curl --fail http://admin:$ADMINPASS@localhost/domjudge/api/contests/1/submissions | python3 -mjson.tool | grep -c '"id":')
+    
+    # Send a general clarification to later test if we see the event.
+    curl $CURLOPTS -F "sendto=" -F "problem=1-" -F "bodytext=Testing" -F "submit=Send" \
```
vmcj, line 203:
```suggestion
    curl $CURLOPTS -F "sendto=" -F "problem=1-" -F "bodytext=Testing" -F "submit=Send" \
```

Like a said, we can test this with the evenfeed still if its received.

gitlab/integration.sh
```diff
@@ -249,7 +273,11 @@ if cat /opt/domjudge/domserver/webapp/var/log/prod.log | egrep '(CRITICAL|ERROR)
 fi
 
 # Check the Contest API:
-$CHECK_API -n -C -e -a 'strict=1' http://admin:$ADMINPASS@localhost/domjudge/api
+if [ $cgroupv1 -ne 0 ]; then
+    $CHECK_API -n -C -e -a 'strict=1' http://admin:$ADMINPASS@localhost/domjudge/api
+else
+    $CHECK_API -n -e -a 'strict=1' http://admin:$ADMINPASS@localhost/domjudge/api
```
vmcj, line 279: Can you add a comment what the `-C` does, its removed for the cgroupv2 (probably for good reasons).

vmcj (approved):
Added some comments which would make it more readable for me but that might be personal preference as those could also look like wrong exitcodes.

gitlab/integration.sh
meisterT: Done.

gitlab/integration.sh
meisterT: GitHub (and I) don't see any changes suggested here. But I assume I addressed the suggestion anyway by moving it outside of the if.

gitlab/integration.sh
meisterT: Done.

gitlab/integration.sh
```diff
@@ -81,11 +81,26 @@ mount
...
 section_end mount
 
+
+section_start check_cgroup_v1 "Checking for cgroup v1 availability"
+grep cgroup$ /proc/filesystems
+if [ $? -eq 0 ]; then
+    cgroupv1=1
+else
+    echo "Skipping tests that rely on cgroup v1"
+    cgroupv1=0
```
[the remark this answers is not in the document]
meisterT: I am testing for cgroup v1 ability, not v2 and using this is a boolean variable in the sense of "is cgroup v1 in the config we need it available" as this is what matters for the integration test.

meisterT:
I merged it now to make CI green again. We can always improve things on top of it

PR #2541, opened 2024-05-11
Add version commands for more languages/executables

vmcj:
This tries to be non-destructive for upgrades but would remove the command if people have picked the same command and they choose to run the `down` migration, in that specific edgecase I assume people know what/(and why) they are doing this.

~~I've not done all the languages yet as I would like some feedback first.~~ it now has all known languages for my installation.


webapp/migrations/Version20240511091916.php
```diff
@@ -0,0 +1,48 @@
...
+final class Version20240511091916 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Fill default version command for compiler/runner.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('UPDATE language SET runner_version_command = \'bash --version\' where langid=\'bash\';');
```
meisterT, line 22: ... `AND runner_version_command IS NULL`?

sql/files/defaultdata/adb/run
```diff
@@ -10,9 +10,3 @@ MAINSOURCE="$1"
 # -static:         Static link with all libraries
 # -bargs -static:  Static link with all libraries
 gnatmake -static -o "$DEST" "$@" -bargs -static
-exitcode=$?
```
meisterT, line 13: Why did you remove these?

generate_sql.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
```
meisterT, line 1: What's the idea behind this script - you created the migration instead, right?

vmcj: The commit will be removed later, its the code to generate the migration to keep quoting correct etc.

sql/files/defaultdata/adb/run
vmcj: See the commit message, I assumed this was from a time where we didn't have the chroot yet so had to do cleanup.

meisterT: @ubergeek42 added this in 2013 in 460c79ab16a0695dca8a120227d319af09cdf0b0

Keith, do you remember why you cleaned up the files?

vmcj:
@meisterT should this just replace what we do in the default data? If you wipe your database we always run the migrations and we always run the defaultdata fixture so one of them is redundant. I think the migration is better as it also helps in case people upgrade.

meisterT:
I would keep both. The migration is clearly necessary now when you upgrade. The defaultdata fixture is way nicer when you add a completely new language (we wouldn't add a migration for that).

vmcj:
> I would keep both. The migration is clearly necessary now when you upgrade. The defaultdata fixture is way nicer when you add a completely new language (we wouldn't add a migration for that).

I think when we add a new language and you upgrade you would indeed not get the new language.

meisterT:
And that is fine. Adding a language is a conscious decision anyway and requires changes to the chroot typically.

doc/manual/install-language.rst
```diff
@@ -0,0 +1,90 @@
+Appendix: Installing the example languages
+==========================================
+
+DOMjudge default ships with some default languages with a default configuration.
+As you might setup contests with those languages we provide how those languages were
+installed in the past as guideline.
+
+Most of those can be found in the table below as there is a package available to install
+inside the judging chroot. Given that you can install your own chroot we only provide the
```
meisterT, line 9: Perhaps mention how it can be installed in the chroot?

example_problems/problems.yaml
```diff
@@ -1,15 +1,15 @@
-- id: hello
-  label: A
+- id: hello2
```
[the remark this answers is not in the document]
vmcj: See the commit message, It will be removed again but I had it for local testing.

doc/manual/install-language.rst
```diff
@@ -0,0 +1,90 @@
...
+     - Default installed in the chroot
+   * - C#
+     - `mono-mcs`
+     -
+   * - Fortran
+     - `gfortran`
+     -
+   * - Haskell
+     - `ghc`
+     - Currently not working
```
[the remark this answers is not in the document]
vmcj: No, there are a couple more where I still need to do some extra work.


The original idea of the PR was to get the version commands, finding out that the language is not usable with those instructions is part 2 IMO.

doc/manual/install-language.rst
```diff
@@ -0,0 +1,93 @@
...
+   * - Lua
+     - `lua5.4`
+     - Ubuntu does not ship a generic meta package (yet).
+   * - Pascal
+     - `fp-compiler`
+     -
+   * - Perl
+     - `perl-base`
+     - Default installed in the chroot
+   * - POSIX
```
eldering, line 60:
```suggestion
   * - POSIX shell
```

doc/manual/install-language.rst
```diff
@@ -0,0 +1,93 @@
...
+   * - Prolog
+     - `swi-prolog-core`
+     -
+   * - Python3
+     - `pypy3`/`python3`
+     - Default installed in the chroot.
+       DOMjudge assumes `pypy3` as it runs generally faster.
+       Consider the `PyPy3 PPA`_ if you need the latest python3 features. PyPy3 does not have 100%
+       compatibility with all non-standard libraries in case this is needed you should reconsider the default
+       CPython implementation.
```
eldering, line 72:
```suggestion
       DOMjudge assumes `pypy3` as it runs faster in general.
       Consider the `PyPy3 PPA`_ if you need the latest python3 features. PyPy3 does not have 100%
       compatibility with all non-standard libraries. In case this is needed you should reconsider the default
       CPython implementation.
```

sql/files/defaultdata/adb/run
eldering: Also, don't remove the `exit $exitcode` as it detects and returns whether compilation was successful.

doc/manual/install-language.rst
vmcj: Yes, will do.

sql/files/defaultdata/adb/run
vmcj: Aah, ok I'll just revert that whole commit.

gitlab/integration.sh
```diff
@@ -70,6 +70,8 @@ cp etc/initial_admin_password.secret "$GITLABARTIFACTS/"
 section_end setup
 
 section_start submit_client "Test submit client"
+# Enable all languages
+echo "UPDATE language SET allow_submit=1;" | mysql domjudge
```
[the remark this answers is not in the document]
vmcj: Kept it in my local branch for another PR.

doc/manual/install-language.rst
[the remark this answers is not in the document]
vmcj: I documented the why here: https://github.com/DOMjudge/domjudge/issues/2551

webapp/migrations/Version20240511091916.php
```diff
@@ -0,0 +1,104 @@
...
+final class Version20240511091916 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Fill default version command for compiler/runner.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('UPDATE language SET compiler_version_command = \'gnatmake --version\' where langid=\'adb\' AND compiler_version_command IS NULL;');
```
meisterT, line 22: For easier readability, I would suggest creating an array something like this:

```
$versionCommands = [
'adb' => 'gnatmake --version',
'awk' => 'awk --version',
...
];
foreach ($versionCommands as $lang => $versionCommand) {
   $this->addSql("UPDATE language SET compiler_version_command = '" . $v ."' WHERE langid='" . $lang . "' AND compiler_version_command IS NULL;')
}
```

And then the same thing for runner versions.

vmcj: Done

sql/files/defaultdata/hs/run
```diff
@@ -10,6 +10,9 @@ MAINSOURCE="$1"
 # Set non-existing HOME variable to make GHC program happy, see:
 # https://ghc.haskell.org/trac/ghc/ticket/11678
 export HOME=/does/not/exist
+# Allow temporary files during compilation, this directory is not
+# available during the submission run.
+export TMPDIR=`pwd`
```
meisterT, line 15: is `$PWD` defined?

meisterT (commented):
I think it looks good after cleaning up the commits.

sql/files/defaultdata/hs/run
vmcj: `pwd` is the command to get the current work directory. Why do you prefer `$PWD` over `pwd`?

vmcj:
> I think it looks good after cleaning up the commits.

I'll wait for @eldering his review and will squash afterwards.

sql/files/defaultdata/hs/run
meisterT: `pwd` spawns a process, `$PWD` just reads an environment variable

doc/manual/install-language.rst
```diff
@@ -0,0 +1,97 @@
+Appendix: Installing the example languages
+==========================================
+
+DOMjudge ships with some default languages with a default configuration.
+As you might setup contests with those languages we provide how those languages were
```
eldering, line 5: Words like "setup" are written together when used as noun, but separate when used as verbs.
```suggestion
As you might set up contests with those languages we provide how those languages were
```

eldering: The text doesn't read so smoothly, but just merge after this change. I'll see if I can go through the whole documentation some time and improve things.

PR #2542, opened 2024-05-11
Improve display of test case run times.

meisterT:
Put it into a card, move the max and sum times into the card header. While we are at it, also make it clear that this is the CPU not the wall time.


meisterT:
This is the first part of https://github.com/DOMjudge/domjudge/pull/2452

Example:
image


The graphs are also removed when we and a potential shadow reported compiler error.

meisterT:
Second screenshot:
image

nickygerritsen (approved):

PR #2545, opened 2024-05-11
Simplify test case run display.

meisterT:
Before:
image

and

image

After:
image

and

image

### DIFF
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -117,21 +117,6 @@ function collapse(x)
     $(x).toggleClass('d-none');
 }
 
-function togglelastruns()
-{
-    var names = {'lastruntime':0, 'lastresult':1, 'lasttcruns':2};
-    for (var name in names) {
-        var cells = document.getElementsByClassName(name);
-        for (var i = 0; i < cells.length; i++) {
-            var style = 'inline';
-            if (name === 'lasttcruns') {
-                style = 'table-row';
-            }
-            cells[i].style.display = (cells[i].style.display === 'none') ? style : 'none';
-        }
-    }
-}
-
 // TODO: We should probably reload the page if the clock hits contest
 // start (and end?).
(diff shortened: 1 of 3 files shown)


meisterT:
Another part of https://github.com/DOMjudge/domjudge/pull/2452

vmcj:
I liked the earlier location for the JS toggle more, what was your argumentation for moving it?

meisterT:
> I liked the earlier location for the JS toggle more, what was your argumentation for moving it?

I would argue it is one of the least used buttons on the page and by moving it to the right it is less in the way of the commonly used flow of reading.

eldering:
> > I liked the earlier location for the JS toggle more, what was your argumentation for moving it?
> 
> I would argue it is one of the least used buttons on the page and by moving it to the right it is less in the way of the commonly used flow of reading when scanning the page.

I agree with @vmcj that I don't like the new position of the JS toggle, especially because it might easily be out of sight.

meisterT:
> especially because it might easily be out of sight.

haha, that was my goal, so success I guess :smiling_face_with_tear:  - but I guess I am in the minority here

I will play around with a couple options.

eldering:
> What about _always_ displaying it, but somehow muted?

Looks good!

Just wondering if the `s211` could be moved to the end (like the `judge remaining` button above) and also make it clear that that's the previous submission, e.g. with `previous: s211`. Not sure that improves things though.

vmcj:
> What about _always_ displaying it, but somehow muted?
> 
> For example: image

I like this already more, but would argue might show the differences less flushed out to directly show the information. To steer someone in seeing the right information.

meisterT:
@vmcj I like that idea, also when we are in shadowing mode, but I suggest to address it in follow-up changes as it has many edge cases.

I moved the text to the back of the row as Jaap requested.

image

vmcj (approved):

PR #2548, opened 2024-05-12
Fix r language

vmcj:
Not sure if this is the R way of doing it but it seems to do the job.

### DIFF
--- a/example_problems/fltcmp/submissions/accepted/fltcmp-test-write-files-between-testcases.R
+++ b/example_problems/fltcmp/submissions/accepted/fltcmp-test-write-files-between-testcases.R
@@ -1,0 +1,73 @@
+# This checks if we can write extra files
+# between different testcase runs. R uses
+# a temporary directory which DOMjudge should
+# make unavailable between runs and in case this
+# is forgotten teams can precalculate extra files for later
+# testcases. We need a working solution to make sure
+# we get to all testcases.
+#
+# This submission is symlinked to make sure we submit it twice
+# to verify that we can't see the TMPDIR of the earlier submission.
+#
+# The script will be WRONG-ANSWER if any of the assumptions
+# are wrong.
+# The script will be NO-OUTPUT if there is a security violation.
+# The verdict RUN-ERROR is expected if we correctly detect the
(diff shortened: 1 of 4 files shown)


example_problems/hello/submissions/compiler_error/test-syntax-error.R
```diff
@@ -0,0 +1,5 @@
+# This should give CORRECT on the default problem 'hello'.
```
meisterT, line 1: This comment is outdated.

sql/files/defaultdata/r/run
```diff
@@ -43,10 +47,11 @@ fi
 # debugging.
 # export ONLINE_JUDGE=1 DOMJUDGE=1
 
+export TMPDIR=./
```
meisterT, line 50: If you know why this is necessary, please add a comment explaining the reason

vmcj: It was needed in the past but seems to now work without.

example_problems/fltcmp/submissions/accepted/fltcmp-test-write-files-between-testcases.R
```diff
@@ -0,0 +1,70 @@
+# This checks if we can write extra files
+# between different testcase runs. R uses
+# a temporary directory which DOMjudge should
+# remove between runs and in case this is forgotten
```
meisterT, line 4: I see where the dir is created, but where is it removed?

vmcj: So I rechecked my filesystem and the files are indeed not removed. The test works because we get a different workingdir so the files can't be accessed anymore.

I wonder if we should also do the cleanup as extra security measure.

vmcj: Opened https://github.com/DOMjudge/domjudge/pull/2550 for that,

meisterT (approved):
LGTM, you probably want to squash some/most of the commits before merging.

PR #2549, opened 2024-05-13
Add keyboard shortcuts.

meisterT:
This is just draft code to demonstrate possible behavior and gauge interest.

Go to a single submission, and try typing `j` or `k` to move to the next or previous submission. This also works on all other pages that end with a number.

Also, try typing `s` followed by a sequence of digits and `Enter`, e.g. `s123<Enter>`, it will navigate to the submission page with the submission 123.

Let me know if something like this is interesting.

TODO:
- We want to restrict this to the jury pages.
- Potentially add an option to disable it.
- If we add it, we could add other shortcuts, e.g. `t42` goes to team 42, etc.
- Unfortunately, `j` is both the natural next key as well as the natural key for judgings.
- I am not sure whether I always reset things correctly.
- We probably want to add `Shift + ?` to display a keyboard shortcut help box.
- Other keyboard shortcuts, e.g. to access source code quickly, are trivial to add.

### DIFF
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -809,3 +809,112 @@ function setupPreviewClarification($input, $previewDiv, previewInitial) {
 $(function () {
     $('[data-toggle="tooltip"]').tooltip();
 });
+
+function initializeKeyboardShortcuts() {
+    var $body = $('body');
+    var ignore = false;
+    $body.on('keydown', function(e) {
+        // Check if the user is not typing in an input field.
+        if (e.target.tagName === 'INPUT' || e.target.tagName === 'TEXTAREA') {
+            return;
+        }
+        var key = e.key.toLowerCase();
+        if (key === '?') {
+            var $keyhelp = $('#keyhelp');
+            if ($keyhelp.length) {
+                $keyhelp.toggleClass('d-none');
+            }
+            return;
+        }
+        if (key === 'escape') {
+            var $keyhelp = $('#keyhelp');
+            if ($keyhelp.length && !$keyhelp.hasClass('d-none')) {
+                $keyhelp.addClass('d-none');
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


webapp/public/js/domjudge.js
```diff
@@ -824,3 +824,113 @@ function setupPreviewClarification($input, $previewDiv, previewInitial) {
...
+    var $body = $('body');
+    var ignore = false;
+    $body.on('keydown', function(e) {
+        // Check if the user is not typing in an input field.
+        if (e.target.tagName === 'INPUT' || e.target.tagName === 'TEXTAREA') {
+            return;
+        }
+        var key = e.key.toLowerCase();
+        if (key === '?') {
+            // Find the element with id keyhelp and make it visible if it wasn't visible before.
```
nickygerritsen, line 838: Can you please remove copilot comments? 🤣

webapp/public/js/domjudge.js
```diff
@@ -824,3 +824,113 @@ function setupPreviewClarification($input, $previewDiv, previewInitial) {
...
+                parts[parts.length - 1] = currentNumber + 1;
+            } else if (key === 'k') {
+                parts[parts.length - 1] = currentNumber - 1;
+            }
+            if (params.length > 1) {
+                parts[parts.length - 1] += '?' + params[1];
+            }
+            window.location = parts.join('/');
+        } else if (!ignore && (key === 's' || key === 't' || key === 'p' || key === 'j' || key === 'c')) {
+            if (e.shiftKey) {
```
nickygerritsen, line 870: Let's add a check for key == 's' as well

nickygerritsen (changes requested):

nickygerritsen (approved):

meisterT:
This is the current list of keyboard shortcuts

image

We probably want to add it the manual once we settled on it.

PR #2552, opened 2024-05-14
Fix swift language

vmcj:


sql/files/defaultdata/swift/run
```diff
@@ -50,6 +50,12 @@ fi
 # Add -DONLINE_JUDGE or -DDOMJUDGE below if you want it make easier for teams
 # to do local debugging.
 
+# Compilation needs to store temporary files. The current workdirectory is not
+# available during the submission run so is expected to be safe. The `HOME`
```
nickygerritsen, line 54: I don’t get the “so …” part of this comment

vmcj: I can't think of a way how you can misuse this to be in your favor what is not possible with default compilation.

We assume any writable directory to be a security risk, for this specific writable dir I don't think this is an additional problem.

vmcj, line 54:
```suggestion
# available during the submission run. The `HOME`
```

nickygerritsen (approved):

sql/files/defaultdata/swift/run
```diff
@@ -50,6 +50,12 @@ fi
 # Add -DONLINE_JUDGE or -DDOMJUDGE below if you want it make easier for teams
 # to do local debugging.
 
+# Compilation needs to store temporary files. The current workdirectory is not
+# available during the submission run. The `HOME`
+# is needed for the linker to work.
+export TMPDIR=`pwd`
```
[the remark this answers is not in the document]
vmcj: Does `pwd` depend on `$PWD`?

PR #2553, opened 2024-05-14
Versions edit button

vmcj:
image

I'm not really sure why the unit tests break (the do consistently so it has to be related) but do we want the button like this?

The unit test fails on `/jury/languages/c/edit` and when I try this locally it does work.
Make sure no contest is selected in the menu, visit the versions page, use the button is what the test should do and is what works for my side.

### DIFF
--- a/webapp/tests/E2E/Controller/ControllerRolesTraversalTest.php
+++ b/webapp/tests/E2E/Controller/ControllerRolesTraversalTest.php
@@ -224,7 +224,16 @@ class ControllerRolesTraversalTest extends BaseTestCase
         $response = $this->client->getResponse();
         self::assertNotEquals(500, $response->getStatusCode(), sprintf('Failed at %s', $url));
         if ($dropdown && !str_contains($url, '/public')) {
-            self::assertSelectorExists('a#navbarDropdownContests:contains("no contest")');
+            try {
+                self::assertSelectorExists('a#navbarDropdownContests:contains("no contest")', "Failed at: " . $url);
+            } catch (\Exception $e) {
+                $gitlabArtifacts = getenv('GITLABARTIFACTS');
+                if ($gitlabArtifacts != '') {
+                    $fileHandler = fopen(sprintf("%s/%s", $gitlabArtifacts, str_replace('/', '_s_', $url)), 'w');
+                    fwrite($fileHandler, $response->getContent());
+                }
+                throw $e;
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


webapp/tests/E2E/Controller/ControllerRolesTraversalTest.php
```diff
@@ -224,7 +224,16 @@ public function visitWithNoContest(string $url, bool $dropdown): void
         $response = $this->client->getResponse();
         self::assertNotEquals(500, $response->getStatusCode(), sprintf('Failed at %s', $url));
         if ($dropdown && !str_contains($url, '/public')) {
-            self::assertSelectorExists('a#navbarDropdownContests:contains("no contest")');
+            try {
+                self::assertSelectorExists('a#navbarDropdownContests:contains("no contest")', "Failed at: " . $url);
+            } catch (\Exception $e) {
+                $gitlabArtifacts = getenv('GITLABARTIFACTS');
```
meisterT, line 230: Should we put this code into a test helper class?

Also, there is a bunch of phpcs errors here

vmcj: I think it's only relevant for the Traversal test so I'm not sure if we'll ever re use it.

webapp/tests/E2E/Controller/ControllerRolesTraversalTest.php
```diff
@@ -224,7 +224,16 @@ public function visitWithNoContest(string $url, bool $dropdown): void
...
+                if ($gitlabArtifacts != '') {
+                    $fileHandler = fopen(sprintf("%s/%s", $gitlabArtifacts, str_replace('/', '_s_', $url)) , 'w');
```
vmcj, line 232:
```suggestion
                    $fileHandler = fopen(sprintf("%s/%s", $gitlabArtifacts, str_replace('/', '_s_', $url)), 'w');
```

webapp/tests/E2E/Controller/ControllerRolesTraversalTest.php
```diff
@@ -224,7 +224,16 @@ public function visitWithNoContest(string $url, bool $dropdown): void
...
+                    fwrite($fileHandler, $response->getContent());
+		}
```
vmcj, line 234:
```suggestion
                }
```

meisterT (approved):

PR #2554, opened 2024-05-14
Set default value is not all invocations use 2 arguments

vmcj:
This unbreaks the language versions page. Visiting the page triggered the functions without the second argument.


meisterT (approved):

eldering:
@vmcj could you please next time try to carefully reread your sentences? The title sentence in this commit/PR does not parse. Even if I replace `is` by `if`, then it still really make sense, because the default value is set independently of arguments being provided.

PR #2558, opened 2024-05-21
Version in the compose file is deprecated.

nickygerritsen:


vmcj (approved):
I checked and it shouldn't hurt to remove it.

PR #2561, opened 2024-05-25
Make problem badges a bit wider by default so they have the same size.

nickygerritsen:


vmcj (approved):
Shouldn't we use `em` for when people have a larger fontsize?

nickygerritsen:
Feel free to do that, I'm not a CSS guy

webapp/src/Twig/TwigExtension.php
```diff
@@ -1080,7 +1080,7 @@ public function problemBadge(ContestProblem $problem): string
         // Pick the foreground text color based on the background color.
         $foreground = ($background[0] + $background[1] + $background[2] > 450) ? '#000000' : '#ffffff';
         return sprintf(
-            '<span class="badge problem-badge" style="background-color: %s; min-width: 28px; border: 1px solid %s"><span style="color: %s;">%s</span></span>',
+            '<span class="badge problem-badge" style="background-color: %s; min-width: 37px; border: 1px solid %s"><span style="color: %s;">%s</span></span>',
```
vmcj, line 1083:
```suggestion
            '<span class="badge problem-badge" style="background-color: %s; min-width: 3em; border: 1px solid %s"><span style="color: %s;">%s</span></span>',
```
I did not check what would be the right value but this is what I meant.

nickygerritsen: I played with it and 2.3 seems ok

mpsijm:
Happend to stumble into this when scrolling through the open PRs, happy to help out a bit with CSS-related stuff :slightly_smiling_face: 

These `<span>`s have two classes:
- `badge`, defined by Bootstrap (`lib/vendor/twbs/bootstrap/scss/_badge.scss`)
- `problem-badge`, defined in `webapp/public/style_domjudge.css`.

Since the class `problem-badge` is only used here, I think it's best to move the constant `min-width` to the global `.css` file.

Additionally, the reason why `2.3em` appears to work, is because the `badge` class has a horizontal padding of `0.65em`, and the width of `M` is `1em` by definition, so `2*0.65 + 1 = 2.3em`.
To my personal taste, the badges become a bit wide now, so I would suggest reducing the horizontal padding to `0.5em`. This reduces the `min-width` to `2em`.

This would result in the following rule in `style_domjudge.css` (and removal of the `min-width` rule from `TwigExtension.php`):
```css
.problem-badge {
    --bs-badge-padding-x: 0.5em;
    font-size: 1em;
    min-width: 2em;
}
```
See https://github.com/mpsijm/domjudge/commit/e0f9a03e8f88f2e2d06389dd973f51a74e201dd5, feel free to cherry-pick, but I can also open a PR if you prefer :slightly_smiling_face: 

| Current situation on this branch | With the suggested changes |
|---|---|
| image | image |

nickygerritsen:
Feel free to open the pr!

vmcj:
Closed with: https://github.com/DOMjudge/domjudge/pull/2579

PR #2564, opened 2024-05-25
When you log in to the main DOMjudge, allow to use the API with the same user

nickygerritsen:


vmcj:
I think we should only merge this after the release.

nickygerritsen:
> I think we should only merge this after the release.

I would agree

webapp/config/packages/security.yaml
```diff
@@ -33,8 +33,8 @@ security:
         # API does Basic Auth and IP address auth
         api:
             pattern: ^/api
+            context: domjudge
             provider: domjudge_db_provider
-            stateless: true
```
vmcj, line 37: Does this have a (measurable) performance penalty because we need to remember the sessions?

webapp/src/EventListener/NoSessionCookieForApiListener.php
```diff
@@ -0,0 +1,24 @@
...
+use Symfony\Component\HttpKernel\Event\ResponseEvent;
+
+// The AbstractSessionListener (which sets the cookie) has a priority of -1000, so we need to
+// set a priority of -1001 to run before it.
+#[AsEventListener(priority: -1001)]
+class NoSessionCookieForApiListener
+{
+    public function __invoke(ResponseEvent $event): void
+    {
+        // We do not want to set the session cookie for API requests. Since the firewall is
```
vmcj, line 15: Would it be hard to use the session for the judgehosts/libcurl (and would it give us an advantage)?

webapp/config/packages/security.yaml
nickygerritsen: It does a `session_start` but that shouldn't have any measurable impact.

webapp/src/EventListener/NoSessionCookieForApiListener.php
nickygerritsen: We would need a way to *DO* keep the session cookie, then use it once to log in and then keep passing the session token.

I don't know how big the advantage is, it means we don't have to check the (encrypted) password for every request. If we want this, I'd say we do it in a separate PR.

vmcj (approved):

PR #2565, opened 2024-05-25
Fix progress display of rejudgings that apply automatically.

### LINKED ISSUE #2141: Rejudging is incorrectly displayed in overview as 0% done
I you create a rejudging where the results are applied automatically, in the overview of rejudgings it will always say 0% done. Probably we need to adjust our query to include already applied judgings in the calculation as well.

meisterT:
Fixes #2141.

### DIFF
--- a/webapp/src/Service/RejudgingService.php
+++ b/webapp/src/Service/RejudgingService.php
@@ -390,31 +390,27 @@ class RejudgingService
     public function calculateTodo(Rejudging $rejudging): array
     {
         // Make sure we have the most recent data. This is necessary to
-        // guarantee that repeated rejugdings are scheduled correctly.
+        // guarantee that repeated rejudgings are scheduled correctly.
         $this->em->flush();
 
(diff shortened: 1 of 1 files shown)


webapp/src/Service/RejudgingService.php
```diff
@@ -390,31 +390,27 @@ public function finishRejudging(Rejudging $rejudging, string $action, ?callable
...
-            ->getSingleScalarResult();
-
-        $done = $this->em->createQueryBuilder()
+        $queryBuilder = $this->em->createQueryBuilder()
             ->from(Judging::class, 'j')
             ->select('COUNT(j)')
             ->andWhere('j.rejudging = :rejudging')
+            ->setParameter('rejudging', $rejudging);
+
+        $clonedQueryBuilder = clone($queryBuilder);
```
nickygerritsen, line 402:
```suggestion
        $clonedQueryBuilder = clone $queryBuilder;
```

nickygerritsen (approved):

PR #2568, opened 2024-05-26
Validate entities when importing JSON/TSV files through the API

### LINKED ISSUE #2567: Import team affiliations with illegal country code should report an error
Currently importing an `organizations.json` file which contains illegal country code won't report an error. The illegal code will be stored in the database, which breaks all relevant API endpoints.

nickygerritsen:
Fixes #2567

### DIFF
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -691,7 +691,7 @@ class ImportExportService
     /**
      * Import groups JSON
      *
-     * @param array<array{id?: string, icpc_id: string, name: string, sortorder?: int,
+     * @param array<array{id?: string, icpc_id: string, name?: string, sortorder?: int,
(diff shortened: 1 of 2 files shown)


meisterT (approved):

webapp/src/Service/ImportExportService.php
```diff
@@ -703,15 +703,15 @@ public function importGroupsJson(array $data, ?string &$message = null, ?array &
             $groupData[] = [
                 'categoryid' => @$group['id'],
                 'icpc_id' => @$group['icpc_id'],
-                'name' => @$group['name'],
+                'name' => $group['name'] ?? '',
```
vmcj, line 706: Wouldn't all properties need to be optional/nullable in this case?

nickygerritsen: Nom name is required, and can't be null since the method doesn't accept null. The other things do

PR #2574, opened 2024-05-27
Add `|` as indication between samples and secret cases.

meisterT:
We did have this already in the individual submission page, but not the overview yet.


meisterT:
This fixes #2574

nickygerritsen (approved):

Issue #2465, opened 2024-04-15
Check all foreign key constraints without `ON DELETE` clause


eldering:
### Description of the problem

In #2464 it was found that we have a foreign key constraint that should have an `ON DELETE CASCADE` clause. In general, almost every FK constraint should have either `ON DELETE CASCADE` or `ON DELETE SET NULL`. We should check and if necessary fix all of these: 

```SQL
CREATE TABLE `debug_package`
  CONSTRAINT `FK_9E17399BE0E4FC3E` FOREIGN KEY (`judgehostid`) REFERENCES `judgehost` (`judgehostid`)

-- This one is ok I think:
CREATE TABLE `executable`
  CONSTRAINT `FK_D68EDA01979A9F09` FOREIGN KEY (`immutable_execid`) REFERENCES `immutable_executable` (`immutable_execid`)

CREATE TABLE `judgetask`
  CONSTRAINT `judgetask_ibfk_1` FOREIGN KEY (`judgehostid`) REFERENCES `judgehost` (`judgehostid`)

CREATE TABLE `judging_run`
  CONSTRAINT `judging_run_ibfk_1` FOREIGN KEY (`testcaseid`) REFERENCES `testcase` (`testcaseid`),

CREATE TABLE `version`
  CONSTRAINT `FK_BF1CD3C32271845` FOREIGN KEY (`langid`) REFERENCES `language` (`langid`),
  CONSTRAINT `FK_BF1CD3C3E0E4FC3E` FOREIGN KEY (`judgehostid`) REFERENCES `judgehost` (`judgehostid`)
```

PR #2578, opened 2024-05-28
Remove dangling judgetasks first

vmcj:
Those can exist when you remove a contest so the submissions are cleaned up but not the judgetasks.

This does directly expose a problem though, its not possible to remove the contest anymore as this migration makes sure that the submission can't be removed as there are still judgetasks using it. I looked into it and this boils down to fixing: https://github.com/DOMjudge/domjudge/issues/2465

### DIFF
--- a/webapp/migrations/Version20230508120033.php
+++ b/webapp/migrations/Version20230508120033.php
@@ -19,6 +19,9 @@ final class Version20230508120033 extends AbstractMigration
 
     public function up(Schema $schema): void
     {
+        // Cleanup judgetasks for removed submissions
+        // This can happen when a full contest has been removed and the full delete cascade has failed.
+        $this->addSql('DELETE from judgetask WHERE submitid not in (SELECT submitid FROM submission)');
         // this up() migration is auto-generated, please modify it to your needs
         $this->addSql('ALTER TABLE judgetask ADD CONSTRAINT FK_83142B703605A691 FOREIGN KEY (submitid) REFERENCES submission (submitid) ON DELETE CASCADE');
     }


webapp/migrations/Version20230508120033.php
```diff
@@ -19,6 +19,7 @@ public function getDescription(): string
 
     public function up(Schema $schema): void
     {
+        $this->addSql('DELETE from judgetask WHERE submitid not in (SELECT submitid FROM submission)');
```
meisterT, line 22: Please add a comment why we are doing this.

vmcj, line 22:
```suggestion
        // Cleanup judgetasks for removed submissions
        // This can happen when a full contest has been removed and the full delete cascade has failed. 
        $this->addSql('DELETE from judgetask WHERE submitid not in (SELECT submitid FROM submission)');
```

meisterT: phpcs complains otherwise LGTM

meisterT (approved):

webapp/migrations/Version20230508120033.php
```diff
@@ -19,6 +19,9 @@ public function getDescription(): string
 
     public function up(Schema $schema): void
     {
+        // Cleanup judgetasks for removed submissions
+        // This can happen when a full contest has been removed and the full delete cascade has failed.
```
vmcj, line 23:
```suggestion
        // This can happen when a full contest has been removed and the full delete cascade has failed.
```

PR #2580, opened 2024-05-28
#2537 Give flags a ring-shadow, to make the white parts visible

### LINKED ISSUE #2537: Flags with white part should get a border
See:
Screenshot from 2024-05-10 13-19-55

The `Adam` teams flag gets lost against the background.

mpsijm:
Fixes #2537.

The issue description suggested to add a border, but since this is seen as part of the width/height of the image, this messes up the ratio of the flag (same goes for `padding` btw, so I changed that to `margin`). So, I added a shadow instead.
Note that the flags are now slightly smaller than they used to be because I used round `rem` values for width and height – I hope that's fine, and else we can still increase them :slightly_smiling_face:
Same goes for the hardness of the shadow – `0.3` looks fine to me, but perhaps others have different opinions :smile: 

| Before | After |
|---|---|
| image | image |
| image | image |

### DIFF
--- a/webapp/public/style_domjudge.css
+++ b/webapp/public/style_domjudge.css
@@ -227,7 +227,7 @@ del {
     right: 0;
     bottom: 0;
 }
-.scoreboard .scoreaf { white-space: nowrap; border: 0; text-align: center; }
+.scoreboard .scoreaf { white-space: nowrap; border: 0; padding-left: 2px; text-align: center; }
 .scoreboard .scoreaf img { vertical-align: middle; }
 .univ {
     font-size: 80%;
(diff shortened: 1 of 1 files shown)


vmcj:
image


It does fix the issue, but it looks strange that its the only item popping out (but probably nothing we can do about that..)

mpsijm:
We could make the shadow smaller to lessen the effect, but there's a fine balance there, because if the shadow is too small, it still doesn't fix the issue. We could also use a non-fading `box-shadow` (like [this so-called "ring"](https://tailwindcss.com/docs/ring-width)), will try that tomorrow :slightly_smiling_face:

mpsijm:
I changed the fading shadow into a solid `1px` "ring"-shadow with hardness `0.2` (again, this can be changed). What do you think? :smile: 
Also note that this ring is now in the margin, so we may want to slightly increase this margin.

image
image

In the screenshot of Demoweb below, I manually changed the styling rules and replaced one of the UK flags with the Polish flag from the browser's Dev Console.
image

vmcj (approved):
I like it, lets give the rest a day or two to look at it and otherwise I'll merge it.

nickygerritsen:
This looks great

nickygerritsen (approved):

PR #2583, opened 2024-05-31
Add commands to compare awards, scoreboard and results

nickygerritsen:


webapp/src/Command/AbstractCompareCommand.php
```diff
@@ -0,0 +1,96 @@
...
+        return $this->displayMessages($style, $messages) ?? Command::SUCCESS;
+    }
+
+    /**
+     * @param Message[] $messages
+     */
+    protected function displayMessages(SymfonyStyle $style, array $messages): ?int
+    {
+        if (empty($messages)) {
+            $style->success('Files match fully');
```
vmcj, line 51:
```suggestion
            $style->success('Files match fully.');
```

webapp/src/Service/Compare/ScoreboardCompareService.php
```diff
@@ -0,0 +1,134 @@
...
+                        $problemForFirst = $problem;
+                        break;
+                    } elseif (preg_match('/^Id = ' . preg_quote($problem->problemId, '/') . '-+\d+$/', $problem2->problemId)) {
+                        $problemForFirst = $problem;
+                        break;
+                    }
+                }
+
+                if ($problemForFirst === null && $problem2->solved) {
+                    $this->addMessage(MessageType::ERROR, sprintf('Row %d: Problem %s solved in second file, but not found in first file', $index, $problem2->problemId));
```
vmcj, line 129:
```suggestion
                    $this->addMessage(MessageType::ERROR, sprintf('Row %d: Problem `%s` solved in second file, but not found in first file', $index, $problem2->problemId));
```
Makes it easier to read IMO.

webapp/src/Service/Compare/ScoreboardCompareService.php
```diff
@@ -0,0 +1,134 @@
...
+                        $this->addMessage(MessageType::INFO, sprintf('Row %d: Problem %s time does not match', $index, $problem->problemId), (string)$problem->time, (string)$problemForSecond->time);
+                    }
+                }
+            }
+
+            foreach ($object2->rows[$index]->problems as $problem2) {
+                $problemForFirst = null;
+
+                foreach ($row->problems as $problem) {
+                    // PC^2 uses different problem ID's. For now also match on `Id = {problemId}-{digits}`
```
vmcj, line 118: Are they correct according to the spec? If not we shouldn't do this as we decided to not have `PC2` specific fixes in the main codebase but have them in a separate branch.

nickygerritsen: They are not. But we will not be able to automate checking if we see this as error since it will produce a non 0 exit code

meisterT (approved):
Let's merge this? It can always be improved later

PR #2587, opened 2024-06-01
Replace data sources with shadow mode

### LINKED ISSUE #2087: Importing of files with data source = 0 broken
When we import files with `data source = 0`, we get errors about duplicate keys.
This is because we store the ID of the teams, organizations and groups as `externalid` on the entity, but then use the 'internal' ID to look up stuff.

We could easily change this to always use the external ID to look up, but there are two issues with that:

1. We don't display the external ID anywhere when `data source = 0`.
2. The API uses the internal ID's everywhere with `data source = 0`.

Since ID's in imports can be any format, we can not store them in our internal, numeric ID column.

I see several options and I don't know which is best:

1. Disallow importing files with `data source = 0`.
2. Always show the external ID field and use that for imports. Don't use it in the API with `data source = 0`.
3. Drop `data source = 0`, making us always require an external ID and thus basically making that the new primary key.

If someone knows a better way I am all ears, since I don't like any of these.

nickygerritsen:
Fixes #2087

This makes the API always use external ID's and use sane defaults for the field if not provided.


doc/manual/config-advanced.rst
```diff
@@ -16,30 +16,25 @@ path of your installation) as follows:
 - *Affiliation logos*: these will be shown with the teams that are
   part of the affiliation, if the ``show_affiliation_logos`` configuration
   option is enabled. They can be placed in
-  `public/images/affiliations/1234.png` where *1234* is the numeric ID
+  `public/images/affiliations/1234.png` where *1234* is the external ID
```
meisterT, line 19: Should we explain what external means at the first mention or somewhere?

webapp/migrations/Version20240601180624.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+final class Version20240601180624 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Move data_source to shadow_mode';
+    }
+
+    public function up(Schema $schema): void
+    {
```
meisterT, line 21: No other migration required?

webapp/src/Controller/API/AbstractApiController.php
```diff
@@ -79,25 +79,16 @@ protected function getContestId(Request $request): int
...
-            ->andWhere(sprintf('c.%s = :cid', $this->getContestIdField()))
+            ->andWhere('c.externalid = :cid')
             ->setParameter('cid', $request->attributes->get('cid'));
 
         /** @var Contest|null $contest */
         $contest = $qb->getQuery()->getOneOrNullResult();
 
         if ($contest === null) {
-            throw new NotFoundHttpException(sprintf('Contest with ID \'%s\' not found', $request->attributes->get('cid')));
+            throw $this->createNotFoundException(sprintf('Contest with ID \'%s\' not found', $request->attributes->get('cid')));
```
meisterT, line 89: Why did this change?

nickygerritsen: My editor suggested it, but I will remove it. We can better do this separately.

doc/manual/config-advanced.rst
nickygerritsen: That would be in import.rst. What do you suggest to add?

webapp/migrations/Version20240601180624.php
vmcj: I think the easy fix is to tell people the query they have to run to check if they would get conflicts? And if we have that query we can already run it for them (hopefully in 1 transaction) and fail the migration if we encounter that situation?

nickygerritsen: How would you propose this? I don't think we can do that in a query but only in code, i.e. load all entities and check if it would give an issue. Then if it does, what then? I don't think we can output log/stdout messages in a migration.

And the event feed still remains.

vmcj: Can't we do a "Select COUNT(externalid) from table t leftjoin t2 on t.id=t.externalid where t.externalid=NULL" (did not check the query) which people manually run and see if they have a problem?

When that's possible we can also check how to do this in a migration (but this should be a cornercase IMO which people can manually fix if they know for which entities to do it). If it happens we can just hardfail and stop migrating.

For the eventfeed we would need to update a lot of events right? Let's handle that case after this "simple" migration.

nickygerritsen: That would maybe work: run that check in the migration, make it fail if there is something wrong. People fix it, rerun the migration.

doc/manual/config-advanced.rst
nickygerritsen: I added a block to the manual and linked to it where we mention it.

webapp/migrations/Version20240601180624.php
nickygerritsen: I cooked up a fancy migration 👀

webapp/migrations/Version20240604202419.php
```diff
@@ -0,0 +1,70 @@
...
+
+    protected function setExternalIds(string $table, string $column, string $prefix = '')
+    {
+        $entries = $this->connection->fetchAllAssociative("SELECT $column FROM $table WHERE externalid IS NULL");
+        foreach ($entries as $entry) {
+            $newExternalId = $prefix . $entry[$column];
+            // Check if any entity already has this external ID
+            $existingEntity = $this->connection->fetchAssociative("SELECT externalid FROM $table WHERE externalid = :externalid", ['externalid' => $newExternalId]);
+            $humanReadableTable = ucfirst(str_replace('_', ' ', $table));
+            $this->abortIf((bool)$existingEntity, "$humanReadableTable entity with external ID $newExternalId already exists, manually set a different external ID");
```
vmcj, line 53: So this would change everything up to that entity?

Can we either:
- check this before, and only run the `UPDATE` if all work
- just skip that one and collect all failing and abort with the full list?

nickygerritsen: This is run in one big transaction, so nope. It either changes all or nothing. I tested this.

vmcj: I'm fine with that state... at least we're not halfway for people changing their database.

webapp/src/Entity/Language.php
```diff
@@ -25,7 +25,8 @@
 #[ORM\UniqueConstraint(name: 'externalid', columns: ['externalid'], options: ['lengths' => [190]])]
 #[UniqueEntity(fields: 'langid')]
 #[UniqueEntity(fields: 'externalid')]
-class Language extends BaseApiEntity
+class Language extends BaseApiEntity implements HasExternalIdInterface,
```
vmcj, line 28:
```suggestion
class Language extends BaseApiEntity implements
    HasExternalIdInterface,
```

webapp/src/Entity/PrefixedExternalIdInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php declare(strict_types=1);
+
+namespace App\Entity;
+
+/**
+ * Entities implementing this interface will have their external ID prefixed with 'dj-'.
+ */
+interface PrefixedExternalIdInterface
```
vmcj, line 8:
```suggestion
interface PrefixedExternalIdWithDJInterface
```

It makes that we can remove the comment but I'm not sure if this is really better.

webapp/src/Entity/Problem.php
```diff
@@ -30,7 +30,8 @@
 #[ORM\Index(columns: ['special_compare'], name: 'special_compare')]
 #[ORM\HasLifecycleCallbacks]
 #[UniqueEntity(fields: 'externalid')]
-class Problem extends BaseApiEntity
+class Problem extends BaseApiEntity implements HasExternalIdInterface,
```
vmcj, line 33:
```suggestion
class Problem extends BaseApiEntity implements
    HasExternalIdInterface,
```

webapp/src/Entity/Problem.php
```diff
@@ -30,7 +30,8 @@
...
+    ExternalIdFromInternalIdInterface, PrefixedExternalIdInterface
```
vmcj, line 34:
```suggestion
    ExternalIdFromInternalIdInterface, 
    PrefixedExternalIdInterface
```

I'm not sure if this is something we want in PHPCS.

webapp/src/Form/Type/AbstractExternalIdEntityType.php
```diff
@@ -24,21 +25,19 @@ public function __construct(protected readonly EventLogService $eventLogService)
...
+        $builder->add('externalid', TextType::class, [
+            'label' => 'External ID',
+            'help' => 'Leave empty to generate automatically',
+            'required' => false,
+            'empty_data' => '',
+            'constraints' => [
+                new Regex(
+                    [
+                        'pattern' => DOMJudgeService::EXTERNAL_IDENTIFIER_REGEX,
+                        'message' => 'Only letters, numbers, dashes, underscores and dots are allowed',
```
vmcj, line 37:
```suggestion
                        'message' => 'Only letters, numbers, dashes, underscores and dots are allowed.',
```

webapp/src/Form/Type/AbstractExternalIdEntityType.php
```diff
@@ -24,21 +25,19 @@ public function __construct(protected readonly EventLogService $eventLogService)
...
-                            'message' => 'Only letters, numbers, dashes, underscores and dots are allowed',
-                        ]
-                    ),
-                    new NotBlank(),
-                ]
-            ]);
-        }
+        $builder->add('externalid', TextType::class, [
+            'label' => 'External ID',
+            'help' => 'Leave empty to generate automatically',
```
vmcj, line 30:
```suggestion
            'help' => 'Leave empty to generate automatically.',
```

Not sure if we do this normally for help.

webapp/tests/Unit/Controller/API/BaseTestCase.php
```diff
@@ -31,7 +31,7 @@ abstract class BaseTestCase extends BaseBaseTestCase
      */
     protected array $expectedObjects = [];
 
-    /** If the class to check uses external IDs in non-local mode, set the class name. */
+    /** The class name to use to get external ID's. */
```
vmcj, line 34:
```suggestion
```

vmcj (commented):
I did not check the Tests at all.

webapp/src/Entity/Problem.php
nickygerritsen: Is this in PSR? I'm not sure if I like it more

webapp/src/Doctrine/ExternalIdAlreadyExistsException.php
```diff
@@ -0,0 +1,20 @@
...
+use RuntimeException;
+
+class ExternalIdAlreadyExistsException extends RuntimeException
+{
+    public function __construct(
+        public readonly string $entityClass,
+        public readonly string $externalid
+    ) {
+        $message = sprintf(
+            'An entity of class %s already exists with the external ID "%s"',
```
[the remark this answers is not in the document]
nickygerritsen: No, we use a different message for the user when we catch the exception, which already suggests to change it I believe. I will check.

webapp/src/Doctrine/ExternalIdAssigner.php
```diff
@@ -0,0 +1,62 @@
...
+
+        if ($entity->getExternalId()) {
+            return;
+        }
+
+        $shadowing = $this->dj->shadowMode();
+        $entityClass = get_class($entity);
+        if ($entity instanceof CalculatedExternalIdInterface) {
+            $externalid = $entity->getCalculatedExternalId();
+        } elseif ($shadowing && $entity instanceof UuidExternalIdWhenShadowingInterface) {
```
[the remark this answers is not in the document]
nickygerritsen: Do you think it's better to use that? I'm fine with either

webapp/src/Entity/Problem.php
vmcj: PHPCS complains so I assume so...

The reason I           like it is that now all implemented things are neatly on the same vertical line,
The reason I don't like it is that this reads very unnatural.

```
Lists of implements and, in the case of interfaces, extends MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one interface per line.
```

https://www.php-fig.org/psr/psr-12/ ctrl+f `implements`

webapp/src/Doctrine/ExternalIdAssigner.php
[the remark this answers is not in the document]
vmcj: If you do direct DB stuff to debug I still like shorter things, and if we ever display it in the future somewhere.

Would be good to make this reservation explicit in `CLICS`?

vmcj:
Overall I think this is good and we just need a contest to prove that it works but it would be good if someone else also takes a good look.

webapp/src/Doctrine/ExternalIdAssigner.php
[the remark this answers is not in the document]
nickygerritsen: I don't think you often need to query the DB for the external ID, but as said, if others like it better as well, I am fine. I don't have a huge preference for UUID's.

I don't think CLICS is the way to get a 'reservation'. You can try but I think it is not the right place to do something like this.

webapp/migrations/Version20240601180624.php
nickygerritsen: I would not fix the event feed indeed, since I see no easy way to do this. Not many people use it, even less who also upgrade DOMjudge. We can bump the version number indeed to indicate a breaking change.

webapp/src/Entity/PrefixedExternalIdInterface.php
nickygerritsen: I don't really like it.

meisterT:
It seems to be possible to modify the external IDs in the UI (might have been the case before). That seems asking for trouble.

nickygerritsen:
> It seems to be possible to modify the external IDs in the UI (might have been the case before). That seems asking for trouble.

This was always possible yes. Are you proposing we disallow this after creation? I.e. only add it for new entities?

meisterT:
That is my proposal indeed, but since you didn't introduce it in this PR, let's do in a separate one if you agree

meisterT (approved):

webapp/src/Entity/User.php
```diff
@@ -28,20 +29,24 @@
 #[ORM\UniqueConstraint(name: 'username', columns: ['username'], options: ['lengths' => [190]])]
 #[ORM\UniqueConstraint(name: 'externalid', columns: ['externalid'], options: ['lengths' => [190]])]
 #[UniqueEntity(fields: 'username', message: "The username '{{ value }}' is already in use.")]
-class User extends BaseApiEntity implements UserInterface, PasswordAuthenticatedUserInterface, EquatableInterface, ExternalRelationshipEntityInterface
+class User extends BaseApiEntity implements UserInterface,
```
vmcj, line 32: @nickygerritsen was this forgotten?

webapp/src/Entity/Team.php
```diff
@@ -26,7 +26,10 @@
 #[ORM\UniqueConstraint(name: 'externalid', columns: ['externalid'], options: ['lengths' => [190]])]
 #[ORM\UniqueConstraint(name: 'label', columns: ['label'])]
 #[UniqueEntity(fields: 'externalid')]
-class Team extends BaseApiEntity implements ExternalRelationshipEntityInterface, AssetEntityInterface
+class Team extends BaseApiEntity implements HasExternalIdInterface,
```
vmcj, line 29: Same here.

webapp/src/Entity/User.php
nickygerritsen: Seems I did, pushing a fix.

vmcj (approved):

PR #2581, opened 2024-05-30
Highlight correct submissions during freeze on jury scoreboard.

meisterT:
Fixes #2572

PR #2590, opened 2024-06-07
Highlight pending submissions during freeze on jury scoreboard

### LINKED ISSUE #2572: Add a way to distinguish solves in the freeze on the jury scoreboard
Either always or with some flag show submissions that are accepted in the freeze differently on the jury scoreboard, so that you know what 'the action' will be.

mpsijm:
Also, during the freeze, also show "x + y tries" for jury.

Supersedes #2581 and fixes #2572.

In the implementation, I've tried to make `showPending` mean "the config allows showing pending submissions _and_ the public scoreboard is frozen. Because when the public scoreboard is frozen, we also want special treatment for the jury scoreboard now :smile:

image

Also, please test whether I didn't break any other scoreboards (e.g. for teams during the contest), because my local test set of contest data is fairily limited :innocent:

### DIFF
--- a/webapp/public/style_jury.css
+++ b/webapp/public/style_jury.css
@@ -196,7 +196,7 @@ table.submissions-table {
 }
 
 .devmode {
-    background-color: #295D8A !important;
+    background-color: #295d8a !important;
 }
 
 .devmode-icon {
--- a/webapp/public/style_jury.css
+++ b/webapp/public/style_jury.css
@@ -48,6 +48,11 @@ tr.summary td { border-top: 1px solid black; }
    clickable in the jury scoreboard and for all team cells */
 .scoreboard_jury td, .scoreboard_jury th { padding: 0; }
 
+/* show pending submissions using a blue corner */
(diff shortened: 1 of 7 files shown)


webapp/public/style_jury.css
```diff
@@ -48,6 +48,11 @@ tr.summary td { border-top: 1px solid black; }
    clickable in the jury scoreboard and for all team cells */
 .scoreboard_jury td, .scoreboard_jury th { padding: 0; }
 
+/* show pending submissions using a blue corner */
+.score_pending.score_correct             { background: linear-gradient(45deg, #60e760 80%, #6666ff 80%); }
```
meisterT, line 52: I personally like 85% better, but I am also fine with 80%

meisterT (approved):

webapp/public/style_jury.css
mpsijm: I already increased it from [75%](https://github.com/DOMjudge/domjudge/issues/2572#issuecomment-2154180229) to 80%, but I agree that 85% looks even better :smile: Changed! ✅

vmcj (approved):
I think some of the logic can be simplified for the backend but we can do that in the future.

PR #2594, opened 2024-06-09
Move webstandard tests to github actions

vmcj:
What surprises me is that there are multiple W3C violations found which we didn't catch on GitLab.


vmcj:
@nickygerritsen was fine with this as it only touches CI.

PR #2596, opened 2024-06-11
For submissions and clarifications, prefix external ID in shadow mode

nickygerritsen:
This is needed to prevent collisions.

We can also use this logic to determine if someone submitted in DOMjudge while being shadow (i.e. an analyst instance) to make sure we judge the submission. This fixes the few breaking tests.

### DIFF
--- a/webapp/src/Entity/Clarification.php
+++ b/webapp/src/Entity/Clarification.php
@@ -33,7 +33,8 @@ use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
 #[UniqueEntity(fields: 'externalid')]
 class Clarification extends BaseApiEntity implements
     HasExternalIdInterface,
-    ExternalIdFromInternalIdInterface
+    ExternalIdFromInternalIdInterface,
+    PrefixedExternalIdInShadowModeInterface
 {
     #[ORM\Id]
     #[ORM\GeneratedValue]
--- a/webapp/src/Entity/Submission.php
+++ b/webapp/src/Entity/Submission.php
@@ -38,7 +38,8 @@ use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
 #[UniqueEntity(fields: 'externalid')]
 class Submission extends BaseApiEntity implements
     HasExternalIdInterface,
-    ExternalIdFromInternalIdInterface
+    ExternalIdFromInternalIdInterface,
+    PrefixedExternalIdInShadowModeInterface
 {
     #[ORM\Id]
     #[ORM\GeneratedValue]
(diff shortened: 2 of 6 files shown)


webapp/src/Entity/Clarification.php
```diff
@@ -33,7 +33,8 @@
 #[UniqueEntity(fields: 'externalid')]
 class Clarification extends BaseApiEntity implements
     HasExternalIdInterface,
-    ExternalIdFromInternalIdInterface
+    ExternalIdFromInternalIdInterface,
+    PrefixedExternalIdInShadowModeInterface
```
vmcj, line 37: Maybe document for which case we get this locally? We always expect to read clarifications from the eventfeed but don't disallow clarifications to be sent from the local instance but are never expected to be used outside of the system. For clarifications we should even not put them in the eventfeed if they are local in case we need to swap systems in the future?

webapp/src/Entity/Submission.php
```diff
@@ -38,7 +38,8 @@
 #[UniqueEntity(fields: 'externalid')]
 class Submission extends BaseApiEntity implements
     HasExternalIdInterface,
-    ExternalIdFromInternalIdInterface
+    ExternalIdFromInternalIdInterface,
+    PrefixedExternalIdInShadowModeInterface
```
vmcj, line 42: As stated above, this case can happen if people resubmit after changing the code, useful for both jury as analysts.

webapp/tests/Unit/Controller/Jury/QueueTaskControllerTest.php
```diff
@@ -144,7 +144,7 @@ public function testLazy(bool $shadowMode, int $globalLazy, int $problemLazy): v
         $config   = self::getContainer()->get(ConfigurationService::class);
         $eventLog = self::getContainer()->get(EventLogService::class);
         $dj       = self::getContainer()->get(DOMJudgeService::class);
-        $config->saveChanges(['lazy_eval_results'=>$globalLazy], $eventLog, $dj);
+        $config->saveChanges(['lazy_eval_results'=>$globalLazy], $eventLog, $dj, treatMissingBooleansAsFalse: false);
```
vmcj, line 147: I don't know why this change is needed?

vmcj (approved):

webapp/tests/Unit/Controller/Jury/QueueTaskControllerTest.php
nickygerritsen: Since we don't resupply the boolean shadow mode, it would assume it is false again without this option

webapp/src/Entity/Clarification.php
nickygerritsen: Added a comment to the interface itself.

vmcj (approved):

Issue #2279, opened 2023-12-13
In the contestant's clarifications block, the shortname and color of the problem are not displayed correctly.


Dup4:
NOTE: If this is a _discussion starter_, you need any _installation help_ or have a _question_ on how to accomplish something,
rather post at [our discussion channel](https://www.domjudge.org/chat) or send an email to our
[DOMjudge-devel mailinglist](https://www.domjudge.org/mailman/postorius/lists/domjudge-devel.domjudge.org/)
instead of filing an issue here.

### Description of the problem

<img width="1109" alt="image" src="https://github.com/DOMjudge/domjudge/assets/38343778/bcec4acf-3e36-431b-b8bc-2237d5b8db9a">

This is the problem I set in the contest, with the correct shortname and color set.

<img width="1690" alt="image" src="https://github.com/DOMjudge/domjudge/assets/38343778/035c14af-d810-4a03-9560-5c6e0c72fce5">

This is the Clarifications page in the administrator interface, showing the correct shortname and color.

<img width="935" alt="image" src="https://github.com/DOMjudge/domjudge/assets/38343778/ab4c2052-7318-4f73-8042-cc431b193930">

But this is the Clarifications page in the contestant's interface, and it doesn't show the correct shortname and color.

This should be the shortname of the problem.

### Your environment

- DOMjudge: 8.2.2
- OS: Ubuntu22.04
- webserver: nginx

### Steps to reproduce

- create a contest.
- send a clarification in contestant's interface.

### Expected behaviour
> Replace this line with what you would expect to happen.

On the Clarifications page in the contestant's interface, the correct shortname and color are shown.

### Actual behaviour
> Replace this line with what happens instead.

The Clarifications page in the contestant's interface currently shows the externalid of the title and the color black.

### Any other information that you want to share?

<img width="1820" alt="image" src="https://github.com/DOMjudge/domjudge/assets/38343778/059ac4f3-3df2-48f4-af5e-8d415d9b53c2">

My data source options are set like this.

nickygerritsen:
I tried reproducing this on 8.2.2, but I can't, even with that data source option. A dump of relevant tables of your database might help.

Dup4:
Maybe it's because I have two contests?

<img width="1793" alt="image" src="https://github.com/DOMjudge/domjudge/assets/38343778/a3395d38-dd77-4029-9de6-b1addd3cf970">

TESTOFFICIAL:

[...]

PR #2600, opened 2024-06-16
fix: no clarfication on team dashboard

cubercsl:


vmcj:
@cubercsl for what type of clarifications was this a fix?

cubercsl:
Clarification associates either problem or contest. The existing code selects `contestproblem` and `contest`, which selects nothing :(. So team will see nothing on their team home, but if they enable notification, they can click it and jump to the clarification URL and see its content.

cubercsl:
This is our hotfix between warmup and contest, but this may not occur in demoweb, I will verify this later.

vmcj:
> Clarification associates either problem or contest. The existing code selects `contestproblem` and `contest`, which selects nothing :(. So team will see nothing on their team home, but if they enable notification, they can click it and jump to the clarification URL and see its content.

From a quick glance at the code I would assume the `LeftJoin` would already have to do this. The only reason for not merging this would be the speed penalty though so let's see if we can reproduce this somewhere and I'll take a look later this week.

cubercsl:
@vmcj I have repro this on demoweb, pls check it.

vmcj (approved):
I'll give the rest a bit of time to look at it but I agree its broken.

webapp/src/Controller/Team/MiscController.php
```diff
@@ -103,8 +103,7 @@ public function homeAction(Request $request): Response
                 ->leftJoin('c.sender', 's')
                 ->leftJoin('c.recipient', 'r')
                 ->select('c', 'cp', 'p')
-                ->andWhere('c.contest = :contest')
-                ->andWhere('cp.contest = :contest')
+                ->andWhere('c.contest = :contest OR cp.contest = :contest')
```
eldering, line 106: General question about the existing code:
When would ever `cp.contest` match `$contest`, but `c.contest` not? That is, why can't we just have
```
->andWhere('c.contest = :contest')
```

nickygerritsen: When you unlink a problem from a contest is the only way I can think of.

cubercsl:
@Dup4 told me that in his shadow (of my instance), clarification (to all) can be displayed correctly on the team homepage.
I suspect that in the current code, the clarification created by the shadow is inconsistent with the behavior created by the user from the UI.

webapp/src/Controller/Team/MiscController.php
moesoha: this was implemented in 0219d8e173540e26b5867d219753dcb23f08fcc5

moesoha:
I found that we misunderstood the motivation of the `->andWhere('cp.contest = :contest')` after `->andWhere('c.contest = :contest')`, so this PR should be altered.

At first, I think `->andWhere('cp.contest = :contest')` is used to support a problem in multiple contests (like one problem exists in Div.1 and Div.2). By this way, clarification can be decoupled with contests, and can be associated with either Contest or Problem. And a clarification associated with a problem can be synced to all contests.

But since I investigated the codes today, I found the commit implements above codes: 0219d8e173540e26b5867d219753dcb23f08fcc5. This code fixes the issue mentioned in #2279, but I don't think this is a solution. Obviously, @vmcj misunderstood the usage of LEFT JOIN, and the fix for this issue is actually an accident. I'd like to draft a new PR  to fix #2279 again.

PR #2603, opened 2024-06-16
ui: problem_list: fix line break

undefined-moe:


vmcj:
Hi @undefined-moe,

Can you show a screenshot from the before and after so we can see why this is better?

cubercsl:
I think `</br>` should be shown a new line, but it  show as itself. You can verify this in demoweb.

cubercsl:
Before:
image
After:
image

vmcj:
Yes that's indeed a bug.

I personally don't like to have Unicode chars for this but can't think of something else for now. Can you add a comment above explaining what it does (and/or explain the issue in your commit message)?

moesoha:
> I personally don't like to have Unicode chars for this but can't think of something else for now.

I recommend using something like `{{ '\n' ~ label }}` to prevent this.

nickygerritsen:
> > I personally don't like to have Unicode chars for this but can't think of something else for now.
> 
> I recommend using something like `{{ '\n' ~ label }}` to prevent this.

Or even just `{{ '\n' }}`? I think I prefer that

vmcj:
@undefined-moe can you squash the commits and rewrite your commit message?

undefined-moe:
well there's "squash and merge" if you expand that dropdown besides "merge pull request" button.

nickygerritsen:
We use merge queues, and then that button is not there anymore

vmcj (approved):
Verified and this fixes the original issue.

It does not work on `/jury/problems/problemset` but we can look into that later.

cubercsl:
> Verified and this fixes the original issue.
> 
> It does not work on `/jury/problems/problemset` but we can look into that later.

@vmcj It looks like ` $('[data-bs-toggle="tooltip"]').tooltip();` is called in `/jury/problems/problemset` but not in `/public`.

vmcj:
@cubercsl, my remark was not meant that we need a fix now (as it was already broken before),

Feel free to create a PR for it. This does add the `<br>` back though? IIRC the newline did not work for the preview problemset page so I think that's that page we should change and not the other ones. Maybe this discussion works better in a new issue though as we're now commenting on a closed PR.

PR #2608, opened 2024-06-19
Escape \ in Python strings

nickygerritsen:
It seems Python complains about this for some Python versions.


vmcj (approved):
Lets see if CI stops complaining.. also why was the script not actually failing on this.

PR #2612, opened 2024-06-21
Use correct API doc root by reading it from the request

### LINKED ISSUE #2611: `/domjudge/api/doc` Unable to load correct server address
image

In 8.2.3, `/domjudge/api/doc` can get the correct server address, as shown in the image above

image

But in 8.3.0, Cannot get the correct server address

### Your environment

- DOMjudge: 8.3.0
- OS: Ubuntu22.04

### Steps to reproduce

To run the 8.3.0 domserver image using docker

nickygerritsen:
Fixes #2611

Also we should not append /api since the swagger already does this.

Should be back-ported to 8.3.

### DIFF
--- a/webapp/config/packages/nelmio_api_doc.yaml
+++ b/webapp/config/packages/nelmio_api_doc.yaml
@@ -1,9 +1,9 @@
 nelmio_api_doc:
     documentation:
         servers:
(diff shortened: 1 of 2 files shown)


vmcj (approved):
I've wrongly assumed this was only for docker.

PR #2616, opened 2024-07-01
Splitoff admin only roles for API

vmcj:
We had this request both during EUC and this would help during next BAPC prelims where we probably ask a BAPC jury member to upload the problems already.

For the 2nd case I would personally trust that person with `admin` access but others might see that differently. Also only giving the needed access is always a good thing and given that this is easy to test I don't think this would be a real feature creep.


webapp/migrations/Version20240629154640.php
```diff
@@ -0,0 +1,44 @@
...
+
+    public function getDescription(): string
+    {
+        return 'Add new roles to the database.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        foreach(self::NEW_ROLES as $role => $description) {
+            $existingRole = $this->connection->fetchAssociative('SELECT * FROM role WHERE role = :role', ['role' => $role]);
```
nickygerritsen, line 23: When would the role already exist? You can't create it in the UI so I'm not sure if we should handle this case?

webapp/config/packages/security.yaml
```diff
@@ -3,10 +3,10 @@
 security:
     role_hierarchy:
         ROLE_JURY: [ROLE_CLARIFICATION_RW, ROLE_API, ROLE_API_READER, ROLE_API_SOURCE_READER]
-        ROLE_ADMIN: [ROLE_JURY, ROLE_JUDGEHOST, ROLE_API_WRITER]
+        ROLE_API_ADMIN: [ROLE_API_PROBLEM_CHANGE, ROLE_API_CONTEST_CHANGE]
```
nickygerritsen, line 6: Why this? Do we need it?

webapp/migrations/Version20240629154640.php
vmcj: I suspected we load the default fixture before the migrations.

nickygerritsen: I don't think so. Since then some tables might still be wrong

vmcj: I checked and you're right. Added a commit for it.

nickygerritsen (approved):

webapp/migrations/Version20240629154640.php
```diff
@@ -0,0 +1,41 @@
...
+                               'api_contest_change' => 'API Contest Changer'];
+
+    public function getDescription(): string
+    {
+        return 'Add new roles to the database.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        foreach(self::NEW_ROLES as $role => $description) {
```
vmcj, line 22:
```suggestion
        foreach (self::NEW_ROLES as $role => $description) {
```

webapp/migrations/Version20240629154640.php
```diff
@@ -0,0 +1,41 @@
...
+            $this->addSql(
+                'INSERT INTO role (`role`, `description`) VALUES (:role, :desc)',
+                ['role' => $role, 'desc' => $description]
+            );
+        }
+    }
+
+    public function down(Schema $schema): void
+    {
+        foreach(array_keys(self::NEW_ROLES) as $role) {
```
vmcj, line 32:
```suggestion
        foreach (array_keys(self::NEW_ROLES) as $role) {
```

webapp/config/packages/security.yaml
```diff
@@ -3,10 +3,10 @@
 security:
     role_hierarchy:
         ROLE_JURY: [ROLE_CLARIFICATION_RW, ROLE_API, ROLE_API_READER, ROLE_API_SOURCE_READER]
-        ROLE_ADMIN: [ROLE_JURY, ROLE_JUDGEHOST, ROLE_API_WRITER]
+        ROLE_ADMIN: [ROLE_JURY, ROLE_JUDGEHOST, ROLE_API_WRITER,
+                     ROLE_API_PROBLEM_CHANGE, ROLE_API_CONTEST_CHANGE]
```
eldering, line 7: I think `editor` would be a better suffix than `change`. Also, is `ROLE_API_CONTEST_CHANGE` a superset of privileges over `ROLE_API_PROBLEM_CHANGE`?

webapp/migrations/Version20240629154640.php
```diff
@@ -0,0 +1,41 @@
...
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20240629154640 extends AbstractMigration
+{
+    private const NEW_ROLES = ['api_problem_change' => 'API Problem Changer',
+                               'api_contest_change' => 'API Contest Changer'];
```
eldering, line 13: Could we have some more detailed descriptions of these roles? How do admin/API writer, contest changer and problem changer relate?

webapp/migrations/Version20240629154640.php
```diff
@@ -0,0 +1,46 @@
...
+
+final class Version20240629154640 extends AbstractMigration
+{
+    private const NEW_ROLES = ['api_problem_editor' => 'API Problem Editor',
+                               'api_contest_editor' => 'API Contest Editor'];
+
+    public function getDescription(): string
+    {
+        return "Add new roles to the database.
+                Problem editor can add/delete/edit anything related to problems; files, testcases.
```
nickygerritsen, line 18: I don't think any (end) user will read this and I think @eldering his feedback is that people now don't know what the roles do without checking the code.

eldering: Exactly. What the roles do should ideally be clear from the role constant names, but at least from the role descriptions .

meisterT:
Two questions:
- Do we need the split in two separate editor roles or is one enough?
- Is the mechanism over which the change happens (Web vs API) relevant?

If the answer is no to both of these questions, then one role `CONTEST_EDITOR` who can edit contests including problems and testdata both via web and API should be enough.

vmcj:
> Two questions:
> 
>     * Do we need the split in two separate editor roles or is one enough?
I think BAPCtools also wants to set the contest, and the judges of EUC wanted to only import the problems and nothing else? I'm fine with merging them as its already better than admin and we can always decide to split them later if there is a real usecase.

> 
>     * Is the mechanism over which the change happens (Web vs API) relevant?
It was for me, I rather not also have to do this in code and if you have the zips already you can also just give them to an admin. I think is is only relevant for automated tools and during the contest I wouldn't want jury members to be able to change things which they might not be aware what are the differences (testcase order etc.).

So its both laziness for the amount of code changed and only giving the rights which are needed. The same distinction was made for locking a contest.

webapp/config/packages/security.yaml
vmcj: No, a contest is a different thing from a problem as we allow problems without a contest.

PR #2617, opened 2024-07-08
Implement new results rules

### LINKED ISSUE #2615: Implement new result rules
- `B` can be dropped (but let's leave it for now?)
- From the SysOps agenda: Teams competing in the ICPC World Finals are eligible for Championship Honors awards (defined as follows):
  - Highest Honors: All medalist teams together with all teams solving the same number of problems as the minimum number of problems solved by any medalist team.
  - High Honors: All teams solving one fewer problem than the minimum number of problems solved by any medalist team.
  - Honors: All teams not receiving Highest Honors or High Honors solving the same number of problems or more than the median scoring team.
  - Honorable Mention: All teams solving fewer than the number of problems solved by the median scoring team.

Honorable Mention is as it is before. The other three groups are new. Note that any of these list can be empty if there are no teams within that group.

nickygerritsen:
Fixes #2615


webapp/src/Service/ImportExportService.php
```diff
@@ -456,16 +457,7 @@ public function getTeamData(): array
     /**
      * Get results data for the given sortorder.
      *
```
vmcj, line 459:
```suggestion
    /**
```

The only relevant bit is the sortorder, so I suggest we remove it.

webapp/src/DataTransferObject/ResultRow.php
```diff
@@ -0,0 +1,40 @@
+<?php declare(strict_types=1);
+
+namespace App\DataTransferObject;
+
+class ResultRow
+{
+    public function __construct(
+        public readonly string $teamId,
+        public readonly ?int $rank,
+        public readonly string $award,
```
vmcj, line 10: Will everyone now always get an award? Or will `null` now be an empty string?

webapp/src/Service/ImportExportService.php
```diff
@@ -542,58 +538,56 @@ public function getResultsData(int $sortOrder, bool $full = false): array
...
+                $rank        = null;
             }
 
-            $groupWinner = "";
+            $groupWinner = null;
             $categoryId  = $teamScore->team->getCategory()->getCategoryid();
             if (!isset($groupWinners[$categoryId])) {
                 $groupWinners[$categoryId] = true;
                 $groupWinner               = $teamScore->team->getCategory()->getName();
             }
```
vmcj, line 558:
```suggestion
            if (isset($groupWinners[$categoryId])) {
                // Winner was encountered in earlier iteration.
                $groupWinner = null;
            } else {
                $groupWinners[$categoryId] = true;
                $groupWinner               = $teamScore->team->getCategory()->getName();
            }
```

vmcj: It had too many negations for me, this adds 1 extra line and is easier for me to read.

webapp/tests/Unit/Service/ImportExportServiceTest.php
```diff
@@ -43,63 +52,72 @@ public function testImportContestDataErrors(mixed $data, string $expectedMessage
...
-            'Can not parse start_time'
+            'Can not parse start_time',
         ];
         yield [
             [
-                'name'                     => 'Test contest',
-                'short-name'               => 'test',
-                'duration'                 => '5:00:00',
-                'start-time'               => '2020-01-01T12:34:56+02:00',
+                'name' => 'Test contest',
```
vmcj, line 83: Is there anything how we can disable those changes for your editor? In this case I find this harder to read and I think PSR has no opinion about this kind of indentation?

webapp/tests/Unit/Service/ImportExportServiceTest.php
```diff
@@ -1149,4 +1182,300 @@ protected function getContest(int|string $cid): Contest
...
+                $teamA = $teamsByIcpcId[$a->teamId] ?? null;
+                $teamB = $teamsByIcpcId[$b->teamId] ?? null;
+                $nameA = $teamA?->getEffectiveName();
+                $nameB = $teamB?->getEffectiveName();
+                $collator = new Collator('en');
+                return $collator->compare($nameA, $nameB);
+            });
+            $expectedResults = array_values($expectedResults);
+        }
+
```
vmcj, line 1471:
```suggestion
```

vmcj (approved):

webapp/src/DataTransferObject/ResultRow.php
nickygerritsen: We already had an award for everyone, since even the lowest people get 'Honorable'.

webapp/tests/Unit/Service/ImportExportServiceTest.php
nickygerritsen: This is me manually pressing a button to auto format a file before save, which is baked into my muscle memory, since I like it. I will undo this for this file.

nickygerritsen:
I added a way to expose both the old (ranked) and new (honors, etc.) way. This made me decide to change the export page to turn the results export into a form: 

image

Note @tuupke that this requires changes to the EOC Ansible.

meisterT (approved):

PR #2618, opened 2024-07-12
Fix forgotten merge

vmcj:


webapp/vendor/Makefile
tom93, line 1: Just a thought -- would it be simpler if this file was merged into webapp/Makefile? (It's possible to use paths such as "vendor/autoload_runtime.php" in make targets, and I think it's easier to follow than the nesting and the "..".)

tom93 (commented):
There are a couple of places that still mention lib/vendor, do they need updating?
```
Makefile:292:  -rm -rf lib/vendor/*
.github/jobs/data/codespellignorefiles.txt:13:./lib/vendor
.github/jobs/syntax-check:26:       -path ./lib/vendor        -prune \
.github/jobs/syntax-check:44:          echo "$i" | grep -qvE ...
```
*Edit:* Oops, I missed .github/jobs/syntax-check:51. I'll fix it in #2635.

webapp/vendor/Makefile
vmcj: It was in the past I specifically choose to take it out. So I think we disagree on the easier to follow.

I'll see if I'll revert that.

webapp/composer.json
```diff
@@ -1,5 +1,6 @@
 {
 	"name": "domjudge/domjudge",
+	"version": "8.4.0DEV",
```
tom93, line 3: Why is the version required? It seems like things work without it -- at commit cd9608857, if I run `make dist` then webapp/vendor/composer/installed.php contains `'version' => 'dev-main'`; and if I first run `git tag 8.3.1` then it contains `'version' => '8.3.1.0'`.

vmcj: I'll check again, I got a local warning which I wanted to get rid of.

tom93: Do you have any tips for spotting such warnings? (When I run the `make` commands there is a lot of output and I don't notice warnings; should I use an IDE?)

vmcj: I think `composer install` triggered it and otherwise `composer dump-autoload`.

I plan to work on this a bit more this week, so when I encounter it again I can pinpoint it a bit more.

What probably works is (from a clean checkout):
```
make dist | less
search for domjudge/domjudge
```

as it mentions that as package.

vmcj: I can't reproduce it with a clean checkout so my checkout might have been in a bad state.

webapp/vendor/Makefile
vmcj: @tom93 feel free to create a PR for the fix,

I can see your argument about `..` although I favor a Makefile per directory to make it easier to see what is happening for that directory. It was originally like you mentioned so it more people did agree with you in the past (probably).

tom93 (commented):
Earlier commits introduced an issue, `make maintainer-install` doesn't actually run `composer-dump-autoload-dev`. I committed a fix and some other cleanup in branch 'fix_forgotten_merge_suggestions' of [my fork](https://github.com/tom93/domjudge) ([commits](https://github.com/tom93/domjudge/compare/aaf78ffa1de83d8fabbbc7c9c0703e08bda5a566...d320b7ee6acf384fba884f0001220b75b31503db)).

webapp/vendor/Makefile
tom93: I have a draft in branch 'merge_webapp_vendor_Makefile' of my fork (it's the last commit). I moved `composer auto-scripts` back into `inplace-install-l` because I think the recent change that moved it into a prereq of `inplace-install` might suffer from ordering issues, but I'm not sure if that's the right approach.

vmcj:
> Earlier commits introduced an issue, `make maintainer-install` doesn't actually run `composer-dump-autoload-dev`. I committed a fix and some other cleanup in branch 'fix_forgotten_merge_suggestions' of [my fork](https://github.com/tom93/domjudge) ([commits](https://github.com/tom93/domjudge/compare/aaf78ffa1de83d8fabbbc7c9c0703e08bda5a566...d320b7ee6acf384fba884f0001220b75b31503db)).

I'll cherry-pick (& alter) some of those.

nickygerritsen (approved):

tom93:
@vmcj, what do you think about a0fd9fd0d7d6714c2585393971db67f2d7932f21 (simplify relative paths -- ../webapp/vendor becomes vendor)? Should I create a separate PR, or just drop it?

vmcj:
> @vmcj, what do you think about [a0fd9fd](https://github.com/DOMjudge/domjudge/commit/a0fd9fd0d7d6714c2585393971db67f2d7932f21)? (That's the simplified relative paths -- _../webapp/vendor_ becomes _vendor_). Should I create a separate PR, or just drop it?

I liked both of the commits but wanted to close this PR first.

So bugfix before refactoring.

tom93:
> I liked both of the commits but wanted to close this PR first.

I created a separate PR (#2635).

webapp/vendor/Makefile
tom93: PR: #2636

PR #2620, opened 2024-07-22
MySQL main container

vmcj:


nickygerritsen (approved):

eldering:
Could you clarify the commit messages. The first "[remove]" is that commit intended to be dropped? Then please do so. The second, it's not clear which module is deprecated and how we're fixing this here: do we keep the old module, or not?

PR #2621, opened 2024-07-23
Fix updates when no contest is set

vmcj:
I think this method of solving is actually wrong as we normally have another behaviour.

| #contests | Cookie selected | Normally shown      | Current behaviour balloonspage |
|-----------|-----------------|---------------------|--------------------------------|
| 0         | -1              | Nothing             | Same                           |
| 2         | 1               | All of contest 1    | Same                           |
| 2         | -1              | All of all contests | **Nothing**                      |

I can see the usecase for having 2 contests like the PacNW situation with divisions where you would still want to hand out balloons to both contests so I think we should refactor to also have the same behavior here (when selectedContest=-1 -> display everything from all active contests).

### DIFF
--- a/webapp/src/Service/DOMJudgeService.php
+++ b/webapp/src/Service/DOMJudgeService.php
@@ -421,16 +421,16 @@ class DOMJudgeService
 
         if ($this->checkrole('balloon')) {
             $balloonsQuery = $this->em->createQueryBuilder()
-            ->select('b.balloonid', 't.name', 't.location', 'p.name AS pname')
-            ->from(Balloon::class, 'b')
-            ->leftJoin('b.submission', 's')
-            ->leftJoin('s.problem', 'p')
-            ->leftJoin('s.contest', 'co')
-            ->leftJoin('p.contest_problems', 'cp', Join::WITH, 'co.cid = cp.contest AND p.probid = cp.problem')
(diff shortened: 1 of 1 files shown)


webapp/src/Service/DOMJudgeService.php
```diff
@@ -419,25 +419,22 @@ public function getUpdates(): array
...
+                ->leftJoin('s.problem', 'p')
+                ->leftJoin('s.contest', 'co')
+                ->leftJoin('p.contest_problems', 'cp', Join::WITH, 'co.cid = cp.contest AND p.probid = cp.problem')
+                ->leftJoin('s.team', 't')
+                ->andWhere('co.cid = :cid')
+                ->setParameter('cid', $contest->getCid())
+                ->andWhere('b.done = 0');
+
+            if (!(bool)$this->config->get('show_balloons_postfreeze')) {
+                $balloonsQuery->andWhere('s.submittime < co.freezetime');
```
nickygerritsen, line 436: If there is no freeze, what does this do?

vmcj: That's a good question. I would say we wouldn't handout balloons at all. I'll check for that (or just use the old code again).

nickygerritsen (approved):

Issue #2462, opened 2024-04-15
Link from card on contests overview page to single contest page


eldering:
### Description of the enhancement request

On the contests overview page each active contest is shown as a "card". There are links to all (not just active) contests in the bottom table, but if there are various active contests, you have to scroll down to find it.
It would be nice to make the header of each card a link to that individual contest page, see the image
![image](https://github.com/DOMjudge/domjudge/assets/4692022/b3a12fc2-6748-4fe3-b15b-1f7f53c7dd90)

PR #2624, opened 2024-07-25
Link from card on contests overview page to single contest page

vmcj:
The issue suggested using the header only but that felt a bit hard for people to find the link so the whole card is used now. The buttons in the card still work with their own links.

Fixes: #2462

```
<a href={{ path('jury_contest', {'contestId': contest.cid}) }}>
```

Is the only interesting piece, the rest is fixing indentation.

### DIFF
--- a/webapp/templates/jury/contests.html.twig
+++ b/webapp/templates/jury/contests.html.twig
@@ -23,6 +23,7 @@
                         {% if contest.locked %}
                             <i class="fas fa-lock"></i>
                         {% endif %}
+                        <a class="fa-regular fa-folder-open" style="float: right; color: black" href={{ path('jury_contest', {'contestId': contest.cid}) }}></a>
                     </div>
                     <div class="card-body">
                         {% if not contest.starttimeEnabled and contest.finalizetime is not empty %}


nickygerritsen:
Shouldn't we only link from the card header instead of the whole card?

vmcj:
> Shouldn't we only link from the card header instead of the whole card?

See my first message, I found that hard to find and I don't see the problem with making the whole card clickable. Do we want a button to attract attention as alternative?

nickygerritsen:
I find it a bit strange that you can now click for example on the checkmark to open the contest, but that might just be me.

vmcj:
> I find it a bit strange that you can now click for example on the checkmark to open the contest, but that might just be me.

Ok, do you want a button as alternative? I think it will be hard for people to find the feature if it only works with the header and I rather not have extra buttons. I do agree that clicking on the checkbox is a little strange. Let's see what the rest thinks.

vmcj:
> What about adding a link to the header with `<i class="fa-solid fa-pen"></i>` as icon?

I would consider that a button, but I'll change the PR with that change.

vmcj:
image

I associate pen with editing so went with another icon but anything from https://fontawesome.com/search?o=r&m=free should be fine.

nickygerritsen (approved):

PR #2627, opened 2024-07-26
Get rid of some deprecations

nickygerritsen:
Will change the deprecation count check in CI after it has run once, so I know the new value.

### DIFF
--- a/webapp/src/Controller/Jury/AnalysisController.php
+++ b/webapp/src/Controller/Jury/AnalysisController.php
@@ -10,6 +10,7 @@ use App\Service\DOMJudgeService;
 use App\Service\StatisticsService;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Query\Expr;
+use Symfony\Bridge\Doctrine\Attribute\MapEntity;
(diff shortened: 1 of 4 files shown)


vmcj (commented):
When do we use this?

nickygerritsen:
When do we use what?

webapp/src/Controller/Jury/AnalysisController.php
```diff
@@ -99,6 +100,7 @@ public function teamAction(Team $team): Response
 
     #[Route(path: '/problem/{probid}', name: 'analysis_problem')]
     public function problemAction(
+        #[MapEntity(id: 'probid')]
```
vmcj, line 103: When do we use this?

nickygerritsen: Map entity makes it so that `$problem` gets injected properly using the `probid` route argument as primary key (i.e. as `id`, not a field called `id`) in the method. This used to be automatic behavior but with `doctrine.orm.controller_resolver.auto_mapping: false`, which will be the default soon, this is not the case anymore

vmcj (approved):

Issue #2563, opened 2024-05-25
Fallback from `lsb_release` to `/etc/os-release`


vmcj:
Not all distributions have `lsb_release` default installed, for all distributions with `systemd` `/etc/os-release` seems to be required.

The script `dj_make_chroot` only uses `lsb_release` but can be extended to use `/etc/os-release`. For an example on fedora:
```
NAME="Fedora Linux"
VERSION="40 (Container Image)"
ID=fedora
VERSION_ID=40
VERSION_CODENAME=""
PLATFORM_ID="platform:f40"
PRETTY_NAME="Fedora Linux 40 (Container Image)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:40"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f40/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=40
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=40
SUPPORT_END=2025-05-13
VARIANT="Container Image"
VARIANT_ID=container
```

For more info see:
https://discussion.fedoraproject.org/t/what-is-the-official-way-to-determine-that-the-remote-linux-is-fedora/72482
http://linuxmafia.com/faq/Admin/release-files.html

PR #2630, opened 2024-07-26
Use /etc/os-release as alternative when lsb_release is not available

vmcj:
Extract the OS/release information

This is not 100% doable with /etc/os-release only. The Debian difference between Release & Codename is not shown in the official containers for example. See: https://hub.docker.com/_/debian

Closes: #2563.


vmcj:
Implemented @thijskh his suggestion.

PR #2635, opened 2024-07-29
Simplify relative paths to webapp/vendor

tom93:
Continued from https://github.com/DOMjudge/domjudge/pull/2618#issuecomment-2242979558.
Also fix a leftover reference to lib/vendor.
Also simplify an unrelated path in webapp/tests/\*.

### DIFF
--- a/.github/jobs/syntax-check
+++ b/.github/jobs/syntax-check
@@ -48,7 +48,7 @@ while read -r i ; do
 	fi
 	if grep -qE "^#\\!.*/bin/(ba)?sh" "$i" && \
 		[ "${i##*.}" != "zip" ] && \
-		echo "$i" | grep -qvE '(^\./(config|autom4te|install-sh|sql/files/defaultdata/hs/run|lib/vendor/|output|judge/judgedaemon))'; then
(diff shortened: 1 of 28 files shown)


vmcj (approved):

tom93:
I was playing with the web demo just before, not sure if it's related to the Sentry issue but thought I should mention anyway. Steps: log in to [demoweb](https://www.domjudge.org/demoweb/login) (8.4.0DEV/c01749d0d) as admin, Configuration settings > Authentication > Auth methods: add "ipaddress", Save all changes.
Result: 500 Internal Server Error, An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '8231-274' for key 'testcaseid'
I don't see how that exception could possibly be related to this patch.

vmcj:
> I was playing with the web demo just before, not sure if it's related to the Sentry issue but thought I should mention anyway. Steps: log in to [demoweb](https://www.domjudge.org/demoweb/login) (8.4.0DEV/c01749d0d) as admin, Configuration settings > Authentication > Auth methods: add "ipaddress", Save all changes. Result: 500 Internal Server Error, An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '8231-274' for key 'testcaseid' I don't see how that exception could possibly be related to this patch.

It is not, the dump in demoweb has some issues and we never take the time to create a new (proper) dump.

PR #2636, opened 2024-07-29
Merge webapp/vendor/Makefile into webapp/Makefile

tom93:
Continued from https://github.com/DOMjudge/domjudge/pull/2618#discussion_r1676778142.


vmcj:
I think this is personal preference as I find this harder to follow as I already stated in the other PR. Let's see what the rest thinks.

webapp/Makefile
```diff
@@ -18,6 +20,16 @@ maintainer-conf: .env.local
 	@echo "# the DOMjudge Symfony application in developer mode. Adjust as needed." >> $@
 	@echo "APP_ENV=dev" >> $@
 
+# Dump autoload dependencies (including plugins)
+# This is needed since symfony/runtime is a Composer plugin that runs while dumping
+# the autoload file.
+# We skip it if autoload_runtime.php already exists, to avoid running composer
+# as root during `sudo make install-domserver`.
+composer-dump-autoload: vendor/autoload_runtime.php
```
vmcj, line 31: I'm not sure if we need `composer-dump-autoload` again. The whole reason I started with the move was to get less phony targets so if possible I rather get rid of that one.

tom93: I added it back for symmetry with composer-dump-autoload-dev, like composer-dependencies(-dev). But it's just a naming choice, happy to change it back.

eldering (approved):
Looks OK to me.

PR #2638, opened 2024-07-30
Follow medal in contest setting categories when export results

cubercsl:
When export the result, follow medal categories in contest setting. This will be consistent with the results generated by the award API.


vmcj:
I think this config is only used with your contests,

Can you extend @nickygerritsen his tests with a ranking of your contest (with teams in the same sortorder with some eligible for medals?), otherwise we always get this kind of PRs.

webapp/src/Service/ImportExportService.php
```diff
@@ -522,13 +526,20 @@ public function getResultsData(
 
             $rank      = $teamScore->rank;
             $numPoints = $teamScore->numPoints;
-            if ($rank <= $contest->getGoldMedals()) {
+            $skip      = false;
```
nickygerritsen, line 529: We don't really need this skip variable right? If it is false we just subtract 0

cubercsl: > We don't really need this skip variable right? If it is false we just subtract 0

If the team is not in the medal category, it should be given a rank or honors. So I use this variable below.

nickygerritsen: But you set it to true exactly when `skippedTeams` is not 0. So you can use that check?

cubercsl: `skippedTeams` counts the total number of skipped teams.
`skip` marks whether the team in the current for loop needs to be skipped.

nickygerritsen: Ahhh now I see it

nickygerritsen (approved):

vmcj (changes requested):
Before we merge this I would like a specific test for such a scoreboard.

cubercsl:
These test data is based on `AwardServiceTest`

https://github.com/DOMjudge/domjudge/blob/6a99afc97cecfec980de4d5eb8a398b1779e5472/webapp/tests/Unit/Service/AwardServiceTest.php#L89-L113

webapp/tests/Unit/Fixtures/sample/results-full-honors.tsv
```diff
@@ -0,0 +1,10 @@
+results	1
+10001	1	Gold Medal	4	36	20	Sample Group A
+10002	2	Gold Medal	2	5	3
+10003	2	Gold Medal	2	5	3
+10004	4	Highest Honors	1	12	12	Sample Group B
```
nickygerritsen, line 5: It's not really defined what order this file should be in, nor what should happen when some categories have no medals. It looks a bit weird like this, but I think this is the best we can do,

nickygerritsen (approved):

vmcj (approved):
Tests were added so I'm fine with it.

PR #2639, opened 2024-08-02
Fix importing a problem archive failed alert with "Array Array Array"

### LINKED ISSUE #2633: Importing a problem archive fails when another archive with the same name has already been uploaded

When importing a problem via an archive, if the uploaded zip has a name that another zip with the same name had been uploaded, then it leads to the following weird error:

```
Array
Array
Array
```

### Your environment

DOMJudge version: 8.2.3
OS: Fedora 36 (Server edition)
Platform: Docker
Webserver: Nginx

### Steps to reproduce

- Create two distinct problems, both with the same name (i.e. `A.zip`).
- Import the first problem archive.
- Import the second problem archive.

### Expected behaviour

Both problems get uploaded.

### Actual behaviour

The first one is imported successfully, but the second one fails with the error above.

There is a similar question [here](https://github.com/DOMjudge/domjudge/issues/1708), but it doesn't seem to be this specific.

cubercsl:
This has already been fixed since a6640d5a5f9357cb0c41d2b51811524efe1fb567 (and released in 8.3). Need backport to **8.2** branch.

Closes #2633.

(cherry picked from commit a6640d5a5f9357cb0c41d2b51811524efe1fb567)

### DIFF
--- a/gitlab/ci_settings.sh
+++ b/gitlab/ci_settings.sh
@@ -13,6 +13,8 @@ export GITSHA
 export PS4='(${BASH_SOURCE}:${LINENO}): - [$?] $ '
 export LOGFILE="/opt/domjudge/domserver/webapp/var/log/prod.log"
 
+CCS_SPECS_PINNED_SHA1='6b11623d586500d11ec20b6c12a4908b44ff0e41'
+
 # Shared storage for all artifacts
 export GITLABARTIFACTS="$DIR/gitlabartifacts"
 mkdir -p "$GITLABARTIFACTS"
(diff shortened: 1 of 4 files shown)


vmcj:
Failing test is the integration, that is to be expected.

cubercsl:
It seems that the CI failed to validate json schema of the `contest` object.

https://github.com/DOMjudge/domjudge-scripts/commit/b853de4db3916497c61cdaf399a6568eda7f5384
https://github.com/DOMjudge/domjudge/pull/1963

But in 8.2 we have no `scoreboard_thaw_time`. Should I change to use a specific commit from domjudge-scripts repo in CI?

vmcj (approved):
Thank you for the PR.

PR #2647, opened 2024-08-11
Add page with language stats

nickygerritsen:
As discussed before.

Screenshots:
Screenshot 2024-08-11 at 15 21 40
Screenshot 2024-08-11 at 15 21 46

There is a button above the language graph to link to this page:
image

I made it two columns next to each other first, but expanding the teams will make a lot of teams two lines, so I didn't do that  in the end.

Feedback on what to show or how to show it is welcome.

### DIFF
--- a/webapp/src/Service/StatisticsService.php
+++ b/webapp/src/Service/StatisticsService.php
@@ -5,6 +5,7 @@ namespace App\Service;
 use App\Entity\Contest;
 use App\Entity\ContestProblem;
 use App\Entity\Judging;
+use App\Entity\Language;
 use App\Entity\Problem;
 use App\Entity\Submission;
 use App\Entity\Team;
@@ -58,6 +59,7 @@ class StatisticsService
                 ->join('t.category', 'tc')
                 ->leftJoin('t.affiliation', 'a')
                 ->join('t.submissions', 'ts')
+                ->join('ts.language', 'l')
                 ->join('ts.judgings', 'j')
                 ->andWhere('j.valid = true')
                 ->join('ts.language', 'lang')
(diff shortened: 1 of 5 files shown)


vmcj:
I would also be interested to see for which problems the language was used:
Lang:
1 team
Used for solving problems: [A][C]

nickygerritsen:
Stealing some of @meisterT his awesome CSS skills to produce this:
image

vmcj:
> Stealing some of @meisterT his awesome CSS skills to produce this: image

This will make that we need to fix the WCAG tests for jury (probably) as this has not enough contrast. As this is by design I'm fine with it.

vmcj:
The only thing missing for Astana is an easy export of this so we can directly share this when we start with EOC with whoever needs this? Is the plan to just save this page or should we also expose this via the API?

meisterT:
Keep it simple: Just print it and hand it over.

webapp/src/Service/StatisticsService.php
```diff
@@ -514,6 +517,103 @@ public function getGroupedProblemsStats(
...
+                }
+                $languageStats[$language->getLangid()]['problems_attempted'][$s->getProblem()->getProbId()] = $s->getContestProblem();
+            }
+        }
+
+        foreach ($languageStats as &$languageStat) {
+            usort($languageStat['teams'], static fn(
+                Team $a,
+                Team $b
+            ) => ($a->getLabel() ?: $a->getExternalid()) <=> ($b->getLabel() ?: $b->getExternalid()));
```
meisterT, line 601: Sort by number of problems solved using this language instead?

nickygerritsen: Do you think that makes more sense than by team ID? Won't it look confusing if the teams are in arbitrary order?

Or do you propose to also show the number of solved problems per team per language? What if a team didn't solve but did try a problem in the language?

nickygerritsen:
After talking to @meisterT I improved it to this:
image

(don't mind the double `in`, I fixed that)

meisterT (approved):

PR #2650, opened 2024-08-16
Fix saving new problems with statements

### LINKED ISSUE #2646: Problem with statement cannot be created in the Jury web interface
[error.pdf](https://github.com/user-attachments/files/16527344/error.pdf)

This happens, because the system tries to do the same insert twice into the `problem_statement_content` table.
```
INSERT INTO problem_statement_content (content, probid) VALUES (?, ?)
```

This problem occurs on latest main branch, both locally and on public demo as well as **on 8.3.0 release tag**!!

nickygerritsen:
Fixes #2646.

We used to persist the entity both before and after calling the lifecycle callbacks ourselves, which would introduce weird behavior where Doctrine wanted to insert some rows twice. By not persisting before calling the callbacks ourselves, this issue goes away.

### DIFF
--- a/webapp/src/Controller/BaseController.php
+++ b/webapp/src/Controller/BaseController.php
@@ -561,7 +561,6 @@ abstract class BaseController extends AbstractController
                         return $response;
                     }
                 } else {
-                    $this->em->persist($entity);
                     $this->saveEntity($entity, null, true);
                 }
                 return $this->redirect($urlGenerator());


vmcj (approved):
Do we have tests for this specific usecase? So creating vs updating of an entity?

CI is green so fine by me.

nickygerritsen:
We don't have any tests to create problems through the web interface (the only place the problem happened) yet, I think.

PR #2652, opened 2024-08-19
docs: Update problem-format.rst

qwedc001:
small fix to correct the outdated problem_package_format url

also working on other good first issues.

### DIFF
--- a/doc/manual/problem-format.rst
+++ b/doc/manual/problem-format.rst
@@ -56,4 +56,4 @@ configured problem by uploading a zip file that contains only testcase files.
 Any jury solutions present will be automatically submitted when ``allow_submit``
 is ``1`` and there's a team associated with the uploading user.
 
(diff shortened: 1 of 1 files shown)


qwedc001:
hmm,just found that version 2023-07-draft has added multiple new parameters to some files.
wondering if legacy(icpc) suits domjudge better

eldering:
Thanks for finding the broken link. Indeed, as you also noticed there were a lot of changes to the `2023-07-draft` problem package format this weekend (I was involved in that). The aim is to release what in `2023-07-draft` as a stable specification around the ICPC World Finals in Astana in September, but for now it would be more accurate to link to the legacy ICPC version indeed.

eldering:
See https://github.com/DOMjudge/domjudge/pull/2653

qwedc001:
okay,so shall I just close this one?

PR #2657, opened 2024-08-25
Extend bats test for new debian versions

vmcj:
Debian testing started using different quotes in the generated configure scripts.

Kinda annoying but changing everything to a regex or only checking for parts of the string is also ugly.

### DIFF
--- a/.github/jobs/configure-checks/all.bats
+++ b/.github/jobs/configure-checks/all.bats
@@ -67,9 +67,9 @@ repo-remove () {
     assert_line "checking for gcc... no"
     assert_line "checking for cc... no"
     assert_line "checking for cl.exe... no"
-    assert_line "configure: error: in \`${test_path}':"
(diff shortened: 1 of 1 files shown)


.github/jobs/configure-checks/all.bats
```diff
@@ -67,9 +67,9 @@ repo-remove () {
     assert_line "checking for gcc... no"
     assert_line "checking for cc... no"
     assert_line "checking for cl.exe... no"
-    assert_line "configure: error: in \`${test_path}':"
+    assert_regex "configure: error: in [\`']${test_path}':"
```
meisterT, line 70: if you want, you can replace the `[\`']` by `.` - this would allow also other characters, but I don't think we care and it is more readable

meisterT: As github screws up my formatting above, this is what I meant: 

```suggestion
    assert_regex "configure: error: in .${test_path}':"
```

PR #2664, opened 2024-08-27
Avoid infinite error handling recursion in runguard.

meisterT:
Previously, there were two scenarios that are problematic:
- while handling an error, we start writing the meta file, and if that also fails, we treat this as handling a new error and never stop;
- while writing the meta file, we encounter an error which in turn tries writing to the meta file causing another error and we never stop.


nickygerritsen (approved):

eldering (approved):
LGTM

PR #2665, opened 2024-08-27
Make PHPstan happy.

meisterT:


meisterT:
Well, turns out it was not happy after all. Anyone has an idea what it is complaining about?

PR #2667, opened 2024-08-30
Allow null `display_name` when importing `teams.json`, fixes #2666

### LINKED ISSUE #2666: Team `name` is not used when `display_name` is empty after importing a JSON

The description of the field *Display name* of a team says "*If provided, will display this instead of the team name in certain places, like the scoreboard.*" However, after creating or modifying a team by importing a `team.json` with no `display_name` key (or assigned to `null`), the scoreboard shows an empty name cell instead of the non-empty team name.

### Your environment
- DOMjudge version: 8.3.0 and current main
- Operating system: does not matter
- Webserver: does not matter

### Steps to reproduce
1. Import a `teams.json` file in *Teams & group* section of the *Import / export* page. For example,
```json
[{"id": "A01", "group_ids": ["A"], "name": "A01", "organization_id": "org"}]
```
2. Add this user to a contest.
3. Navigate to the scoreboard.

### Expected behaviour

`A01` should appear on the left of the row for this new user.

### Actual behaviour

An empty name cell appears instead.

The problem is that the absence of value (or `null` value) is replaced by an empty string when importing, so the `""` `display_name` is shown instead of the team name. I will submit a pull request that fixes it.

ningit:
In the method `ImportExportService::importTeamsJson`, use `null` instead of the empty string as default `display_name`, so that it does not override the team name as explained in #2666.

I have tested it and it actually solves #2666.

### DIFF
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -1012,7 +1012,7 @@ class ImportExportService
                     'label' => $team['label'] ?? null,
                     'categoryid' => $team['group_ids'][0] ?? null,
                     'name' => $team['name'] ?? '',
-                    'display_name' => $team['display_name'] ?? '',
+                    'display_name' => $team['display_name'] ?? null,
                     'publicdescription' => $team['public_description'] ?? $team['members'] ?? '',
                     'location' => $team['location']['description'] ?? null,
                 ],


nickygerritsen:
This was broken in e1f78cadc9f4c606942488d9b477831f626efb9d which is a revert of 9974cec5e96e557450f47f6ac5771f6bcc6733a9.

I think your code makes sense, to do the null thing for display name but not for name.

@vmcj what do you think?

nickygerritsen (approved):
This seems like the proper fix

ningit:
Thank you for the quick reply and merge.

PR #2668, opened 2024-08-31
preg_match_all always returns an array of arrays

vmcj:
Even if we cannot match anything we will have an array containing an empty array. That array indicates that nothing was matched.

Hopefully this fixes what https://github.com/DOMjudge/domjudge/pull/2665 tried to do.

### DIFF
--- a/webapp/src/Twig/TwigExtension.php
+++ b/webapp/src/Twig/TwigExtension.php
@@ -1023,8 +1023,8 @@ EOF;
         if (is_null($col)) {
             return $text;
         }
-        preg_match_all("/[0-9A-Fa-f]{2}/", $col, $m);
-        if (!count($m)) {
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

meisterT (commented):
After reading https://www.php.net/manual/en/function.preg-match-all.php#refsect1-function.preg-match-all-returnvalues my suggestion is to capture the return value and check it as well.

vmcj:
> After reading https://www.php.net/manual/en/function.preg-match-all.php#refsect1-function.preg-match-all-returnvalues my suggestion is to capture the return value and check it as well.

from the commit message:
```
We should now fail when:
- preg_match_all fails
- preg_match_all returns it found 0 matches
- the redundant case (as safeguard) when the array has 0 matches
```

meisterT (approved):

PR #2672, opened 2024-09-07
Simplify submission page and re-order information in a better way.

meisterT:
Before:
image

After:
image

Another after example:
image


eldering (approved):
Github is not great at showing moved code in a great way, but I trust you just moved things around. LGTM as far as I can see.

nickygerritsen:
Same. It looks good so probably fine

nickygerritsen (approved):

PR #2673, opened 2024-09-10
Add answers to questions from Luxor to manual/clars

vmcj:
I'll squash after approval.

No reason to merge it now, but I found it in the GDrive folder.

### DIFF
--- a/doc/manual/team.rst
+++ b/doc/manual/team.rst
@@ -158,7 +158,7 @@ which problems you attempted and solved. Via the menu you can view the public
 scoreboard page with the scores of all teams. Many cells will show
 additional "title text" information when hovering over them. The
 score column lists the number of solved problems and the total time including
-penalty time. Each cell in a problem column lists the number of submissions,
+penalty time, the first team to solve the problem will be in dark green. Each cell in a problem column lists the number of submissions,
 and if the problem was solved, the time of the first correct
 submission in minutes since contest start. This is included in your
 total time together with any penalty time incurred for previous
--- a/doc/manual/team.rst
+++ b/doc/manual/team.rst
@@ -418,6 +418,8 @@ for different programming languages.
 Improvements to DOMjudge
 ------------------------
 
-The DOMjudge team would like your feedback. We do not receive much feedback from participants.
-If you find something lacking or have improvement ideas, please report them. See https://www.domjudge.org/development.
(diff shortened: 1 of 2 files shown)


etc/db-config.yaml
```diff
@@ -186,6 +186,12 @@
             default_value:
                 - No comment.
                 - Read the problem statement carefully.
+                - The hardware of the machines evaluating your submissions is identical to the team machines.
+                  They run the same operating system and base image but are tuned to make submission run times
+                  more consistent, for example we pin your submission to one core.
+                - DOMjudge does not define any special macros like ONLINE_JUDGE.
+                - For interactive problems, only your code runs inside of the sandbox and other processes
+                  including the interactive problem interactor do not count towards memory or time limits for the submission.
```
eldering, line 194: I don't really understand what these do here in the default clarification answers (I think they should have been general clarifications sent out at the beginning of WFs), but they should not go into `main`.

doc/manual/team.rst
```diff
@@ -261,6 +261,8 @@ your program).
 On the contest control system your program enters a queue, awaiting compilation,
 execution and testing on one of the autojudges.
 
+Beware that if you submit multiple times in a short time interval the DOMjudge will determine which submission to judge based on both the submission time and how many other submission you have in queue to penalize spamming.
```
eldering, line 264:
```suggestion
Beware that if you submit multiple times in a short time interval, then DOMjudge will determine which submission to judge based on both the submission time and how many other submission you have in queue to penalize spamming.
```

eldering: Also, I'm not really sure this should go into the manual in `main`.

doc/manual/team.rst
```diff
@@ -410,6 +418,8 @@ for different programming languages.
 Improvements to DOMjudge
 ------------------------
 
-The DOMjudge team would like your feedback. We do not receive much feedback from participants.
-If you find something lacking or have improvement ideas, please report them. See https://www.domjudge.org/development.
+The DOMjudge team would like your feedback. We do not
+receive much feedback from participants. If you find
+something lacking or have improvement ideas, please report
+them. See https://www.domjudge.org/development.
```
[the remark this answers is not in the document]
vmcj: Only reformatting.

PR #2691, opened 2024-09-17
Preserve information whether attachments are executable.

meisterT:
This is relevant for testing tools which are often uploaded as attachments for interactive problems. When we download the samples (via the web interface or API), we should mark these files as executable iff they were executable when uploading.


nickygerritsen (commented):
The migration is missing.

meisterT:
Thanks, added.

PR #2695, opened 2024-09-18
Check that last seen event ID exists while emitting the event feed

eldering:
This will throw an error, for example, when the database gets reloaded while there is a running process serving the event feed to a connected client. We encountered this problem at WF Astana.

### DIFF
--- a/webapp/src/Controller/API/ContestController.php
+++ b/webapp/src/Controller/API/ContestController.php
@@ -39,6 +39,7 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

Kevinjil (approved):

webapp/src/Controller/API/ContestController.php
```diff
@@ -721,19 +723,31 @@ public function getEventFeedAction(
...
 
                 /** @var Event[] $events */
                 $events = $q->getResult();
 
+                if ($lastIdExists) {
+                    if (count($events) == 0 || $events[0]->getEventid() !== $lastIdSent) {
+                        throw new HttpException(500, sprintf('Cannot find event %d in database anymore', $lastIdSent));
+                    }
+                    // Remove the previously last sent event. We just fetched
+                    // it to make sure it's there.
```
meisterT, line 747:
```suggestion
                    // it to make sure it's still there.
```

meisterT (approved):

Issue #2689, opened 2024-09-17
Consider replacing ace editor with monaco


meisterT:
Monaco (https://microsoft.github.io/monaco-editor/) is more feature rich, in particular supports nice diff views (side by side etc).

We started looking into this because the current way we handle diff views for edit/resubmit (via https://github.com/sebastianbergmann/diff) does not support line ending changes very well.

This will then also make it easier to allow embedding an editor in the team interface.

PR #2696, opened 2024-09-18
Add two features: the ability to view the first failed test case and to copy and paste code directly.

as6325400:
## Motivation:

1. I have served as a teaching assistant for several courses, such as Data Structures and Programming, where DOMjudge was used in class. For beginners, it's often the case that a simple mistake or a small oversight of edge cases can lead to many "wrong answer" results. I believe this can be quite frustrating for those just learning to program. On platforms like Codeforces, you can see which test case your solution failed on, and on LeetCode, when a solution is not accepted (non-AC), the incorrect test case is often provided. Therefore, I believe these two features would be very useful for educational purposes.

2. As for the process of uploading code as files, I often find it a bit inconvenient. In online judges like Codeforces and Luogu, there are features allowing users to paste code directly. I believe this is a very convenient feature to have.

## Implementation:

These features can all be toggled on or off in the configuration settings.

Most of them are adapted from the jury interface.

1. Teams can view the status of each test case from their interface.
截圖 2024-09-19 凌晨12 19 31

2. For non-correct and non-compile error submissions, the first failed test case will be displayed.截圖 2024-09-19 凌晨12 18 26

3. A navigation tab is provided for users to choose between uploading a file or pasting code.
<p align="center">
  
  
</p>

4. Configuration setting
截圖 2024-09-19 凌晨12 28 07


meisterT:
General comment: this PR contains a lot of unrelated code changes including passwords that you probably should change. 

As for the additional features: I don't think we should incorporate sharing results with contestants in main and it's probably ok to keep this functionality in your fork. Also note that with parallel judging you are exposing more to contestants than you likely want to expose. 

The source code editor behind a toggle sounds like a reasonable feature. We might switch at a later point from Ace editor to Monaco (https://github.com/DOMjudge/domjudge/issues/2689) but this shouldn't make it more complex. Can you please try to extract your changes to add this functionality in a separate PR?

as6325400:
> General comment: this PR contains a lot of unrelated code changes including passwords that you probably should change.
> 
> As for the additional features: I don't think we should incorporate sharing results with contestants in main and it's probably ok to keep this functionality in your fork. Also note that with parallel judging you are exposing more to contestants than you likely want to expose.
> 
> The source code editor behind a toggle sounds like a reasonable feature. We might switch at a later point from Ace editor to Monaco (#2689) but this shouldn't make it more complex. Can you please try to extract your changes to add this functionality in a separate PR?

Alright, I will submit a separate PR for the source code editor later. Thank you!

PR #2697, opened 2024-09-19
Draft: Split configure tests

eldering:
This splits the tests in the ones that make changes to system packages and users and those that don't. The former are a bit messy as they don't return to a clean, original state for each tests and that will introduce conflicts with further changes.


vmcj:
@eldering shall I work on this or are you planning to finish this one?

PR #2707, opened 2024-09-21
Fix error handling / messaging on import validation errors.

### LINKED ISSUE #2705: Import of organizations.json broken
If you have an organizations.json with empty names (e.g. https://github.com/WISVCH/icpc-logos/releases/download/latest/organizations.json), DOMjudge seems to import no organization at all and also does not show any error (neither on the command line nor the web UI).

meisterT:
Fixes #2705.

While we are at it, also improve the usefulness of the error messages.

Example:

image

Command line:
```
webapp/bin/console api:call -m POST -f json=../Downloads/organizations.json users/organizations
04:17:58 ERROR     [request] Uncaught PHP Exception Symfony\Component\HttpKernel\Exception\BadRequestHttpException: "Error while adding organizations: Organization at index 36 ({"externalid":"U-3397","shortname":"","name":"Chalmers University of Technology","country":"SWE","icpc_id":null}) has errors:
  • `shortname`: This value should not be blank.

Organization at index 42 ({"externalid":"U-3758","shortname":"","name":"King's College London","country":"GBR","icpc_id":null}) has errors:
  • `shortname`: This value should not be blank.

Organization at index 67 ({"externalid":"U-8102","shortname":"","name":"University of Passau","country":"DEU","icpc_id":null}) has errors:
  • `shortname`: This value should not be blank.

" at UserController.php line 131 ["exception" => Symfony\Component\HttpKernel\Exception\BadRequestHttpException^ { …}]
```

### DIFF
--- a/webapp/src/Service/ImportExportService.php
+++ b/webapp/src/Service/ImportExportService.php
@@ -268,7 +268,7 @@ class ImportExportService
             $messages = [];
             /** @var ConstraintViolationInterface $error */
             foreach ($errors as $error) {
-                $messages[] = sprintf('%s: %s', $error->getPropertyPath(), $error->getMessage());
+                $messages[] = sprintf('  • `%s`: %s', $error->getPropertyPath(), $error->getMessage());
             }
 
             $errorMessage = sprintf("Contest has errors:\n\n%s", implode("\n", $messages));
@@ -547,7 +547,7 @@ class ImportExportService
                 );
                 continue;
             }
-        
+
             if (!$skip && $rank - $skippedTeams <= $contest->getGoldMedals()) {
                 $awardString = 'Gold Medal';
                 $lowestMedalPoints = $teamScore->numPoints;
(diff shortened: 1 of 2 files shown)


nickygerritsen (approved):

eldering (commented):
Just a comment that the spec does not have `shortname`, only `name` and `formal_name`, see https://ccs-specs.icpc.io/draft/contest_api#organizations, so the error message is a bit confusing since it refers to internal attribute names, not the ones from the spec.

meisterT:
> Just a comment that the spec does not have `shortname`, only `name` and `formal_name`, see https://ccs-specs.icpc.io/draft/contest_api#organizations, so the error message is a bit confusing since it refers to internal attribute names, not the ones from the spec.

Yep, since error handling is a little bit opaque it is kinda non-trivial to change though. I will merge this as it is clearly better than a silent failure and then we can discuss with Nicky when he is back how to best improve it.

PR #2712, opened 2024-09-21
add Paste code to submit feature

as6325400:
Sorry, I accidentally closed the PR due to some GitHub operation error. I have resubmitted it. Apologies for the inconvenience.

## Modify the old PR
[Add two features: the ability to view the first failed test case and to copy and paste code directly.](https://github.com/DOMjudge/domjudge/pull/2696)

## Motivation:
As for the process of uploading code as files, I often find it a bit inconvenient. In online judges like Codeforces and Luogu, there are features allowing users to paste code directly. I believe this is a very convenient feature to have.

## Implementation:
1. A navigation tab is provided for users to choose between uploading a file or pasting code.
<p align="center">
  
  
</p>
2. Configuration setting

截圖 2024-09-22 凌晨2 57 34


nickygerritsen:
For python the entry point is going to be weird, since it needs to be the filename, which is dynamically generated. We should come up with something smart here

as6325400:
> For python the entry point is going to be weird, since it needs to be the filename, which is dynamically generated. We should come up with something smart here

I have tested Python, and there is no issue. I believe the Paste code feature can only be used with a single file, so when it's just one file, the Python entry point shouldn't be a problem, right?

nickygerritsen:
> > For python the entry point is going to be weird, since it needs to be the filename, which is dynamically generated. We should come up with something smart here
> 
> 
> 
> I have tested Python, and there is no issue. I believe the Paste code feature can only be used with a single file, so when it's just one file, the Python entry point shouldn't be a problem, right?

I think this depends on if you enable "require entry point" for python. If you do it might break

as6325400:
> > > For python the entry point is going to be weird, since it needs to be the filename, which is dynamically generated. We should come up with something smart here
> > 
> > 
> > I have tested Python, and there is no issue. I believe the Paste code feature can only be used with a single file, so when it's just one file, the Python entry point shouldn't be a problem, right?
> 
> I think this depends on if you enable "require entry point" for python. If you do it might break

Alright, I'm not familiar with Python, so I didn't consider that.

Perhaps we could set up an option that allows the administrator to decide whether to enable the paste code feature. Do you think this would be a good approach?

nickygerritsen:
I was thinking maybe to either:
- ask the user how to name the file
- do something consistent with the filename and special case python

as6325400:
> I was thinking maybe to either:
> - ask the user how to name the file
> - do something consistent with the filename and special case python 

I don't quite understand in what situations Python would have issues. Could you provide me with an example problem using Python?

Thanks.

nickygerritsen:
> > I was thinking maybe to either:
> > - ask the user how to name the file
> > - do something consistent with the filename and special case python 
> 
> I don't quite understand in what situations Python would have issues. Could you provide me with an example problem using Python?
> 
> Thanks.

Can you try to enable "require entry point" for the python language in the jury and try to submit? Let's see what happens. Maybe it's fine

as6325400:
> "require entry point"

If enabling the "require entry point" in Python causes issues, would it be feasible to design the system so that, when using Python and the paste_code feature, the code for such cases is specifically saved with a designated filename? Thank you.

nickygerritsen:
> > "require entry point"
> 
> 
> 
> If enabling the "require entry point" in Python causes issues, would it be feasible to design the system so that, when using Python and the paste_code feature, the code for such cases is specifically saved with a designated filename? Thank you.

That was my option 2. If you can make it nicely I think that's ok

as6325400:
> > > "require entry point"
> > 
> > 
> > If enabling the "require entry point" in Python causes issues, would it be feasible to design the system so that, when using Python and the paste_code feature, the code for such cases is specifically saved with a designated filename? Thank you.
> 
> That was my option 2. If you can make it nicely I think that's ok

I have modified my code to force the entry point to equal the filename when using paste_code, since there's only one file in that case. I've tested it, and no issues occurred. Thanks!

meisterT:
You probably want to change the merge target to the `main` branch.

as6325400:
> You probably want to change the merge target to the `main` branch.

Thank you for pointing that out! I've already changed the merge target to the main branch.

vmcj:
I think you want to do a rebase against `main` first, you're now adding commits which should stay `8.3.1` specific.

as6325400:
> I think you want to do a rebase against `main` first, you're now adding commits which should stay `8.3.1` specific.

I have made the changes and submitted a new [PR](https://github.com/DOMjudge/domjudge/pull/2720)

PR #2714, opened 2024-09-22
download file when submission is single source

as6325400:
When there is only one uploaded file, download the file directly instead of downloading a compressed file.

### DIFF
--- a/webapp/src/Controller/Team/SubmissionController.php
+++ b/webapp/src/Controller/Team/SubmissionController.php
@@ -242,6 +242,10 @@ class SubmissionController extends BaseController
                 $submitId));
         }
 
+        if ($this->submissionService->getSubmissionFileCount($submission) === 1) {
+            return $this->submissionService->getSubmissionFileResponse($submission);
+        }
+
         return $this->submissionService->getSubmissionZipResponse($submission);
     }
 }
--- a/webapp/src/Service/SubmissionService.php
+++ b/webapp/src/Service/SubmissionService.php
@@ -768,4 +768,31 @@ class SubmissionService
 
         return Utils::streamZipFile($tmpfname, 's' . $submission->getSubmitid() . '.zip');
     }
+
+    public function getSubmissionFileResponse(Submission $submission): StreamedResponse
+    {
+        /** @var SubmissionFile[] $files */
+        $files = $submission->getFiles();
+        
+        if (count($files) !== 1) {
+            throw new ServiceUnavailableHttpException(null, 'Submission does not contain exactly one file.');
+        }
+
+        $file = $files[0];
+        $filename = $file->getFilename();
+        $sourceCode = $file->getSourcecode();
+
+        return new StreamedResponse(function () use ($sourceCode) {
+            echo $sourceCode;
+        }, 200, [
+            'Content-Type'        => 'text/plain',
+            'Content-Disposition' => 'attachment; filename="' . $filename . '"',
+            'Content-Length'      => strlen($sourceCode),
+        ]);
+    }
+
+    public function getSubmissionFileCount(Submission $submission): int
+    {
+        return count($submission->getFiles());
+    }
 }


webapp/src/Service/SubmissionService.php
```diff
@@ -768,4 +768,33 @@ public function getSubmissionZipResponse(Submission $submission): StreamedRespon
...
+        }
+
+        $file = $files[0];
+        $filename = $file->getFilename();
+        $sourceCode = $file->getSourcecode();
+
+        return new StreamedResponse(function () use ($sourceCode) {
+            echo $sourceCode;
+        }, 200, [
+            'Content-Type'        => 'application/octet-stream',
```
meisterT, line 788: I wonder whether this should be `text/plain` or something different?

webapp/src/Service/SubmissionService.php
```diff
@@ -768,4 +768,33 @@ public function getSubmissionZipResponse(Submission $submission): StreamedRespon
...
+            'Content-Disposition' => 'attachment; filename="' . $filename . '"',
+            'Content-Length'      => strlen($sourceCode),
+        ]);
+    }
+
+
+    public function getSubmissionFileNums(Submission $submission): int
```
meisterT, line 795: nit: I would prefer the name `getSubmissionFileCount` (and I assume we can mark it as private)

webapp/src/Controller/Team/SubmissionController.php
```diff
@@ -236,6 +236,10 @@ public function downloadAction(int $submitId): Response
                 $submitId));
         }
 
+        if($this->submissionService->getSubmissionFileNums($submission) == 1){
```
meisterT, line 239: nit: add spaces before/after the parentheses

webapp/src/Service/SubmissionService.php
as6325400: Ok, Should I modify it once more, or will you handle the changes when merging?

webapp/src/Service/SubmissionService.php
as6325400: application/octet-stream is more commonly used when downloading files, but I just tested it, and the effect here is the same as text/plain.

webapp/src/Service/SubmissionService.php
```diff
@@ -768,4 +768,33 @@ public function getSubmissionZipResponse(Submission $submission): StreamedRespon
...
+
+        return new StreamedResponse(function () use ($sourceCode) {
+            echo $sourceCode;
+        }, 200, [
+            'Content-Type'        => 'text/plain',
+            'Content-Disposition' => 'attachment; filename="' . $filename . '"',
+            'Content-Length'      => strlen($sourceCode),
+        ]);
+    }
+
```
vmcj, line 793:
```suggestion
```
small nitpick

webapp/src/Service/SubmissionService.php
```diff
@@ -768,4 +768,33 @@ public function getSubmissionZipResponse(Submission $submission): StreamedRespon
...
+
+    public function getSubmissionFileCount(Submission $submission): int
+    {
+        $files = $submission->getFiles();
+        return count($files);
```
vmcj, line 798:
```suggestion
        return count($submission->getFiles());
```
I think this is still readable and saves a bit of space on our screen.

webapp/src/Controller/Team/SubmissionController.php
```diff
@@ -236,6 +236,10 @@ public function downloadAction(int $submitId): Response
                 $submitId));
         }
 
+        if ($this->submissionService->getSubmissionFileCount($submission) == 1) {
```
vmcj, line 239:
```suggestion
        if ($this->submissionService->getSubmissionFileCount($submission) === 1) {
```

If we can be very exact I prefer that. It helps steering PHPStan to better results.

vmcj (commented):
This will remove the `x` bit from such files, whether that is fine I leave up to the rest.

meisterT (approved):

PR #2717, opened 2024-09-22
Count in-contest submissions on submission list

eldering:
This is to make comparing with a shadow CCS easier, as we don't want to include jury submissions here.

(cherry picked from commit 96af89e2ab11ac143373b71fb8975a4f5bf1723b)

### DIFF
--- a/webapp/src/Service/SubmissionService.php
+++ b/webapp/src/Service/SubmissionService.php
@@ -282,6 +282,14 @@ class SubmissionService
             ->andWhere($countQueryExtras['queued'])
             ->getQuery()
             ->getSingleScalarResult();
+        $counts['inContest'] = (clone $queryBuilder)
+            ->select('COUNT(s.submitid)')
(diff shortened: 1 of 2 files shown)


webapp/src/Service/SubmissionService.php
```diff
@@ -282,6 +282,12 @@ public function getSubmissionList(
             ->andWhere($countQueryExtras['queued'])
             ->getQuery()
             ->getSingleScalarResult();
+        $counts['inContest'] = (clone $queryBuilder)
+            ->select('COUNT(s.submitid)')
+            ->join('s.contest', 'c')
+            ->andWhere('s.submittime BETWEEN c.starttime AND c.endtime')
```
meisterT, line 288: I think we also want to filter out submissions from teams in invisible categories?

meisterT (approved):

meisterT:
Please squash the commits into one, then we can merge it.

PR #2719, opened 2024-09-22
Ignore start status event

eldering:
This is a CDS specific extension to the contest API and irrelevant for us.

### DIFF
--- a/webapp/src/DataTransferObject/Shadowing/EventType.php
+++ b/webapp/src/DataTransferObject/Shadowing/EventType.php
@@ -18,6 +18,7 @@ enum EventType: string
     case PROBLEMS = 'problems';
     case RUNS = 'runs';
     case STATE = 'state';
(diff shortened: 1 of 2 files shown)


webapp/src/DataTransferObject/Shadowing/EventType.php
```diff
@@ -28,7 +29,12 @@ public static function fromString(string $value): EventType
             return EventType::CONTESTS;
         }
 
-        return EventType::from($value);
+	// When encountering an error for an unknown event
+	// Consider if we need to verify this as shadow or
+	// ignore it by adding the case in:
+	// webapp/src/DataTransferObject/Shadowing/EventType.php
+	// webapp/src/Service/ExternalContestSourceService.php
+	return EventType::from($value);
```
Kevinjil, line 37: Indentation seems off?

Kevinjil (changes requested):

PR #2720, opened 2024-09-23
add Paste code to submit feature

as6325400:
### Modify PR merge to **main**

[add Paste code to submit feature](https://github.com/DOMjudge/domjudge/pull/2712)


vmcj:
Hi, 

I think the rebasing still didn't work correctly, for example https://github.com/DOMjudge/domjudge/pull/2720/commits/92e3ad705cef4b6e4db2033751384327437e4a9a should stay specific for the 8.3 branch.

Can you run:
```
git remote add upstream https://github.com/DOMjudge/domjudge
git fetch upstream
git rebase -i upstream/main
```

And only use your own commits there?

as6325400:
> Hi,
> 
> I think the rebasing still didn't work correctly, for example [92e3ad7](https://github.com/DOMjudge/domjudge/commit/92e3ad705cef4b6e4db2033751384327437e4a9a) should stay specific for the 8.3 branch.
> 
> Can you run:
> 
> ```
> git remote add upstream https://github.com/DOMjudge/domjudge
> git fetch upstream
> git rebase -i upstream/main
> ```
> 
> And only use your own commits there? Alternatively you enabled the option that we can do this, if you want I can try it later this week if you're fine with it.

I think I've got it right now. Could you please take a look?

Thanks!

webapp/templates/team/submit_modal.html.twig
```diff
@@ -80,4 +123,31 @@
             filesSelected.removeClass('d-none');
         });
     </script>
+    <style>
+        .container {
```
vmcj, line 127: Can you add this to the `.ccs` files and not directly in the page?

webapp/src/Controller/Team/SubmissionController.php
```diff
@@ -60,34 +61,111 @@ public function createAction(Request $request, ?Problem $problem = null): Respon
...
+                        'success',
+                        'Submission done! Watch for the verdict in the list below.'
+                    );
+                } else {
+                    $this->addFlash('danger', $message);
+                }
+                return $this->redirectToRoute('team_index');
+            }
+        } elseif ($formPaste->isSubmitted() && $formPaste->isValid()) {
+            if ($contest === null) {
```
vmcj, line 119: A lot of those now get duplicated so maybe merge those where possible? Either by a helper function or using something like:
`if ($formUpload->isSubmitted() || $formPaste->isSubmitted())` and do the `if/elseif` one branch deeper.

etc/db-config.yaml
```diff
@@ -201,6 +201,14 @@
 -   category: Display
     description: Options related to the DOMjudge user interface.
     items:
+        -   name: default_submission_code_mode
```
vmcj, line 204: I wonder if this isn't more a preference per user so that having a cookie remembering the last used method is more convenient for the team.

webapp/templates/team/submit_modal.html.twig
as6325400: Ok

webapp/src/Form/Type/SubmitProblemPasteType.php
```diff
@@ -0,0 +1,99 @@
...
+        ]);
+        $builder->addEventListener(FormEvents::PRE_SET_DATA, function (FormEvent $event) use ($problemConfig) {
+            $data = $event->getData();
+            if (isset($data['problem'])) {
+                $problemConfig += ['row_attr' => ['class' => 'd-none']];
+                $event->getForm()->add('problem', EntityType::class, $problemConfig);
+            }
+        });
+    }
+}
```
vmcj, line 81: Missing newline.

webapp/templates/team/submit_modal.html.twig
```diff
@@ -19,21 +19,64 @@
                     {% include 'partials/alert.html.twig' with {'type': 'danger', 'message': 'Submissions (temporarily) disabled.'} %}
                 </div>
             {% else %}
-                {{ form_start(form) }}
+                {% set active_tab = defaultSubmissionCodeMode == 0 ? 'paste' : 'upload' %}
+
+                <!-- Bootstrap Nav Tabs for Switching -->
```
vmcj, line 24: I don't think this comment is needed.

vmcj (commented):
Rebase seemed to have worked.

webapp/templates/base.html.twig
```diff
@@ -14,6 +14,8 @@
     <script src="{{ asset("js/bootstrap.bundle.min.js") }}"></script>
 
     <script src="{{ asset("js/domjudge.js") }}"></script>
+    <script src="{{ asset('js/ace/ace.js') }}"></script>
```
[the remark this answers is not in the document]
as6325400: 截圖 2024-09-27 下午3 14 32
I think it's possible. Currently, besides the newly added paste_code, there are four other files using js/ace/ace.js. Have you considered including all of them in base.html.twig? If so, should I handle it, or should I wait for it to be merged and then have your team handle it?

webapp/src/Form/Type/SubmitProblemPasteType.php
as6325400: ok

webapp/templates/team/submit_modal.html.twig
as6325400: ok

etc/db-config.yaml
meisterT: I prefer the configuration option. Making this user controlled could still happen secondary.

as6325400: I think we can start by using a configuration option initially, and then use cookies to track once the user starts using it. What do you think?

webapp/src/Controller/Team/SubmissionController.php
as6325400: I think I've made it much clearer. Is this okay now?

etc/db-config.yaml
as6325400: I have already modified it according to the method I mentioned above.

as6325400:
I have mostly finished the modifications as mentioned above. Is there anything else that needs to be changed? Thank you.

webapp/public/style_domjudge.css
```diff
@@ -699,3 +699,25 @@ blockquote {
...
+    margin-top: 10px;
+    margin-bottom: 10px;
+    background-color: #fafafa;
+    max-height: 400px;
+    overflow: auto;
+}
+
+.editor-container:hover {
+    box-shadow: 0 4px 10px rgba(0, 0, 0, 0.2);
+}
```
vmcj, line 723: The `EOF` is missing.

webapp/templates/team/submit_modal.html.twig
```diff
@@ -19,21 +19,61 @@
...
+                            <div class="modal-footer">
+                                <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Cancel</button>
+                                <button type="submit" class="btn btn-success">
+                                    <i class="fas fa-cloud-upload-alt"></i> Submit Upload
+                                </button>
+                            </div>
+                            {{ form_end(formupload) }}
+                        </div>
+
+                        <!-- Paste Code Tab -->
```
vmcj, line 49:
```suggestion
```

webapp/templates/team/submit_modal.html.twig
```diff
@@ -19,21 +19,61 @@
...
+                        <div class="tab-pane fade {% if active_tab == 'paste' %}show active{% endif %}" id="paste" role="tabpanel" aria-labelledby="paste-tab">
+                            {{ form_start(formpaste) }}
+                            {{ form_widget(formpaste.code_content) }}
+                            <label for="codeInput">Paste your code here:</label>
+                            <div class="editor-container">
+                                {{ "" | codeEditor(
+                                    "_team_submission_code",
+                                    "c_cpp",
```
vmcj, line 57: What do you think of moving the language selection above the editor? This way you can also use it here to fix syntax highlighting in the editor (as idea).

vmcj (approved):
Looks already greag, some small remarks.

I'll wait for someone else to also approve (as is the custom) but this looks good already.

webapp/templates/team/submit_modal.html.twig
as6325400: I think it's a good idea. When I first implemented it, I wasn't sure how to dynamically adjust the syntax highlighting in the editor by changing the language in the menu. Is that achievable?

webapp/public/style_domjudge.css
as6325400: I've added the EOF. Thanks for pointing that out!

webapp/templates/team/submit_modal.html.twig
vmcj: Ah, that depends on the twig template.

Let's leave that for a follow up PR in that case.

webapp/templates/team/submit_modal.html.twig
```diff
@@ -19,21 +19,60 @@
...
+                    <ul class="nav nav-tabs container text-center" id="submissionTabs" role="tablist" style="width: 100%">
+                        <li class="nav-item" role="presentation" >
+                            <a class="nav-link {% if active_tab == 'upload' %}active{% endif %}" id="upload-tab" data-bs-toggle="tab" href="#upload" role="tab" aria-controls="upload" aria-selected="{% if active_tab == 'upload' %}true{% else %}false{% endif %}" onclick="setCookie('active_tab', 'upload')" >Upload File</a>
+                        </li>
+                        <li class="nav-item text-center" role="presentation">
+                            <a class="nav-link {% if active_tab == 'paste' %}active{% endif %}" id="paste-tab" data-bs-toggle="tab" href="#paste" role="tab" aria-controls="paste" aria-selected="{% if active_tab == 'paste' %}true{% else %}false{% endif %}" onclick="setCookie('active_tab', 'paste')">Paste Code</a>
+                        </li>
+                    </ul>
+                    <div class="tab-content" id="submissionTabsContent" style="margin-top: 20px;">
+                        <!-- File Upload Tab -->
```
vmcj, line 32:
```suggestion
```

as6325400:
I found the issue and successfully passed the unit tests.

vmcj:
Can you already squash?

as6325400:
> Can you already squash?

ok 

> Can you already squash?

Ok, I already squashed it

webapp/src/Controller/Team/SubmissionController.php
```diff
@@ -60,47 +61,111 @@ public function createAction(Request $request, ?Problem $problem = null): Respon
...
+                    $uploadedFile = new UploadedFile(
+                        $tempFilePath,
+                        $tempFileName,
+                        'application/octet-stream',
+                        null,
+                        true
+                    );
+        
+                    $files = [$uploadedFile];
+                    $entryPoint = $tempFileName;
```
meisterT, line 127: The filename as entry point does not work for Kotlin, nor for Java.

See https://github.com/DOMjudge/domjudge/blob/main/webapp/templates/team/partials/submit_scripts.html.twig#L3 or https://github.com/DOMjudge/domjudge/blob/main/submit/submit#L206

webapp/src/Controller/Team/SubmissionController.php
```diff
@@ -60,47 +61,111 @@ public function createAction(Request $request, ?Problem $problem = null): Respon
...
+                        date('Y-m-d_H-i-s'),
+                        $language->getExtensions()[0]
+                    );
+                    $tempFileName = preg_replace('/[^a-zA-Z0-9_.-]/', '_', $tempFileName);
+                    $tempFilePath = $tempDir . DIRECTORY_SEPARATOR . $tempFileName;
+                    file_put_contents($tempFilePath, $codeContent);
+        
+                    $uploadedFile = new UploadedFile(
+                        $tempFilePath,
+                        $tempFileName,
```
meisterT, line 120: I think we should use a simpler name here (and then of course use that as base for the entry point calculation). A nice simple naming scheme could be the `shortName` of the `contestProblem`.

etc/db-config.yaml
```diff
@@ -201,6 +201,14 @@
...
+            type: int
```
meisterT, line 205: Let's make this an `array_val`, then one can select which ones are enabled. By default let's enable upload for now.

webapp/templates/team/submit_modal.html.twig
```diff
@@ -19,21 +19,59 @@
...
+                                    formpaste.language.vars.value
+                                ) }}
+                            </div>
+                                {{ form_row(formpaste.problem) }}
+                                {{ form_row(formpaste.language) }}
+                                {{ form_row(formpaste.entry_point) }}
+                            <div class="modal-footer">
+                                <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Cancel</button>
+                                <button type="submit" class="btn btn-primary">
+                                    <i class="fas fa-paste"></i> Submit Paste
```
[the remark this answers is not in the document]
as6325400: If both "Upload" and "Paste" are available at the same time, would it be weird or cause confusion if both buttons are labeled "Submit"?

webapp/src/Controller/Team/SubmissionController.php
as6325400: Have you considered ignoring the entry in Paste mode, since only one file can be uploaded with Paste?

etc/db-config.yaml
as6325400: Okay, but in that case, the cookies will also need a major revamp.

webapp/src/Controller/Team/SubmissionController.php
meisterT: One example is Java, where the entry point basically specifies the class. If you would make the name of the file clear in the UI (or even an input that can be changed after selecting the problem) we could do the same auto-detection (with possibility to change it) with the entry point as for the uploaded file.

as6325400: I think it's feasible in terms of implementation, but Paste code is essentially designed for convenience with a single file. If we allow changes to the file name, it somewhat defeats the original purpose of this feature. Alternatively, we could try handling languages where the entry point is not the file name with special treatment? (Actually, this seems similar to the original approach for uploads.)

as6325400: By the way, I'm currently teaching a data structures class where only C language is allowed, and the usage of Paste is significantly higher than that of upload.

webapp/templates/team/submit_modal.html.twig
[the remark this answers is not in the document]
meisterT: I don't think it is confusing as in both cases you are submitting what you filled in in the form you are currently seeing.

webapp/src/Controller/Team/SubmissionController.php
meisterT: Assume I am coding Java in my IDE and the IDE automatically adds a `package foo;` statement on top of my file with the class `Bar` and `Goo` in it. Now I **need** to be able to configure the  entry point after copy/pasting to say `foo.Bar` or `foo.Goo` (depending where the main method is defined, or if in both which one I want to be calling).

as6325400: Alright, this is a bit difficult for me. I don't know Java and Kotlin. I think I need to learn more about these two languages.

webapp/templates/team/submit_modal.html.twig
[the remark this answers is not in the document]
as6325400: I understand now, I misunderstood what you meant.

etc/db-config.yaml
as6325400: You should prevent the array from being empty here. How should I handle it when someone submits the form with an empty array?

Should I use an alert?

I haven’t seen any other similar menus to refer to for this implementation.

Sorry about that.

webapp/src/Controller/Team/SubmissionController.php
as6325400: I made a modification so that when the language is Java or Kotlin, the entry_point is always required, regardless of whether the language itself is set to require an entry_point. For other languages, the entry_point will still default to the filename as originally designed. I've tested it myself and it seems to work fine. Is this modification acceptable?

webapp/src/Controller/Team/SubmissionController.php
as6325400: I still want the pasted submission to have a special identifier in the filename, to differentiate it from uploads. Is it possible to do this?

webapp/src/Controller/Team/SubmissionController.php
nickygerritsen: I don't think this is the desired approach. Better would be to mimic our javascript logic: https://github.com/DOMjudge/domjudge/blob/5153f91192dcac133985cb36d7c8689387c7ac31/webapp/templates/team/partials/submit_scripts.html.twig#L42-L52

This still means if you copy something with a package it won't work, but that issue exists also in the current code with autodetection.

as6325400:
Sorry to bother you, but I was wondering if there are any improvements needed for this?

Once this is merged, I should be able to work on the issue of replacing it with Monaco.

vmcj:
> Sorry to bother you, but I was wondering if there are any improvements needed for this?
> 
> Once this is merged, I should be able to work on the issue of replacing it with Monaco.

I think you should address @eldering  his question, and we should address some of your questions still.

etc/db-config.yaml
```diff
@@ -201,6 +201,14 @@
 -   category: Display
     description: Options related to the DOMjudge user interface.
     items:
+        -   name: default_submission_code_mode
+            type: array_val
+            default_value: [upload]
+            public: true
+            description: Select the default submission method for the team
```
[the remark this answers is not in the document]
vmcj: > what is submit client?

This: https://raw.githubusercontent.com/DOMjudge/domjudge/refs/heads/main/submit/submit

etc/db-config.yaml
vmcj: I would check that the array doesn't get empty and if it does refuse to save or use a `danger` flash with the message. Alternative is to have a message in the `config-check` page.

webapp/src/Controller/Team/SubmissionController.php
vmcj: > Additionally, I'd like to ask how I can reference code in a comment.

https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-a-permanent-link-to-a-code-snippet#linking-to-code

etc/db-config.yaml
as6325400: I think implementing the required restriction in the backend can be done in the next PR, as it's more general. Including it in this commit would make things confusing.

can i do it this way?

etc/db-config.yaml
vmcj, line 208:
```suggestion
            description: Select the default submission method for the team in the webinterface
```

webapp/src/Controller/Team/SubmissionController.php
```diff
@@ -60,47 +61,135 @@ public function createAction(Request $request, ?Problem $problem = null): Respon
...
+                        date('Y-m-d_H-i-s'),
+                        $language->getExtensions()[0]
+                    );
+                    $tempFileName = preg_replace('/[^a-zA-Z0-9_.-]/', '_', $tempFileName);
+
+                    if ($language->getExtensions()[0] == 'java' || $language->getExtensions()[0] == 'kt') {
+                        $entryPoint = $formPaste->get('entry_point')->getData() ?: null;
+                        // Check for invalid characters in entry point name
+                        $invalidChars = '/[<>:"\/\\|?*]/';
+                        if(preg_match($invalidChars, $entryPoint)) {
```
vmcj, line 120: There are some warnings on https://github.com/DOMjudge/domjudge/pull/2720/files which you should still handle.

webapp/src/Controller/Team/SubmissionController.php
vmcj: What is the reason the way of submitting is important? You could test with adding `fileupload-` to the name? You would need to test yourself if this still works.

as6325400: Why is it "fileupload"? I thought it would be something like "Paste."

However, I now think using "shortName" directly is fine too.

webapp/src/Controller/Team/SubmissionController.php
as6325400: done

vmcj: I still see them:
image

maybe you have to toggle `a` to see them.

webapp/src/Controller/Team/SubmissionController.php
```diff
@@ -60,47 +61,136 @@ public function createAction(Request $request, ?Problem $problem = null): Respon
...
+                    $tempFileName = sprintf(
+                        'submission_%s_%s_%s.%s',
+                        $user->getUsername(),
+                        $problem->getName(),
+                        date('Y-m-d_H-i-s'),
+                        $language->getExtensions()[0]
+                    );
+                    $tempFileName = preg_replace('/[^a-zA-Z0-9_.-]/', '_', $tempFileName);
+
+                    if ($language->getExtensions()[0] == 'java' || $language->getExtensions()[0] == 'kt') {
```
meisterT, line 116: Instead of hardcoding Java and Kotlin here, use the `$language->getRequireEntryPoint()` instead.

Similarly, use `$language->getEntryPointDescription()` for a description to display.

meisterT: We probably want to use what we built here so that the entry point works the same regardless of the upload methos: https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Form/Type/SubmitProblemType.php#L66

webapp/src/Controller/Team/SubmissionController.php
meisterT: If you think the short name as given by the contest problem is ok, let's use that.

as6325400: Currently, the short name allows various characters. I will still use ```preg_replace('/[^a-zA-Z0-9_.-]/', '_', $tempFileName);``` to handle file naming using the short name.

webapp/src/Controller/Team/SubmissionController.php
as6325400: Sorry, I don't understand how to use $language->getRequireEntryPoint(). For example, in the case of Java, when using paste_code, the entry_point might still be no. In this case, how can I use this to determine the file name standard?

vmcj: I think the remark here is that instead of listing the current languages which need an entrypoint, use that function @meisterT linked and use it from there?

as6325400: > I think the remark here is that instead of listing the current languages which need an entrypoint, use that function @meisterT linked and use it from there?

The function should be checking whether the language in the form requires an entrypoint, which seems to be different from my current situation, right?

vmcj: Why would it be different? You check for 2 languages here and determine the entrypoint for those. If there are other languages like this we would also want to pick them up. People can add more languages besides those 2.

webapp/src/Form/Type/SubmitProblemPasteType.php
```diff
@@ -0,0 +1,108 @@
...
+            'required' => false,
+            'help' => 'The entry point for your code.',
+            'row_attr' => ['data-entry-point' => ''],
+            'constraints' => [
+                new Callback(function ($value, ExecutionContextInterface $context) {
+                    /** @var Form $form */
+                    $form = $context->getRoot();
+                    /** @var Language $language */
+                    $language = $form->get('language')->getData();
+                    $langId = strtolower($language->getExtensions()[0]);
```
vmcj, line 78:
```suggestion
                    $langId = strtolower($language->getExtensions()[0]);
```

webapp/src/Controller/Team/SubmissionController.php
as6325400: @vmcj 
截圖 2024-11-16 晚上8 10 24
I'm not sure if my understanding is correct, but are the methods **$language->getRequireEntryPoint()** and **$language->getEntryPointDescription()** linked to the part I highlighted in the image?

If so, there might be an issue.

as6325400: For example, in the case of Java, it can be configured to not require an entry point. In this scenario, under the original upload mode, the uploaded file itself becomes the entry point. However, due to the characteristics of the Java language, the file name must match the main class name.

In the case of the paste code mode, regardless of whether an entry point is required or not, the user must be allowed to specify the main class name. Otherwise, I won’t be able to save the file correctly, which will result in a compile error.

meisterT:
This has been implemented in https://github.com/DOMjudge/domjudge/pull/3526

Issue #2703, opened 2024-09-19
Change submissions over time to bar plot


eldering:
On the `/jury/analysis` page it would be nice to replace the current submissions over time plot (see below) by a bar plot with submissions in given time interval buckets, so it's easier to see how many submissions arrived at each point/interval of time. Currently that's difficult to read off from the slope of the plot.

![image](https://github.com/user-attachments/assets/315097db-c125-4d16-9006-f79b5f0bcdea)

as6325400:
Could you provide me with a dataset of submissions evenly distributed throughout a contest that I can import?

I think I can try modifying it, but I don't have the data. Thanks!

eldering:
> Could you provide me with a dataset of submissions evenly distributed throughout a contest that I can import?
> 
> I think I can try modifying it, but I don't have the data. Thanks!

I can share the demoweb SQL dump with you. Can you ping me privately in slack?

as6325400:
> > Could you provide me with a dataset of submissions evenly distributed throughout a contest that I can import?
> > I think I can try modifying it, but I don't have the data. Thanks!
> 
> I can share the demoweb SQL dump with you. Can you ping me privately in slack?

OK

as6325400:
this issue can be close?

PR #2722, opened 2024-09-26
Change the total number of submissions to the number of submissions p…

as6325400:
## Tracking issue
[issue](https://github.com/DOMjudge/domjudge/issues/2703)

## Description
Change the cumulative number of submissions over time to the number of submissions per time interval.

## Screenshots
截圖 2024-09-26 下午1 00 47

### DIFF
--- a/webapp/templates/jury/analysis/contest_overview.html.twig
+++ b/webapp/templates/jury/analysis/contest_overview.html.twig
@@ -356,26 +356,55 @@ nv.addGraph(function() {
 // x-axis is contest time
 // y axis is # of submissions
 
+var max_submissions_per_minute = 0;
+
 var submission_stats = [
   {% for result in ['correct', 'wrong-answer', 'timelimit', 'run-error', 'compiler-error', 'no-output'] %}
   {
     key: "{{result}}",
     color: "{{colors[result]}}",
-    values: [
-      {# TODO: make sure these are actually ordered by submittime #}
-      {# TODO: also make sure these submissions are in the same contest #}
-      [0,0],
-      {% set count = 0 %}
-      {% for submission in submissions | filter(submission => submission.result) %}
-        {% if submission.result == result %}{% set count = count +1 %}{% endif %}
-        [ {{ (submission.submittime - current_contest.starttime)/60.0 }},
-          {{ count }}
-        ],
-        {% endfor %}
-    ]
+    values: []
   },
   {% endfor %}
 ];
+
+var submissions = [
+  {% for submission in submissions %}
+  {
+    result: "{{ submission.result }}",
+    submittime: {{ submission.submittime }},
+    starttime: {{ current_contest.starttime }}
+  }{{ loop.last ? '' : ',' }}
+  {% endfor %}
+];
+
+var contest_duration_minutes = Math.ceil(({{ current_contest.endtime }} - {{ current_contest.starttime }}) / 60);
+
+submission_stats.forEach(stat => {
+  stat.values = Array.from({ length: contest_duration_minutes }, (_, i) => [i, 0]);
+});
+
+const statMap = submission_stats.reduce((map, stat) => {
+  map[stat.key] = stat;
+  return map;
+}, {});
+
+submissions.forEach(submission => {
+  let submission_minute = Math.floor((submission.submittime - submission.starttime) / 60);
+  let stat = statMap[submission.result];
+  if (stat && submission_minute >= 0 && submission_minute < contest_duration_minutes) {
+    stat.values[submission_minute][1]++;
+  }
+});
+
+for (let minute = 0; minute < contest_duration_minutes; minute++) {
+  let this_minute_submission_nums = 0;
+  submission_stats.forEach(stat => {
+    this_minute_submission_nums += stat.values[minute][1];
+  });
+  max_submissions_per_minute = Math.max(max_submissions_per_minute, this_minute_submission_nums);
+}
+
 nv.addGraph(function() {
   var chart = nv.models.stackedAreaChart()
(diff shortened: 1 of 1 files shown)


as6325400:
I'm not very familiar with Twig, but I believe there would be a more efficient approach in terms of time complexity if implemented in the controller. However, given the scale of the competition's data, the difference may not be significant.

as6325400:
## Optimize time complexity

| 2000 submissions      | time (ms)  |
|---------------|--------------|
| Origin  | 9859 ms      |
| Optimize | 277 ms       |

原始執行時間
_9859 ms_

優化後執行時間
_277 ms_

meisterT:
This is from this year's finals:
image

I wonder how this would look with log scale (I tried making that work but didn't succeed, perhaps someone else has an idea).

There is another minor thing we might want to improve: in case the contest duration is not just 5 hours, but say 1 week or even more we create too many data points. We could decide to always use a fixed number of buckets, e.g. 60 - in a 5h contest this would group things together in 5min buckets. That should work well for most contest durations.

as6325400:
> This is from this year's finals: image
> 
> I wonder how this would look with log scale (I tried making that work but didn't succeed, perhaps someone else has an idea).
> 
> There is another minor thing we might want to improve: in case the contest duration is not just 5 hours, but say 1 week or even more we create too many data points. We could decide to always use a fixed number of buckets, e.g. 60 - in a 5h contest this would group things together in 5min buckets. That should work well for most contest durations.

I think the number of buckets can be fixed, and it's not difficult to change this. Should we make this change in this PR itself? Or should we discuss it later? I feel that 60 might be a bit too few.

mpsijm:
If it's going to be buckets (whether they're gonna be 1 minute or 5 minutes or some other size), maybe a bar chart would work better than a line graph? There's no continuity between each subsequent pair of buckets :slightly_smiling_face:

as6325400:
> If it's going to be buckets (whether they're gonna be 1 minute or 5 minutes or some other size), maybe a bar chart would work better than a line graph? There's no continuity between each subsequent pair of buckets 🙂

I think this is a good approach, although in very dense situations, these two charts might look quite similar.

meisterT:
I'm fine with doing the bucketing in a separate PR as this is already an improvement.

I also wondered what a (stacked) bar chart might look like. Do you want to give this a try now or later?

as6325400:
> I'm fine with doing the bucketing in a separate PR as this is already an improvement.
> 
> I also wondered what a (stacked) bar chart might look like. Do you want to give this a try now or later?

Since I'm not sure if the bar chart would be better, I'll implement the bar chart version in another branch and then submit a PR for you to compare the differences. Does that work for you?

meisterT:
:+1:

webapp/templates/jury/analysis/contest_overview.html.twig
```diff
@@ -356,26 +356,47 @@ nv.addGraph(function() {
...
+  stat.values = Array.from({ length: contest_duration_minutes }, (_, i) => [i, 0]);
+});
+
+const statMap = submission_stats.reduce((map, stat) => {
+  map[stat.key] = stat;
+  return map;
+}, {});
+
+submissions.forEach(submission => {
+  let submission_minute = Math.round((submission.submittime - submission.starttime) / 60);
```
eldering, line 393: Small detail: this should not be rounded, but the floor taken. A submission at 57 seconcs into the contest is in minute 0, not minute 1. Above `contest_duration_minutes` should be taken the `ceil` though, to make sure that `0 <= submission_minute < contest_duration_minutes` always holds.

webapp/templates/jury/analysis/contest_overview.html.twig
```diff
@@ -388,7 +409,7 @@ nv.addGraph(function() {
       .showYAxis(true)        //Show the y-axis
       .showXAxis(true)        //Show the x-axis
       .forceX([0, {{ (current_contest.endtime - current_contest.starttime) / 60 }}])
-      .forceY([0, {{ submissions|length *1.10 }}])
+      .forceY([0, {{ max_submissions_per_minute * 1.10 }}])
```
eldering, line 412: Maybe I overlook something, but where is `max_submissions_per_minute` ever updated?

eldering:
> I'm fine with doing the bucketing in a separate PR as this is already an improvement.
> 
> I also wondered what a (stacked) bar chart might look like. Do you want to give this a try now or later?

I think a bar chart would be clearer, but I'm fine with doing that in a second iteration.

webapp/templates/jury/analysis/contest_overview.html.twig
as6325400: If a contest lasts 300 minutes, can a submission be made at 299 minutes and 59 seconds? Is it possible to submit during the 300th minute? If not, then should contest_duration_minutes be taken as the floor (rounded down)?

as6325400: Oh, sorry, I understand now. It was my mistake.

webapp/templates/jury/analysis/contest_overview.html.twig
as6325400: That was my oversight. I've fixed it.

as6325400:
Do I need to squash them to a single commit?

eldering:
> Do I need to squash them to a single commit?

Preferably yes, at least for bugfixes. You can do that either locally with `git rebase --interactive` or if there's really only one single logical commit necessary (which I think makes sense here), then tick the box in github.

eldering (approved):

meisterT (approved):

PR #2727, opened 2024-09-29
Change the line chart to a bar chart.

as6325400:
### Bar Chart
Bar Chart

### Origin Chart
Origin Chart

It has been changed to a bar chart. Personally, I don't think the difference is significant.

### DIFF
--- a/webapp/templates/jury/analysis/contest_overview.html.twig
+++ b/webapp/templates/jury/analysis/contest_overview.html.twig
@@ -406,18 +406,17 @@ for (let minute = 0; minute < contest_duration_minutes; minute++) {
 }
 
 nv.addGraph(function() {
-  var chart = nv.models.stackedAreaChart()
+  var chart = nv.models.multiBarChart()
       // .margin({left: 100})  //Adjust chart margins to give the x-axis some breathing room.
       // .useInteractiveGuideline(true)  //We want nice looking tooltips and a guideline!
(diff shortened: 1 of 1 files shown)


mpsijm:
I do prefer the bar chart, makes it clearer that we're looking at individual submissions per bin :smile: Talking about bins, what was the idea again for contests that run for (way) longer than 5 hours? :slightly_smiling_face: 

Also, the x-axis looks off: why does it use multiples of 17 instead of 15? :thinking:

as6325400:
> I do prefer the bar chart, makes it clearer that we're looking at individual submissions per bin 😄 Talking about bins, what was the idea again for contests that run for (way) longer than 5 hours? 🙂
> 
> Also, the x-axis looks off: why does it use multiples of 17 instead of 15? 🤔

Regarding the x-axis, it's likely due to RWD (Responsive Web Design). The original version also adjusts the ticks automatically, so I'll take a look into it later.

As for the bins, I think we can set it to max(300, number of minutes) bins, since most contests don't exceed 5 hours.

as6325400:
Is it acceptable to use max(300, number_of_minutes) as the interval count?

Should I submit this change in this PR?

meisterT (approved):

meisterT:
Merging this as is - let's do it max bucketing and ticks in a separate PR.

as6325400:
> Merging this as is - let's do it max bucketing and ticks in a separate PR.

ok

Issue #2518, opened 2024-05-01
Implement partial scoring


meisterT:
* With partial scoring testcases are grouped in testcase groups.
* Each testcase group has a number of points assigned.
* If the submission solves **all** testcases in the group, the points for that group are awarded.

Then there are two variants:

1. The points for a submission is determined by the sum of the awarded points per testcase group. The points of the problem are determined by taking the maximum of the points of all submissions to that problem.
2. Multiple submissions can solve different testcase groups, so you take the maximum per group over all submission and then sum up.

If we would implement the first variant, the second variant could be worked around by splitting it up into multiple problems.

The kattis problem format suggests adding a `testdata.yml` per testcase group: https://icpc.io/problem-package-format/spec/problem_package_format#test-data-groups

meisterT:
From https://cms.readthedocs.io/en/v1.4/Configuring%20a%20contest.html#computation-of-the-score:

> “Use the sum over each subtask of the best result for that subtask across all submissions” is the score mode that follows the rules of IOI since 2017. It is intended to be used with tasks that have a group score type, like “GroupMin” (note that “group” and “subtask” are synonyms). The score on the task is the sum of the best score for each subtask, over all submissions. The difference with the previous score mode is that here a contestant can achieve the maximum score on the task even when no submission gets the maximum score (for example if each subtask is solved by exactly one submission).

alfphaderp:
In each case, how will penalty time be calculated?

eldering:
Related to my comments in https://github.com/DOMjudge/domjudge/issues/2525#issuecomment-2091757210, I'm not a big fan option 2 to let the score for a problem depend in such a complex way depend on the score of each individual submission. I'd rather stay with: the score for a problem is always the maximum over the scores of each team's submissions for that problem.

meisterT:
I think this is mostly implemented.

What's missing:
- should there be tie-breakers for scoring problems?
- check `on_reject_continue` handling, I think it doesn't work under all circumstances
- range bounds are not enforced at all levels
- should any hierarchy be displayed to teams?

PR #2730, opened 2024-10-02
Hotjar csp

zapotocnylubos:


zapotocnylubos:
sorry, wrong default repository target...

zapotocnylubos:
@vmcj we did not stick to any standard, so the main problem is that you wanted to adhere to some standard, but we needed it quite fast, and it is not yet 100% integrated, for example, we did not have time to change the scoreboard behavior.
I know about https://github.com/DOMjudge/domjudge/issues/2518, our solution uses percentage based scoring, because you can change the number of point in the contest - so for example contest 1 has problem 1 for 100 points, but contest 2 has problem 1 only for 20 points. Then we define testcase groups, each having percentage assigned - summing up all testcase groups should result in 100%. 
So basically this
> With partial scoring testcases are grouped in testcase groups.
> If the submission solves all testcases in the group, the points for that group are awarded.

The main requirement for partial scoring is that lazy evaluation needs to be turned off so that each testcase group can be correctly computed and the resulting points summed up.
We show the points from the "best" submission on the submission list.

Adhering to this basically

> The points for a submission is determined by the sum of the awarded points per testcase group. The points of the problem are determined by taking the maximum of the points of all submissions to that problem.


This is the PR we implemented our solution if you want to have a look at the code and screenshots https://github.com/delta-cs/seminar-domjudge/pull/101 (its in czech, but google translator should help)

PR #2732, opened 2024-10-05
Limit the labels on the x-axis

as6325400:
## Old PR
https://github.com/DOMjudge/domjudge/pull/2727

## Change
Since the `forceX` method is no longer available after using `multibarchart`, it causes the x-axis labels to be misaligned, and I can't set the competition end time as the final label. Therefore, I force interpolation to fill in the x-axis labels (I haven't thought of a better solution yet).

In addition to 0 and the end time, I will insert 9 additional points. This should handle all competitions where the duration is a multiple of 10.

## Next

Additionally, according to the last PR, the data size of the table uses `min(minutes, 300)`. How should I create the labels in this case? For example, if there is a competition that lasts for a week, should the unit of my labels still be in minutes?

### DIFF
--- a/webapp/templates/jury/analysis/contest_overview.html.twig
+++ b/webapp/templates/jury/analysis/contest_overview.html.twig
@@ -381,7 +381,7 @@ var submissions = [
 var contest_duration_minutes = Math.ceil(({{ current_contest.endtime }} - {{ current_contest.starttime }}) / 60);
 
 submission_stats.forEach(stat => {
-  stat.values = Array.from({ length: contest_duration_minutes }, (_, i) => [i, 0]);
+  stat.values = Array.from({ length: contest_duration_minutes + 1 }, (_, i) => [i, 0]);
 });
 
 const statMap = submission_stats.reduce((map, stat) => {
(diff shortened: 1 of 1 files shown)


webapp/templates/jury/analysis/contest_overview.html.twig
```diff
@@ -397,14 +397,30 @@ submissions.forEach(submission => {
...
-for (let minute = 0; minute < contest_duration_minutes; minute++) {
+for (let minute = 0; minute <= contest_duration_minutes; minute++) {
   let this_minute_submission_nums = 0;
   submission_stats.forEach(stat => {
     this_minute_submission_nums += stat.values[minute][1];
   });
   max_submissions_per_minute = Math.max(max_submissions_per_minute, this_minute_submission_nums);
 }
 
+function generate_values(start_point, end_point, number_of_points) {
```
meisterT, line 408: I think we can this both simpler and nicer:

This will make sure that the ticks we use are still limited but are nice. In the default 5h long contest, every 15 minutes there is a tick, for longer contests, it increases to 30mins, 60mins and so on.

```
// Pick a nice round tickDelta and tickValues
var tickDelta = 15;
while (contest_duration_minutes / tickDelta > 15) {
  tickDelta *= 2;
}
var tickValues = Array.from({ length: Math.ceil(contest_duration_minutes / tickDelta) + 1 }, (_, i) => i * tickDelta);
```

as6325400: So, the unit will always be in minutes?

If we take a classroom scenario, it's quite common for a contest to last an entire semester.

meisterT: You are right that we also would want to fall back to larger units than minutes when the contest duration is that long - I assume we should do that in the same change that adds the limitation to 300 buckets/bars. Do you rather want to do this in this PR or separately?

as6325400: Let's handle it separately.

as6325400: alredy update

meisterT (approved):

Issue #2307, opened 2024-01-24
Multipass problem Support


mzuenni:
### Description of the enhancement request
In multipass problems the submission gets started multiple times and is fed with the output of the output valdiator from the previous run (but besides that the new run has no knowledge of any prior run). This allows some new types of problems about information theory and some other fields.

### Expected behaviour
- The submission is run on the test input (like for normal problems)
- The output Validator is run on the submission output (like for normal problems)
- The output Validator either accepts, rejects or indicates that the submission should be run again with a new input that it provides

### Any other information that you want to share?
Multipass seems to also get added to the kattis problem package format: https://www.kattis.com/problem-package-format/spec/2023-07-draft.html#multi-pass-validation
Maybe it would also be nice to run the output validator with an argument which indicates which iteration this is or provide a way for the output validator to keep information between invocations.

vmcj:
@mzuenni as discussed in person, I'll close this issue as this requires the problemspec to first have a definition for this. Please re-open when that is in a reasonable state so we can start implementing this.

mzuenni:
Just a small update:

> A multi-pass validator can be used for problems that should run the submission multiple times sequentially, using a new input generated by output validator during the previous invocation of the submission.
> 
> To signal that the submission should be run again, the output validator must exit with code `42` and output the new input in the file `nextpass.in` in the feedback directory. Judging stops if no nextpass.in was created, or the output validator exited with any other code.
> 
> It is a judge error to create the `nextpass.in` file and exit with any other code than 42.

- `multipass` can be combined with `validation` set to `custom` or `custom interactive`.
- Regarding the timelimit, it should be checked per invocation of the submission.
- The `judgemessage.txt` and `judgeerror.txt` should not be removed between invocations of the output validator i.e. the validator is responsible for appending/overwriting to the files in there

Comments:
- If the validator uses `stdout` and `stderr` you likely want to display that per invocation

Open questions:
- Can the output validator (inside feedback?) write other files that persist between invocations?

Are there more things that need to be properly defined?

mzuenni:
Maybe these are helpful for testing:
https://github.com/RagnarGrootKoerkamp/BAPCtools/tree/feat/multipass/test/problems/multipass
https://github.com/RagnarGrootKoerkamp/BAPCtools/tree/feat/multipass/test/problems/interactivemultipass

meisterT:
Example interactive problem on Kattis: https://open.kattis.com/problems/guessinggame2

[...]

PR #2733, opened 2024-10-05
Allow changing the multi-pass problems attributes in the web UI.

meisterT:
Part of #2307.

### DIFF
--- a/webapp/src/Form/Type/ProblemType.php
+++ b/webapp/src/Form/Type/ProblemType.php
@@ -81,6 +81,15 @@ class ProblemType extends AbstractExternalIdEntityType
             'label' => 'Use run script as compare script.',
             'required' => false,
         ]);
+        $builder->add('multipassProblem', CheckboxType::class, [
+            'label' => 'Multi-pass problem',
+            'required' => false,
+        ]);
+        $builder->add('multipassLimit', IntegerType::class, [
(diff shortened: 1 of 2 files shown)


meisterT:
cc @cubercsl

nickygerritsen (approved):

webapp/src/Form/Type/ProblemType.php
```diff
@@ -81,6 +81,15 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             'label' => 'Use run script as compare script.',
             'required' => false,
         ]);
+        $builder->add('multipassProblem', CheckboxType::class, [
+            'label' => 'Multi-pass problem',
+            'required' => false,
+        ]);
+        $builder->add('multipassLimit', IntegerType::class, [
```
nickygerritsen, line 88: We could hide this with some JavaScript but I will do this later

nickygerritsen (approved):

webapp/src/Form/Type/ProblemType.php
cubercsl: Do you think we should check this value? 
For multi-pass problems, the passlimit must be >= 2, and for traditional problem it must be equal to 1. (I noticed that `getMultiPassLimit` checks for traditional problem, but the value stored in the db may be inconsistent)

Issue #2728, opened 2024-09-29
Streamed event feed does not send the latest events


ningit:
### Description of the problem

Contest event feeds are not sending the latest events when `stream=true`, but they do when `stream=false`. Moreover, the output stops in the middle of the JSON object. This looks like a flushing issue on the DOMjudge side.

### Your environment

* DOMjudge version: 8.3.0.
* Operating system: Debian 12 (bookworm)
* Web server: Apache 2.4.62

### Steps to reproduce

The issue could be reproduced (with some instance of DOMjudge 8.3.0, I have also tried with [demoweb](https://www.domjudge.org/demoweb/public), but the `event-feed` method always produces `{"code":500,"message":"Internal Server Error"}`) with the following commands:

```
curl -v --no-buffer --user 'user:passwd' \
    "https://<server name>/domjudge/api/v4/contests/<contest>/event-feed?stream=false&since_token=33874"
... request headers ...
< HTTP/1.1 200 OK
< Server: Apache/2.4.62 (Debian)
< Cache-Control: no-cache, private
< X-Accel-Buffering: no
< X-Current-Contest:
< Transfer-Encoding: chunked
< Content-Type: application/x-ndjson
{"token":"33875","id":"4301", ...}
... some more messages ...
{"token":"34174","id":"4323", ...}
```

The previous request sets the `stream` parameter to `false`, so the command finishes and outputs all the events. However, if we set `stream` to `true`, the command keeps waiting as expected, but it stops in the middle of an older event.

```
curl -v --no-buffer --user 'user:passwd' \
    "https://<server name>/domjudge/api/v4/contests/<contest>/event-feed?stream=true&since_token=33874"
... same headers and events ...
{"token":"34101","id
```

The stream does not progress unless there is more activity in the contest. Hence, events will arrive with some (unbounded) delay or will not arrive at all if there is no more interaction.

The problem is not likely on the client side because `curl` is run with the `--no-buffer` flag. Moreover, I have checked with Wireshark that the last chunk stops at `"id` and then only keep-alive messages are received.

### Expected behaviour

The stream stops after the last event (in real time).

### Actual behaviour

The stream stops in the middle of an older event.

meisterT:
We have not seen this before, I wonder if this is an artifact of Apache or general webserver setup.

Just to double check - you see the same behavior as with curl also with `httpie`?

meisterT:
I am using Apache locally and am not seeing this problem.

[...]

PR #2738, opened 2024-10-06
Fix event feed last event ID lookup.

meisterT:
This has been added in https://github.com/DOMjudge/domjudge/commit/ef8c832e63ff4c2538b595ca91bd4389d9e59387

The first `eventid` will be 1. If the event feed reading client does not pass in a `since_token` or `since_id`, we set `since_id` to `-1` to signal start reading from the beginning. In that case there is no previous event that can be expected to exist.

This was also reported as broken in a side comment of https://github.com/DOMjudge/domjudge/issues/2728


nickygerritsen (approved):

eldering (approved):

eldering:
Looks like a typo I overlooked and that wasn't tested.

Kevinjil (approved):
Confirmed while testing #2740

Issue #367, opened 2018-04-19
Add image visualization as type of script to generate testcase images


eldering:
Currently we support having testcase images, but only loaded statically. We should have a new type of script to both be able generate those during import and also to generate them from team output, in case the visualization depends on the output and then also display those in the jury interface for visual comparison.

meisterT:
I seem to remember that you have a prototype of this, is that correct?

eldering:
I started looking into it, but no prototype yet. All I have is https://github.com/eldering/domjudge/tree/testcase-image-gen

nickygerritsen:
Some points:
* We want to visualize after the result is known to the team (and jury)
* We want to run the visualizer on the judgehost

meisterT:
In the discussion just now, it came up that this could a judgetask.

vmcj:
@eldering looking at the problem spec it seems to hint that a `generator` for testcases should be run before import, I would say that a generator for the testcases should by the same logic also be ran before the import.

So should we only visualize the output of teams? I see an usecase for that or do you want both features?

eldering:
> @eldering looking at the problem spec it seems to hint that a `generator` for testcases should be run before import, I would say that a (testcase)visualization_generator should by the same logic also be ran before the import.
> 
> So should we only visualize the output of teams? I see an usecase for that or do you want both features?

Yes, this only applies to team output in the current context.

PR #2744, opened 2024-10-14
Output vizualizer

vmcj:
I tried to implement https://github.com/DOMjudge/domjudge/issues/367. This is still a bit rough but I rather first get some feedback instead of continuing with this if this is not what @eldering  envisioned.

The main problem I currently have is which testcases to visualize, do we only want the failing one and if so only the `wrong-answer` one or do we just want to try all of them and accept that the script might fail as the team output might not conform to the expected output. (For example: the team might have stopped halfway through the problem or might still have debug output etc.)

The visualized team output for boolfind is ofcourse not very helpful but as this is only a demo this is the best for now.

When implementing this I found out we could just always visualize as a task with lower priority and update the visualization task with the judgehost which should have the needed files. I wrongly assume that the file will always be at the judgehost to keep this simple but that should still be fixed if we want to continue with this.


vmcj:
image

webapp/migrations/Version20241018061817.php
```diff
@@ -0,0 +1,56 @@
...
+        $this->addSql('ALTER TABLE visualization DROP FOREIGN KEY FK_E0936C40D360BB2B');
+        $this->addSql('DROP TABLE visualization');
+        $this->addSql('ALTER TABLE judging DROP visualization');
+        $this->addSql('ALTER TABLE `judgetask`
+                       MODIFY COLUMN `type` ENUM(\'judging_run\', \'generic_task\', \'config_check\', \'debug_info\', \'prefetch\') DEFAULT \'judging_run\' NOT NULL COMMENT \'Type of the judge task.(DC2Type:judge_task_type)\'');
+        $this->addSql('ALTER TABLE judgetask DROP output_visualizer_script_id');
+        $this->addSql('ALTER TABLE problem DROP FOREIGN KEY FK_D7E7CCC819F5352E');
+        $this->addSql('DROP INDEX special_output_visualizer ON problem');
+        $this->addSql('ALTER TABLE problem DROP special_output_visualizer, CHANGE multipass_limit multipass_limit INT UNSIGNED DEFAULT NULL COMMENT \'Optional limit on the number of rounds for multi-pass problems; defaults to 2 if not specified.\'');
+
```
vmcj, line 49:
```suggestion
```

webapp/migrations/Version20241018061817.php
```diff
@@ -0,0 +1,56 @@
...
+        $this->addSql('CREATE TABLE visualization (visualization_id INT UNSIGNED AUTO_INCREMENT NOT NULL COMMENT \'Visualization ID\', judgingid INT UNSIGNED DEFAULT NULL COMMENT \'Judging ID\', judgehostid INT UNSIGNED DEFAULT NULL COMMENT \'Judgehost ID\', testcaseid INT UNSIGNED DEFAULT NULL COMMENT \'Testcase ID\', filename VARCHAR(255) NOT NULL COMMENT \'Name of the file where we stored the visualization.\', INDEX IDX_E0936C40E0E4FC3E (judgehostid), INDEX IDX_E0936C40D360BB2B (testcaseid), INDEX judgingid (judgingid), PRIMARY KEY(visualization_id)) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_unicode_ci` ENGINE = InnoDB COMMENT = \'Team output visualization.\' ');
+        $this->addSql('ALTER TABLE visualization ADD CONSTRAINT FK_E0936C405D5FEA72 FOREIGN KEY (judgingid) REFERENCES judging (judgingid) ON DELETE CASCADE');
+        $this->addSql('ALTER TABLE visualization ADD CONSTRAINT FK_E0936C40E0E4FC3E FOREIGN KEY (judgehostid) REFERENCES judgehost (judgehostid) ON DELETE SET NULL');
+        $this->addSql('ALTER TABLE visualization ADD CONSTRAINT FK_E0936C40D360BB2B FOREIGN KEY (testcaseid) REFERENCES testcase (testcaseid) ON DELETE CASCADE');
+
```
vmcj, line 33:
```suggestion
```

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -1474,10 +1529,11 @@ private function getTestcaseFiles(string $id): array
     )]
     public function getJudgeTasksAction(Request $request): array
     {
-        if (!$request->request->has('hostname')) {
+        /*if (!$request->request->has('hostname')) {
```
vmcj, line 1532:
```suggestion
        if (!$request->request->has('hostname')) {
```

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -1474,10 +1529,11 @@ private function getTestcaseFiles(string $id): array
...
             throw new BadRequestHttpException('Argument \'hostname\' is mandatory');
         }
-        $hostname = $request->request->get('hostname');
+        $hostname = $request->request->get('hostname');*/
```
vmcj, line 1535:
```suggestion
        $hostname = $request->request->get('hostname');
```

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -1667,13 +1744,21 @@ private function serializeJudgeTasks(array $judgeTasks, Judgehost $judgehost): a
...
+                $assignedJudgehost = $judgeTask->getJudgehost();
+                if ($assignedJudgehost === $judgehost || $assignedJudgehost === null) {
+                    $judgetaskids[] = $judgeTask->getJudgetaskid();
+                }
             }
         }
 
         $now = Utils::now();
-        $numUpdated = $this->em->getConnection()->executeStatement(
+        $numUpdated = sizeof($judgeTasks);/*$this->em->getConnection()->executeStatement(
```
vmcj, line 1761:
```suggestion
        $numUpdated = $this->em->getConnection()->executeStatement(
```

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -1688,7 +1773,7 @@ private function serializeJudgeTasks(array $judgeTasks, Judgehost $judgehost): a
         if ($numUpdated == 0) {
             // Bad luck, some other judgehost beat us to it.
             return [];
-        }
+        }*/
```
vmcj, line 1776:
```suggestion
        }
```

webapp/src/Controller/Jury/UserController.php
```diff
@@ -167,6 +173,107 @@ public function indexAction(): Response
...
+                'UPDATE team SET judging_last_started = :starttime WHERE teamid = :teamid',
+                [
+                    'starttime' => $now,
+                    'teamid' => $teamid,
+                ]
+            );
+        }
+
+        // Just return everything here
+        return $judgeTasks;
```
vmcj, line 249:
```suggestion
```

webapp/src/Entity/Visualization.php
```diff
@@ -0,0 +1,86 @@
...
+        return $this->judgehost;
+    }
+
+    public function setJudgehost(Judgehost $judgehost): Visualization
+    {
+        $this->judgehost = $judgehost;
+        return $this;
+    }
+
+    public function getTestcase(): Judgehost
```
vmcj, line 76:
```suggestion
    public function getTestcase(): Testcase
```

judge/judgedaemon.main.php
```diff
@@ -768,7 +771,7 @@ function fetch_executable_internal(
             $judgehosts = request('judgehosts', 'GET');
             if ($judgehosts !== null) {
                 $judgehosts = dj_json_decode($judgehosts);
-                $judgehost = array_filter($judgehosts, fn($j) => $j['hostname'] === $myhost);
+                $judgehost = array_values(array_filter($judgehosts, fn($j) => $j['hostname'] === $myhost))[0];
```
nickygerritsen, line 774: This is part of your other PR

judge/judgedaemon.main.php
```diff
@@ -872,6 +877,50 @@ function fetch_executable_internal(
...
+                $run_config = dj_json_decode($judgeTask['run_config']);
+                $tmpfile = tempnam(TMPDIR, 'output_visualization_');
+                [$runpath, $error] = fetch_executable_internal(
+                    $workdirpath,
+                    'output_visualization',
+                    $judgeTask['output_visualizer_script_id'],
+                    $run_config['hash']
+                );
+                if (isset($error)) {
+                    // FIXME
```
nickygerritsen, line 897: Fix me 😛

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -550,6 +553,58 @@ public function addDebugInfo(
...
+        $judging = $this->em->getRepository(Judging::class)->find($judgeTask->getJobId());
+        if ($judging === null) {
+            throw new BadRequestHttpException(
+                'Inconsistent data, no judging known with judgingid = ' . $judgeTask->getJobId() . '.');
+        }
+        if ($tempFilename = tempnam($this->dj->getDomjudgeTmpDir(), "visual-")) {
+            $debug_package = base64_decode($request->request->get('visual_output'));
+            file_put_contents($tempFilename, $debug_package);
+        }
+        // FIXME: error checking
```
nickygerritsen, line 596: Another fixme

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -1478,6 +1533,7 @@ public function getJudgeTasksAction(Request $request): array
             throw new BadRequestHttpException('Argument \'hostname\' is mandatory');
         }
         $hostname = $request->request->get('hostname');
+        $hostname = 'Computer';
```
nickygerritsen, line 1536: Whoops

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -550,6 +553,58 @@ public function addDebugInfo(
...
+        if (!$judgehost) {
+            throw new BadRequestHttpException("Who are you and why are you sending us any data?");
+        }
+
+        $judging = $this->em->getRepository(Judging::class)->find($judgeTask->getJobId());
+        if ($judging === null) {
+            throw new BadRequestHttpException(
+                'Inconsistent data, no judging known with judgingid = ' . $judgeTask->getJobId() . '.');
+        }
+        if ($tempFilename = tempnam($this->dj->getDomjudgeTmpDir(), "visual-")) {
```
nickygerritsen, line 592: This means it doesn't work with 2 servers in HA mode. We should store this in the DB as a file.

webapp/templates/jury/submission.html.twig
```diff
@@ -532,6 +532,12 @@
                                 {{ runs | displayTestcaseResults(judgingDone) }}
                             </td>
                             <td>
+                                {% if hasOutputVisualizer and judgingDone %}
```
nickygerritsen, line 535: Maybe add `and not visualization is defined`?

webapp/src/Controller/API/JudgehostController.php
vmcj: This is loosely based on what we do with debug packages, so we have the same problem there.

nickygerritsen: Ah I didn't know. Then maybe for now that's fine but we should fix it at some point?

judge/judgedaemon.main.php
vmcj: Loosely based on what we do for the debug tasks, I'll see if I can merge the shared parts and possibly fix this one.

meisterT:
Getting this error when clicking the "visualize" button:
image

example_problems/boolfind/output_visualizer/boolfind_visual/visualize.py
```diff
@@ -0,0 +1,20 @@
...
+my_name  = sys.argv[0]
+my_input = sys.argv[1]
+my_feedback = sys.argv[2]
+
+with open(my_input, 'r') as f:
+  lines = f.readlines()
+  vals = []
+  for line in lines:
+    if 'READ' in line:
+      vals.append(int(line.split(' ')[-1]))
```
meisterT, line 17: This breaks if the submission writes bogus output, e.g. `READ domjudge`, how do we handle cases like this?

judge/judgedaemon.main.php
```diff
@@ -872,6 +877,50 @@ function fetch_executable_internal(
...
+                    continue;
+                }
+
+                $teamoutput = $workdir . "/testcase" . sprintf('%05d', $judgeTask['testcase_id']) . '/1/program.out';
```
meisterT, line 901: where does the `/1/` here come from?

judge/judgedaemon.main.php
```diff
@@ -872,6 +877,50 @@ function fetch_executable_internal(
...
+                $visual_cmd = implode(' ', array_map('dj_escapeshellarg',
+                    [$runpath, $teamoutput, $tmpfile]));
+                system($visual_cmd, $retval);
+                if ($retval !== 0) {
+                    error("Running '$runpath' failed.");
```
meisterT, line 906: let's rather report an internal error like we do in other cases (and not crash judgedaemons)

judge/judgedaemon.main.php
vmcj: Getting the output of the first run, so in other words, this does not work for multipass problems.

webapp/src/Service/ImportProblemService.php
```diff
@@ -926,36 +938,61 @@ public function importProblemFromRequest(Request $request, ?int $contestId = nul
...
     /**
      * @param array{danger: string[], info: string[]} $messages
      */
     private function searchAndAddValidator(ZipArchive $zip, ?array &$messages, string $externalId, string $validationMode, ?Problem $problem): bool
     {
-        $validatorFiles = [];
+        $programStrings = [];
+        $programStrings['package_dir'] = 'output_validators/';
+        $programStrings['type'] = 'output validator';
+        $programStrings['clash'] = 'cmp';
```
meisterT, line 961: what does `clash` stand for?

example_problems/boolfind/problem.yaml
```diff
@@ -1,3 +1,4 @@
 name: Boolean switch search
 
 validation: custom interactive
+visualization: default
```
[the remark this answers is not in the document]
vmcj: No, this is not something which exists in the spec.

meisterT:
Do we limit upload filesize anywhere?

webapp/src/Service/ImportProblemService.php
vmcj: naming clash, this is copied mostly from how the output_validator is uploaded.

example_problems/boolfind/output_visualizer/boolfind_visual/visualize.py
vmcj: I would say this is something which is left up to the jury.

I wonder if there is something we can do with the output_validator & the judgemessage for communication. Either an extra exitcode or otherwise the jury has to parse the output in this script.

judge/judgedaemon.main.php
meisterT: Ah, right - then please add a TODO and consider making the one a local variabled, e.g. `$pass = 1;` and then using that in the path construction.

example_problems/boolfind/problem.yaml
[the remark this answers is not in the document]
meisterT: I was asking because I think this should be `custom` instead of `default` since there is no default validator.

example_problems/boolfind/output_visualizer/boolfind_visual/visualize.py
meisterT: Yes, it's left to the jury. In this case you are the jury by implementing the visualizer :-)

Since people are going to model after examples we give them, we should make it robust and not crash.

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -550,6 +553,58 @@ public function addDebugInfo(
...
+            $debug_package = base64_decode($request->request->get('visual_output'));
```
meisterT, line 593: nit: update name (no longer `debug_package`)

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -550,6 +553,58 @@ public function addDebugInfo(
...
 
+    /**
+     * Add visual team output.
+     */
+    #[IsGranted('ROLE_JUDGEHOST')]
+    #[Rest\Post('/add-visual/{hostname}/{judgeTaskId<\d+>}')]
+    #[OA\Response(response: 200, description: 'When the visual output has been added')]
+    public function addVisualization(
+        Request $request,
+        #[OA\PathParameter(description: 'The hostname of the judgehost that wants to add the debug info')]
```
meisterT, line 564: nit: this is not `debug info`

example_problems/boolfind/output_visualizer/boolfind_visual/visualize.py
meisterT: Adding to this, I think we should return 42 in the script if we were able to visualize the output, and 43 if not.

webapp/src/Entity/Visualization.php
```diff
@@ -0,0 +1,86 @@
...
+ * A team output on a specific testcase.
+ */
+#[ORM\Entity]
+#[ORM\Table(options: [
+    'collation' => 'utf8mb4_unicode_ci',
+    'charset' => 'utf8mb4',
+    'comment' => 'Team output visualization.',
+])]
+#[ORM\Index(columns: ['judgingid'], name: 'judgingid')]
+class Visualization
```
meisterT, line 16: If we allow a visualizer to fail gracefully (see exit code 43 message above), we need to be able to record this here (and perhaps a message?).

meisterT:
One implication of current implementation (which is reasonable) is that if you shut down a judgehost, you are not going to  get the visualization. We should probably think about a good way to message this to the user.

meisterT:
Out of scope for this PR, we could have a toggle at the problem level to visualize the output right after judging a test case (perhaps filtered down by verdict, e.g. only for wrong-answer).

vmcj:
> One implication of current implementation (which is reasonable) is that if you shut down a judgehost, you are not going to get the visualization. We should probably think about a good way to message this to the user.

That problem is also there for the debug package I think. It wouldn't be that hard to extend here to optionally calculate the result on the testcase but if a submission is non-deterministic we would get a different visualization as what the original team-output had. Your remark https://github.com/DOMjudge/domjudge/pull/2744#issuecomment-2429912162 would prevent the problem but would give a problem when visualization takes a some time and there is a backlog.

vmcj:
> Out of scope for this PR, we could have a toggle at the problem level to visualize the output right after judging a test case (perhaps filtered down by verdict, e.g. only for wrong-answer).

I assume still with the priority of first the submission work and only if there is nothing do the visualization or do you want to extend the queued work to have a {Run+Visual}[x] with x as the length of the worklist?

judge/judgedaemon.main.php
vmcj: What would we want to do in this case? I assume we want to visualize actually the last pass?

example_problems/boolfind/problem.yaml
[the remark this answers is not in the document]
vmcj: any string will do at this moment, would be nice to get @eldering his opinion as I just picked something. Will change it to `custom` for now.

vmcj:
Spec changes too much.

PR #2745, opened 2024-10-20
Implement judgedaemon fixme's

vmcj:


nickygerritsen (approved):

judge/judgedaemon.main.php
```diff
@@ -486,7 +486,6 @@ function fetch_executable_internal(
 }
 
 $options = getopt("dv:n:hVe:j:t:", ["diskspace-error"]);
-// FIXME: getopt doesn't return FALSE on parse failure as documented!
```
eldering, line 486: Where does it say that this bug has been fixed? Would be good to either have an explicit link to the bugfix or have tested it.

vmcj: If you have a way to reproduce this? I tested some bogus invocations and didn't get a failure.

eldering: That's exactly reproducing problem. If you can the line
```
echo "Error: parsing options failed.\n";
```
triggered, then this comment can be removed.

vmcj: Ok, I'll play around with it a bit more but have removed the commit for now. Let's discuss next weekend.

judge/judgedaemon.main.php
```diff
@@ -334,7 +334,10 @@ function fetch_executable_internal(
     $execrunpath     = $execbuilddir . '/run';
     $execrunjurypath = $execbuilddir . '/runjury';
     if (!is_dir($execdir) || !file_exists($execdeploypath)) {
-        system('rm -rf ' . dj_escapeshellarg($execdir) . ' ' . dj_escapeshellarg($execbuilddir));
+        system('rm -rf ' . dj_escapeshellarg($execdir) . ' ' . dj_escapeshellarg($execbuilddir), $retval);
+        if ($retval !== 0) {
+            error("Deleting '$execdir' or '$execbuilddir' was unsuccessful.");
```
meisterT, line 339: Can we create an internal error disabling this judgehost instead?

judge/judgedaemon.main.php
```diff
@@ -837,14 +840,16 @@ function fetch_executable_internal(
                     $run_config['hash']
                 );
                 if (isset($error)) {
-                    // FIXME
-                    continue;
+                    $scriptId = $judgeTask['run_script_id'];
+                    error("Retrieving/storing debug script '$scriptId' failed.");
```
meisterT, line 844: similar here, prefer internal error over crashing

judge/judgedaemon.main.php
vmcj: I assume you also want it for the `mkdir` in the next step? I've added that now.

judge/judgedaemon.main.php
```diff
@@ -838,14 +842,16 @@ function fetch_executable_internal(
                     $run_config['hash']
                 );
                 if (isset($error)) {
-                    // FIXME
-                    continue;
+                    $scriptId = $judgeTask['run_script_id'];
+                    disable('judgehost', 'hostname', $myhost, "Retrieving/storing debug script '$scriptId' failed.");
```
meisterT, line 846: This should disable the script not the judgehost. I wonder why we are not calling `fetch_executable` above (instead of the `_internal` variant which would do this for us if I am not mistaken).

judge/judgedaemon.main.php
```diff
@@ -838,14 +842,16 @@ function fetch_executable_internal(
...
                 }
 
                 $debug_cmd = implode(' ', array_map('dj_escapeshellarg',
                     [$runpath, $workdir, $tmpfile]));
                 system($debug_cmd, $retval);
-                // FIXME: check retval
+                if ($retval !== 0) {
+                    disable('judgehost', 'hostname', $myhost, "Running '$runpath' failed.");
```
meisterT, line 853: I am wondering (less strong opinion than above) what the right behavior is here: if executing a debug task fails, it is either a problem with the script or something has been cleaned up. I think I would disable the script for now.

judge/judgedaemon.main.php
vmcj: So in normal cases you'll not get `false`, see:
https://bugs.php.net/bug.php?id=81352 (we run PHP>8.0 so will not encounter this anymore)

You might be able to still get it by not registering $argv (https://stackoverflow.com/a/59785550) or using `unset($argv)` which was enough to encounter this.

Also see the note:
```
The parsing of options will end at the first non-option found, anything that follows is discarded.
```
So it passes a string and if it finds anything which is not an option it will stop. So if we would want to handle this perfectly we would need to count the length of the argument $argv and the index where parsing stopped and compare that? I think this would only complicate the code.

eldering: This shows that `getopt` is buggy:
```
#!/usr/bin/env php
<?php declare(strict_types=1);

$options = getopt("dv:n:hVe:j:t:", ["diskspace-error"], $index);
// FIXME: getopt doesn't return FALSE on parse failure as documented!
if ($options===false) {
    echo "Error: parsing options failed.\n";
    usage();
}
var_dump($index, $options, $argv);
```

vmcj: I agree this is buggy. I don't think there is really more testing we can do apart from writing our own implementation.

vmcj: I've added a wrapper which handles most (if not all?) of the buggy behaviours of the `getopt` function.

judge/judgedaemon.main.php
```diff
@@ -16,6 +16,72 @@
 $endpoints = [];
 $domjudge_config = [];
 
+function dj_getopt(string $short_options, array $long_options = []): array
+{
+    global $argv;
+    define('GETOPT_REGEX', "/^([a-zA-Z0-9]:{0,2})*$/");
+    $options = getopt($short_options, $long_options);
+    if (preg_match(GETOPT_REGEX, $short_options) !== 1) {
```
meisterT, line 24: You probably want to do the check in line 24 before the `getopt` call in line 23.

judge/judgedaemon.main.php
```diff
@@ -16,6 +16,72 @@
...
+        echo "Error: short options format specified is invalid.\n";
+        usage();
+    }
+    if ($options===false || !is_array($argv)) {
+        echo "Error: parsing options failed.\nPlease check: `register_argc_arg` in php.ini.\n";
+        usage();
+    }
+    check_unknown_cli_options($options);
```
meisterT, line 32: I think this function can be simplified (and then perhaps inlined?) to something like:
```
foreach (array_slice($argv, 1) as $arg) {
  if (str_starts_with($arg, '-') && !array_key_exists(ltrim($arg, '-'), $options)) {
    echo "Error: Unknown option: $arg\n";
    usage();
  }
}
```

meisterT (commented):
This is just reviewing commit 936fc63fa262ff610826a35a7d0c0a9fd4de6ced - I think it would be good to split this PR into multiple smaller ones that are independent.

judge/judgedaemon.main.php
vmcj: Done.

judge/judgedaemon.main.php
```diff
@@ -16,6 +16,32 @@
...
+    if (preg_match(GETOPT_REGEX, $short_options) !== 1) {
+        echo "Error: short options format specified is invalid.\n";
+        usage();
+    }
+    $options = getopt($short_options, $long_options);
+    if ($options===false || !is_array($argv)) {
+        echo "Error: parsing options failed.\nPlease check: `register_argc_arg` in php.ini.\n";
+        usage();
+    }
+    $unknown = false;
```
eldering, line 32: I think this would be clearer:
```suggestion
    $unknown_option = false;
```

judge/judgedaemon.main.php
```diff
@@ -16,6 +16,32 @@
...
+    foreach (array_slice($argv, 1) as $arg) {
+        if (str_starts_with($arg, '-') && !array_key_exists(ltrim($arg, '-'), $options)) {
+            echo "Error: Unknown option: $arg\n";
```
eldering, line 35: This won't work if you pass multiple options like `-df`.

judge/judgedaemon.main.php
```diff
@@ -16,6 +16,52 @@
...
+    foreach (array_slice($argv, 1) as $arg) {
+        if ($arg === '--') {
+            // By convention we can stop processing arguments.
+            break;
+        }
+        if (str_starts_with($arg, '--')) {
+            foreach($long_options as $long_option) {
+                $stripped_arg = substr($arg, 2);
+                if (str_starts_with($long_option, $stripped_arg) && $stripped_arg !== $long_option) {
+                    echo "Error: Shorten long options is not supported: $arg\n";
```
eldering, line 42:
```suggestion
                    echo "Error: shortening long options is not supported: $arg\n";
```

judge/judgedaemon.main.php
```diff
@@ -16,6 +16,52 @@
...
+                    $unknown_option = true;
+                } elseif (!array_key_exists($stripped_arg, $options)) {
```
eldering, line 44: I'm not sure this will work with something like `--my-long-option=foo`.

vmcj:
This should be fixed in PHP itself.

PR #2749, opened 2024-10-21
Add warning about cgroup v1 force flag

nickygerritsen:
Systemd 257 (currently used on arch at least) stops users from booting if you don't set this flag. Although cgroup v2 support is on its way, we might want to warn current users.

### DIFF
--- a/doc/manual/install-judgehost.rst
+++ b/doc/manual/install-judgehost.rst
@@ -172,7 +172,8 @@ any other tasks on the same CPU core the judgedaemon is using:
 
 On modern distros (e.g. Debian bullseye and Ubuntu Jammy Jellyfish) which have
 cgroup v2 enabled by default, you need to add ``systemd.unified_cgroup_hierarchy=0``
-as well. Then run ``update-grub`` and reboot.
+as well. If you are running systemd v257, you also need to add `SYSTEMD_CGROUP_ENABLE_LEGACY_FORCE=1`.
+Then run ``update-grub`` and reboot.
 After rebooting check that ``/proc/cmdline`` actually contains the
 added kernel options. On VM hosting providers such as Google Cloud or
 DigitalOcean, ``GRUB_CMDLINE_LINUX_DEFAULT`` may be overwritten


vmcj (approved):
Let's also do it for 8.2?

doc/manual/install-judgehost.rst
```diff
@@ -172,7 +172,8 @@ any other tasks on the same CPU core the judgedaemon is using:
 
 On modern distros (e.g. Debian bullseye and Ubuntu Jammy Jellyfish) which have
 cgroup v2 enabled by default, you need to add ``systemd.unified_cgroup_hierarchy=0``
-as well. Then run ``update-grub`` and reboot.
+as well. If you are running systemd v257, you also need to add `SYSTEMD_CGROUP_ENABLE_LEGACY_FORCE=1`.
```
meisterT, line 175: `v257 or higher` I assume?

nickygerritsen: They say they will drop it fully in 248, so no

mpsijm: I'm currently running v256 and I can't run with cgroupsv1 at all.

If I remember correctly, only v255 requires this extra command-line parameter, and in v256, support was dropped entirely, so the flag is no longer needed there.

nickygerritsen: https://github.com/systemd/systemd/blob/main/NEWS seems to suggest that for `257`:

>         * Support for cgroup v1 ('legacy' and 'hybrid' hierarchies) is now
>           considered obsolete and systemd by default will refuse to boot under
>           it. To forcibly reenable cgroup v1 support,
>           SYSTEMD_CGROUP_ENABLE_LEGACY_FORCE=1 must be set on kernel command
>           line. The complete removal of cgroup v1 is scheduled for v258.

You are saying this is wrong?

mpsijm: Oh, never mind, I misremembered :slightly_smiling_face: Looking back at https://github.com/DOMjudge/domjudge/issues/1072#issuecomment-2191908883, enabling cgroup v1 in v256 makes it _harder_ to boot (i.e. they've added a 30 second timeout) :sweat_smile: 

On the other hand, @RagnarGrootKoerkamp couldn't boot his Arch at all anymore (https://github.com/DOMjudge/domjudge/issues/1072#issuecomment-2213535688). I assume he also ran v256 back then, though (because v257 is not released yet) :thinking:

PR #2750, opened 2024-10-22
Cherry-pick to fix CI

vmcj:
If CI becomes green I'll merge this.


vmcj:
The failure of the other test will be handled in another PR.

Issue #2481, opened 2024-04-18
Add configuration option to make disallow answering clarifications without claiming


meisterT:
Currently claiming a clarification is purely decorative and doesn't enforce anything. The jury at the finals requested to make it actually impossible to reply to a clarification which is claimed by another jury member. 

This could be a configuration setting.

thijskh:
Is the issue that people replied to things they didn't notice were claimed by someone else? Would it already be helpful to have a warning at the reply box: "You are now replying to a claification claimed by X"?

meisterT:
Multiple jury members answered the same clarification. So a dialog like you propose that checks for whether the question has been claimed or answered sounds useful.

as6325400:
It seems like this issue hasn't been resolved yet?

I think we could add a 'strong claim' option.

This option would prevent other jury members from responding.

And for regular claims, we could add an additional alert notification, just like mentioned above."

eldering:
I don't think a "strong claim" option is great, since we always allow someone to respond to a clarification, even if a response has already been sent. So blocking that would require adding extra complexity.I think @thijskh's suggestion would be good enough to prevent people from answering inadvertently, both if the clarification was claimed or replied to in the mean time.

PR #2752, opened 2024-10-23
Add a warning log when clarifications are claimed.

as6325400:
## issue
https://github.com/DOMjudge/domjudge/issues/2481

## display
截圖 2024-10-23 晚上11 34 24


Kevinjil (changes requested):
To be effective in an actual contest situation, this check must happen on the backend.

as6325400:
I want to try making a GET request to the backend when pressing a button to check if anyone other than myself has claimed it. What do you think of this solution?

PR #2761, opened 2024-10-26
Always show MySQL variables, also if check is ok

eldering:
This helps quickly inspecting settings, in my case for example the MySQL transaction isolation level.
We should probably do this for all configuration settings.

### DIFF
--- a/webapp/src/Service/CheckConfigService.php
+++ b/webapp/src/Service/CheckConfigService.php
@@ -219,37 +219,43 @@ class CheckConfigService
         $max_inout = max($max_inout, $output_limit);
 
         $result = 'O';
-        $desc = '';
+        $desc = sprintf("max_connections is set to %s.\n", $vars['max_connections']);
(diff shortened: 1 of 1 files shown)


webapp/src/Service/CheckConfigService.php
```diff
@@ -222,34 +222,42 @@ public function checkMysqlSettings(): ConfigCheckItem
         $desc = '';
         if ($vars['max_connections'] < 300) {
```
meisterT, line 223: I would write these to reduce duplication:
```
$desc .= sprintf("max_connections is set to %s.\n", $vars['max_connections']);
if ($vars['max_connections'] < 300) {
            $result = 'W';
            $desc .= sprintf("In our experience you need at least 300, but better 1000 connections to prevent connection refusal during the contest.\n", $vars['max_connections']);
}
```

eldering: Changed.

PR #2777, opened 2024-11-03
Make output of import contest more readable.

### LINKED ISSUE #2342: import-contest should create an upload log
When you upload more problems than fit in your terminal window it is hard to check if you got any danger/warning messages. I would like it if we log all choices made and the returned warnings/errors which were printed (or alternative just all interactions)

(I've updated the requirement)

meisterT:
Print the output of `zip` line by line, overwriting the last line and removing all output of it on success.

Also pretty-print the JSON response on problem output.

That makes it much easier to see warnings or errors.

Fixes #2342.

### DIFF
--- a/misc-tools/import-contest.in
+++ b/misc-tools/import-contest.in
@@ -15,12 +15,13 @@ Part of the DOMjudge Programming Contest Jury System and licensed
 under the GNU GPL. See README and COPYING for details.
 '''
 
-import json
 from os import listdir
+from typing import List
+import json
 import os.path
 import re
+import subprocess
 import sys
-from typing import List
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

vmcj:
To make sure I understand it correctly, can you send a demonstration (gif/movie)?

meisterT:
https://asciinema.org/a/BhtgH3Sj2x4m9eWPlFGMZwDSW

vmcj (approved):
This is indeed already a huge improvement. I'll see with my next contest if this is enough, you didn't have any errors here right?

PR #2781, opened 2024-11-06
Update Symfony after security release

nickygerritsen:
Symfony just released new patch versions fixing some security issues. Better update asap.


nickygerritsen:
I don't know why integration fails all of a sudden on the `no-output` cases

nickygerritsen:
> > I don't know why integration fails all of a sudden on the `no-output` cases
> 
> I think the entry point detection fails, but I've seen that before so this might actually be random.

Then let's do https://github.com/DOMjudge/domjudge/pull/2788

Kevinjil (approved):

PR #2789, opened 2024-11-06
Make shadowing a bit more robust

nickygerritsen:
All these were found while shadowing PC^2.

### DIFF
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -1027,8 +1027,11 @@ class ExternalContestSourceService
 
         $toCheckProblem = [
             'name'      => $data->name,
-            'timelimit' => $data->timeLimit,
         ];
+
+        if ($data->timeLimit) {
+            $toCheckProblem['timelimit'] = $data->timeLimit;
+        }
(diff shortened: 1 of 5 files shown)


Kevinjil (approved):

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -1097,7 +1100,10 @@ protected function validateAndUpdateTeam(Event $event, EventData $data): void
         if (!empty($data->organizationId)) {
             $affiliation = $this->em->getRepository(TeamAffiliation::class)->findOneBy(['externalid' => $data->organizationId]);
             if (!$affiliation) {
+                // Affiliation does not exist. Create one with a dummy name so we can continue.
                 $affiliation = new TeamAffiliation();
+                $affiliation->setName($data->organizationId);
+                $affiliation->setShortname($data->organizationId);
```
vmcj, line 1106: Can externalid be more chars than the shortname?

nickygerritsen: It seems it can. We need to limit it to 32 characters so I will take the first 32.

PR #2802, opened 2024-11-16
Add quotes for some env vars.

meisterT:


nickygerritsen (approved):

judge/compile.sh
```diff
@@ -139,10 +139,10 @@ logmsg $LOG_INFO "starting compile"
 
 # shellcheck disable=SC2236
 if [ ! -z "$ENTRY_POINT" ]; then
-	ENVIRONMENT_VARS="-V ENTRY_POINT=$ENTRY_POINT"
+	ENVIRONMENT_VARS="-V ENTRY_POINT=\"$ENTRY_POINT\""
```
vmcj, line 142: I think we can use single-quotes? I like that more than escaping.

meisterT:
Ah, this is annoying and it doesn't work. We should re-think our stance on POSIX vs bash. That would simplify things all around.

vmcj:
> Ah, this is annoying and it doesn't work. We should re-think our stance on POSIX vs bash. That would simplify things all around.

Last hackathon the consensus was that we can use `bash` on the judgehost but try to keep POSIX compliant scripts on the domserver.

PR #2804, opened 2024-11-17
runguard: Allow passing `-V` multiple times.

meisterT:
The previous behavior was that the last arg wins. IMHO, this is confusing and we even fell into this trap ourselves in https://github.com/DOMjudge/domjudge/blame/main/judge/compile.sh#L145 where we would shadow any previously passed env var (such as the entry point).


judge/runguard.cc
```diff
@@ -363,7 +365,8 @@ Run COMMAND with restrictions.\n\
   -s, --streamsize=SIZE  truncate COMMAND stdout/stderr streams at SIZE kB\n\
   -E, --environment      preserve environment variables (default only PATH)\n\
   -V, --variable         add additional environment variables\n\
-                           (in form KEY=VALUE;KEY2=VALUE2)\n\
+                           (in form KEY=VALUE;KEY2=VALUE2); may be passed\n\
+			   multiple times\n\
```
eldering, line 369: Indentation is off here because of tabs.

judge/runguard.cc
```diff
@@ -1175,7 +1178,7 @@ int main(int argc, char **argv)
 			preserve_environment = 1;
 			break;
 		case 'V': /* set environment variable */
-			environment_variables = strdup(optarg);
+			environment_variables.push_back(optarg);
```
eldering, line 1181: Nitpick: maybe make the implicit `string()` cast/constructor that creates a copy explicit?

judge/runguard.cc
```diff
@@ -802,8 +805,8 @@ void setrestrictions()
 	}
 
 	/* Set additional environment variables. */
-	if (environment_variables != nullptr) {
-		char *token = strtok(environment_variables, ";");
+	for (const auto &env_var : environment_variables) {
+		char *token = strtok(strdup(env_var.c_str()), ";");
```
eldering, line 809: Nitpick: `env_var` can actually contain multiple variables separated by `;`, so maybe rename to something like
```suggestion
	for (const auto &tokens : environment_variables) {
		char *token = strtok(strdup(tokens.c_str()), ";");
```

eldering (commented):
Overall LGTM, makes writing scripts to call this easier (even when using bash).

judge/runguard.cc
meisterT: Done

judge/runguard.cc
meisterT: Done

judge/runguard.cc
meisterT: Done.

eldering (approved):

Issue #2525, opened 2024-05-02
Redesign to support other scoring methods


eldering:
In the past there have been various requests to support other scoring methods than pass/fail with penalty time, some have been listed in https://github.com/DOMjudge/domjudge/labels/scoring

Before we try to support any of these, we should consider what a common (internal) interface/data structures would look like to generically support these. This would include making a trade-off where the more weird scoring methods we'd support, the broader and less well-defined this interface would become. This ticket is to discuss that interface.

We should then also refactor our code so that all scoring related code is hidden behind implementations of this interface. Currently there is plenty of code scattered around the code base.

eldering:
Some comments/questions for this:

1. I think we should not allow a scoring method to take arbitrary data from the system to determine the score (like: the current CPU load on the server, to name something silly). So what should be the set of allowed input to a score for a single submission? For team team on a problem? For a team in the whole contest?
2. Related, what is the data format for the score for a submission, problem, etc? A float, or can it be any more complex structure like, `{"points": 123, "time": 6534.061, "penalty": 20, "foobar": [1, 2, 3]}`?

meisterT:
Some thoughts here: 

- We probably want to try to implement the current score in seconds and runtime as tie breaker using whatever new mechanism we design.
- We need to check what the implications on score cache and rank cache are.
- Partial scoring seems to be then the easiest to add.
- As data format I'm thinking of a list of (named) floats, it should be doable to represent any scoring function with that. Then this list of floats can be used for sorting/ranking teams (that implies it's maximized and you need to negate one of the floats if you want to minimize). In addition, we probably want to add a text representation (e.g. for the scoreboard).

MathisHammel:
Hi!
I would be interested in financially supporting the development of such a feature, mainly to include optimization problems in the small contests we run. Let me know if we can make this happen, contact details are on my GitHub profile if needed.

PR #2812, opened 2024-11-21
Balloon list: only add 'first in contest' if there is no earlier unju…

### LINKED ISSUE #2810: Balloon list may have problem on awarding the "first to solve"

I'm intersted in the logic of balloon.

I found that when the submission judged as Accepted, the balloon will be created immediately. https://github.com/DOMjudge/domjudge/blob/7c890e8ddf0ed13c6164e091d903408736f465a8/webapp/src/Controller/API/JudgehostController.php#L1078

But if we have two submission on the same problem (they are the eariliest submissions of the problem) , maybe we call them A and B, and the submit time of A is earlier than B, and A's judge finish time is later, the balloon of submission A will be created after submission B. But when we looking at the balloon list, we temporally found that the balloon of submission B is remindered as the first balloon.
https://github.com/DOMjudge/domjudge/blob/7c890e8ddf0ed13c6164e091d903408736f465a8/webapp/src/Service/BalloonService.php#L92  
In the collectBalloonTable function, we can see that award is determined by the balloon list

The above is my understanding of the balloon logic.

I don't know whether my worry is right or not. If it is, I think it is a bug or if you can tell me how can I temporaly solve this problem (maybe we should send balloons only if there submit time is much earlier than now to avaid eariler submission is not judged yet ?)

### Steps to reproduce

Create a problem with a long time limit. 
Team A submit a code correct but run very slow first. 
Team B submit a code correct and run fast later. 

### Expected behaviour

It's obvious Team A is the first to solve.

### Actual behaviour

When Team B's submission is judged as correct but Team A's submission is still judging, we can found the balloon list maked Team B is the first to solve the problem. 

Sorry that currently I have no machine to test it. But according to my understing, I think this bug truely happens.

meisterT:
…dged submission.

The 'first for problem' is already working correctly as it re-uses the corresponding scoreboard data.

Fixes #2810.

Example submission list:
image

Corresponding balloon list:
image

Previously we would have awarded the first balloon a "first in contest".

### DIFF
--- a/webapp/src/Service/BalloonService.php
+++ b/webapp/src/Service/BalloonService.php
@@ -141,7 +141,25 @@ class BalloonService
             }
             // Keep overwriting this - in the end it'll
             // contain the ID of the first balloon in this contest.
+            $submittime = $balloonsData[0]->getSubmission()->getSubmittime();
             $AWARD_BALLOONS['contest'] = $balloonsData[0]->getBalloonId();
+        }
+        $countBefore = $em->createQueryBuilder()
+            ->from(Judging::class, 'j')
+            ->leftJoin('j.submission', 's')
+            ->select('COUNT(s.submitid)')
+            ->where('s.contest = :cid')
+            ->andWhere('s.submittime < :submittime')
+            ->andWhere('s.valid = 1')
+            ->andWhere('j.valid = 1')
+            ->andWhere('j.result IS NULL')
+            ->setParameter('cid', $contest->getCid())
+            ->setParameter('submittime', $submittime)
+            ->getQuery()->getSingleScalarResult();
+        if ($countBefore > 0) {
+            // We don't have complete information about 'first in contest' yet as there are unjudged submissions that
+            // could influence the result.
+            $AWARD_BALLOONS['contest'] = -1;
         }
 
         // Loop again to construct table.


nickygerritsen (approved):

eldering:
How will this deal with the scenario where team A submits, then team B submits, B gets judged as correct, and then A gets judged as wrong answer?

eldering:
> How will this deal with the scenario where team A submits, then team B submits, B gets judged as correct, and then A gets judged as wrong answer?
> 
>     1. Will the balloon already be marked for team B even when A's submission is still pending?
> 
>     2. I so, will B's balloon retro-actively be updated to be marked as `first in contest`?

I think the ideal scenario is that a balloon will only be added once, never updated, and so a balloon should only be added after it is known that it's contents (including first in contest/for problem) won't change anymore.
Otherwise there's a high risk that users of the balloon interface hand out the wrong balloons.

meisterT:
In the same way as on the scoreboard as of today. We mark it as solved, but not first to solve while we have incomplete information. Then we have more information and you revisit that page later it gets marked as first to solve.

We could decide to hold the balloon back or mark it as potentially first to solve. Both things are more complex than this but of course can be done. Since this is fixing a clear bug and not making anything worse, I think we should address that separately.

meisterT:
@nickygerritsen how is this btw handled in ICPC tools?

nickygerritsen:
> @nickygerritsen how is this btw handled in ICPC tools?

🤷 I can check the code to see if I can determine this, but I think it's easier there since it is a long running process that has state.

eldering:
> In the same way as on the scoreboard as of today. We mark it as solved, but not first to solve while we have incomplete information. Then we have more information and you revisit that page later it gets marked as first to solve.
> 
> We could decide to hold the balloon back or mark it as potentially first to solve. Both things are more complex than this but of course can be done. Since this is fixing a clear bug and not making anything worse, I think we should address that separately.

I'm a bit wary of various such fixes we've been making over time that all patch a bug by doing their own DB querying, leading to inconsistent logic and making it harder to eventually implement other scoring strategies as we'd have to refactor each of these.

But ok, merge it. Also tagging #2525 for visibility.

meisterT:
That's a fair point with the DB query. We currently don't have first in contest in the award service, otherwise I would have checked whether we can reuse that. But this doesn't address the other point that you are making about immutability. Let's discuss in person before merging anything.

nickygerritsen:
Will be discussed onsite in Baku

PR #2822, opened 2024-11-22
Python code style and typing

Kevinjil:


submit/submit
```diff
@@ -364,7 +377,8 @@ parser.add_argument('-c', '--contest', help='''submit for contest with ID or sho
 parser.add_argument('-p', '--problem', help='submit for problem with ID or label PROBLEM', default='')
 parser.add_argument('-l', '--language', help='submit in language with ID LANGUAGE', default='')
 parser.add_argument('-e', '--entry_point', help='set an explicit entry_point, e.g. the java main class')
-parser.add_argument('-v', '--verbose', help='increase verbosity', choices=loglevels.keys(), nargs='?', const='INFO', default='WARNING')
+parser.add_argument('-v', '--verbose',
+                    help='increase verbosity', choices=loglevels.keys(), nargs='?', const='INFO', default='WARNING')
```
eldering, line 381: I'd argue that we should not break the line here: keeping it consistently on a single long line is better for readability than breaking it because it's longer than 120 characters.

submit/submit
```diff
@@ -248,7 +259,7 @@ drop-down box in the web interface.'''
         language_part = 'For LANGUAGE use the ID or a common extension.'
     else:
         language_part = 'For LANGUAGE use one of the following IDs or extensions:'
-        max_length = max([len(l['name']) for l in languages])
+        max_length = max([len(L['name']) for L in languages])
```
eldering, line 262: Let's not use single capital letters for variables. If you think that a single lowercase `l` is confusing, then let's rename it to `lang` or so.

submit/submit
Kevinjil: I'll check how easy it is to make CI ignore this

misc-tools/dj_utils.py
```diff
@@ -79,15 +79,17 @@ def do_api_request(name: str, method: str = 'GET', jsonData: dict = {}):
         url = f'{domjudge_webapp_folder_or_api_url}/{name}'
         parsed = urlparse(domjudge_webapp_folder_or_api_url)
         auth = None
-        if parsed.username or parsed.password:
+        if parsed.username and parsed.password:
```
eldering, line 82: This looks like an actual bugfix, could you put that in a separate commit?

eldering (commented):
Other than single comment, LGTM

PR #2825, opened 2024-11-22
Last gitlab ci

vmcj:
I'll squash the commits about the unit-tests after approval. Split them out to keep the flow understandable.


webapp/src/Form/Type/RejudgingType.php
```diff
@@ -25,7 +25,7 @@
 
 class RejudgingType extends AbstractType
 {
-    public function __construct(protected readonly DOMJudgeService $dj, protected readonly EntityManagerInterface $em)
+    public function __construct(protected readonly ConfigurationService $config, protected readonly EntityManagerInterface $em)
```
nickygerritsen, line 28: This commit needs to be dropped but you are probably already aware.

nickygerritsen (approved):

vmcj:
Also added the unit tests as required for merge queues

PR #2826, opened 2024-11-22
Allow to ❤️ on mobile scoreboard

### LINKED ISSUE #2762: Add ability to favorite / heart teams in the mobile view
Currently the mobile view does not support the ability to favorite teams.

Align mobile view with desktop view.

### Expected behaviour
1. Go to the scoreboard with mobile view
2. see the heart symbol
3. have it work
4. ???
5. profit

Desktop:
image

Mobile:
image

nickygerritsen:
Fixes #2762.


meisterT (approved):

webapp/public/js/domjudge.js
```diff
@@ -297,69 +325,99 @@ function toggle(id, show)
...
+                    }
+                    if (rank.trim().length === 0) {
+                        // make rank explicit in case of tie
+                        getRank(toAdd[cntFound]).innerHTML += lastRank;
+                    }
+                    scoreboard[j].style.background = "lightyellow";
+                    const scoretn = scoreboard[j].querySelector('.scoretn');
+                    if (scoretn && scoretn.classList.contains('cl_FFFFFF')) {
+                        scoretn.classList.remove('cl_FFFFFF');
+                        scoretn.classList.add('cl_FFFFE0');
```
mpsijm, line 376: Well, I guess that's one way to fix the white colour :joy: 

Is the extra `cl_FFFFE0` class needed, when the wrapping `<div>` will receive the background colour `lightyellow` anyway? :slightly_smiling_face:

nickygerritsen: Yes, that cl_ thing is there because if you set a fixed team column width, we do a 'fade out'. For that to work we need an actual background colour always. So I'm just switching it out. Not the most beautiful but it works.

webapp/templates/partials/scoreboard_table.html.twig
```diff
@@ -106,7 +106,7 @@
     <tbody>
     {% set previousSortOrder = -1 %}
     {% set previousTeam = null %}
-    {% set backgroundColors = {"#FFFFFF": 1} %}
+    {% set backgroundColors = {"#FFFFFF": 1, '#FFFFE0': 1} %}
```
mpsijm, line 109: What's this for? Is it still needed after resolving the previous comment? (also on line 359)

nickygerritsen: I do think we don't have to do the loop twice though. let me fix that

PR #2838, opened 2024-11-23
Replace Ace with Monaco editor

### LINKED ISSUE #2689: Consider replacing ace editor with monaco

nickygerritsen:
Fixes #2689.

Beautiful screenshots:
monaco
image


webapp/src/Controller/Jury/ExecutableController.php
```diff
@@ -523,28 +523,28 @@ protected function dataForEditor(Executable $executable): array
...
             'executableBits' => $executableBits,
         ];
     }
 
-    private function getAceFilename(string $filename, string $content): string
+    private function getEditorFilename(string $filename, string $content): string
     {
         if (!str_contains($filename, '.')) {
             // If the file does not contain a dot, see if we have a shebang which we can use as filename.
             // We do this to hint the ACE editor to use a specific language.
```
meisterT, line 544: nit: no longer ace

webapp/templates/jury/partials/submission_diff.html.twig
```diff
@@ -34,7 +34,7 @@
 
         <div class="tab-pane fade {% if loop.first %}show active{% endif %}" id="diff-{{ filePair.1.submitfileid }}"
              role="tabpanel">
-            <pre class="output_text">{{ filePair.0 | showDiff(filePair.1) }}</pre>
+            {{ showDiff("diff" ~  filePair.1.submitfileid,filePair.0,filePair.1) }}
```
meisterT, line 37: super nit: add a space after commas

meisterT (commented):
also missing: readme mentions ace 

Perhaps do a `git grep ace`?

webapp/templates/jury/base.html.twig
```diff
@@ -21,6 +21,12 @@
     {% include 'jury/menu.html.twig' %}
 {% endblock %}
 {% block footer %}
+    <script src="{{ asset('js/monaco/vs/loader.js') }}"></script>
+    <script>
+        {# A bit ugly to do the replace, but Symfony adds a version to  #}
```
meisterT, line 26: adds a version to what?

meisterT (approved):

PR #2840, opened 2024-11-23
Warn jury if clarification is claimed concurrently

### LINKED ISSUE #2481: Add configuration option to make disallow answering clarifications without claiming

Kevinjil:
This is an alternative implementation of #2766 which performs the validation in the same request as the form submission by using a Symphony form callback.

fixes #2481
closes #2766

### DIFF
--- a/webapp/templates/jury/partials/clarification_form.html.twig
+++ b/webapp/templates/jury/partials/clarification_form.html.twig
@@ -10,6 +10,7 @@
         </div>
     </div>
 </div>
+{{ form_errors(form) }}
 <div class="row">
     <div class="col-sm">
         <div class="mb-3">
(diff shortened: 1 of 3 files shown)


webapp/templates/jury/partials/clarification_form.html.twig
```diff
@@ -32,4 +33,7 @@
     </div>
 
 </div>
+{% if jurymember is defined %}
+    <input type="hidden" name="{{ field_name(form.jurymember) }}" value="{{jurymember}}">
```
nickygerritsen, line 37:
```suggestion
    <input type="hidden" name="{{ field_name(form.jurymember) }}" value="{{ jurymember }}">
```

nickygerritsen (approved):

PR #2841, opened 2024-11-23
Respond in JSON for URL's below API if 404

### LINKED ISSUE #2685: Give API error responses in JSON format
Currently, a 404 on an API endpoints gives a HTML page back. This is both wasteful in rendering and annoying to parse on the client / CLI.

For endpoints which normally return JSON, also return JSON in the non-happy flow.

nickygerritsen:
Fixes #2685.

### DIFF
--- a/webapp/src/EventListener/RespondInJsonForApiListener.php
+++ b/webapp/src/EventListener/RespondInJsonForApiListener.php
@@ -1,0 +1,27 @@
+<?php declare(strict_types=1);
+
+namespace App\EventListener;
+
+use Symfony\Component\EventDispatcher\Attribute\AsEventListener;
+use Symfony\Component\HttpKernel\Event\ControllerEvent;
+
+#[AsEventListener]
+class RespondInJsonForApiListener
+{
+    public function __invoke(ControllerEvent $event): void
+    {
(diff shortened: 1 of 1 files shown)


Kevinjil (approved):

webapp/src/EventListener/RespondInJsonForApiListener.php
```diff
@@ -0,0 +1,27 @@
...
+    {
+        // For API requests, if no accept header has been set, or it is text/html or */*,
+        // set it to application/json.
+
+        $request = $event->getRequest();
+        if ($request->attributes->get('_fos_rest_zone')) {
+            $acceptHeader = $request->headers->get('accept');
+
+            if (!$acceptHeader
+                || str_starts_with($acceptHeader, 'text/html')
```
vmcj, line 21: Is this allowed by the specs? I think if someone explicit requests `text/html` only we should still return `text/html`.
Can we do the alternative of when the acceptHeader contains `app/json` or is `*/*` set it to `app/json` only?

nickygerritsen: To me it's annoying if you open a wrong api in your browser you don't get json. 
Seeks an n=1 test with GitHub also returns json: https://api.github.com/bla

PR #2843, opened 2024-11-23
Remove transaction from fetch work API.

meisterT:
We have seen the transaction to fail, resulting in exceptions/500s.

There is also no need to have a transaction at all. We now do check after the update whether we won instead and if not, tell the judgehost to try again.


tuupke (approved):
LGTM

PR #2847, opened 2024-11-23
Print files as team using the API

Kevinjil:
This includes:
- [x] sharing some code between controllers
- [x] the API endpoint
- [ ] the submit client for workstations


webapp/src/Controller/API/PrintController.php
```diff
@@ -0,0 +1,143 @@
...
+
+        if (file_put_contents($tempFilename, $contents) === false) {
+            throw new ServiceUnavailableHttpException(
+                null,
+                sprintf("Could not write printfile to temporary file '%s'.", $tempFilename)
+            );
+        }
+
+        // Handle validation errors
+        if (!empty($errors)) {
```
nickygerritsen, line 123: There can only be one error it seems (line 108). Should we just return there?

Kevinjil: We could do that. I felt like following the same structure as other places might be nice.

nickygerritsen (approved):

Kevinjil:
As the language is not required in the web UI, I also made it optional in the API.

PR #2850, opened 2024-11-24
Add ? to nullable non-mixed properties

nickygerritsen:


webapp/tests/Unit/Controller/API/BaseTestCase.php
```diff
@@ -87,7 +87,7 @@ protected function verifyApiResponse(
         int $status,
         ?string $user = null,
         mixed $jsonData = null,
-        array $files = [],
+        ?array $files = [],
```
eldering, line 90: Is this meant to be nullable?

webapp/src/Service/ImportExportService.php
```diff
@@ -327,7 +327,7 @@ public function importContestData(mixed $data, ?string &$errorMessage = null, st
      * @param string[]|null $ids
      * @param array<string, string[]> $messages
      */
-    public function importProblemsData(Contest $contest, array $problems, array &$ids = null, ?array &$messages = []): bool
+    public function importProblemsData(Contest $contest, array $problems, ?array &$ids = null, ?array &$messages = []): bool
```
eldering, line 330: I guess the `?` should be removed from `&$messages` here.

webapp/tests/Unit/Controller/API/BaseTestCase.php
nickygerritsen: Nope, fixed.

PR #2854, opened 2024-11-24
Shadow mode related fixes

### LINKED ISSUE #2779: Shadow differences table: do not show submissions that don't have an external ID
The shadow differences table and submission list shows submissions that don't have an external ID, i.e. jury submissions we import.
We should not display those.

### LINKED ISSUE #2791: Shadowing: loading pending events does not decode event
[Here](https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Service/ExternalContestSourceService.php#L1881-L1881) we set the event on a pending warning, but it gets JSON encoded.

However, [here](https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Service/ExternalContestSourceService.php#L1881-L1881) it doesn't get unencoded, so the code crashes.

nickygerritsen:
Fixes #2791
Fixes #2779

### DIFF
--- a/webapp/src/Service/SubmissionService.php
+++ b/webapp/src/Service/SubmissionService.php
@@ -89,7 +89,9 @@ class SubmissionService
         }
 
         if ($restrictions->withExternalId ?? false) {
-            $queryBuilder->andWhere('s.externalid IS NOT NULL');
+            $queryBuilder
+                ->andWhere('s.externalid IS NOT NULL')
+                ->andWhere('s.expected_results IS NULL');
         }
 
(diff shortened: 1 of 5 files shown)


webapp/src/Service/SubmissionService.php
```diff
@@ -145,6 +147,8 @@ public function getSubmissionList(
             }
         }
 
+        dump($restrictions);
```
meisterT, line 150: leftover dump

meisterT (approved):

vmcj (approved):

PR #2856, opened 2024-11-24
Create verdict groups that selectively can be used

eldering:
Instead of manually adding various verdicts to the list in different places. This is not completely equivalent to before, but the changes should be fine. The original behaviour of calling `getVerdicts` without arguments is unchanged and `getVerdicts(mergeExternal: true)` maps to `getVerdicts(['final', 'external'])`.

### DIFF
--- a/webapp/src/Service/DOMJudgeService.php
+++ b/webapp/src/Service/DOMJudgeService.php
@@ -1485,14 +1485,19 @@ class DOMJudgeService
     /**
      * @return array<string, string>
      */
-    public function getVerdicts(bool $mergeExternal = false): array
+    public function getVerdicts(array $groups = ['final']): array
     {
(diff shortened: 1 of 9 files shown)


webapp/src/Service/DOMJudgeService.php
```diff
@@ -1485,14 +1485,19 @@ public function getCompileConfig(Submission $submission): string
     /**
      * @return array<string, string>
      */
-    public function getVerdicts(bool $mergeExternal = false): array
+    public function getVerdicts(array $groups = ['final']): array
```
nickygerritsen, line 1488: If we create a (string backed?) enum for the groups, we can change it here.

Then we can add

```php
/**
 * @param VerdictGroup[] $groups
 */
```

PR #2872, opened 2024-12-03
Clean up DOMJudgeService

eldering:
Remove a bunch of stuff that doesn't have to be here and that causes unnecessary dependencies.

And fix bug in getting verdicts as reported in slack.

### DIFF
--- a/webapp/src/Service/ConfigurationService.php
+++ b/webapp/src/Service/ConfigurationService.php
@@ -396,4 +396,39 @@ EOF;
         }
         return $item;
     }
+
+    /**
+     * Returns all possible judgement (run) verdicts, both internally
+     * hardcoded and from configured judgement types. Depending on the
+     * requirements of the context, the following groups of verdicts can be
+     * requests:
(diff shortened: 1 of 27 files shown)


meisterT (approved):

webapp/src/Service/ConfigurationService.php
```diff
@@ -395,4 +395,39 @@ public function addOptions(ConfigurationSpecification $item): ConfigurationSpeci
...
+     *
+     * @return array<string, string>
+     */
+    public function getVerdicts(array $groups = ['final']): array
+    {
+        $verdictsConfig = $this->etcDir . '/verdicts.php';
+        $verdictGroups  = include $verdictsConfig;
+
+        $verdicts = [];
+        foreach( $groups as $group ) {
```
vmcj, line 421:
```suggestion
        foreach ($groups as $group) {
```

vmcj (commented):
The mistake was already in the original code but still good to keep it consistent.

nickygerritsen (approved):

PR #2874, opened 2024-12-05
Prevent import-examples to fail

vmcj:
The mariadb images seems to now allow a smaller maximum packet by default. This resulted in problems for the boolfind problem which we didn't detect as all checks would still pass correctly.

We now also capture the SQL logs to detect this easier and increase the maximum packetsize (1G).

**When this is accepted I'll move the installation of the `docker.io` to the runner image and remove that line for a small speedup.**


nickygerritsen (approved):

vmcj:
The other errors will be fixed in https://github.com/DOMjudge/domjudge/pull/2876

PR #2877, opened 2024-12-08
Move all inline style to dedicated endpoint

vmcj:
We rendered the <style> within the <body> which is not valid. By moving it to a dedicated endpoint we also save a bit of bandwith as this is content which stays static during the contest.

### DIFF
--- a/webapp/src/Controller/PublicController.php
+++ b/webapp/src/Controller/PublicController.php
@@ -100,6 +100,14 @@ class PublicController extends BaseController
         return $this->dj->getScoreboardZip($request, $requestStack, $contest, $this->scoreboardService);
     }
 
+    #[Route(path: '/dynamic-css', name: 'get_dynamic_css')]
+    public function dynamicCSS(): Response {
+        $response = new Response();
+        $response->headers->set('Content-Type', 'text/css');
+        $response->setContent($this->renderView('public/dynamic.css.twig', $this->dj->getDynamicCSS()));
+	return $response;
+    }
+
     /**
      * Get the contest from the request, if any
      */
--- a/.github/jobs/webstandard.sh
+++ b/.github/jobs/webstandard.sh
@@ -113,7 +113,7 @@ if [ "$TEST" = "w3cval" ]; then
     unzip -q vnu.linux.zip
     section_end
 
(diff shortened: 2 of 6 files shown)


webapp/src/Controller/PublicController.php
```diff
@@ -100,6 +100,14 @@ public function scoreboardDataZipAction(
         return $this->dj->getScoreboardZip($request, $requestStack, $contest, $this->scoreboardService);
     }
 
+    #[Route(path: '/dynamic-css', name: 'get_dynamic_css')]
+    public function dynamicCSS(): Response {
+        $response = new Response();
+        $response->headers->set('Content-Type', 'text/css');
```
nickygerritsen, line 106: We should maybe set some expire headers? Otherwise browsers won't cache it.
Drawback is that it will change once the colors change, so maybe use an etag header instead?

webapp/src/Controller/PublicController.php
```diff
@@ -100,6 +100,14 @@ public function scoreboardDataZipAction(
         return $this->dj->getScoreboardZip($request, $requestStack, $contest, $this->scoreboardService);
     }
 
+    #[Route(path: '/dynamic-css', name: 'get_dynamic_css')]
```
nickygerritsen, line 103: The name is too broad. It should be something like `scoreboardCategoryColorCss` or something.

vmcj: And rename if there are more?

nickygerritsen: More dynamic CSS? I would put them in separate actions, since otherwise you get one big if here, which doesn't make sense IMHO.

webapp/src/Controller/PublicController.php
vmcj: We discussed and I'll try to implement a cache and keep the etag there.

webapp/src/Controller/PublicController.php
vmcj: I'll change it for now, not sure if we have more pages like this so if the discussion will actually ever be non-theoretical.

webapp/src/Controller/PublicController.php
vmcj: I wonder if the current implementation is not already enough, we save on the bandwidth but still need to calculate the "whole" page (as we did before).

nickygerritsen (approved):

PR #2880, opened 2024-12-13
Add a toggle to enable side-by-side diff viewer

### LINKED ISSUE #2863: [Feature request] Add an option to switch to an inline diff editor.

Now that the editor has been changed to monaco, then we can add an option to swich to inline diff editor in submission page: `jury/submissions/<sid>/source`

Add a toggle button and/or a global preference (like editor theme) to switch between side by side and inline diff editor.

The inline diff editor can be created by option `renderSideBySide: false`

image

Kevinjil:
Store the last toggled diff viewer state in the local storage. The way diffs are viewed is mostly a user preference, but depending on the file and type of changes, the option to quickly toggle it can come in handy.

Note that for multi-file submissions, a per-file toggle is included in the tab, but on reload all tabs will have the state of the last toggled switch.

Fixes #2863

### DIFF
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -124,6 +124,20 @@ function applyEditorTheme(theme = undefined, isExternal = false)
     });
 }
 
+function isDiffSideBySide()
+{
+    let sideBySide = localStorage.getItem('domjudge_editor_side_by_side');
+    if (sideBySide === undefined) {
+        return true;
+    }
+    return sideBySide == 'true';
+}
+
+function setDiffSideBySide(value)
+{
+    localStorage.setItem('domjudge_editor_side_by_side', value);
+}
+
 // Send a notification if notifications have been enabled.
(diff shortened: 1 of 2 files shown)


nickygerritsen (approved):
We could do the same as with the theme to change them all, but then the toggle should go in the menu since we only need one. I don't think that's better

webapp/public/js/domjudge.js
```diff
@@ -124,6 +124,21 @@ function applyEditorTheme(theme = undefined, isExternal = false)
...
 }
 
+function isDiffSideBySide()
+{
+    let sideBySide = localStorage.getItem('domjudge_editor_side_by_side');
+    if (sideBySide === undefined) {
+        return true;
+    } else {
+        return sideBySide == 'true';
+    }
```
vmcj, line 134:
```suggestion
    }
    return sideBySide == 'true';
```

PR #2882, opened 2024-12-15
Add pagination to submission lists

### LINKED ISSUE #2511: Add pagination to pages with long lists
When visiting: https://onlinejudge.icpc.global/jury/judgehosts/1 with more than 2000 submissions handled the page is slow, we should do a similar thing to the general submission page and only display the last `n` entries to make the pageload time consistent and not dependent on the number of submissions in the contest.

nickygerritsen:
Fixes #2511

Some notes:

- I used a well-known bundle that takes care of most of the things we want, so the changes are minimal.
- We paginate all submissions pages by default, but exclude the team interface and the prometheus metrics. The former should not have that many submissions and the latter needs all.
- The query parameter for pages is always `page`. We can't currently override this, see https://github.com/KnpLabs/KnpPaginatorBundle/issues/815.
- I changed the jury submissions filter form to a PHP/backend filter form. This requires you to press a new "Apply" button, but makes it such that filtering happens server side and thus pagination is honored. This also allowed us to get rid of quite a bit of javascript and code in the Twig (but some logic in the controller action came in return).
- I removed the "new" view type in two spots, since we don't need it anymore.
- A page contains 50 submissions, just like the "new" filter did before.
- We always show testcases for the main submissions page now, since we only show 50 submissions.
- I needed to add filters for restrictions with arrays of ID's. Those were trivial.
- I also added a filter for an array of results. That one is less trivial. I tested the "normal" results, judging and queued. import-error should also work, so I think we have all cases.

Screenshots:
image
image

### DIFF
--- a/webapp/src/Controller/Jury/SubmissionController.php
+++ b/webapp/src/Controller/Jury/SubmissionController.php
@@ -36,6 +36,7 @@ use Doctrine\ORM\Exception\ORMException;
 use Doctrine\ORM\NonUniqueResultException;
 use Doctrine\ORM\NoResultException;
 use Doctrine\ORM\Query\Expr\Join;
+use Knp\Component\Pager\Pagination\PaginationInterface;
 use Symfony\Bridge\Doctrine\Attribute\MapEntity;
 use Symfony\Bridge\Doctrine\Form\Type\EntityType;
 use Symfony\Component\Form\Extension\Core\Type\SubmitType;
@@ -79,7 +80,7 @@ class SubmissionController extends BaseController
         #[MapQueryParameter(name: 'view')]
         ?string $viewFromRequest = null,
     ): Response {
-        $viewTypes = [0 => 'newest', 1 => 'unverified', 2 => 'unjudged', 3 => 'judging', 4 => 'all'];
+        $viewTypes = [0 => 'all', 1 => 'unverified', 2 => 'unjudged', 3 => 'judging'];
         $view      = 0;
         if (($submissionViewCookie = $this->dj->getCookie('domjudge_submissionview')) &&
             isset($viewTypes[$submissionViewCookie])) {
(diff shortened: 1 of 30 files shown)


vmcj:
>     * I changed the jury submissions filter form to a PHP/backend filter form. This requires you to press a new "Apply" button, but makes it such that filtering happens server side and thus pagination is honored. This also allowed us to get rid of quite a bit of javascript and code in the Twig (but some logic in the controller action came in return).

Why can't we `POST` on `mouseclick/buttonup`?

nickygerritsen:
> > ```
> > * I changed the jury submissions filter form to a PHP/backend filter form. This requires you to press a new "Apply" button, but makes it such that filtering happens server side and thus pagination is honored. This also allowed us to get rid of quite a bit of javascript and code in the Twig (but some logic in the controller action came in return).
> > ```
> 
> Why can't we `POST` on `mouseclick/buttonup`?

We can but that will reload the page every time, which looks very ugly.

What we could do is somehow do some AJAX submit and update in the background. However, I didn't want to go there since that makes it even more complex and I don't think hitting one button is that big of an issue.

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -147,10 +200,10 @@ public function indexAction(
             'submissions' => $submissions,
             'submissionCounts' => $submissionCounts,
             'showContest' => count($contests) > 1,
-            'hasFilters' => !empty($filters),
+            'hasFilters' => $hasFilters,
             'results' => $results,
             'showExternalResult' => $this->dj->shadowMode(),
-            'showTestcases' => count($submissions) <= $latestCount,
+            'showTestcases' => true,
```
vmcj, line 206: Is this used anywhere else?

webapp/src/Form/Type/SubmissionsFilterType.php
```diff
@@ -127,7 +128,12 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             "attr" => ["data-filter-field" => "result"],
         ]);
 
-        $builder->add("clear", ButtonType::class, [
+        $builder->add("apply", SubmitType::class, [
+            "label" => "Apply filters",
+
```
vmcj, line 133:
```suggestion
```

webapp/src/Controller/Jury/SubmissionController.php
nickygerritsen: We only show it on the general submission page and not others (like team page, judgehost page, etc.). So yes.

vmcj (approved):
Nothing more to ask about it.

PR #2884, opened 2024-12-21
Fix the upload of the unit test results

vmcj:
We expect the file to be named per matrix job but this was not updated in the code so the upload action couldn't find the files.

First part to fix: https://github.com/DOMjudge/domjudge/pull/2879


vmcj:
Will also need: https://github.com/EnricoMi/publish-unit-test-result-action/blob/v2.18.0/README.md#support-fork-repositories-and-dependabot-branches

PR #2895, opened 2025-02-25
Free C-string allocated by strdup after use

eldering:
This resource leak was found by Coverity scan.


meisterT (approved):

meisterT (commented):
Wait, this is breaking related tests, but I don't immediately see what's wrong. Can have a closer look tomorrow evening

PR #2897, opened 2025-02-26
Installation fails without a Session

vmcj:
It fixes my problem, but I'm not sure if there is not an underlying problem.

As we create a request it makes sense we also need a session, but does that session also need other things? I don't know enough about Symfony to be sure that this is besides a solution for my problem also the correct fix for the issue where no session is set.

I hope @nickygerritsen has some ideas?


nickygerritsen (approved):
I think maybe some upgrade requires us to have a session. If this works it works I'd say

PR #2899, opened 2025-02-26
Replace medals legend with actual medals in the scoreboard.

meisterT:
# Before

image

image

# After

image

### DIFF
--- a/webapp/templates/partials/scoreboard_table.html.twig
+++ b/webapp/templates/partials/scoreboard_table.html.twig
@@ -135,7 +135,10 @@
         {% endif %}
         <tr class="{{ classes | join(' ') }}" data-team-id="{{ score.team.teamid }}">
             {% if enable_ranking %}
-                <td class="scorepl {{medalColor}}">
+                <td class="scorepl">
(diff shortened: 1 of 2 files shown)


nickygerritsen (approved):
Best feature ever. I like it

webapp/templates/partials/scoreboard_table.html.twig
```diff
@@ -135,7 +135,10 @@
         {% endif %}
         <tr class="{{ classes | join(' ') }}" data-team-id="{{ score.team.teamid }}">
             {% if enable_ranking %}
-                <td class="scorepl {{medalColor}}">
+                <td class="scorepl">
+                    {% if medalColor != '' %}
+                        <i class="fa fa-medal {{medalColor}}" style="font-size:150%;"></i>
```
vmcj, line 140: I'm not 100% sure but is it ok that we mix `em` and `%` throughout the code?

meisterT: Good point, using `rem` instead now.

PR #2900, opened 2025-02-27
Render contest validation errors more nicely. Fixes #2891

### LINKED ISSUE #2891: Classname in user exposed config page (8.3)
See:
https://files.slack.com/files-pri/THT5URKCM-F08DG6BAWUF/image.png

Mentioned by a user, I did not check if this can be reproduced yet.

### Your environment
DOMjudge 8.3

### Steps to reproduce
Probably enable process medals, make the categories with medals empty

meisterT:


vmcj (approved):

webapp/src/Service/CheckConfigService.php
```diff
@@ -479,9 +479,15 @@ public function checkContestsValidate(): ConfigCheckItem
...
         $desc = '';
         foreach ($contesterrors as $cid => $errors) {
-            $desc .= "Contest: c$cid: " .
-                /* @phpstan-ignore-next-line */
-                    (count($errors) == 0 ? 'no errors' : (string)$errors) ."\n" .$cperrors[$cid];
+            $desc .= "Contest: c$cid: ";
+            if (count($errors) == 0) {
+                $desc .= "no errors\n";
+            } else {
+                $desc .= "errors:\n";
```
vmcj, line 486:
```suggestion
                $desc .= "errors:\n" . ($cperrors[$cid] ? $cperrors[$cid] . '\n' : '');
```

Did not check if it works, but as you mentioned last night the $cperrors are missing.

meisterT: Also added it to the if statement above and there is no newline necessary as it adds it already

PR #2901, opened 2025-02-28
Do not collapse errored items in config check.

meisterT:


meisterT:
Example:

image

vmcj (approved):
Why did we ever choose to collapse those?

meisterT:
We turned at some point from everything expanded to everything collapsed, now that's the right balance, IMHO.

eldering (approved):

nickygerritsen (approved):

PR #2909, opened 2025-02-28
Removes unneeded transactions.

tuupke:


tuupke:
Just the first three transactions. But can be merged. Probably easiest to check commit-by-commit.

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -910,37 +910,22 @@ public function internalErrorAction(Request $request): ?int
      */
     protected function giveBackJudging(int $judgingId, ?Judgehost $judgehost): void
     {
+        // Reset the judgings without using Doctrine, it has no support for update queries containing a join.
```
meisterT, line 910: doctrine :exploding_head:

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -866,32 +866,24 @@ public function internalErrorAction(Request $request): ?int
...
+                ->andWhere('jt.judgehost_id IS NULL')
+                ->andWhere('jt.valid = 1')
+                ->setParameter('error', $error)
+                ->setParameter('id', $disabled_id)
+                ->distinct()
+                ->getQuery()
+                ->getArrayResult();
+
+            if ($rows == 0) {
+                // TODO, handle this case. Nothing was updated.
```
meisterT: Is there anything to be handled? If there is no outstanding that we would disable here, we don't have an issue, do we?

tuupke: Fair point, I'll remove.

Reading this back, this cannot be correct! I've had to create a manual query for updates with joins. 🤯 

Apparently this case is never hit in the testsuite either.

nickygerritsen (approved):

vmcj:
Discussed with @tuupke this one shouldn't be merged yet.

PR #2915, opened 2025-02-28
SVG test case visualisation for jury submission

### LINKED ISSUE #2875: [Feature request] Allow SVG as test case visualization
Test cases can already have PNG/JPG/GIF images attached to them, see [here](https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Service/ImportProblemService.php#L404). It would be nice to also allow support for SVG.

In the Slack discussion, it was mentioned that SVG allows potentially executable code, so it would be wise to sanitize them during the import. On the other hand, these SVGs are only created by (usually trusted) jury members (perhaps except in the case of output visualization, see #2744).

For jury members, it would give more options to create a visualization. Automatically generating PNGs requires some library code, and SVGs can simply be written as plain text.

### Expected behaviour
Expectation: SVGs are not rejected during import.

Kevinjil:
Fixes #2875

### DIFF
--- a/webapp/src/Utils/Utils.php
+++ b/webapp/src/Utils/Utils.php
@@ -3,6 +3,7 @@ namespace App\Utils;
 
 use DateTime;
 use Doctrine\Inflector\InflectorFactory;
+use enshrined\svgSanitize\Sanitizer as SvgSanitizer;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 
@@ -715,6 +716,15 @@ class Utils
         }
 
         return [$width, $height, $width / $height];
+    }
+
+    public static function sanitizeSvg(string $svgContents): string | false
+    {
+        $sanitizer = new SvgSanitizer();
+        $sanitizer->removeRemoteReferences(true);
+        $sanitizer->minify(true);
+        
+        return $sanitizer->sanitize($svgContents);
     }
 
     /**
--- a/webapp/templates/jury/submission.html.twig
+++ b/webapp/templates/jury/submission.html.twig
@@ -742,7 +742,7 @@
                         <span style="float:right; border: 3px solid #438ec3; margin: 5px; padding: 5px;">
                             {% set imgUrl = path('jury_problem_testcase_fetch', {'probId': submission.problem.probid, 'rank': run.rank, 'type': 'image'}) %}
                             <a href="{{ imgUrl }}">
-                                <img src="data:image/{{ run.imageType }};base64,{{ runsOutput[runIdx].image_thumb | base64 }}"/>
+                                <img src="data:{{ run.imageType | extensionToMime }};base64,{{ runsOutput[runIdx].image_thumb | base64 }}"/>
                             </a>
                         </span>
                     {% endif %}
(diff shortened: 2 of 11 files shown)


webapp/src/Utils/Utils.php
```diff
@@ -717,6 +718,19 @@ public static function getImageSize(string $filename): array
         return [$width, $height, $width / $height];
     }
 
+    /**
+     * Sanitize the given SVG file contents and return the sanitized SVG file contents.
```
vmcj, line 722: I think we don't need the explanation here so I rather have that the name implies this instead of this DOCBLOCK

webapp/templates/jury/submission.html.twig
```diff
@@ -742,7 +743,7 @@
                         <span style="float:right; border: 3px solid #438ec3; margin: 5px; padding: 5px;">
                             {% set imgUrl = path('jury_problem_testcase_fetch', {'probId': submission.problem.probid, 'rank': run.rank, 'type': 'image'}) %}
                             <a href="{{ imgUrl }}">
-                                <img src="data:image/{{ run.imageType }};base64,{{ runsOutput[runIdx].image_thumb | base64 }}"/>
+                                <img src="data:{{ EXTENSION_TO_MIMETYPE[run.imageType] }};base64,{{ runsOutput[runIdx].image_thumb | base64 }}"/>
```
vmcj, line 746: Does it not work with the variable in between? I think you can just use PHP here.

webapp/tests/Unit/Utils/UtilsTest.php
```diff
@@ -713,6 +713,54 @@ public function provideTestGetImageSize(): Generator
         yield [__DIR__ . '/../../../public/images/DOMjudgelogo.svg', 510, 1122];
     }
 
+    public function testSanitizeSvg(): void
+    {
+        $dirty = <<<EOF
```
vmcj, line 718: Can you add the URL to the original location so we can update if there is more nasty stuff?

vmcj (approved):

webapp/templates/jury/submission.html.twig
```diff
@@ -742,7 +747,8 @@
                         <span style="float:right; border: 3px solid #438ec3; margin: 5px; padding: 5px;">
                             {% set imgUrl = path('jury_problem_testcase_fetch', {'probId': submission.problem.probid, 'rank': run.rank, 'type': 'image'}) %}
                             <a href="{{ imgUrl }}">
-                                <img src="data:image/{{ run.imageType }};base64,{{ runsOutput[runIdx].image_thumb | base64 }}"/>
+                                {% set EXTENSION_TO_MIMETYPE = constant('App\\Service\\DOMJudgeService::EXTENSION_TO_MIMETYPE') %}
```
meisterT, line 750: let's use a normal twig filter for this?

webapp/tests/Unit/Utils/UtilsTest.php
meisterT: Isn't the URL a line above your comment? :-D

meisterT (commented):
If you upload a large SVG, how wide will that appear then?

Kevinjil:
> If you upload a large SVG, how wide will that appear then?

Since the last commit, the configured max thumbnail size is enforced in CSS (200x200 PX by default)

meisterT (approved):

PR #2919, opened 2025-02-28
Fix displaying medals on scoreboard

nickygerritsen:
They sometimes were displaying next to each other. By using two table cells, this will never happen.


nickygerritsen:
image

meisterT (approved):

PR #2920, opened 2025-02-28
Split display of problems in current and other contests.

meisterT:
This helps especially when there are many problems loaded.

-----

Before:

image

After:

image


vmcj (approved):
I like it.

PR #2921, opened 2025-02-28
Improve hearts on scoreboard by using separate column

nickygerritsen:
Before:
Screenshot 2025-02-28 at 18 07 46

After:
Screenshot 2025-02-28 at 18 07 37

Ignore the medal, that is fixed in #2919


vmcj (approved):

meisterT (approved):
Thanks, I also tested locally that hearting the heart works still.

PR #2927, opened 2025-03-01
Do not clutter unit test output on CI with deprecation notices.

meisterT:
If we actually care we should set up a separate test for it.


vmcj:
> If we actually care we should set up a separate test for it.

We've done this for quite a while now and I rather not do an extra test for this as CI is already slow.
We use it here: https://github.com/DOMjudge/domjudge/pull/2927/files#diff-10407f9f203257ec9a7d9297775e7d7085e36ce06faeaf17450c25a1b6949a20L49 does that usecase still work?

PR #2928, opened 2025-03-01
Resize editor to adjust to window size.

meisterT:
Otherwise the editor pane is typically way to small.


nickygerritsen (approved):

Kevinjil:
Why don't we just add a CSS rule instead of js?

PR #2931, opened 2025-03-01
Change size of mobile flags

nickygerritsen:
Before:
Screenshot 2025-03-01 at 14 36 24

After:
Screenshot 2025-03-01 at 14 36 17


webapp/public/style_domjudge.css
```diff
@@ -243,6 +243,11 @@ del {
 }
 .scoreboard .scoreaf { white-space: nowrap; border: 0; padding-left: 2px; text-align: center; }
 .scoreboard .scoreaf img { vertical-align: middle; }
+.mobile-scoreboard .countryflag {
+    height: 1rem;
+    width: 1.333333rem;
```
meisterT, line 248: I assume we don't need so much precision here :-D

meisterT (approved):

PR #2933, opened 2025-03-01
Show flags on the pre-contest page if they are enabled.

meisterT:


vmcj:
I think your upload of the screenshot failed.

nickygerritsen (approved):

meisterT:
image

vmcj (approved):

meisterT:
Also fixed bg color now:

image

vmcj:
Not sure if I like this big color block,

Can we do it subtle in the header or with one border around the whole group?

nickygerritsen (approved):

meisterT:
> Not sure if I like this big color block,
> 
> Can we do it subtle in the header or with one border around the whole group?

I tried both and they also don't look great. Let's use this is as incremental improvement and improve from here.

PR #2936, opened 2025-03-02
Record validator metadata.

meisterT:
This can be especially useful when the validator takes more resources than expected or is misbehaving in other ways.

Example:

image

and

image

### DIFF
--- a/webapp/src/Controller/API/JudgehostController.php
+++ b/webapp/src/Controller/API/JudgehostController.php
@@ -608,7 +608,12 @@ class JudgehostController extends AbstractFOSRestController
                     ),
                     new OA\Property(
                         property: 'metadata',
-                        description: 'The (base64-encoded) metadata',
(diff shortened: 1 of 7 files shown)


meisterT:
(Fixed the unnecessary clock in the second screenshot already)

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -919,7 +925,8 @@ private function addSingleJudgingRun(
         string $outputRun,
         ?string $teamMessage,
         string $metadata,
-        ?string $testcasedir
+        ?string $testcasedir,
+        ?string $compare_meta,
```
nickygerritsen, line 929: Nit:
```suggestion
        ?string $compareMeta,
```
to be consistent with the other arguments

nickygerritsen (approved):

PR #2938, opened 2025-03-02
Highlight whole team name column for categories

nickygerritsen:


vmcj:
How does this look for the default DOMjudge category? I think it will look quite ugly together with the default colors?

nickygerritsen:
> How does this look for the default DOMjudge category? I think it will look quite ugly together with the default colors?

The pink one? 
Before:
Screenshot 2025-03-02 at 11 07 36

After:
Screenshot 2025-03-02 at 11 07 54

It's not worse IMHO, since we don't display a flag. If we do, it would be ugly but then just not use that color.

vmcj (approved):

meisterT (approved):

PR #2941, opened 2025-03-02
Only display affected judgings if there is more than one.

meisterT:
If there is only one, it already is the "related" judging.


nickygerritsen (approved):

vmcj (approved):
Merge conflict and no screenshots but I guess this is an improvement.

PR #2944, opened 2025-03-02
Use adaptive buckets for the submission graph

### LINKED ISSUE #2934: Limit bucket count for submission graph
In #2722 we mention that we should limit the number of buckets in the submission graph, which we have not yet done. Today we encountered this with a page-freeze in most popular browsers when opening a year-long contest.

Kevinjil:
We determine a suitable unit based on the contest length and some min/max bucket bounds. We convert all submission times in seconds to this unit and determine nice locations to put ticks on the x-axis.

Scaling for e.g. 1-year contest:
image

Fixes #2934

### DIFF
--- a/webapp/templates/jury/analysis/contest_overview.html.twig
+++ b/webapp/templates/jury/analysis/contest_overview.html.twig
@@ -365,8 +365,6 @@ nv.addGraph(function() {
 // x-axis is contest time
 // y axis is # of submissions
 
-var max_submissions_per_minute = 0;
-
 var submission_stats = [
   {% for result in ['correct', 'wrong-answer', 'timelimit', 'run-error', 'compiler-error', 'no-output'] %}
   {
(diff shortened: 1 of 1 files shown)


webapp/templates/jury/analysis/contest_overview.html.twig
```diff
@@ -399,27 +421,30 @@ const statMap = submission_stats.reduce((map, stat) => {
...
-var tickDelta = 15;
-while (contest_duration_minutes / tickDelta > 15) {
-  tickDelta *= 2;
-}
-var tickValues = Array.from({ length: Math.ceil(contest_duration_minutes / tickDelta) + 1 }, (_, i) => i * tickDelta);
+// Pick a nice round tickDelta and tickValues based on the step size of units.
+// We want whole values in the unit, and the ticks MUST match a corresponding bucket otherwise the resulting
+// coordinate will be NaN.
+const convert_factor = seconds_per_bucket / unit.convert;
+const maxTicks = Math.min(bucket_count, contest_duration / unit.step, 31)
```
meisterT, line 444: Where does this 31 come from?

Kevinjil: The idea was to limit the max number of ticks shown to some reasonable number. Changed it to the minimum bucket count as it looks pleasant.

meisterT (approved):

nickygerritsen (approved):

PR #2950, opened 2025-03-03
Code refactor for printing UI, API and client

cubercsl:
This PR does two things.
1. The current UI allows us to select any available language for printing (including plain text) So I removed the code in the form that retrieves the available languages.

jury
team

2. Make the API available in any available language (plain text if no language is available) and make appropriate modifications to the print client so that it can submit any language for printing.

But I think putting submitting and printing in one command is not a wise choice, which makes the code logic confusing.

Here is our implementation of a command line printing client that we used in a recent ICPC Regional for reference. If anyone thinks a separate print client implementation would be more appropriate, I can open another PR.

https://gist.github.com/cubercsl/809fcfdb65a7f88d6322cf0ab3505aa4

### DIFF
--- a/submit/submit
+++ b/submit/submit
@@ -506,7 +506,7 @@ elif contests:
 languages: list = []
 problems: list = []
 if my_contest and baseurl:
-    if 'allow_submit' in my_contest and not my_contest['allow_submit']:
+    if not args.print and 'allow_submit' in my_contest and not my_contest['allow_submit']:
         warn_user('Submissions for contest (temporarily) disabled')
         exit(1)
     languages = read_languages()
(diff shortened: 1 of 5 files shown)


Kevinjil (approved):

Kevinjil:
> But I think putting submitting and printing in one command is not a wise choice, which makes the code logic confusing.
> 
> Here is our implementation of a command line printing client that we used in a recent ICPC Regional for reference. If anyone thinks a separate print client implementation would be more appropriate, I can open another PR.
> 
> https://gist.github.com/cubercsl/809fcfdb65a7f88d6322cf0ab3505aa4

The choice for not creating a separate script was intentional: we want to share the code and not depend on imports for easy deployment.

submit/submit
```diff
@@ -610,23 +610,25 @@ for problem in problems:
 if not my_problem and not args.print:
     usage('No known problem specified or detected.')
 
+if not args.print:
 # Guess entry point if not already specified.
```
vmcj, line 614: nit: can you also indent that line?

PR #2954, opened 2025-03-09
Introduce the concept of sort keys.

meisterT:
Progress towards https://github.com/DOMjudge/domjudge/issues/2525.

The basic idea is to replace a lot of scattered scoring logic (which
might become more complex when we introduce more scoring options such as
optimization problems) by encoding sufficient information within the
existing rank cache.

The information is encoded into a sort key that is designed to be sorted
descendingly. The sort key is a tuple (serialized as string into the
database) of fixed precision decimals. Each tuple uses 9 decimals and is
left padded with `0`s, enabling sorting via standard database query
operations. For ascendingly sorted tuple entries (e.g. penalty time),
values are subtracted from a very large constant - this ensures that
that the key can be sorted as a whole.

For example, with ICPC scoring, the top 3 teams from NWERC 2024 receive
these sort keys:
```
00000000000000000000013.000000000,99999999999999999998725.000000000,99999999999999999999711.000000000
00000000000000000000011.000000000,99999999999999999998918.000000000,99999999999999999999701.000000000
00000000000000000000011.000000000,99999999999999999998916.000000000,99999999999999999999706.000000000
```

This mechanism should facilitate the implementation of other planned
scoring methods, particularly partial scoring and optimization problems,
with reasonable complexity in the business logic.

We are using `bcmath` with fixed precision to avoid numerical precision
issues caused by the order of floating point operations.  While not
critical currently, this will be essential when handling non-integers,
such as those in optimization problems.

The new mechanism also caches some computations and thus improves
scoreboard computation efficiency.

### DIFF
--- a/webapp/migrations/Version20250309122806.php
+++ b/webapp/migrations/Version20250309122806.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20250309122806 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add sort keys to rankcache, allowing us to support different scoring functions efficiently and elegantly';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE rankcache ADD sort_key_public TEXT DEFAULT \'\' NOT NULL COMMENT \'Opaque sort key for public audience.\', ADD sort_key_restricted TEXT DEFAULT \'\' NOT NULL COMMENT \'Opaque sort key for restricted audience.\'');
+        $this->addSql('CREATE INDEX sortKeyPublic ON rankcache (sort_key_public)');
+        $this->addSql('CREATE INDEX sortKeyRestricted ON rankcache (sort_key_restricted)');
+    }
+
+    public function down(Schema $schema): void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->addSql('DROP INDEX sortKeyPublic ON rankcache');
+        $this->addSql('DROP INDEX sortKeyRestricted ON rankcache');
+        $this->addSql('ALTER TABLE rankcache DROP sort_key_public, DROP sort_key_restricted');
+    }
+
+    public function isTransactional(): bool
+    {
+        return false;
+    }
+}
--- a/webapp/src/Service/ScoreboardService.php
+++ b/webapp/src/Service/ScoreboardService.php
@@ -19,6 +19,7 @@ use App\Utils\Scoreboard\Scoreboard;
 use App\Utils\Scoreboard\SingleTeamScoreboard;
 use App\Utils\Scoreboard\TeamScore;
 use App\Utils\Utils;
+use Doctrine\Common\Collections\Order;
 use Doctrine\DBAL\ArrayParameterType;
 use Doctrine\DBAL\Exception as DBALException;
 use Doctrine\ORM\EntityManagerInterface;
@@ -67,14 +68,15 @@ class ScoreboardService
             return null;
         }
 
-        $teams      = $this->getTeams($contest, $jury && !$visibleOnly, $filter);
+        $teams      = $this->getTeamsInOrder($contest, $jury && !$visibleOnly, $filter);
         $problems   = $this->getProblems($contest);
         $categories = $this->getCategories($jury && !$visibleOnly);
         $scoreCache = $this->getScorecache($contest);
+        $rankCache  = $this->getRankcache($contest);
 
         return new Scoreboard(
             $contest, $teams, $categories, $problems,
-            $scoreCache, $freezeData, $jury || $forceUnfrozen,
+            $scoreCache, $rankCache, $freezeData, $jury || $forceUnfrozen,
             (int)$this->config->get('penalty_time'),
             (bool)$this->config->get('score_in_seconds'),
         );
@@ -93,7 +95,7 @@ class ScoreboardService
     {
         $freezeData = new FreezeData($contest);
 
-        $teams = $this->getTeams($contest, true, new Filter([], [], [], [$teamId]));
+        $teams = $this->getTeamsInOrder($contest, true, new Filter([], [], [], [$teamId]));
         if (empty($teams)) {
             return null;
         }
@@ -101,7 +103,7 @@ class ScoreboardService
         $problems   = $this->getProblems($contest);
         $rankCache  = $this->getRankcache($contest, $team);
         $scoreCache = $this->getScorecache($contest, $team);
-        $teamRank   = $this->calculateTeamRank($contest, $team, $rankCache, $freezeData, true);
+        $teamRank   = $this->calculateTeamRank($contest, $team, $freezeData, true);
 
         return new SingleTeamScoreboard(
             $contest, $team, $teamRank, $problems,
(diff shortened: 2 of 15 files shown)


webapp/src/Utils/Scoreboard/Scoreboard.php
```diff
@@ -33,10 +34,12 @@ class Scoreboard
      */
     public function __construct(
         protected readonly Contest $contest,
+        // TODO: document somewhere that this is already sorted.
```
vmcj, line 37: I suggest renaming the class so we can expect it in the future.

Kevinjil:
What is the benefit we get from storing a fixed-precision decimal as string instead of using a native `decimal` type or integer representation of a fixed-precision decimal?

meisterT:
> What is the benefit we get from storing a fixed-precision decimal as string instead of using a native `decimal` type or integer representation of a fixed-precision decimal?

If the sort key would consist of a simple number, then we could use the native `decimal` type. But we are using a sort key which is essentially a tuple of fixed precision decimals, allowing to generalize away from ICPC scoring (while making code less scattered and more efficient at the same time). This tuple is of "unknown" length if we want to support more than one scoring method, so we represent it as string.

Side note: `bcmath` which is the recommended way to work with fixed precision decimal types in PHP, is also using `string` as underlying representation for individual numbers.

Side note on the side note: separately from this PR, we should look into whether it makes sense using `bcmath` where we are using `decimal(32,9)` today, since the mapping from/to `float` loses precision.

webapp/tests/Unit/Service/ScoreboardServiceTest.php
```diff
@@ -0,0 +1,148 @@
...
+
+namespace App\Tests\Unit\Service;
+
+use App\Service\ScoreboardService;
+use Doctrine\Common\Collections\Order;
+use Exception;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+class ScoreboardServiceTest extends KernelTestCase
+{
```
eldering, line 11: Not sure the tests here are very relevant as they test internal details. I'd focus on the behaviour tests at the end.

webapp/migrations/Version20250309122806.php
```diff
@@ -0,0 +1,40 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20250309122806 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add sort keys to rankcache, allowing us to support different scoring functions efficiently and elegantly';
```
eldering, line 17: I'd consider going a step further and not use these opaque sort keys directly to determine an ordering, but maybe a fuction of that data for flexibility. But then again, that might be not to hard to extend to later.

eldering:
> > What is the benefit we get from storing a fixed-precision decimal as string instead of using a native `decimal` type or integer representation of a fixed-precision decimal?
> 
> If the sort key would consist of a simple number, then we could use the native `decimal` type. But we are using a sort key which is essentially a tuple of fixed precision decimals, allowing to generalize away from ICPC scoring (while making code less scattered and more efficient at the same time). This tuple is of "unknown" length if we want to support more than one scoring method, so we represent it as string.
> 
> Side note: `bcmath` which is the recommended way to work with fixed precision decimal types in PHP, is also using `string` as underlying representation for individual numbers.
> 
> Side note on the side note: separately from this PR, we should look into whether it makes sense using `bcmath` where we are using `decimal(32,9)` today, since the mapping from/to `float` loses precision.

Repeating my code comment here: an alternative would be to let the sort key really be an opaque blob, even without the assumption that it is sorted (lexicographically), but instead have

1. a separate function (depending on the scoring type) that maps this opaque blob into something that defines a rank ordering
2. a compare function to define an order these opaque blobs

Then the opaque blob could e.g. be some JSON encoding the scoring state.

OTOH, in case 1 this function just maps into something similar to what's currently stored in the sorting key already, so nothing is really gained. I guess 2 could be a bit more generic, but I do like the fact that we now declare an ordering simply by comparing these sort keys. That indeed simplifies the code a lot. So probably stick to that unless we need more flexibility.

webapp/tests/Unit/Service/ScoreboardServiceTest.php
meisterT: They helped me uncover a small bug, so I think they were useful. If at one point they become tedious to update because how it is implemented internally frequently (I don't think so, but who knows) we can always delete them.

meisterT:
> > > What is the benefit we get from storing a fixed-precision decimal as string instead of using a native `decimal` type or integer representation of a fixed-precision decimal?
> 
> I don't think we always have to use these fixed length decimal strings: the sort keys can be different per different contest/scoring type. So in an ICPC scoring setting I think we can just use zero-padded integers.

If you care I am happy to change it, but it requires more code (not much of course) for almost no benefit. Who is going to inspect sort keys?


> > If the sort key would consist of a simple number, then we could use the native `decimal` type. But we are using a sort key which is essentially a tuple of fixed precision decimals, allowing to generalize away from ICPC scoring (while making code less scattered and more efficient at the same time). This tuple is of "unknown" length if we want to support more than one scoring method, so we represent it as string.
> > Side note: `bcmath` which is the recommended way to work with fixed precision decimal types in PHP, is also using `string` as underlying representation for individual numbers.
> > Side note on the side note: separately from this PR, we should look into whether it makes sense using `bcmath` where we are using `decimal(32,9)` today, since the mapping from/to `float` loses precision.
> 
> Repeating my code comment here: an alternative would be to let the sort key really be an opaque blob, even without the assumption that it is sorted (lexicographically), but instead have
> 
> 1. a separate function (depending on the scoring type) that maps this opaque blob into something that defines a rank ordering
> 2. a compare function to define an order these opaque blobs
> 
> Then the opaque blob could e.g. be some JSON encoding the scoring state.
> 
> OTOH, in case 1 this function just maps into something similar to what's currently stored in the sorting key already, so nothing is really gained. I guess 2 could be a bit more generic, but I do like the fact that we now declare an ordering simply by comparing these sort keys. That indeed simplifies the code a lot. So probably stick to that unless we need more flexibility.

I started out with having a separate function and a JSON blob indeed, but then came up with this idea and it was quite a bit less code. If at one point we are realize that this is not powerful enough, it is trivial to change to what you describe. I do like about the current code that sorting when retrieving the scoreboard is purely a database operation, no business logic required.


But this discussion reminds me that I have to add to the migration a scoreboard refresh so that the data is correct.

meisterT:
> But this discussion reminds me that I have to add to the migration a scoreboard refresh so that the data is correct.

Well, seems like this is not really possible: https://github.com/doctrine/DoctrineMigrationsBundle/pull/559/files

cc @nickygerritsen in case you have ideas

-------

At least we recommend refreshing the cache here: https://www.domjudge.org/docs/manual/8.2/upgrading.html#upgrading

nickygerritsen:
> > But this discussion reminds me that I have to add to the migration a scoreboard refresh so that the data is correct.
> 
> Well, seems like this is not really possible: [doctrine/DoctrineMigrationsBundle#559 (files)](https://github.com/doctrine/DoctrineMigrationsBundle/pull/559/files)
> 
> cc @nickygerritsen in case you have ideas
> 
> At least we recommend refreshing the cache here: [domjudge.org/docs/manual/8.2/upgrading.html#upgrading](https://www.domjudge.org/docs/manual/8.2/upgrading.html#upgrading)

Nope I don't think there is an easy way. What we could do is create a CLI command to refresh the scoreboard cache and then run that on `make install` or in the `upgrade` command of the SQL script.

meisterT:
> What we could do is create a CLI command to refresh the scoreboard cache and then run that on `make install` or in the `upgrade` command of the SQL script.

Good idea, I will do that (separately from this PR).

meisterT:
see https://github.com/DOMjudge/domjudge/pull/2971

nickygerritsen (approved):
I like it. It looks like it makes things a bit less complex and allows for expansion

webapp/src/Service/ScoreboardService.php
```diff
@@ -1103,16 +1096,20 @@ protected function getScorecache(Contest $contest, ?Team $team = null): array
      * Get the rank cache for the given team.
      * @throws NonUniqueResultException
      */
```
vmcj, line 1098:
```suggestion
     returns RankCache[]
     */
```
(I suspect that's the return type, please verify)

I wonder why there is no test complaining about this

webapp/tests/Unit/Service/ScoreboardServiceTest.php
```diff
@@ -0,0 +1,148 @@
...
+        self::assertEquals($teamA, $teamB);
+    }
+
+    public function testOneTeamEmpty(): void
+    {
+        $teamA = ScoreboardService::getICPCScoreKey(0, 0, 0);
+        $teamB = ScoreboardService::getICPCScoreKey(7, 666, 420);
+        self::assertTrue($teamA < $teamB);
+    }
+
```
vmcj, line 127:
```suggestion
    public function testOneTeamEmpty(): void
    {
        $teamA = ScoreboardService::getICPCScoreKey(0, 0, 0);
        $teamB = ScoreboardService::getICPCScoreKey(7, 666, 420);
        self::assertTrue($teamA < $teamB);
    }

    public function testOneTeamSolvedMore(): void
    {
        $teamA = ScoreboardService::getICPCScoreKey(1, 333, 210);
        $teamB = ScoreboardService::getICPCScoreKey(7, 666, 420);
        self::assertTrue($teamA < $teamB);
    }

```

We had all the special cases, but I think nothing to make sure that the normal case works. So the case where 1 sortcondition was higher but the others were lower was not handled yet.

vmcj (approved):
Did this make any parts faster or slower? How will scoreboards with 1000 teams be affected?

webapp/src/Service/ScoreboardService.php
meisterT: I assume you meant `@return RankCache[]`?

> I wonder why there is no test complaining about this

Why would tests complain about a missing non-concrete return type?

meisterT:
> Did this make any parts faster or slower? How will scoreboards with 1000 teams be affected?

It is more or less the same, and other parts dominate the total time, going into some detail below.

-----------

Theoretically it does make updates of the cache a tiny bit slower - the good thing here is that this is constant overhead (to construct the sort key) and doesn't scale with the number of teams.

In practice: when refreshing the full NWERC scoreboard from last year it takes in prod mode on my laptop (as median of 11 runs each):
- on `main`: ~20.4s 
- this PR: ~20.5s

(But there is a lot of noise, so it is likely not statistically significant.)

-----------

Theoretically, it does make retrieving the scoreboard a tiny bit faster as there is less business logic to execute.

In practice, when running
```
time ( for i in {1..100}; do curl -n -N http://localhost/domjudge/api/contests/nwerc2024/scoreboard > /dev/null; done )
```
I see:
- on `main`: ~18.0s 
- this PR: ~17.9s

(Again there is a lot of noise, so it is likely not statistically significant.)

eldering:
> > Did this make any parts faster or slower? How will scoreboards with 1000 teams be affected?
> 
> It is more or less the same, and other parts dominate the total time, going into some detail below.
> 
> Theoretically it does make updates of the cache a tiny bit slower - the good thing here is that this is constant overhead (to construct the sort key) and doesn't scale with the number of teams.

I guess you could make it faster by keeping the sort keys smaller. For example for the standard ICPC scoring, we don't need decimals, so the string length can be roughly cut in half.

meisterT:
> I guess you could make it faster by keeping the sort keys smaller. For example for the standard ICPC scoring, we don't need decimals, so the string length can be roughly cut in half.

I tried this, also with ints, and there is no difference in timing.

webapp/src/Service/ScoreboardService.php
vmcj: > I assume you meant `@return RankCache[]`?
> 
Yes, the `returns` should be `return` and I suspect the function returns an array of `RankCache`.

> > I wonder why there is no test complaining about this
> 
> Why would tests complain about a missing non-concrete return type in a docstring?

I thought we had PHPStan configured to also complain if the return type is not 100% clear as it uses it for further static analysis.

meisterT: Ok, that's done. (Never seen such a PHPstan comment though.)

webapp/migrations/Version20250309122806.php
meisterT: I responded in the main conversation thread to this as well - we can always add the function if we need it later (I suspect we don't).

PR #2956, opened 2025-03-15
Share analysis graph code for submissions

Kevinjil:
The code for the contest analysis was updated in #2944, but not yet for the problem analysis. As these graphs should behave the same, the code for creating the graph has been moved from twig to static JavaScript and only the data is rendered in twig. This simplifies sharing the graph creation code and slightly reduces server rendering load.

Smaller problem-submission graph:
image


meisterT (approved):
Great!

PR #2960, opened 2025-03-16
Disallow negative numbers on the scoreboard.

meisterT:
If submissions happen before contest start - realistically these can only happen for jury - they are moved up to the contest start and represented as 0.

(They also did break the scoreboard UI in the past with large negative numbers if you set up the contest a few days in advance)


vmcj (approved):

meisterT:
This requires https://github.com/DOMjudge/domjudge/pull/2962 to be merged first.

Issue #2946, opened 2025-03-02
Penalty time parsing from external data source in new draft spec


Kevinjil:
The CDS already implements the new draft spec, and exposes the penalty time as a relative time string instead of an integer in seconds.

We probably want to create a new DTO event version for the new spec with the changed parsing. A temporary fix has been pushed to [fix/penaltytime](https://github.com/Kevinjil/domjudge/tree/fix/penaltytime)

Kevinjil:
Note for everyone that uses a data source with the new draft spec: [fix/penaltytime](https://github.com/Kevinjil/domjudge/tree/fix/penaltytime) has been rebased after merging #2969, such that is can easily be cherry-picked on top of `main` again.

vmcj:
After https://github.com/icpctools/icpctools/commit/8803babd5d00b2cf971d368e0975e7d150ae064c we can make that DTO now.

PR #2969, opened 2025-03-17
Fix reltime to seconds parsing

Kevinjil:
Extracted the refactor and bug fix of #2946 to a separate PR. The change in the API spec is not included in this change: I will rebase the temporary fix on `main` after merging this.

In the original code, the modifier was only applied to the hours of the reltime string. Note that test `testRelTimeToSecondsWithNegative` fails without the fix commit:
```
Failed asserting that -3477.0 matches expected -3723.
```

### DIFF
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -690,41 +690,16 @@ class ExternalContestSourceService
             ->getRepository(Contest::class)
             ->find($this->getSourceContestId());
 
-        // We need to convert the freeze to a value from the start instead of
-        // the end so perform some regex magic.
-        $duration     = $data->duration;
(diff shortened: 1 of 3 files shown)


nickygerritsen:
Does this still work with the non reltime format which we still will have in the current ICPC season?

Kevinjil:
> Does this still work with the non reltime format which we still will have in the current ICPC season?

Yes: the spec change is not in this PR, only the refactoring and the bugfix. I will update my branch with the spec change ("ugly" fix) such that we can cherry-pick it for contests if we need it for as long as we don't implement the new spec properly.

Kevinjil:
Nevermind, I'm stupid and cherry-picked too much

meisterT (approved):

PR #2975, opened 2025-03-23
Extend autoconf tests

vmcj:
This tries to split tests where extra packages get installed to different runs.

It does add some small improvements to the configure script to make testing easier.


vmcj:
The PR should be fine without the `make distclean` but I added it to test that we get a good tarball after it.

Not sure why https://github.com/DOMjudge/domjudge/actions/runs/14092975984/job/39473975481#step:4:1010 fails, @eldering shouldn't it find the target because of the Makefile.global? Other files also don't have an explicit dependency for distclean to depend on clean so what is it complaining about?

.github/jobs/configure-checks/all.bats
```diff
@@ -3,77 +3,108 @@
...
+        args=${args/texlive-latex-extra/texlive-collection-latexextra}
+        args=${args/texlive-latex-recommended/texlive-collection-latexrecommended}
+    fi
     if [ "$distro_id" = "ID=fedora" ]; then
-        repo-install httpd
+        args=${args/libcgroup-dev/libcgroup-devel}
+        args=${args/tex-gyre/texlive-tex-gyre}
+        args=${args/python3-sphinx-rtd-theme/python3-sphinx_rtd_theme}
+        # Debian/Ubuntu need 2 packages, fedora some more. As we always install those together we just
+        # add them as replacement for and skip the other.
```
eldering, line 48: This sentence is not written correctly:
```suggestion
        # Debian/Ubuntu need 2 packages, fedora some more. As we always install those together we just
        # add them as replacement and skip the second.
```

.github/jobs/configure-checks/all.bats
```diff
@@ -3,77 +3,108 @@
...
+
+run_configure () {
+    su $u -c "./configure $*"
 }
 
 compiler_assertions () {
-    run run_configure
+    # This test is ran with multiple containers, as we only care for the C/C++ test
+    # a bogus value is provided for the webserver as for example arch doesn't have
+    # a default www-data like group.
```
eldering, line 106:
```suggestion
    # This test is ran with multiple containers. Since we only care for the C/C++ test,
    # a bogus value is provided for the webserver group, as for example arch doesn't
    # have a default www-data like group.
```

.github/jobs/configure-checks/all.bats
```diff
@@ -110,7 +141,169 @@ compile_assertions_finished () {
...
+        repo-install httpd
+    fi
+    if [ "$distro_id" = 'ID="opensuse-leap"' ]; then
+        repo-install apache2
+    fi
+    repo-install gcc g++ libcgroup-dev composer acl
+}
+
+multiple_compilers() {
+    # all of the tests ran before will have a subset of thos already
```
eldering, line 170: Typo in `thos`, but more importantly, it's not really cleat to me what that refers to.

.github/jobs/configure-checks/all.bats
```diff
@@ -3,77 +3,108 @@
...
-    assert_regex "See [\`']config.log' for more details"
+docs_required_install () {
+    repo-install python3-sphinx-rtd-theme python3-sphinx rst2pdf tex-gyre latexmk texlive-latex-recommended texlive-latex-extra
+}
+
+run_configure () {
+    su $u -c "./configure $*"
 }
 
 compiler_assertions () {
```
eldering, line 103: It would help to rename this function to clearly indicate what it does (and for example include a verb in the function name). Same with `multiple_compilers`. Probably something starting with `assert_...`

.github/jobs/configure-checks/all.bats
```diff
@@ -3,77 +3,108 @@
...
     fi
-    repo-install gcc g++ libcgroup-dev composer
-}
-
-run_configure () {
-    su $u -c "./configure $*"
+    echo "$args"
 }
 
 repo-install () {
```
eldering, line 55: Maybe clearer to rename this to `install-package` or so, same with `repo-remove` below.

.github/jobs/configure-checks/all.bats
```diff
@@ -203,258 +433,334 @@ compile_assertions_finished () {
     assert_line " * webserver group.....: nginx"
 }
 
+@test "Run as root discouraged" {
```
eldering, line 436: I think various test descriptions could be written a bit more clearly/explicitly. For example here:
```suggestion
@test "A warning is shown when you implicitly configure with domjudge user=root" {
```
So say *what* is tested for.

.github/jobs/configure-checks/all.bats
```diff
@@ -203,258 +433,334 @@ compile_assertions_finished () {
...
-  assert_regex "^ \* webserver group\.\.\.\.\.: (www-data|apache|nginx)$"
-  assert_line " * judgehost...........: /opt/domjudge/judgehost"
-  assert_line " * runguard group......: domjudge-run"
-  run make domserver
-  assert_success
-  run make judgehost
-  assert_success
+# Cleanup for next runs
+rm -rf /opt/domjudge &>/dev/zero
+make clean &>/dev/zero
```
eldering, line 651: Can this be a test setup/teardown action? Adding it here manually makes this code fragile.

.github/jobs/configure-checks/distclean.bats
```diff
@@ -0,0 +1,303 @@
+#!/usr/bin/env bats
```
eldering, line 1: It looks like this file shares a lot of duplicate code with `all.bats`. Can common functionality be extracted?

meisterT:
I assume this needs to be redone at HEAD?

vmcj:
I hope we can merge https://github.com/DOMjudge/domjudge/pull/3566 first and I can close this one after merging that one.

PR #2977, opened 2025-03-25
Distclean should remove what `make clean` removes and configure artifacts

vmcj:
See: https://www.gnu.org/software/automake/manual/automake.html#Clean

Currently we would leave the `tex` output after a `make dist` which does get removed after `make clean` so `make distclean` was not a superset of `make clean`. Easiest way of achieving that is a recursive target on top. We remove nothing in that folder so I'm not sure if we want to set it lower in the subdirectories (what is the convention?).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -316,7 +316,7 @@ clean-l:
 # Remove Coverity scan data:
 	-rm -rf cov-int domjudge-scan.t* coverity-scan.tar.xz cov-submit-data-version.sh
 
-distclean-l: clean-autoconf
+distclean-l: clean-autoconf clean
 	-rm -f paths.mk
 
 maintainer-clean: inplace-uninstall


eldering (commented):
There's already https://github.com/DOMjudge/domjudge/blob/main/Makefile.global#L90, so I think the correct way to do this is add a `clean-l` target and let `clean` depend on that. That keeps the whole setup consistent.

vmcj:
> There's already https://github.com/DOMjudge/domjudge/blob/main/Makefile.global#L90, so I think the correct way to do this is add a `clean-l` target and let `clean` depend on that. That keeps the whole setup consistent.

So replace in every `Makefile`, the target `clean` with `clean-l`?

vmcj:
Digging a bit further it seems that we do this correctly for most (or all) directories. For the `doc/manual/build` we leave too many files. I think we should clean those with a `make clean`, what are your thoughts @eldering?

eldering:
> Digging a bit further it seems that we do this correctly for most (or all) directories. For the `doc/manual/build` we leave too many files. I think we should clean those with a `make clean`, what are your thoughts @eldering?

Yeah, probably call the clean targets in the sphinx generated makefiles from `doc/manual/Makefile`. There's actually  already some of that there.

PR #2979, opened 2025-03-29
Allow selecting new problem types as part of the problem entity.

meisterT:
While you can select the new types, they won't function yet.

Part of https://github.com/DOMjudge/domjudge/issues/2525

Problem types are defined here: https://icpc.io/problem-package-format/spec/2023-07-draft.html#type

### DIFF
--- a/webapp/migrations/Version20250323190305.php
+++ b/webapp/migrations/Version20250323190305.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20250323190305 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE problem ADD types INT NOT NULL COMMENT \'Bitset of problem types, default is pass-fail.\'');
+        $this->addSql('UPDATE problem SET types = 1');
+        $this->addSql('UPDATE problem SET types = 5 WHERE is_multipass_problem = 1');
+        $this->addSql('UPDATE problem SET types = 9 WHERE combined_run_compare = 1');
+        $this->addSql('UPDATE problem SET types = 13 WHERE combined_run_compare = 1 AND is_multipass_problem = 1');
+        $this->addSql('ALTER TABLE problem DROP combined_run_compare, DROP is_multipass_problem');
+    }
+
+    public function down(Schema $schema): void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE problem ADD combined_run_compare TINYINT(1) DEFAULT 0 NOT NULL COMMENT \'Use the exit code of the run script to compute the verdict\', ADD is_multipass_problem TINYINT(1) DEFAULT 0 NOT NULL COMMENT \'Whether this problem is a multi-pass problem.\'');
+        $this->addSql('UPDATE problem SET combined_run_compare = 1 WHERE types = 9 OR types = 13');
+        $this->addSql('UPDATE problem SET is_multipass_problem = 1 WHERE types = 5 OR types = 13');
+        $this->addSql('ALTER TABLE problem DROP types');
+    }
+
+    public function isTransactional(): bool
+    {
+        return false;
+    }
+}
--- a/webapp/src/Entity/Problem.php
+++ b/webapp/src/Entity/Problem.php
@@ -86,13 +86,6 @@ class Problem extends BaseApiEntity implements
     #[Serializer\Exclude]
     private ?string $special_compare_args = null;
 
-    #[ORM\Column(options: [
-        'comment' => 'Use the exit code of the run script to compute the verdict',
-        'default' => 0,
-    ])]
-    #[Serializer\Exclude]
-    private bool $combined_run_compare = false;
-
     #[Assert\File]
     #[Serializer\Exclude]
     private ?UploadedFile $problemstatementFile = null;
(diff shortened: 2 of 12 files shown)


webapp/src/Entity/Problem.php
```diff
@@ -108,12 +101,15 @@ class Problem extends BaseApiEntity implements
     #[Serializer\Exclude]
     private ?string $problemstatement_type = null;
 
-    #[ORM\Column(options: [
-        'comment' => 'Whether this problem is a multi-pass problem.',
-        'default' => 0,
-    ])]
+    public const TYPE_PASS_FAIL = 1;
```
nickygerritsen, line 104: Should we use an int backed enum for this?

nickygerritsen (approved):

webapp/src/Entity/Problem.php
meisterT: It can have multiple types at once, can I do this with a standard enum?

nickygerritsen: Ah fair point. There is maybe a enum bitset package but I can't find it. Let’s stick to constants for now

webapp/src/Entity/Problem.php
```diff
@@ -287,26 +283,82 @@ public function getSpecialCompareArgs(): ?string
...
+        foreach ($types as $type) {
+            if ($type === 'pass-fail') {
+                $this->types |= self::TYPE_PASS_FAIL;
+            } elseif ($type === 'scoring') {
+                $this->types |= self::TYPE_SCORING;
+            } elseif ($type === 'multi-pass') {
+                $this->types |= self::TYPE_MULTI_PASS;
+            } elseif ($type === 'interactive') {
+                $this->types |= self::TYPE_INTERACTIVE;
+            }
```
eldering, line 300: I think this should raise an exception on a fall through case.

Also, why not use `switch ... case` here?

webapp/src/Entity/Problem.php
```diff
@@ -287,26 +283,82 @@ public function getSpecialCompareArgs(): ?string
...
+        }
+        if ($this->isScoringProblem()) {
+            $types[] = 'scoring';
+        }
+        if ($this->isMultipassProblem()) {
+            $types[] = 'multi-pass';
+        }
+        if ($this->isInteractiveProblem()) {
+            $types[] = 'interactive';
+        }
```
eldering, line 319: Why not do this the same as `getTypesForForm` below? What's the difference between these except for returning an array of strings or constants? This could easily be mapped from one to the other.

webapp/src/Service/ImportProblemService.php
```diff
@@ -277,6 +277,35 @@ public function importZippedProblem(
...
+                    }
+                    if (in_array('submit-answer', $types)) {
+                        if (in_array('multi-pass', $types)) {
+                            $messages['danger'][] = "Invalid problem type: 'submit-answer' and 'multi-pass' are mutually exclusive.";
+                            return null;
+                        }
+                        if (in_array('interactive', $types)) {
+                            $messages['danger'][] = "Invalid problem type: 'submit-answer' and 'interactive' are mutually exclusive.";
+                            return null;
+                        }
```
eldering, line 302: I think it would be better to move this validation into the problem class itself: we should never have an inconsistent problem types bitfield.

webapp/src/Entity/Problem.php
meisterT: reworked the code according to your second suggestion below, also added exceptions

webapp/src/Entity/Problem.php
meisterT: yeah, that is much better, see new code

webapp/src/Service/ImportProblemService.php
meisterT: done

webapp/migrations/Version20250323190305.php
```diff
@@ -0,0 +1,41 @@
...
+        $this->addSql('UPDATE problem SET types = 5 WHERE is_multipass_problem = 1');
+        $this->addSql('UPDATE problem SET types = 9 WHERE combined_run_compare = 1');
+        $this->addSql('UPDATE problem SET types = 13 WHERE combined_run_compare = 1 AND is_multipass_problem = 1');
+        $this->addSql('ALTER TABLE problem DROP combined_run_compare, DROP is_multipass_problem');
+    }
+
+    public function down(Schema $schema): void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE problem ADD combined_run_compare TINYINT(1) DEFAULT 0 NOT NULL COMMENT \'Use the exit code of the run script to compute the verdict\', ADD is_multipass_problem TINYINT(1) DEFAULT 0 NOT NULL COMMENT \'Whether this problem is a multi-pass problem.\', DROP types');
```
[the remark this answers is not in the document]
meisterT: You mean in general? Or at least say that `down` is lossy. I also wouldn't know what the `down` migration path is used for. A typical upgrade should do a backup, try `up` and be done with it and never look back.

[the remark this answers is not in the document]
eldering: I agree that the `down` path should be expected to be lossy and that there's no common use case for it. OTOH, since we offer it, I think we should make an effort to make it non-lossy where (reasonably) possible. Maybe someone wants to use it, if only to switch branches during development or so.
It would not be too hard to do the reverse here, right?

webapp/src/Entity/Problem.php
```diff
@@ -108,12 +101,23 @@ class Problem extends BaseApiEntity implements
     #[Serializer\Exclude]
     private ?string $problemstatement_type = null;
 
-    #[ORM\Column(options: [
-        'comment' => 'Whether this problem is a multi-pass problem.',
-        'default' => 0,
-    ])]
+    public const TYPE_PASS_FAIL = 1;
```
eldering, line 104: Maybe add a comment for clarity that these are bitfields (even though the powers of 2 kinda reveal it)?

webapp/src/Entity/Problem.php
```diff
@@ -287,26 +291,84 @@ public function getSpecialCompareArgs(): ?string
...
-    public function setMultipassProblem(bool $isMultipassProblem): Problem
+    public function setTypes(array $types): Problem
     {
-        $this->isMultipassProblem = $isMultipassProblem;
+        $types = array_unique($types);
+        $this->types = 0;
+        foreach ($types as $type) {
+            $this->types |= $type;
+        }
+        if ($this->types & self::TYPE_PASS_FAIL && $this->types & self::TYPE_SCORING) {
```
eldering, line 340: Wouldn't this be better?
```suggestion
        if (($this->types & self::TYPE_PASS_FAIL) xor ($this->types & self::TYPE_SCORING)) {
```

meisterT: I must be missing something. If I am not mistaken, my code showed the warning if both pass-fail and scoring bits are set while your code shows the warning in case exactly one is set?

webapp/migrations/Version20250323190305.php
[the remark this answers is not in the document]
meisterT: Done.

webapp/src/Entity/Problem.php
meisterT: Done.

webapp/src/Entity/Problem.php
eldering: Indeed. But isn't what I propose the correct constraint to enforce? A problem should always be either pass-fail or scoring. Never neither or both? But maybe I'm missing context of how this is handled in our code.

meisterT: Understood, I changed your code suggestion by negating it, and adjusting the error message.

eldering (approved):

webapp/src/Entity/Problem.php
```diff
@@ -287,26 +292,84 @@ public function getSpecialCompareArgs(): ?string
...
     {
-        $this->isMultipassProblem = $isMultipassProblem;
+        $types = array_unique($types);
+        $this->types = 0;
+        foreach ($types as $type) {
+            $this->types |= $type;
+        }
+        if (!($this->types & self::TYPE_PASS_FAIL) xor ($this->types & self::TYPE_SCORING)) {
+            throw new Exception("Invalid problem type: must be exactly one of 'pass-fail' or 'scoring'.");
+        }
```
mpsijm, line 343: Note that `pass-fail` is implied when it is not provided. For example, `type: interactive` implies `type: [pass-fail, interactive]`. I noticed this when doing a test upload today :joy: 

See also: https://github.com/Kattis/problem-package-format/issues/440

EDIT: Looks like this was fixed in [`2b7adab` (#2999)](https://github.com/DOMjudge/domjudge/pull/2999/commits/2b7adab1d06c4e89600b257e01ed5107c76b0c18), thanks! :heart: 

It's now also in the spec, see https://github.com/Kattis/problem-package-format/pull/471 :smile:

PR #2980, opened 2025-03-29
Only show warning on jury problemset page before contest start.

meisterT:


vmcj (approved):

nickygerritsen (approved):

webapp/templates/jury/problemset.html.twig
```diff
@@ -11,6 +11,6 @@
         problem_statement_path: 'jury_problem_statement',
         problem_attachment_path: 'jury_attachment_fetch',
         problem_sample_zip_path: 'jury_problem_sample_zip',
-        show_jury_warning: true
+        show_jury_warning: not current_contest.state or not current_contest.state.started
```
vmcj, line 14:
```suggestion
        show_jury_warning: not current_contest or not current_contest.state or not current_contest.state.started
```

This still failed for when there is no contest active, I wonder if ` not current_contest or not current_contest.state.started` should also be enough. So if any current_contest will also have the state already. Looking at the `getState()` I think it does. But I think I prefer to protect against this without this extra knowledge and go for the first suggestion.

```suggestion
        show_jury_warning: not current_contest or not current_contest.state.started
```

meisterT: Yeah, I simplified this condition even more.

PR #2983, opened 2025-04-07
Improve shadowing from PC2

nickygerritsen:
- When no (local) API info is found, don't crash the UI.
- For local contest source, check for local submission files in two known places.
- Only set problem color from RGB when preset.

### DIFF
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -200,7 +200,7 @@ class ExternalContestSourceService
             throw new LogicException('The contest source is not valid');
         }
 
-        return $this->cachedApiInfoData->version;
+        return $this->cachedApiInfoData?->version;
     }
 
     public function getApiVersionUrl(): ?string
@@ -209,7 +209,7 @@ class ExternalContestSourceService
             throw new LogicException('The contest source is not valid');
         }
 
-        return $this->cachedApiInfoData->versionUrl;
+        return $this->cachedApiInfoData?->versionUrl;
     }
 
     public function getApiProviderName(): ?string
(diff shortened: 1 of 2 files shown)


webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -1017,7 +1017,7 @@ protected function validateAndUpdateProblem(Event $event, EventData $data): void
             );
             $contestProblem->setShortname($data->label);
         }
-        if ($contestProblem->getColor() !== ($data->rgb)) {
+        if ($data->rgb && $contestProblem->getColor() !== $data->rgb) {
```
Kevinjil, line 1020: Why is this change made? The RGB data is nullable and should be in the notification if it is not null?

nickygerritsen: PC2 sends `color` but not `rgb`. In that case we made everything white, which doesn't seem OK. We **could** use `color` in that case, but that would change our colors (more than) slightly, so that didn't feel good either?

Kevinjil: If I read https://ccs-specs.icpc.io/2023-06/contest_api#problems correctly, the presence of `color` should imply the existence of the `rgb` value, as it is merely the human-readable description of the `rgb` field.

nickygerritsen: I don't think anything in that file specifies either or none are required.

https://ccs-specs.icpc.io/2023-06/ccs_system_requirements#contest-api does however. I will check with PC^2, but I guess they won't be sending it anytime soon.

nickygerritsen: Regardless, sending rgb *is* optional in the spec, so I'd say if we get no color but we do have one, it's better to stick to our color

nickygerritsen: If the CCS we are shadowing from would not support `rgb` but we do (which could be a valid scenario), I don't think it makes sense to remove our value. But maybe I'm in the minority here and if so I'm OK with reverting this change.

nickygerritsen: Turns out it was a bug in the YAML (an unquoted rgb string starting with `#` which PC2 saw as a comment and we somehow_ parsed as a string, not sure how). I'm fine reverting this.

Kevinjil (approved):

PR #2992, opened 2025-04-22
Install awk on Fedora.

meisterT:


.github/workflows/autoconf-check-different-distro.yml
```diff
@@ -18,6 +18,8 @@ jobs:
     steps:
       - name: Install git so we get the .github directory
         run: dnf install -y git
```
vmcj, line 20:
```suggestion
      - name: Install needed tools
        run: dnf install -y git awk
```

vmcj: `dnf` is a lot faster if it can do everything in 1 pass,

PR #2993, opened 2025-04-30
Shadow fixes

nickygerritsen:
These are fixes I made when doing the smoke test with PC2 for NAC.

### DIFF
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -1435,6 +1435,10 @@ class ExternalContestSourceService
                 $zipUrl = $data->files[0]->href;
                 if (preg_match('/^https?:\/\//', $zipUrl) === 0) {
                     // Relative URL, prepend the base URL.
+                    // If the base URL ends with a slash and the zip URL starts with one, remove the slash.
+                    if (str_ends_with($this->basePath, '/') && str_starts_with($zipUrl, '/')) {
+                        $zipUrl = substr($zipUrl, 1);
+                    }
                     $zipUrl = ($this->basePath ?? '') . $zipUrl;
                 }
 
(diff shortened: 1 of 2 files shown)


webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -833,6 +833,10 @@ protected function validateLanguage(Event $event, EventData $data): void
             ]);
         } else {
             $this->removeWarning($event->type, $data->id, ExternalSourceWarning::TYPE_DATA_MISMATCH);
+
+            $toCheck = ['extensions' => $data->extensions];
```
meisterT, line 836: Aren't the extensions standardized?

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -1468,26 +1468,31 @@ protected function importSubmission(Event $event, EventData $data): void
...
-                            if ($response->getStatusCode() !== 200) {
-                                // TODO: Retry a couple of times.
+                        $tries = 1;
+                        do {
+                            try {
+                                $response = $this->httpClient->request('GET', $zipUrl);
+                                $ziphandler = fopen($zipFile, 'w');
+                                if ($response->getStatusCode() !== 200) {
+                                    $this->addOrUpdateWarning($event, $data->id, ExternalSourceWarning::TYPE_SUBMISSION_ERROR, [
+                                        'message' => "Cannot download ZIP from $zipUrl after trying $tries times",
```
meisterT, line 1478: Is this the same error we saw with the CDS a couple years back, where they expose a submission before they expose the file?

webapp/src/Service/ExternalContestSourceService.php
meisterT: No, it is fine to take them from the primary. We should just make sure we publish the right extensions in any docs.

webapp/src/Service/ExternalContestSourceService.php
nickygerritsen: Basically yes. PC2 fixed it on their side now but I think hardening doesn't hurt

nickygerritsen: And that the CCS'es use said extensions 🫠

webapp/src/Service/ExternalContestSourceService.php
nickygerritsen: And that the CCS'es use said extensions 🫠

Issue #2519, opened 2024-05-01
Implement optimization problems


meisterT:
* This is meant for problems where there is no "correct" answer.
* The grader spits out a (float) number that is representative how "good" the submission is.
* The maximum of all submissions for that problem is taken as points for this problem.

(There could be variant where you take the minimum instead, but I would argue that the problem/grader could be changed so that it's always the maximum.)

It's not really clear to me how the Kattis problem format handles these, e.g. what is the spec for how to output points. Example problems welcome!

meisterT:
Possible future extensions:
- allow combining with partial scoring: https://github.com/DOMjudge/domjudge/issues/2518
- allow combining with interactive problems

as6325400:
I think we can discuss how to implement this?

When I took the data structures course, there was a professor who would always include NP-hard problems in the assignments for students to compete on.

At that time, the professor had to create an additional form to track everyone's scores.

If we had a feature that automatically updates the scoreboard with the scores, it would be much more convenient.

meisterT:
Does anyone have example problems that fit into this category?

meisterT:
I consider this one complete (except the more general remaining issues mentioned in #2518).

PR #2998, opened 2025-05-21
Implement optimization problems

as6325400:
As part of a feature I needed for my class, I went ahead and implemented it myself. I'm not sure if this is something the DOMjudge team would be interested in adopting, but I thought I'd share it in case it can serve as a useful reference.

Issue: https://github.com/DOMjudge/domjudge/issues/2519

To use this feature, simply have the compare script output to stderr a line starting with `OPT_SCORE=`. Currently, I'm storing the value as a float, so both integers and decimals are supported.

截圖 2025-05-21 晚上9 24 44
截圖 2025-05-21 晚上9 30 26

On the judgehost side, the score for each testcase is recorded in the judging_run table and then propagated to scorecache and rankcache.

To enable using the optscore as a contest tiebreaker, go to the contest settings and set `"Opt score as score tiebreaker" `to yes. This will reveal two options:

- Larger is best
- Smaller is best

You can then configure whether a higher or lower score is considered better for optimization problems.

截圖 2025-05-21 晚上9 22 30
截圖 2025-05-21 晚上9 23 00

Note: It is asserted that Runtime as score tiebreaker cannot be used at the same time as Opt score as score tiebreaker.
截圖 2025-05-21 晚上9 45 46

In the team interface, if Opt score as score tiebreaker is enabled, the optscore will be shown in the submission list for each submission. For problems that do not report an optscore, a value of 0 will be shown on the scoreboard and - in the submission list.
截圖 2025-05-21 晚上9 35 57

Regardless of whether the tiebreaker setting is enabled, jury members will always be able to see the optscore for each submission.
截圖 2025-05-21 晚上9 39 05

Standard problems (without optimization scores) are unaffected by this feature.
截圖 2025-05-21 晚上9 39 45

This implementation is essentially a minimum viable product (MVP). I’m not sure if outputting OPT_SCORE= directly in the compare script is the best long-term approach, but it works well for now.

Since my class is actively using this judge system, I’ve implemented the changes on version 8.3.1. If this feature is of interest, I can port the code to the main branch.

I’ve also set up a live demo judge at: https://domjudge.as6325400.org/
problem: https://domjudge.as6325400.org/public/problems/4/statement

If you’d like to try it out, feel free to message me on Slack for a login account.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,7 +19,7 @@ services:
     volumes:
       - /var/lib/mysql
   domjudge:
-    image: docker.io/domjudge/domjudge-contributor
+    image: as6325400/domjudge-contributor:8.3.1
     hostname: domjudge-contributor
     volumes:
       - /sys/fs/cgroup:/sys/fs/cgroup
--- a/etc/db-config.yaml
+++ b/etc/db-config.yaml
@@ -244,6 +244,11 @@
                 2: always
             regex: /^\d+$/
             error_message: A value between 0 and 2 is required.
+        -   name: show_test_results
+            type: bool
+            default_value: false
+            public: true
+            description: Show what kinds of testcases to user?
         -   name: show_sample_output
             type: bool
             default_value: false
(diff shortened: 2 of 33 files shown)


nickygerritsen:
As discussed on Slack we are working on our own version. Thanks for the code though, we might use it for some inspiration.

PR #2999, opened 2025-05-22
Refactor YAML parsing when importing problems

meisterT:
This allows us to add a unit test. No functional changes intended, these will be added later.


meisterT:
Reviewing each commit individually should make this easier than the PR as a whole

PR #3001, opened 2025-05-23
When none of pass-fail, scoring is given, add pass-fail as default.

nickygerritsen:


meisterT:
See my draft PR for a more complete fix, still need to do some minor stuff

nickygerritsen:
PR #2999 is better, closing this one.

PR #3002, opened 2025-05-23
Add endpoint to fetch submissions by internal ID for domlogo

nickygerritsen:


nickygerritsen:
The other alternatives are:
* Add an endpoint that converts the external submission ID to the internal submission ID.
* Add the internal submission ID to the `Working directory: ` line in the judgedaemon.

vmcj:
Why do we actually get this problem? I would prefer not exposing our "internal" id anywhere.

So I think option1 is better, or making sure our internal ID is also the external ID, to remove such doublings.

nickygerritsen:
Another option is to use external ID's in the folder structure on the server, then it automatically works.

I remember we had a discussion we didn't want this, but I'm not sure why anymore.

vmcj:
I think `domlogo.py` should work mostly as "external tool", so I would prefer that the judgedaemon only exposes parts as `external` to stay in line. So I think we should reconsider that we just use the externalid.

But @meisterT / @eldering might know what was this discussion?

meisterT:
Is there any limitation on the character set of external IDs?

nickygerritsen:
> Is there any limitation on the character set of external IDs?

> `[a-zA-Z0-9_.-]` of length at most 36 and not starting with a `-` (dash) or `.` (dot) or ending with a `.` (dot)

Sounds like we can use them as folder names easily.

nickygerritsen:
After discussing IRL we will move to using external IDs in the judgehost API. I will implement this.

PR #3004, opened 2025-05-25
When we shadow, trust the filenames and file sizes.

nickygerritsen:


meisterT (approved):

nickygerritsen:
I removed it from the merge queue for now. We need to think about it maybe more, but currently we can't import these submissions and thus have differences.

Maybe we should import them as `import-error`, which should make them not judge in DOMjudge.

nickygerritsen:
In fact I like that: if one of the things from the shadow don't follow our settings, mark it as import error and don't create a judging. Wdyt?

vmcj (approved):
Let's also consider making it an enum.

vmcj (approved):

Issue #2687, opened 2024-09-16
"Deactivate now" button still visible after deactivation


meisterT:
On an individual contest page, e.g. `/jury/contests/2` we show buttons for actions such as unfreeze, finalize, deactivate, etc. After the action happened (either by button press or time has passed) we no longer show it.

This seems to work for all buttons, but the deactivate one.

misha-khar:
Going to try to work on this as one of my first contributions!

meisterT:
Great, let us know if you have any questions: https://www.domjudge.org/chat

as6325400:
> Going to try to work on this as one of my first contributions!

Hi! I noticed there hasn't been any activity on this issue for a while. If it's alright, I'd love to take a stab at it. Let me know if anyone is still working on it — otherwise, I'll go ahead and give it a try.

PR #3008, opened 2025-06-14
hide “Deactivate now” button once deactivated

as6325400:
Fixes https://github.com/DOMjudge/domjudge/issues/2687

Use getDeactivatetime() to check whether the `Deactivate time` field is set, and decide accordingly whether to display the `Deactivate now` button.


nickygerritsen:
Now the button doesn't show if a deactivate time is set, even if it is in the future. Is that what we want? I'd say we only hide it if the deactivate time is in the past?

as6325400:
> Now the button doesn't show if a deactivate time is set, even if it is in the future. Is that what we want? I'd say we only hide it if the deactivate time is in the past?

Sorry, I didn’t notice that—thanks for pointing it out! I’ve updated the condition so that the button only hides if the deactivation time is in the past.

PR #3013, opened 2025-07-04
Fix all PHPStan reported errors

nickygerritsen:


vmcj:
Setting it as draft so we can't merge it,

Hijacking it to fix the other PR.

Issue #2252, opened 2023-11-25
Push notification for submission results


mpsijm:
See title :slightly_smiling_face: (@vmcj approves :+1:)

eldering:
Clarifying the title: we do have push notifications for clarifications and also for some jury things, but not for team receiving the result of a submission. A team complained about this at NWERC 2023 and said that this was working at BAPC. We could indeed reproduce that notifications about clarifications work, but not for submission results.

However, looking at the code at https://github.com/DOMjudge/domjudge/blob/main/webapp/public/js/domjudge.js, there is no call to `sendNotification` for submission results, so this has never worked. It would be a good feature though.

Scarletto:
I would like to try to implement this

meisterT:
Great, let us know (e.g. via https://domjudge-org.slack.com/) if you need any help.

vmcj:
> I would like to try to implement this

@Scarletto are you still planning to work on this? Anything we can help with?

nickygerritsen:
Semi related to #1840

PR #3018, opened 2025-07-05
feat: add submission notification

cubercsl:
Related to #2252, just for a POC now.

Lack of testing, and I'm not sure how much impact such a query would have on performance when there are a large number of teams polling for updates.

### DIFF
--- a/webapp/src/Service/DOMJudgeService.php
+++ b/webapp/src/Service/DOMJudgeService.php
@@ -321,6 +321,53 @@ class DOMJudgeService
     }
 
     /**
+     * @return array<array{submissionid: int, judgingid: int, body: string}>
+     */
+
+    public function getUnreadJudgements(): array
+    {
+        $user = $this->getUser();
+        $team = $user->getTeam();
+        $contest = $this->getCurrentContest($team->getTeamId());
+        $unreadJudgements = [];
+        if ($contest === null) {
+            return $unreadJudgements;
+        }
+
+        $queryBuilder = $this->em->createQueryBuilder()
+            ->from(Judging::class, 'j')
+            ->select('j, c, s')
+            ->leftJoin('j.contest', 'c')
+            ->leftJoin('j.submission', 's')
+            ->groupBy('j.judgingid')
+            ->orderBy('j.judgingid')
+            ->andWhere('j.contest = :cid')
+            ->setParameter('cid', $contest->getCid())
+            ->andWhere('j.result IS NOT NULL')
+            ->andWhere('s.team = :team')
+            ->setParameter('team', $team)
+            ->andWhere('s.submittime < c.endtime')
+            ->andWhere('j.valid = 1')
+            ->andWhere('j.seen = 0');
+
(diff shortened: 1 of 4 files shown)


webapp/src/Service/DOMJudgeService.php
```diff
@@ -320,6 +320,53 @@ public function getUnreadClarifications(): array
...
+            ->groupBy('j.judgingid')
+            ->orderBy('j.judgingid')
+            ->andWhere('j.contest = :cid')
+            ->setParameter('cid', $contest->getCid())
+            ->andWhere('j.result IS NOT NULL')
+            ->andWhere('s.team = :team')
+            ->setParameter('team', $team)
+            ->andWhere('s.submittime < c.endtime')
+            ->andWhere('j.valid = 1')
+            ->andWhere('j.seen = 0');
```
meisterT, line 353: is there a migration missing? `seen` doesn't exist yet, does it?

cubercsl: https://github.com/DOMjudge/domjudge/blob/4398ae5b89796c7bd5f13d94d955454edbad2638/webapp/src/Entity/Judging.php#L118-L120

But this may not be important, because the logic of whether to send notifications does not look at this field, (which is at `sendNotification` @ `domjudge.js`)

cubercsl: BTW, this SQL query is the same as the one a user access `/api/contests/:cid/judgements`, except it filters out entries they have ever seen (it will be marked as seen when they click the submission). 

Therefore, it is also feasible to implement it directly on the frontend js instead of adding a function to the `/update` endpoint.

Dup4 (approved):

cubercsl:
Have tested in some ICPC Asia East Reginal Contests this month.

Here is a reaction desktop video for selected team open this feature.

[reaction-desktop.webm](https://github.com/user-attachments/assets/95bcda86-5815-4e32-b8c8-2bed846c5452)

webapp/src/Service/DOMJudgeService.php
nickygerritsen: When are we setting the `seen` to true? I assume we already do this.

cubercsl: > When are we setting the `seen` to true? I assume we already do this.

https://github.com/DOMjudge/domjudge/blob/4a991717abdabaad68aa203135b04d9110f7312a/webapp/templates/team/partials/submission_list.html.twig#L33-L49

When click submission and see detail in team home page.

BTW: We track notification sent status by cookie. It might be unnecessary to use `seen` for filtering in this place.

cubercsl:
After discussing with some of my friend yesterday, we think that we can provide notifications only for API submissions. Most of the contestants who need notifications use API submissions. Contestants who submit through the web almost always just refresh the page to check the results.

Since we have already maintained the submission source (via API or web) in the submission, this is also easy to do. If someone thinks this is reasonable, I can update this PR.

PR #3020, opened 2025-07-05
Doctype fixes for new PHPstan

nickygerritsen:


webapp/src/Controller/Team/SubmissionController.php
```diff
@@ -78,7 +78,7 @@ public function createAction(Request $request, ?Problem $problem = null): Respon
                 $problem = $form->get('problem')->getData();
                 /** @var Language $language */
                 $language = $form->get('language')->getData();
-                /** @var UploadedFile[] $files */
+                /** @var UploadedFile[]|UploadedFile $files */
```
vmcj, line 81: Shouldn't we fix somewhere that we always get an array instead of this behaviour?

nickygerritsen: We allow both since when sourcefiles_limit=1 we show one, otherwise multiple.

PR #3025, opened 2025-07-06
Implement analyst evaluation mode

Kevinjil:
The idea behind this mode is that we are not really shadowing and trust the event feed results. However, we want to judge "interesting" runs locally to get useful information without the judging capacity to judge all testcases due to limited judgehost assignment.

We do not consider 'TLE' or 'AC' interesting, as rerunning will not yield much more information. We consider 'WA' very interesting and prioritize the judging, but allow manual judging to overtake the priority. We consider 'CE' somewhat interesting, but downprioritize them a lot. For other verdicts, keep the normal priority.


webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -1839,9 +1840,50 @@ protected function importRun(Event $event, EventData $data): void
         if ($persist) {
             $this->em->persist($run);
         }
+
+        $lazyEval = $this->config->get('lazy_eval_results');
+        if ($lazyEval === DOMJudgeService::EVAL_ANALYST) {
+            // Check if we want to judge this testcase locally to provide useful information for analysts
+            $priority = $this->getAnalystRunPriority($run);
+            if ($priority !== null) {
+                // Make the judgetask valid and assign running priority if no judgehost has picked it up yet.
```
vmcj, line 1869: What do we do if someone has pressed `judge submission` this might increase the order?


I think we should revisit the `judge submission` for this mode and add those with another prio?

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -1076,7 +1076,9 @@ private function addSingleJudgingRun(
                     throw new BadMethodCallException('internal bug: the evaluated result changed during judging');
                 }
 
-                if ($lazyEval !== DOMJudgeService::EVAL_FULL) {
+                if ($$lazyEval === DOMJudgeService::EVAL_ANALYST) {
```
tuupke: Is `$$lazyEval` intentional? Doesn't look right to me.

tuupke (changes requested):

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -1858,9 +1860,59 @@ protected function importRun(Event $event, EventData $data): void
...
+     * @return int The judging priority if it should be run locally, null otherwise.
+     */
+    protected function getAnalystRunPriority(ExternalRun $run): int | null {
+        return match ($run->getResult()) {
+            // We will not get any new useful information for TLE testcases, while they take a lot of judgedaemon time.
+            'timelimit' => null,
+            // We often do not get new useful information for judging correct testcases.
+            'correct' => null,
+            // Wrong answers are interesting for the analysts, assign a high priority but below manual judging.
+            'wrong-answer' => -9,
```
tuupke: Probably better to use the constants here. Otherwise there is a hidden dependency between the 'actual' constants, and these constants. So instead of `-9` use `JudgeTask::PRIORITY_HIGH + 1` and instead of `9` use `JudgeTask::PRIORITY_LOW - 1`.

webapp/src/Service/ExternalContestSourceService.php
vmcj: I think my point was that the `judge submission` should also have been added with high prio. But as the system could easily keep up we can leave this as is.

PR #3029, opened 2025-07-06
Demonstrate PiP for team elements

vmcj:
This allowes a team to move part of our interface over their editor. It's still a bit hacky but we should think about:
- Do we want this
- Which info to display

And a screenshot:
image


eldering:
I think it's interesting to offer in a similar way as notifications. The most relevant things for it, I think are:

- updates on your submissions
- new clarifications
- maybe contest timer

PR #3030, opened 2025-07-06
Demo using bootstrap in team page

vmcj:
Discussed with @nickygerritsen, a small demo on this idea. I think this looks a bit better and without it taking up valuable space.

I duplicated the old and the new stuff to show the difference, I think we can create more space by moving the `request clarification` button to the header. Not sure if we want to bring back the old color scheme? Next step would be to take away the clarification lists padding (and see if we have the same layout in the jury interface). 

image

### DIFF
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -617,7 +617,7 @@ function enableRefresh($url, $after, usingAjax) {
         }
     };
     refreshHandler = setTimeout(refresh, $after * 1000);
-    refreshEnabled = true;
+    refreshEnabled = false;
     setCookie('domjudge_refresh', 1);
 
     if(window.location.href == localStorage.getItem('lastUrl')) {
--- a/webapp/templates/base.html.twig
+++ b/webapp/templates/base.html.twig
@@ -48,17 +48,6 @@
     {% endif %}
     var markdownPreviewUrl = "{{ path('markdown_preview') }}";
     $(function () {
-        /* toggle refresh if set */
-        {% if refresh is defined and refresh %}
-        $('#refresh-navitem').on('click', function () {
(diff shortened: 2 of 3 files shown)


eldering:
I think the old one (bottom with blue headers) looks cleaner actually. Is there a technical reason to change this, or just to have the standard bootstrap layout?

nickygerritsen:
I do like the blue, but maybe we can use the cards but then with a blue header? I would also like to see us use the default bootstrap table styles, just to see how it looks.

I'm not sure yet what's better, but I do think the bootstrap style is more consistent.

vmcj:
There is no technical reason, it was mostly to see if we can get more consistent between the pages.

I tried with the blue in the header and moving the button for clarification requests:
image

I centered only one to show the difference, I think centered is better though.

eldering:
> I centered only one to show the difference, I think centered is better though.

I don't have a strong preference on centering, although if the request clarification button goes on the right side in the header, then I think the title should be left-aligned.

I think the blue header is a bit better, especially because there's less clutter from the margins around the tables. I also think that the table titles in the original are a bit clearer because of the font (including size/thickness).

meisterT:
Multiple suggestions to try out, which are all slightly unrelated to the PR, but if we do them, I think it will look nicer (both without, but especially with this PR):

- the striped tables don't provide a lot of additional value (anymore?), and I think the gray bites with table header - I think we should try removing `table-striped` from these tables, at least on the team side, perhaps later also in general
- I think the bold-ing of the submissions that have not been clicked is unnecessary - there is no real additional content that we want teams to acknowledge. I would remove it completely, it doesn't carry any value. We probably should use the unread column in the database for notifications instead
- can we try out one card for submissions and one card for clarifications (both from the jury and requests)?
- we should remove the from/to columns for the clarification side and also remove the word "problem" in the subject column. from/to makes sense for jury interface but not so much for team

Issue #3034, opened 2025-07-07
Judgehost stuck: deadlock in runguard


rry-je:
### Description of the problem
During GCPC, one of our judgehosts got stuck in a deadlock and stopped judging submissions.

### Your environment
DOMjudge version: `b32fc5c8c5a84160e6e26203228fbbe1ec8444e9` with `028995f9c00e7897ec863283986ef995661e38b9` cherry-picked on top of that
Operating system / Linux distribution and version: Debian GNU/Linux 12 (bookworm), 6.1.0-27-amd64
Webserver: nginx/1.22.1

### Steps to reproduce
We don't have a reproducer. Probably timing-dependent

### Expected behaviour
Judgehost should judge submission and continue judging afterwards.

### Actual behaviour
Judgehost stops judging and the web interface shows a warning.

### Any other information that you want to share?
Here's a stacktrace from the runguard process:

```gdb
(gdb) bt
#0  0x00007fa4dc0a50d6 in __lll_lock_wait_private () from target:/lib/x86_64-linux-gnu/libc.so.6
#1  0x00007fa4dc07da45 in ?? () from target:/lib/x86_64-linux-gnu/libc.so.6
#2  0x00007fa4dc135a7f in __fprintf_chk () from target:/lib/x86_64-linux-gnu/libc.so.6
#3  0x000055cc552c6621 in fprintf (__fmt=0x55cc552cb004 "%s: warning: ", __stream=<optimized out>) at /usr/include/x86_64-linux-gnu/bits/stdio2.h:79
#4  warning (format=format@entry=0x55cc552cbc08 "timelimit exceeded (hard wall time): aborting command") at runguard.cc:203
#5  0x000055cc552c6d8d in terminate (sig=14) at runguard.cc:693
#6  <signal handler called>
#7  0x00007fa4dc07d9ff in ?? () from target:/lib/x86_64-linux-gnu/libc.so.6
#8  0x00007fa4dc135a7f in __fprintf_chk () from target:/lib/x86_64-linux-gnu/libc.so.6
#9  0x000055cc552c6621 in fprintf (__fmt=0x55cc552cb004 "%s: warning: ", __stream=<optimized out>) at /usr/include/x86_64-linux-gnu/bits/stdio2.h:79
#10 warning (format=format@entry=0x55cc552ccf28 "timelimit exceeded (hard cpu time)") at runguard.cc:203
[...]

PR #3035, opened 2025-07-07
Remove functions that are not async-signal-safe from signal handler.

meisterT:
Fixes https://github.com/DOMjudge/domjudge/issues/3034


judge/runguard.cc
```diff
@@ -226,6 +227,22 @@ void verbose(const char *format, ...)
...
+void verbose_from_signalhandler(const char* msg)
+{
+	if (!be_quiet) {
+		write(STDERR_FILENO, msg, strlen(msg));
+	}
+}
+
+void warning_from_signalhandler(const char* msg)
+{
+	if (!be_quiet && be_verbose) {
```
eldering: The `be_verbose` should be part of the condition in `verbose_from_signalhandler`, not here.

judge/runguard.cc
```diff
@@ -226,6 +227,22 @@ void verbose(const char *format, ...)
 	va_end(ap);
 }
 
+void verbose_from_signalhandler(const char* msg)
```
eldering, line 233: Maybe good to add a general comment here why we have these `{verbose,warning}_from_signalhandler` functions?

judge/runguard.cc
```diff
@@ -674,43 +691,45 @@ void terminate(int sig)
...
 	}
 
 	received_signal = sig;
 
 	/* First try to kill graciously, then hard.
 	   Don't report an already exited process as error. */
-	verbose("sending SIGTERM");
+	verbose_from_signalhandler("sending SIGTERM");
 	if ( kill(-child_pid,SIGTERM)!=0 && errno!=ESRCH ) {
-		error(errno,"sending SIGTERM to command");
```
eldering, line 704: I think it would be good to keep a warning message with "sending SIGTERM to command" so that at least we get some more info what went wrong if it does. Also below.

meisterT: Done

judge/runguard.cc
meisterT: Done

judge/runguard.cc
meisterT: How embarrassing... fixed.

eldering (approved):

PR #3038, opened 2025-07-13
Fixes to API access endpoint and update some exposed properties

eldering:
List all properties (also DOMjudge specific ones) in the [access](https://ccs-specs.icpc.io/draft/contest_api#access) API endpoint, and add endpoints `accounts` and `clarifications` there.

Also remove exposing `problems.types_to_string` and add `judgements.current` (as alias of `valid`).

### DIFF
--- a/webapp/src/Controller/API/AccessController.php
+++ b/webapp/src/Controller/API/AccessController.php
@@ -100,9 +100,13 @@ class AccessController extends AbstractApiController
                         'name',
                         'formal_name',
                         'start_time',
+                        'countdown_pause_time',
                         'duration',
                         'scoreboard_freeze_duration',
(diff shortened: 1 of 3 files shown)


nickygerritsen:
I already had a local change for `current` that only exposes it if you enable the new api and removes it from the event feed as well, but if we merge this I can rebase and change my code

eldering:
I can also remove `current` from my PR. I guess that's easier for you?

nickygerritsen:
> I can also remove `current` from my PR. I guess that's easier for you?

Sure that would work indeed

eldering:
> > I can also remove `current` from my PR. I guess that's easier for you?
> 
> Sure that would work indeed

Done.

PR #3040, opened 2025-07-20
Fix user add and update api endpoints

LuukBlankenstijn:
Found some bugs while using the API, and since I need the endpoints, fixed them.

Problem I encountered:
Endpoint /api/v4/users/{id} PUT request returned{ 400, Body data exceeded php.ini's 'post_max_size' directive (currently set to 512M)}.
Could not get it to work whatever I tried. When looking at the code I also discovered some other issues in the case of updating the user with App\Controller\API\UserController.php addOrUpdateUser(). 
What I did is changing the content type of the api-doc attribute to application/json, and separated the add and update method in to dedicated methods. I realize this is a breaking change in the api spec and not ideal.

I got the errors on both the DOMjudge instance running on the GEHACK server in Eindhoven, and my own local instance.

This exception is generated in App\EventListener\BodyToBigListener.php. This happens because the amount of parameters on the request is 0. (I get this even when using /api/doc on a running domjudge instance). I first tried fixing the request, but could not get it to work. Because it made more sense to me for a request like this to accept json instead of formdata I decided to change it to that. After I did this the request now worked, but i found some flaws in the update logic for user. 

For example: name, username and id are required to even make an update by the documentation. I would have expected to just need an id and being able to update any field. Then, on line 150 in App\Controller\Api\Usercontroler:
``` 
protected function addOrUpdateUser(AddUser $addUser, Request $request): Response
    {
        if ($addUser instanceof UpdateUser && !$addUser->id) {
            throw new BadRequestHttpException('`id` field is required');
        }

        if ($this->em->getRepository(User::class)->findOneBy(['username' => $addUser->username])) {
            throw new BadRequestHttpException(sprintf("User %s already exists", $addUser->username));
        }
        $user = new User();
              if ($addUser instanceof UpdateUser) {
                  $existingUser = $this->em->getRepository(User::class)->findOneBy(['externalid' => $addUser->id]);
                  if ($existingUser) {
                      $user = $existingUser;
                  }
              }

        $user
            ->setUsername($addUser->username)
            ->setName($addUser->name)
            ->setEmail($addUser->email)
            ->setIpAddress($addUser->ip)
            ->setPlainPassword($addUser->password)
            ->setEnabled($addUser->enabled ?? true);

       // rest of method
}
```
So if the user already exists we get a BadRequestException. This effectively means that for every update we want to do to a user, the username has to be changed.
Then also we cannot do partial update by leaving fields in the request null, since all properties of the UpdateUser dto are assigned to the existing user, even when null.

### DIFF
--- a/webapp/src/Controller/API/UserController.php
+++ b/webapp/src/Controller/API/UserController.php
@@ -23,7 +23,10 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Attribute\MapRequestPayload;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\Security\Http\Attribute\IsGranted;
+use Symfony\Component\Validator\Exception\ValidationFailedException;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
 
 /**
  * @extends AbstractRestController<User, User>
@@ -41,7 +44,8 @@ class UserController extends AbstractRestController
         DOMJudgeService $dj,
         ConfigurationService $config,
         EventLogService $eventLogService,
-        protected readonly ImportExportService $importExportService
+        protected readonly ImportExportService $importExportService,
+        protected readonly ValidatorInterface $validator,
     ) {
         parent::__construct($entityManager, $dj, $config, $eventLogService);
     }
@@ -86,9 +90,9 @@ class UserController extends AbstractRestController
         $message = null;
         $result = -1;
         if ((($tsvFile && ($result = $this->importExportService->importTsv('groups', $tsvFile, $message))) ||
-             ($jsonFile && ($result = $this->importExportService->importJson('groups', $jsonFile, $message)))) &&
+            ($jsonFile && ($result = $this->importExportService->importJson('groups', $jsonFile, $message)))) &&
             $result >= 0) {
-             return "$result new group(s) successfully added.";
+            return "$result new group(s) successfully added.";
         } else {
             throw new BadRequestHttpException("Error while adding groups: $message");
         }
@@ -171,7 +175,7 @@ class UserController extends AbstractRestController
         }
         $message = null;
         if ((($tsvFile && ($result = $this->importExportService->importTsv('teams', $tsvFile, $message))) ||
-             ($jsonFile && ($result = $this->importExportService->importJson('teams', $jsonFile, $message)))) &&
+                ($jsonFile && ($result = $this->importExportService->importJson('teams', $jsonFile, $message)))) &&
             $result >= 0) {
             // TODO: better return all teams here?
             return "$result new team(s) successfully added.";
@@ -235,7 +239,7 @@ class UserController extends AbstractRestController
 
         $message = null;
         if ((($tsvFile && ($result = $this->importExportService->importTsv('accounts', $tsvFile, $message))) ||
-             ($jsonFile && ($result = $this->importExportService->importJson('accounts', $jsonFile, $message)))) &&
+                ($jsonFile && ($result = $this->importExportService->importJson('accounts', $jsonFile, $message)))) &&
             $result >= 0) {
             // TODO: better return all accounts here?
             return "$result new account(s) successfully added.";
@@ -291,13 +295,12 @@ class UserController extends AbstractRestController
      * Add a new user.
      */
     #[IsGranted('ROLE_API_WRITER')]
-    #[Rest\Post]
+    #[Rest\Post()]
     #[OA\RequestBody(
         required: true,
         content: [
-            new OA\MediaType(
-                mediaType: 'multipart/form-data',
-                schema: new OA\Schema(ref: new Model(type: AddUser::class))
+            new OA\JsonContent(
+                ref: new Model(type: AddUser::class)
             ),
         ]
     )]
@@ -307,7 +310,7 @@ class UserController extends AbstractRestController
         content: new Model(type: User::class)
     )]
     public function addAction(
-        #[MapRequestPayload(validationFailedStatusCode: Response::HTTP_BAD_REQUEST)]
+        #[MapRequestPayload(validationFailedStatusCode: Response::HTTP_BAD_REQUEST, validationGroups: ['add'])]
         AddUser $addUser,
         Request $request
     ): Response {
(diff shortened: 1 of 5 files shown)


vmcj:
Can you explain which problem you encountered?

LuukBlankenstijn:
Updated the description

vmcj:
> Found some bugs while using the API, and since I need the endpoints, fixed them.
> 
> Problem I encountered: Endpoint /api/v4/users/{id} PUT request returned{ 400, Body data exceeded php.ini's 'post_max_size' directive (currently set to 512M)}.

Let's discuss in Slack how to fix this one, I assume you posted something which is larger than 512M? Did you raise the `post_max_size`? You can check for the value detected in the config checker.


> ... When looking at the code I also discovered some other issues in the case of updating the user with App\Controller\API\UserController.php addOrUpdateUser(). What I did is changing the content type of the api-doc attribute to application/json,

I can see that you did this, but I'm not sure yet why it wants a form but I suspect this is legacy.

> and separated the add and update method in to dedicated methods. I realize this is a breaking change in the api spec and not ideal.

So we're working on a major release so now would be the time to make such choices but let's see if we can prevent it still.

webapp/src/Controller/API/UserController.php
```diff
@@ -291,12 +291,12 @@ public function singleAction(Request $request, string $id): Response
      * Add a new user.
      */
     #[IsGranted('ROLE_API_WRITER')]
-    #[Rest\Post]
+    #[Rest\Put("")]
```
vmcj: I think it's fine to also allow PUT here, but dropping POST doesn't make sense to me.

webapp/src/DataTransferObject/AddUser.php
```diff
@@ -8,10 +8,9 @@
 #[OA\Schema(required: ['username', 'name', 'roles'])]
 class AddUser
 {
-    /**
-     * @param array<string> $roles
```
vmcj, line 12: PHPstan does not agree with removing this.

vmcj (changes requested):

vmcj:
@LuukBlankenstijn can you provide the files you were using which hit this error?

eldering:
I've not looked into the changes in detail (and I guess @nickygerritsen might be a better reviewer), but from the description it's not clear to me what these changes are fixing and why/how. Because this seems to have nothing to do with a maximum post size of a request. Can you clarify that?

LuukBlankenstijn:
I updated the description again, I'm sorry for lack initial lack of explanation

webapp/src/DataTransferObject/AddUser.php
LuukBlankenstijn: Yeah sorry, I forgot to run PHPstan

nickygerritsen:
I think we can do something smart to still have both update and create in one action, will think about it.

Although that is not very REST, so maybe we shouldn't?

As form json vs formdata, ideally it'd accept both, so I will investigate if we can get that to work.

LuukBlankenstijn:
> I think we can do something smart to still have both update and create in one action, will think about it.

I have a, imo, nice idea for this, I can implement it tommorrow if you want.

nickygerritsen:
> > I think we can do something smart to still have both update and create in one action, will think about it.
> 
> I have a, imo, nice idea for this, I can implement it tommorrow if you want.

Sounds good

eldering:
Some general comments:

- Our API follows the Contest API specification, specifically following https://ccs-specs.icpc.io/draft/contest_api#http-methods partial updates should only be allowed using PATCH, not PUT.
- Nitpick: we don't use semantic commit messages, so just as well leave `fix(api)` out
- Would be good to also write the details of what was broken and how it's changed/fixed in the commit message, so it gets recorded in git history.

LuukBlankenstijn:
Sorry for the delay, but I finally found some time to work on it. I cannot seem to figure out how to run the tests and how to fix code-style, any help on that would be appreciated.

webapp/src/Controller/API/UserController.php
```diff
@@ -1,4 +1,6 @@
-<?php declare(strict_types=1);
+<?php
+
+declare(strict_types=1);
```
vmcj: Any reason you wanted to change this? I don't think we want the newlines there.

webapp/src/Controller/API/UserController.php
```diff
@@ -86,9 +92,10 @@ public function addGroupsAction(Request $request): string
         $message = null;
         $result = -1;
         if ((($tsvFile && ($result = $this->importExportService->importTsv('groups', $tsvFile, $message))) ||
-             ($jsonFile && ($result = $this->importExportService->importJson('groups', $jsonFile, $message)))) &&
-            $result >= 0) {
-             return "$result new group(s) successfully added.";
+                ($jsonFile && ($result = $this->importExportService->importJson('groups', $jsonFile, $message)))) &&
```
vmcj: I don't think you change anything here. Can either first run your linter on the file and commit or remove the changes your linter made (my preference)?

webapp/src/Controller/API/UserController.php
```diff
@@ -335,58 +346,55 @@ public function addAction(
...
+        string $id,
         Request $request
     ): Response {
-        return $this->addOrUpdateUser($updateUser, $request);
+        /** @var User|null $user */
+        $user = $this->em->getRepository(User::class)->findOneBy(['externalid' => $id]);
+        if ($user === null) {
+            throw new NotFoundHttpException(sprintf("User with id %s not found", $id));
+        }
+        return $this->addOrUpdateUser($mutateUser, $user, $request);
```
vmcj: The name `addOrUpdate` is now misleading. You only allow for update here.

vmcj:
> Sorry for the delay, but I finally found some time to work on it. I cannot seem to figure out how to run the tests and how to fix code-style, any help on that would be appreciated.

I dropped you a message in the DOMjudge slack to help. The tests will only run when we press the button as you haven't merged anything yet.

LuukBlankenstijn:
This should fix everything except the tests, will try again on those later this week

vmcj:
> This should fix everything except the tests, will try again on those later this week

This is also a lot of formatting changes which makes discussing/reviewing a lot harder. Can you make a separate PR for those first as some are also a bit opinionated and would take away the discussion from what you're trying to achieve.

webapp/src/Controller/API/UserController.php
```diff
@@ -307,86 +310,86 @@ public function singleAction(Request $request, string $id): Response
...
-            ->setEnabled($addUser->enabled ?? true);
-
-        if ($addUser instanceof UpdateUser) {
-            $user->setExternalid($addUser->id);
+        if ($mutateUser->email !== null) {
+            $user->setEmail($mutateUser->email);
         }
-
-        if ($addUser->teamId) {
+        if ($mutateUser->ip !== null) {
```
Kevinjil, line 367: All these `if` conditions mean that we can no longer remove e.g. the IP address from an user using the API. For this, we would either still need a `PUT` endpoint or differentiate between the absence of properties in the request vs explicit `null` values.

vmcj: > All these `if` conditions mean that we can no longer remove e.g. the IP address from an user using the API. For this, we would either still need a `PUT` endpoint or differentiate between the absence of properties in the request vs explicit `null` values.

The last is not allowed in CLICS, leaving a property out or setting it to `null` are the same action.

cubercsl:
I have met some similar when updating user.

Minimal reproducible method in demoweb:

image

PR #3041, opened 2025-07-20
Changes for the 2026-01 CCS spec

### LINKED ISSUE #2946: Penalty time parsing from external data source in new draft spec

nickygerritsen:
This adds the new fields in the new API spec, as well as change some `int` fields to `reltime` when using that version.

Fixes #2946


webapp/src/Controller/API/AccessController.php
```diff
@@ -90,7 +94,11 @@ public function getStatusAction(Request $request): Access
         }
         if ($this->dj->checkrole('team') && $this->dj->getUser()->getTeam()) {
             $capabilities[] = 'team_submit';
-            $capabilities[] = 'team_clar';
+            if ($ccsApiVersion->usePostClar()) {
```
meisterT: What does `usePostClar` mean? Can we change it to a descriptive name?

webapp/src/DataTransferObject/ContestProblemWrapper.php
```diff
@@ -10,6 +11,12 @@ class ContestProblemWrapper
     public function __construct(
         #[Serializer\Inline]
         protected readonly ContestProblem $contestProblem,
+        #[Serializer\Groups([ARC::GROUP_NONSTRICT, '2025-draft'])]
```
meisterT: What does this annotation do?

webapp/src/Entity/Judging.php
```diff
@@ -335,6 +335,14 @@ public function getValid(): bool
         return $this->valid;
     }
 
+    // Note: we can't use CCSApiVersion::Format_2025_DRAFT->value in PHP 8.1, only in 8.2+
```
meisterT: We just set the min PHP version to 8.2+

webapp/src/Serializer/ContestVisitor.php
```diff
@@ -41,12 +42,14 @@ public function onPreSerialize(ObjectEvent $event): void
...
-            Contest::class,
-            'penalty_time',
-            null
-        );
-        $contest->setPenaltyTimeForApi((int)$this->config->get('penalty_time'));
+        $penaltyTime = $this->config->get('penalty_time');
+        /** @var CcsApiVersion $ccsApiVersion */
+        $ccsApiVersion = $this->config->get('ccs_api_version');
+        if ($ccsApiVersion->useRelTimes()) {
+            $contest->setPenaltyTimeForApi(Utils::relTime((int)$penaltyTime * 60, true));
```
meisterT: nit: add `floored:` before `true`

webapp/src/Service/EventLogService.php
```diff
@@ -855,4 +856,27 @@ public function endpointForEntity($entity): ?string
...
+    public function applyCcsVersionChanges(string $endpointType, array $event): array
+    {
+        /** @var CcsApiVersion $ccsApiVersion */
+        $ccsApiVersion = $this->config->get('ccs_api_version');
+
+        if ($endpointType == 'contests') {
+            $penaltyTime = $event['penalty_time'];
+            $penaltyTimeIsRelative = is_string($penaltyTime) && Utils::isRelTime($penaltyTime);
+            if ($ccsApiVersion->useRelTimes() && !$penaltyTimeIsRelative) {
+                $event['penalty_time'] = Utils::relTime($penaltyTime * 60, true);
```
meisterT: same as above

webapp/src/DataTransferObject/ContestProblemWrapper.php
nickygerritsen: It only exposes it for those serializer groups. Non strict means 'expose it unless you do ?strict=1' and the draft group means: 'only expose it for this draft version'.

I will change the draft version to a constant though.

webapp/src/Entity/Judging.php
nickygerritsen: We did not yet on git, but indeed we can once we merge that PR. I will wait before merging this so I can change it.

nickygerritsen:
This PR should now contain everything from the 2026-01 spec, except:

- Support for clarifications to more than one team or a group. Do we want this? This would require us to support this in DOMjudge (with 2 new tables). I do like it and it's not too hard, but then we need to think how we handle unread clarification support when sending to a group.
- Support for file tags. I think we could add this, announcing all our images being light mode. What do others think?
- Submissions without a team. The idea is that jury submissions could be without a team, but we always have a jury team for this currently. I think we should keep it like that? But that does mean we can't read submissions from a feed without a team.

vmcj:
> This PR should now contain everything from the 2026-01 spec, except:
> 
>     * Support for clarifications to more than one team or a group. Do we want this? This would require us to support this in DOMjudge (with 2 new tables). I do like it and it's not too hard, but then we need to think how we handle unread clarification support when sending to a group.

When is this relevant? I would think any time you remark something non-trivial about the problem it should be sent to everyone and if it's considered trivial it's only to the team? So I would prefer not adding this as it's currently unclear to my why the spec allows for this.
> 
>     * Support for file tags. I think we could add this, announcing all our images being light mode. What do others think?

:+1: , assuming we'll just hardcode this in the response and not make it a field in the database?
> 
>     * Submissions without a team. The idea is that jury submissions could be without a team, but we always have a jury team for this currently. I think we should keep it like that? But that does mean we can't read submissions from a feed without a team.

I still think the spec over-complicates things here. This is so much easier to verify when submissions belong to teams because that way we can also properly verify an event-feed for inconsistent data. If everything is just a many to many relation it's very hard to miss actual problems and everything is just a parsing problem IMO.

nickygerritsen:
> > This PR should now contain everything from the 2026-01 spec, except:
> > ```
> > * Support for clarifications to more than one team or a group. Do we want this? This would require us to support this in DOMjudge (with 2 new tables). I do like it and it's not too hard, but then we need to think how we handle unread clarification support when sending to a group.
> > ```
> 
> When is this relevant? I would think any time you remark something non-trivial about the problem it should be sent to everyone and if it's considered trivial it's only to the team? So I would prefer not adding this as it's currently unclear to my why the spec allows for this.

Sending a message to all teams in group "Eindhoven" that lunch is ready, for a multi site contest.

> 
> > ```
> > * Support for file tags. I think we could add this, announcing all our images being light mode. What do others think?
> > ```
> 
> 👍 , assuming we'll just hardcode this in the response and not make it a field in the database?

Yeah that was my idea and then (later) also update the importer to read light mode files if both are there.
> 
> > ```
> > * Submissions without a team. The idea is that jury submissions could be without a team, but we always have a jury team for this currently. I think we should keep it like that? But that does mean we can't read submissions from a feed without a team.
> > ```
> 
> I still think the spec over-complicates things here. This is so much easier to verify when submissions belong to teams because that way we can also properly verify an event-feed for inconsistent data. If everything is just a many to many relation it's very hard to miss actual problems and everything is just a parsing problem IMO.

Whatever we do, the spec has it now and the people on the spec call wanted it. So either we accept we do not import submissions without a team (and thus handle this case with a log message or something) or we do allow it.

vmcj:
> > > ```
> > > * Support for clarifications to more than one team or a group. Do we want this? This would require us to support this in DOMjudge (with 2 new tables). I do like it and it's not too hard, but then we need to think how we handle unread clarification support when sending to a group.
> > > ```
> > 
> > 
> > When is this relevant? I would think any time you remark something non-trivial about the problem it should be sent to everyone and if it's considered trivial it's only to the team? So I would prefer not adding this as it's currently unclear to my why the spec allows for this.
> 
> Sending a message to all teams in group "Eindhoven" that lunch is ready, for a multi site contest.
> 

Given that we also support filtering on such sites (well institutions) I think this is indeed a valid usecase to support. But I would like it if we extend it to allow for sending it to teams in institutions/locations and make that an implicit group in a follow up issue?

nickygerritsen:
I now added support for image tags, supporting light mode for all images and light+dark for country logos.

I (with the help of Claude) also added support for the tags in `import_contest`, as well as for reading file refs from the JSON files.

misc-tools/import_contest_test.py
```diff
@@ -0,0 +1,251 @@
+#!/usr/bin/env python3
```
vmcj, line 1: Can you move this to `misc-tests`?

webapp/src/DataFixtures/Test/AnotherVisibleTeamSubmissionFixture.php
```diff
@@ -0,0 +1,63 @@
+<?php declare(strict_types=1);
```
vmcj, line 1: I think there is already another fixture which does the same thing. Did you check if you can extend on that one?

webapp/src/DataTransferObject/Shadowing/SubmissionEvent.php
```diff
@@ -11,11 +11,13 @@ public function __construct(
         public string  $id,
         public string  $languageId,
         public string  $problemId,
-        public string  $teamId,
+        // CCS 2026-01 made team_id optional and added account_id as an alternative.
```
vmcj, line 14: Can we somehow pass this as a tuple so either of the two fields has a value? I don't think we can force this with an array notation but would like PHPStan to be aware of this restriction so the best I can think of is a tuple.

nickygerritsen: I don't think we can use a tuple, since the constructor has to match the JSON. Otherwise we need a custom deserializer and that sounds way too complex for this.

Three options:

- Add ` * @param ($accountId is null ? string : string|null) $teamId` (suggested by Claude). However if you pass two `?string` args, it won't know. It will only complain if you pass 2 literal `null`s.
- Add a check in the constructor and throw, i.e. move it to runtime.
- Do nothing and accept it this way, since it's only used for parsing JSONs from an external API.

webapp/src/Serializer/Shadowing/EventDenormalizer.php
```diff
@@ -124,12 +124,13 @@ public function getSupportedTypes(?string $format): array
...
      * @param array{api_version?: string} $context
      */
-    protected function getEventFeedFormat(array $event, array $context): EventFeedFormat
+    protected function getCcsApiVersion(array $event, array $context): CcsApiVersion
     {
         return match ($context['api_version']) {
-            '2020-03', '2021-11' => EventFeedFormat::Format_2020_03,
-            '2022-07', '2023-06' => EventFeedFormat::Format_2022_07,
-            default => isset($event['op']) ? EventFeedFormat::Format_2020_03 : EventFeedFormat::Format_2022_07,
+            '2020-03' => CcsApiVersion::Format_2020_03,
```
vmcj, line 130: I assume `'2020-03', '2021-11' => EventFeedFormat::Format_2020_03,` got changed by design?

nickygerritsen: Yes, the 2023 format is better to use for the 2021-11 version of the spec.

webapp/src/Controller/API/AccessController.php
```diff
@@ -252,6 +256,7 @@ public function getStatusAction(Request $request): Access
                         'end_time',
                         'end_contest_time',
                         'max_run_time',
+                        'current',
```
vmcj, line 259: What is the reason for doing this? In which situations would the last judgement not be the current judgement?

nickygerritsen: Is this a question for DOMjudge or for the CCS spec? Assuming the latter: this came from https://github.com/icpc/ccs-specs/issues/53 and I think that thread contains reasoning.

webapp/src/Entity/Submission.php
```diff
@@ -365,6 +366,15 @@ public function getApiTeamId(): string
         return $this->getTeam()->getExternalid();
     }
 
+    #[OA\Property(nullable: true)]
+    #[Serializer\Groups([ARC::GROUP_NONSTRICT, CcsApiVersion::Format_2026_01->value])]
+    #[Serializer\VirtualProperty]
+    #[Serializer\SerializedName('account_id')]
+    public function getApiAccountId(): ?string
+    {
+        return $this->getUser()?->getExternalid();
```
vmcj, line 375: Do we now also output this for all team submissions?

nickygerritsen: If we fill the field (which I think we do if you upload) and we use the new API then we do output this. I think this is a feature, not a bug

webapp/src/Controller/API/AccessController.php
vmcj: It was for the CCS spec.

PR #3043, opened 2025-08-01
Add category types and allow teams to be in more than one category

### LINKED ISSUE #2937: Allow teams to be in more than one category
Currently teams can be in only one category, while the CCS spec allows more than one group. I believe we traditionally only allow one category since we use the categories on the scoreboard for ranking purposes and it doesn't make sense if a team is in more than one.

However, there are reasons why more than one group would be useful. For example, at EUC we wanted to display what teams already prequalified for the World Finals Baku. We solved this by putting them in a separate category with the same sort order, but that UX isn't very clear.

After talking onsite we came up with roughly this idea:

* Add settings to categories to define the type of category, which can be:
  * A background color (like we currently do).
  * A badge for the top team of the category (like we currently do, but split it up from the background color).
  * Some label / icon for each team in the category.
  * Whether it's used for scoring and if so, what the sort order is.

We then add validation that each team should be at most one category that is used for sorting. Or should it be exactly one? What happens if they are in 0 categories?

This allows us to be more CCS spec compatible and adds features that to me sound useful.

nickygerritsen:
This fixes #2937

I added validation to make sure each team is in at most one scoring category and a config checker to make sure they are in at least one (since this is not an error but still weird).

There are some minor changes to the scoreboard compared to before:

* We used to only show the badges for top teams if there was more than one category. Now we show them when you have any category with the `TYPE_BADGE_TOP` type.
* Similar for the colors, we only showed them if there was more than one color, also taking into account the 'non-color' as color. Now we display them as soon as you have one color, ignoring categories without color. We don't display the names of categories without colors anymore, since a team can be in multiple of those categories. We could change this to something smart if all categories that have a color also have a scoring type, but I'm not sure if that's worth it.

I decided that filtering on the scoreboard should be possible with all category types, since I think this makes sense (i.e. one can filter on all teams that advanced to finals before or all teams that didn't).

I also decided that we can have categories without any type. The above mentioned example is a good use case: all teams NOT advancing to finals from EUC could be put in that category, so you can filter on it.

The PR is quite big, since it's quite a lot of changes. I split some of the work up in separate commits, but plan to squash before merging.


webapp/migrations/Version20250620082406.php
```diff
@@ -0,0 +1,38 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20250620082406 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Change comments to reflect entities';
```
vmcj, line 17: I assume this is unrelated to your PR but you wanted it in to cleanup the code?

webapp/src/Controller/API/ScoreboardController.php
```diff
@@ -171,7 +171,7 @@ public function getScoreboardAction(
         $scoreIsInSeconds = (bool)$this->config->get('score_in_seconds');
 
         foreach ($scoreboard->getScores() as $teamScore) {
-            if ($teamScore->team->getCategory()->getSortorder() !== $sortorder) {
+            if ($teamScore->team->getSortorder() !== $sortorder) {
```
vmcj, line 174: Did not read all the code yet but if we don't have the sortorder per category anymore wouldn't this make the scoreboard calculation very expensive if we have a lot of categories?

webapp/migrations/Version20250620082406.php
nickygerritsen: Yes so I can generate migrations

webapp/src/Controller/API/ScoreboardController.php
nickygerritsen: The scoreboard calculation only works with categories with sort orders so no real change there

vmcj: I think I misinterpret the `sortorder` it was the sortorder of the category not the rank of the team in that specific sortorder.

webapp/src/Controller/API/TeamController.php
```diff
@@ -305,14 +307,16 @@ protected function getQueryBuilder(Request $request): QueryBuilder
         $queryBuilder = $this->em->createQueryBuilder()
             ->from(Team::class, 't')
             ->leftJoin('t.affiliation', 'ta')
-            ->leftJoin('t.category', 'tc')
+            ->leftJoin('t.categories', 'tc')
+            ->leftJoin('t.categories', 'tcc', Join::WITH, 'BIT_AND(tcc.types, :scoring) = :scoring')
```
vmcj: Why do we need the extra `tcc`, aren't you joining this to be the same category?

webapp/src/Service/DOMJudgeService.php
```diff
@@ -1433,7 +1433,7 @@ public function loadTeam(string $teamId, Contest $contest): Team
         $queryBuilder = $this->em->createQueryBuilder()
             ->from(Team::class, 't')
             ->select('t')
-            ->leftJoin('t.category', 'tc')
+            ->leftJoin('t.categories', 'tc')
```
vmcj, line 1500: phpcs below

webapp/templates/partials/scoreboard_table.html.twig
```diff
@@ -586,7 +625,7 @@
     <p><br/><br/></p>
 
     {# only print legend when there's more than one category #}
-    {% if limitToTeamIds is null and usedCategories | length > 1 and hasDifferentCategoryColors %}
+    {% if limitToTeamIds is null and colorCategories | length > 0 and hasDifferentCategoryColors %}
```
vmcj, line 624: The comment is now not in line with the code. Why do we need to print the legend with only 1 category?

webapp/src/Service/ScoreboardService.php
```diff
@@ -988,9 +999,14 @@ protected function getTeamsInOrder(Contest $contest, bool $jury = false, ?Filter
             }
 
             if ($filter->categories) {
+                // Use a new join, since we need both the other two category joins for other logic already
```
vmcj, line 1002: I think you don't limit `tcc` so can re use it here,

webapp/src/Form/Type/ContestType.php
```diff
@@ -125,6 +126,11 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             'multiple' => true,
             'choice_label' => fn(TeamCategory $category) => $category->getName(),
             'help' => 'List of team categories that will receive medals for this contest.',
+            'query_builder' => function(EntityRepository $er) {
```
vmcj:
```suggestion
            'query_builder' => function (EntityRepository $er) {
```

webapp/src/Form/Type/ExportResultsType.php
```diff
@@ -21,6 +21,8 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             ->from(TeamCategory::class, 'c', 'c.categoryid')
             ->select('c.sortorder, c.name')
             ->where('c.visible = 1')
+            ->andWhere('BIT_AND(c.types, :scoring) = :scoring')
+            ->setParameter('scoring', TeamCategory::TYPE_SCORING)
```
vmcj, line 25: Probably good to add an assertion in the Unit tests for the results with a CSS only category and that we don't see it in the result.

webapp/src/Service/ImportExportService.php
```diff
@@ -853,11 +871,29 @@ protected function importGroupData(
                 }
                 $added = true;
             }
+            /** @var list<int> $types */
+            $types = $groupItem['types'] ?? [TeamCategory::TYPE_SCORING, TeamCategory::TYPE_BADGE_TOP];
+            $sortOrder = $groupItem['sortorder'] ?? null;
+            if (in_array(TeamCategory::TYPE_SCORING, $types, true) && $sortOrder === null) {
+                $sortOrder = 0;
+            }
+            if ($sortOrder !== null && !in_array(TeamCategory::TYPE_SCORING, $types, true)) {
```
vmcj:
```suggestion
            if (!($sortOrder === null || in_array(TeamCategory::TYPE_SCORING, $types, true)) {
```

I think we always do this, so let's keep that convention.

webapp/src/Service/ImportExportService.php
```diff
@@ -853,11 +871,29 @@ protected function importGroupData(
...
+                $types = [TeamCategory::TYPE_SCORING];
+            }
+            if (isset($groupItem['color']) && !in_array(TeamCategory::TYPE_BACKGROUND, $types, true)) {
+                $types[] = TeamCategory::TYPE_BACKGROUND;
+            }
+            if (isset($groupItem['css_class']) && !in_array(TeamCategory::TYPE_CSS_CLASS, $types, true)) {
+                $types[] = TeamCategory::TYPE_CSS_CLASS;
+            }
```
vmcj, line 888: A couple of those statements are checks if something is set and if we already have it in the array. I rather read that as just always adding it to the list and afterwards just cleaning the array with `array_unique` as it makes it easier to read IMO. So basically treating the array as a sorted set by always adding and making it unique in the end.

Alternative is to do something as $arr[TeamCategory::TYPE_SOMETHING] = bool and getting the keys afterwards if that's faster.

webapp/src/Service/ImportExportService.php
```diff
@@ -1192,7 +1228,7 @@ public function importAccountsJson(array $data, ?string &$message = null, ?array
      * Import team data from the given array.
      *
      * @param array<array{team: array{teamid: string|null, icpcid: string|null, label?: string|null,
-     *                                categoryid: string|null, name: string|null, display_name?: string,
+     *                                categoryids: string[]|null, name: string|null, display_name?: string,
```
vmcj: Any (good) reason why this cant be `categoryids: string[]`?

webapp/src/Utils/Scoreboard/Scoreboard.php
```diff
@@ -76,6 +78,27 @@ public function getCategories(): array
         return $this->categories;
     }
 
+    /**
+     * @param int[] $limitToTeamIds
```
vmcj: Shouldn't this also list the `null`?
```suggestion
     * @param int[]|null $limitToTeamIds
```
or set it to `$limitToTeamIds = []`

webapp/templates/partials/scoreboard_table.html.twig
```diff
@@ -514,10 +535,28 @@
         {% endif %}
         <a {% if extra is not null %}{{ extra | raw }}{% endif %} {% if link is not null %}href="{{ link }}"{% endif %}>
                     <span class="forceWidth">
-                        {% if false and usedCategories | length > 1 and scoreboard.bestInCategory(score.team, limitToTeamIds) %}
-                            <span class="badge text-bg-warning category-best">
-                            {{ score.team.category.name }}
-                        </span>
+                        {# TODO: how to display badges on mobile scorebaord #}
```
vmcj:
```suggestion
                        {# TODO: how to display badges on mobile scoreboard #}
```

webapp/templates/partials/scoreboard_table.html.twig
```diff
@@ -514,10 +535,28 @@
...
+                        {% if false %}
```
vmcj: We can just comment this right?

webapp/templates/partials/scoreboard_table.html.twig
```diff
@@ -239,11 +242,26 @@
                 {% endif %}
                 <a {% if extra is not null %}{{ extra | raw }}{% endif %} {% if link is not null %}href="{{ link }}"{% endif %}>
                     <span class="forceWidth">
-                        {% if usedCategories | length > 1 and scoreboard.bestInCategory(score.team, limitToTeamIds) %}
-                            <span class="badge text-bg-warning category-best">
-                            {{ score.team.category.name }}
-                        </span>
-                        {% endif %}
+                        {% set hasBadges = score.team.badgeCategories | length > 0 %}
```
vmcj, line 245: Do we know the reason why we didn't display those if there is only 1 category?

webapp/templates/partials/scoreboard_table.html.twig
nickygerritsen: I think if you set up colors you always want to display a legend.

webapp/src/Service/ScoreboardService.php
nickygerritsen: We use `tcc` to select the categories, so we can't reuse it.

webapp/templates/partials/scoreboard_table.html.twig
nickygerritsen: We assumed if you had one category, all have the same color. Nowe we can have some having a color category and some having nothing.

webapp/templates/partials/scoreboard_table.html.twig
nickygerritsen: I just enabled it now

webapp/src/Utils/Scoreboard/Scoreboard.php
nickygerritsen: This can't be `[]` since we check the difference to render the summary. Will update docblock

webapp/migrations/Version20250801124024.php
```diff
@@ -0,0 +1,50 @@
...
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql(<<<SQL
+            ALTER TABLE team_category
+            ADD types INT NOT NULL DEFAULT 1 COMMENT 'Bitmask of category types, default is scoring.' AFTER name,
+            CHANGE sortorder sortorder TINYINT UNSIGNED DEFAULT NULL COMMENT 'Where to sort this category on the scoreboard',
+            ADD css_class VARCHAR(255) DEFAULT NULL COMMENT 'CSS class to apply to scoreboard rows (only for TYPE_CSS_CLASS)' AFTER allow_self_registration
+            SQL);
+
+        // Now update the types for existing categories based on whether the color is set
+        $this->addSql('UPDATE team_category SET types = 7 WHERE color IS NOT NULL');
```
meisterT, line 33: nit: add a comment what 7 and 5 mean?

webapp/src/Entity/TeamCategory.php
```diff
@@ -60,13 +61,48 @@ class TeamCategory extends BaseApiEntity implements
...
+        self::TYPE_BADGE_TOP => 'badge-top',
+        self::TYPE_BADGE_ALL => 'badge-all',
+        self::TYPE_CSS_CLASS => 'css-class',
+    ];
+
+    /**
+     * @var array<int, string>
+     */
+    public const TYPES_TO_HUMAN_STRING = [
+        self::TYPE_SCORING => 'Scoring',
```
meisterT, line 86: What does `Scoring` mean?

webapp/src/Service/ScoreboardService.php
```diff
@@ -951,17 +958,20 @@ protected function getTeamsInOrder(Contest $contest, bool $jury = false, ?Filter
     {
         $queryBuilder = $this->em->createQueryBuilder()
             ->from(Team::class, 't', 't.teamid')
-            ->innerJoin('t.category', 'tc')
+            // Join on categories twice: once to determine the sort order (tc) and once to get all categories the team belongs to (tcc)
```
meisterT, line 961: I am not sure I understand why we need to join twice here

webapp/src/Entity/TeamCategory.php
nickygerritsen: That it's used for the scoreboard, i.e. that is what makes sort order required / useful. Each team can be in at most one scoring category as to only have one order on the scoreboard

nickygerritsen: That it's used for the scoreboard, i.e. that is what makes sort order required / useful. Each team can be in at most one scoring category as to only have one order on the scoreboard

webapp/src/Service/ScoreboardService.php
nickygerritsen: `tc` is used to determine the sort order on the scoreboard for the team basically and `tcc` is used to get ALL categories of the team to use to determine what UI we need to display (the colors, badges, etc.).

I don't see how we can combine them in one join.

webapp/src/Entity/TeamCategory.php
meisterT: can be in at most one, or does have to be exactly one?

nickygerritsen: I will double check, I thought at most (and if 0 they won't appear on the scoreboard)

nickygerritsen: I will double check, I thought at most (and if 0 they won't appear on the scoreboard)

meisterT: won't appear on the scoreboard could be super confusing

nickygerritsen: Do you suggest to make it 'exactly 1'?

nickygerritsen: I did this.

PR #3044, opened 2025-08-02
Improve cgroup error message when not configured correctly.

meisterT:


vmcj (changes requested):
Looks like a typo?

meisterT:
> Looks like a typo?

Unfortunately,  github doesn't tell me what looks like a typo, not even the line it is attached to.

vmcj:
> > Looks like a typo?
> 
> Unfortunately, github doesn't tell me what looks like a typo, not even the line it is attached to.

It was about: `cgroup_error_and_usage "Error: Cgroups not configured properly, missing systemd slice under /proc/self/cgroup. If running under docker, make sure to set cgroupns=host."` but `cgroupns` seems to be correct.

PR #3045, opened 2025-08-02
Explicit handle ignoring the return code for failed error messages

vmcj:
This gave a compiler warning.


judge/runguard.cc
```diff
@@ -233,7 +233,9 @@ void verbose(const char *format, ...)
 void verbose_from_signalhandler(const char* msg)
 {
 	if (!be_quiet && be_verbose) {
-		write(STDERR_FILENO, msg, strlen(msg));
+		if (write(STDERR_FILENO, msg, strlen(msg) != 0)) {
```
meisterT: I think I would prefer just an explicit cast to `(void)` instead of these if's.

vmcj: Casting to `void` is indeed better (TIL).

But that doesn't work because `write` has `warn_unused_result`. So we either have to do this, or wrap it in a custom function. Alternative is using a pragma but that would be even more lines.

Using the custom function would open the risk that we use it more often, and unintentionally not check the result. So I think here my current implementation is the right one (although I was not aware of that before).

eldering: See https://stackoverflow.com/questions/40576003/ignoring-warning-wunused-result

I think doing
```
std::ignore = write(STDERR_FILENO, msg, strlen(msg));
```
is the cleanest and most explicit option. This requires C++ 11, but we set that.

meisterT (approved):

eldering (approved):

PR #3046, opened 2025-08-02
Maintainer fedora

vmcj:
I was testing on my fedora install and needed a couple of packages. Doing the PR to make this easier in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,9 @@ REC_TARGETS=build domserver install-domserver judgehost install-judgehost \
 # Global Makefile definitions
 include $(TOPDIR)/Makefile.global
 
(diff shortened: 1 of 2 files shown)


nickygerritsen (approved):

eldering (approved):

vmcj:
@nickygerritsen to set the SELinux contest in the right way I can limit where php-fmp can write. Apart from webapp/var/cache what other locations would we have? Probably the locations for logo's & team pictures?

nickygerritsen:
> @nickygerritsen to set the SELinux contest in the right way I can limit where php-fmp can write. Apart from webapp/var/cache what other locations would we have? Probably the locations for logo's & team pictures?

Yeah webapp/public/images I'd say. And webapp/var/log, so I'd say whole webapp/var

PR #3047, opened 2025-08-03
Basic implementation of partial scoring.

meisterT:
This follows the legacy spec in:
https://icpc.io/problem-package-format/spec/legacy.html#graders

We might decided at a later point whether we want to support this or just the new (currently draft) spec, but basic constructs of the change should be the same.

Not yet implemented:
- any form of testing
- anything on the team interface
- shadowing

Part of https://github.com/DOMjudge/domjudge/issues/2518

### DIFF
--- a/webapp/templates/jury/partials/submission_list.html.twig
+++ b/webapp/templates/jury/partials/submission_list.html.twig
@@ -172,7 +172,7 @@
                     </a>
                 </td>
                 {% if is_scoring %}
-                    <td class="right">
+                    <td class="{{ tdExtraClass }} right">
                         {% if submission.score is not null %}
                             {{ submission.score | number_format(2, '.', '') }}
                         {% else %}
@@ -292,12 +292,21 @@
                             {% endif %}
                         </a>
                     </td>
+                    {% if is_scoring %}
+                        <td class="right">
+                            {% if externalJudgement.score is not null %}
+                                {{ externalJudgement.score | number_format(2, '.', '') }}
+                            {% else %}
+                                {{ '-' }}
+                            {% endif %}
+                        </td>
+                    {% endif %}
                     <td class="testcase-results">
                         {{- submission | testcaseResults(true) -}}
                     </td>
                 </tr>
                 <tr>
-                    <td colspan="3" class="inline-verify-form">
+                    <td colspan="{% if is_scoring %}4{% else %}3{% endif %}" class="inline-verify-form">
                         {% if externalJudgement is not null and (not submission.judgings.first or submission.judgings.first.result != externalJudgement.result) %}
                             {% include 'jury/partials/verify_form.html.twig' with {
                                 label: 'Shadow difference verified',


meisterT:
Some notes: 
- there's a to-do already, but we need to talk about internal precision and precision in presentation
- we should add to the jury submission page some info that allows following the result calculation with test case groups (should be easy to do with the existing function)
- the number of tries on a scoring scoreboard should really be the number of total tries (not until the first correct attempt like for pass-fail)
- it seems we are not handling the accept_if_any_accepted case always correctly, needs more testing
- while we can ingest results from a primary CCS now, we don't compare the scores yet for shadow differences
- the layout of the shadow differences table needs to be fixed 
- optimization problems are by their very nature less deterministic as they incentivize teams to do stuff like try this approach for that amount of time, then this other approach for that amount of time and generally the use of RNG is much more prevalent

webapp/src/Entity/Problem.php
```diff
@@ -211,6 +211,11 @@ class Problem extends BaseApiEntity implements
     #[Serializer\Exclude]
     private Collection $languages;
 
+    #[ORM\ManyToOne(inversedBy: 'problems')]
```
nickygerritsen, line 214: I'm not seeing this reverse relation? And should it be manytoone? When does a testcasegroup have many problems?

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -1855,6 +1859,10 @@ protected function importRun(Event $event, EventData $data): void
             ->setRuntime($runTime)
             ->setResult($judgementTypeId === null ? null : $verdictsFlipped[$judgementTypeId]);
 
+        if (isset($data->score)) {
+            $run->setScore($data->score);
```
nickygerritsen, line 1877: This comment from PHPStan is not lying 😛

webapp/src/Service/ScoreboardService.php
```diff
@@ -259,6 +260,8 @@ public function calculateScoreRow(
 
         $contestStartTime = $contest->getStarttime();
 
+        $pointsJury      = "0";
```
nickygerritsen, line 263: Why not `$scoreJury`? We have points already for multipoint problems, so this might be confusing?

webapp/src/Service/SubmissionService.php
```diff
@@ -56,6 +59,142 @@ public function __construct(
         protected readonly PaginatorInterface $paginator,
     ) {}
 
+    /**
+     * Returns a two-element array:
+     *   - The score for the testcase group, or null if not all results are ready.
+     *   - The result for the testcase group, or null if not all results are ready.
+     * @return array{string|null, string|null}
+     */
```
nickygerritsen, line 67: I would return a string indexed array to make this more clear:

```suggestion
    /**
     * Returns a two-element array:
     *   - The score for the testcase group, or null if not all results are ready.
     *   - The result for the testcase group, or null if not all results are ready.
     * @return array{score: string|null, result: string|null}
     */
```

webapp/src/Entity/Problem.php
nickygerritsen: And for me calling it "parent testcase group" sounds weird, maybe "root testcase group"?

nickygerritsen:
> Some notes:
> 
> * the number of tries on a scoring scoreboard should really be the number of total tries (not until the first correct attempt like for pass-fail)

I changed this to show the total number of submissions. I wonder if this is what we want? Say you have 5 submissions:
1. AC, score = 50
2. AC, score = 75
3. AC, score = 100
4. WA
5. WA

Do we want 3 or 5 tries? And what if the fourth one is AC, score = 100?

Also please check my logic. Basically I do not stop processing submissions after the first correct one for scoring problems (and have changed the FTS logic since this was dependent on this assumption).

> * the layout of the shadow differences table needs to be fixed

Fixed

I also fixed some other things:

* I added the reverse relation for problem->testcaseGroup.
* We now correctly import (and export) the scoreboard type from contest.yaml/json (but for this I have renamed the `SCORING` enum case to `SCORE` as that is what the spec says).
* I fixed the single team scoreboard row, it didn't show the scores.
* I added the score to the team submission list and detail modal.
* I fixed an issue where if we calculated the scorecache for a submission without judgement, it would crash.
* I now check the scoreboard type when starting up the shadow script.
* I fixed some logic for importing an event with no items, as that is what we had in the replay for the NAC challenge.
* I also fixed importing the ZIPs from that challenge that contained folders.

webapp/src/Entity/Problem.php
nickygerritsen: I did push the change for the reverse relation.

meisterT:
Closing in favor of https://github.com/DOMjudge/domjudge/pull/3337

PR #3051, opened 2025-08-06
Verify that timestrings can be parsed

vmcj:
In case of "unknown" timezones we would silently set the starttime to 0 but leave the startTimeString so the user would not be aware that the starttime is set to a completely different value.

### DIFF
--- a/webapp/src/Entity/Contest.php
+++ b/webapp/src/Entity/Contest.php
@@ -1260,6 +1260,12 @@ class Contest extends BaseApiEntity implements
         return new FreezeData($this);
     }
 
+    public function checkValidTimeString(string $timeString): void {
+        if (preg_match("/^[+-]?\d+:\d\d(:\d\d(\.\d{1,6})?)?$/", $timeString) !== 1) {
+            $date = new DateTime($timeString);
+        }
+    }
+
     #[ORM\PrePersist]
     #[ORM\PreUpdate]
     public function updateTimes(): void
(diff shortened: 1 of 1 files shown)


webapp/src/Entity/Contest.php
```diff
@@ -1260,6 +1260,17 @@ public function getFreezeData(): FreezeData
         return new FreezeData($this);
     }
 
+    public function checkValidTimeString(string $timeString): bool {
+        if (preg_match("/^[+-]?\d+:\d\d(:\d\d(\.\d{1,6})?)?$/", $timeString) !== 1) {
+            try {
+                $date = new DateTime($timeString);
+            } catch (Exception) {
```
meisterT: does the exception here tell us what's wrong? If so, can we extract it (or get it from somewhere else) and return it to the user?

vmcj: image

Ugly message but more info yes.

webapp/src/Entity/Contest.php
```diff
@@ -1272,6 +1278,18 @@ public function updateTimes(): void
     #[Assert\Callback]
     public function validate(ExecutionContextInterface $context): void
     {
+        foreach (['Activate', 'Deactivate', 'Start', 'End', 'Freeze', 'Unfreeze'] as $timeString) {
+            $tmpValue = $this->{'get' . $timeString . 'timeString'}();
+            if ($tmpValue === null) continue;
+            try {
+                $this->checkValidTimeString($tmpValue));
```
meisterT:
```suggestion
                $this->checkValidTimeString($tmpValue);
```

webapp/src/Entity/Contest.php
```diff
@@ -1272,6 +1278,18 @@ public function updateTimes(): void
...
+            } catch (Exception $e) {
+                $context
+                    ->buildViolation("Can't parse this time," . $e->message)
```
meisterT:
```suggestion
                    ->buildViolation("Can't parse this time: " . $e->message)
```

PR #3052, opened 2025-08-06
Support alpha in hexadecimal colors

vmcj:
This is as a basis to detect problematic colors coming from the primary.


webapp/src/Utils/Utils.php
```diff
@@ -349,6 +352,12 @@ public static function convertToColor(string $hex): ?string
      */
     public static function parseHexColor(string $hex): array
     {
+        // Ignore alpha in hexstrings
```
meisterT: Why silently ignore the alpha channel?

vmcj: It might indeed a bit harsh to do this silently. Do we ever think we need the alpha channel? Otherwise I think I'll add a optional boolean for this behaviour (optionally also returning `rgba` instead of `rgb`), alternative is to check all calling functions and handle this there.

meisterT: It's unlikely that we need it. But I'm either for supporting it properly or not allowing it at all to avoid user confusion.

vmcj: The spec allows for sending such a string "#ff00ffaa" so we need to store it. I think my fix already makes that we never pick it up but I'll think about this some more. Maybe just supporting it is easier in that case.

webapp/src/Utils/Utils.php
```diff
@@ -343,19 +343,25 @@ public static function convertToColor(string $hex): ?string
...
     /**
-     * Parse a hex color into it's three RGB values.
+     * Parse a hex color into it's four RGBA values.
      *
-     * @return array{int, int, int}
+     * @return array{int, int, int, int}
      */
     public static function parseHexColor(string $hex): array
     {
+        $tmp = substr($hex, 1);
```
meisterT:
```suggestion
        $tmp = substr($hex, 1);
        // For a RGB hex value (without A), add a fully opaque alpha channel.
        // e.g. convert #ABC to #ABCF, and #123456 to #123456FF
```

meisterT (approved):
Next to my minor comment, there are some phpcs errors left to fix. Otherwise good to merge.

PR #3053, opened 2025-08-10
Interactive extra time

vmcj:
Mostly as PoC,

I think this is helpful for the current state where interactive problems have overhead.

If we want to do this:
- I think lib.misc.php can mostly be moved to utils for this
- I made some things static, I assume this is an improvement but I'm not 100% on the performance gain/impact


eldering (commented):
I think it best to first discuss how we best address this issue with the interactive validator sometimes requiring a lot of extra wallclock time. Ideally we solve that or just add wallclock extra overhead for interactive problems, not generic (also CPU time) overhead.

A side note: there are some problem color changes in this PR.

etc/db-config.yaml
```diff
@@ -101,8 +101,10 @@
             regex: /^[1-9]\d*$/
             error_message: A positive number is required.
         -   name: timelimit_overshoot
```
meisterT, line 103: I would keep this separate from `timelimit_overshoot`. Let's discuss in person?

PR #3055, opened 2025-08-23
Autocheck for new DOMjudge releases

vmcj:
Discussed during the NWERC hackathon in 2023.

We allow admins to toggle an autoquery to domjudge.org for 2 reasons:
- Alerting users that a new version might exist, helping us in case of security releases.
- Giving a gentle nudge for people to upgrade making support easier.
- Getting some information on what installations are out there.

Default this doesn't happen, so the admin chooses to enable this themselves so this is different from normal telemetry.

### DIFF
--- a/etc/db-config.yaml
+++ b/etc/db-config.yaml
@@ -393,6 +393,12 @@
             description: Time in seconds after an external contest source reader last checked in before showing its status as `critical`.
             regex: /^\d+$/
             error_message: A non-negative number is required.
+        -   name: check_new_version
+            type: enum
+            enum_class: App\Utils\UpdateStrategy
+            default_value: none
+            public: false
+            description: Automatically check and notify for new DOMjudge releases?
         -   name: adminer_enabled
             type: bool
             default_value: false
--- a/webapp/src/Service/DOMJudgeService.php
+++ b/webapp/src/Service/DOMJudgeService.php
@@ -31,6 +31,7 @@ use App\Entity\TeamCategory;
 use App\Entity\Testcase;
 use App\Entity\User;
 use App\Utils\FreezeData;
+use App\Utils\UpdateStrategy;
 use App\Utils\Utils;
 use DateTime;
 use Doctrine\ORM\EntityManagerInterface;
@@ -38,9 +39,11 @@ use Doctrine\ORM\NonUniqueResultException;
 use Doctrine\ORM\NoResultException;
 use Doctrine\ORM\Query\Expr\Join;
 use Doctrine\ORM\QueryBuilder;
+use Exception;
 use InvalidArgumentException;
 use Psr\Log\LoggerInterface;
 use ReflectionClass;
+use Symfony\Component\Cache\Adapter\FilesystemAdapter;
 use Symfony\Component\DependencyInjection\Attribute\Autowire;
 use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
 use Symfony\Component\HttpFoundation\Cookie;
@@ -62,6 +65,7 @@ use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInt
 use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
 use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 use Symfony\Component\Security\Core\User\UserInterface;
+use Symfony\Contracts\Cache\ItemInterface;
 use Twig\Environment;
 use ZipArchive;
 
@@ -69,6 +73,8 @@ class DOMJudgeService
 {
     protected ?Executable $defaultCompareExecutable = null;
     protected ?Executable $defaultRunExecutable = null;
+
+    private string $localVersionString = '';
 
     final public const EVAL_DEFAULT = 0;
     final public const EVAL_LAZY = 1;
@@ -108,6 +114,10 @@ class DOMJudgeService
         protected string $projectDir,
         #[Autowire('%domjudge.vendordir%')]
         protected string $vendorDir,
+        #[Autowire('%domjudge.version%')]
+        protected readonly string $domjudgeVersion,
+        #[Autowire('%domjudge.installmethod%')]
+        protected readonly string $domjudgeInstallMethod,
     ) {}
 
     /**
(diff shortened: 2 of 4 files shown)


webapp/src/Service/DOMJudgeService.php
```diff
@@ -1671,4 +1675,59 @@ public function getAllowedLanguagesForContest(?Contest $contest) : array {
             ->getQuery()
             ->getResult();
     }
+
+    public function checkNewVersion(): string|false {
+        if (!$this->config->get('check_new_version')) {
+            return false;
+	}
+        $versionLocalString = explode("/", str_replace("DEV", "-prerelease", $this->domjudgeVersion))[0];
+        $versionLocal = Version::parse($versionLocalString, false);
```
nickygerritsen: Why tarball in the user agent and not DOMjudge like we do when shadowing?

webapp/src/Service/DOMJudgeService.php
```diff
@@ -1671,4 +1675,59 @@ public function getAllowedLanguagesForContest(?Contest $contest) : array {
...
+        $response = @file_get_contents($versionUrl, false, $context);
+	if ($response === false) {
+            return false;
+        }
+        $versions = json_decode($response, true);
+        /* Steer towards to the latest patch first 
+         * the user can see on the website if there is a new Major/minor themselves
+         * otherwise the latest minor, or Major release. So the user might make the upgrade path:
+         * DJ6.0.0 -> DJ6.0.6 -> DJ6.6.0 -> DJ9.0.0 instead of
+         *         -> DJ6.0.[1..6] -> DJ6.[1..6] -> DJ[7..9].0.0
```
nickygerritsen: This whole loop could also be done by sorting the versions lexicographically and taking the next one higher than the local version. This could replace all three checks?

webapp/src/Service/DOMJudgeService.php
vmcj: We also wanted to know which installmethods are out there, so next step is a config variable but I had no idea on how to set that (yet) I suspect a make option..

webapp/src/Service/DOMJudgeService.php
vmcj: Wouldn't that give the path over all patches, minor plus patches in that minor etc?

Basically the 2nd path I mentioned.

nickygerritsen: Ah yeah you might be right but we could solve that by only exposing specific versions on the versions site?

vmcj: Even if the array would only the latest patch release per `M.m` release we would still need to filter somewhere.

But I get the point that we should do something smarter so the code is less duplicated.

vmcj: We still need the loop as at best we can only clean the list to always have the highest patch version in the list, 

We could ofcourse do this math on the version site and basically only let the server do a lookup there so:
{'1.1.5': '1.5.5',
  '1.2.3': '1.2.5',
  '1.5.5': '5.5.5'}
where 5 is always the highest version in that release.

But that also feels expensive, it just removes the calculation to another server.


I did some cleanup though and I think this is already more what you want. Although I'm not 100% if it will now not also alert for the current version.

etc/db-config.yaml
```diff
@@ -392,6 +392,11 @@
             description: Time in seconds after an external contest source reader last checked in before showing its status as `critical`.
             regex: /^\d+$/
             error_message: A non-negative number is required.
+        -   name: check_new_version
+            type: bool
+            default_value: false
+            public: false
+            description: Automatically check for new DOMjudge release?
```
meisterT:
```suggestion
            description: Automatically check and notify for new DOMjudge releases?
```

webapp/src/Service/DOMJudgeService.php
```diff
@@ -1671,4 +1674,42 @@ public function getAllowedLanguagesForContest(?Contest $contest) : array {
             ->getQuery()
             ->getResult();
     }
+
+    public function checkNewVersion(): string|false {
+        if (!$this->config->get('check_new_version')) {
```
meisterT:
```suggestion
        if (!$this->config->get('check_new_version', false)) {
```

webapp/src/Service/DOMJudgeService.php
```diff
@@ -1671,4 +1674,42 @@ public function getAllowedLanguagesForContest(?Contest $contest) : array {
...
+            return false;
+        }
```
meisterT, line 1754: Nit: add a comment here like, that will make understanding the code easier:

```
// Version is something like "x.y.z / commit hash", e.g. "6.1.2 / 4e25adb13"
```

webapp/src/Service/DOMJudgeService.php
```diff
@@ -1671,4 +1674,42 @@ public function getAllowedLanguagesForContest(?Contest $contest) : array {
...
+        $versionLocalString = explode("/", $this->domjudgeVersion)[0];
+        $patch = "/" . substr($this->domjudgeVersion, 0, strrpos($this->domjudgeVersion, '.')) . ".\d/";
```
meisterT: replace these with `explode` on `.`?

Also verify that there are three elements and silently `return false` if not

webapp/src/Service/DOMJudgeService.php
```diff
@@ -1671,4 +1674,42 @@ public function getAllowedLanguagesForContest(?Contest $contest) : array {
             ->getQuery()
             ->getResult();
     }
+
+    public function checkNewVersion(): string|false {
```
meisterT, line 1751: nit: add a comment what this function should return

webapp/src/Service/DOMJudgeService.php
```diff
@@ -1671,4 +1674,42 @@ public function getAllowedLanguagesForContest(?Contest $contest) : array {
...
+        $major = "/\d.\d.\d/";
+
+        $versionUrl = 'https://versions.domjudge.org';
+        $options = ['http' => ['method' => 'GET', 'header' => "User-Agent: tarball/" . $versionLocalString . "\r\n"]];
+        $context = stream_context_create($options);
+        $response = @file_get_contents($versionUrl, false, $context);
+        if ($response === false) {
+            return false;
+        }
+        $versions = json_decode($response, true);
```
meisterT: nit: add a comment to explain what format you expect the server to return

webapp/src/Service/DOMJudgeService.php
meisterT: Should we not return just the newest version here?

webapp/templates/jury/menu.html.twig
```diff
@@ -93,6 +93,13 @@
                 </li>
             </ul>
         {% endif %}
+        {% if new_version_available %}
+            <ul class="navbar-nav">
+                <li class="nav-item nav-link">
+                    <i class="fa-solid fa-ship fa-2x"></i> New release {{ new_version_available }} available.
```
meisterT: Add link?

vmcj: To do that I need to make it unraw, I'll need @nickygerritsen for that if we want this.

webapp/src/Service/DOMJudgeService.php
vmcj: I've now implemented both strategies as I can see a case for both and the code was already there. We can remove either in the future after some feedback assuming it gets used.

etc/db-config.yaml
meisterT: Did you mark this as resolved because you don't like it (which is fine) or by accident?

webapp/src/Service/DOMJudgeService.php
```diff
@@ -1675,4 +1681,79 @@ public function getAllowedLanguagesForContest(?Contest $contest) : array {
...
+            return false;
+        }
+        // Assume we get a one-level unordered JSON list with the released versions e.g. ["10.0.0", "9.11.0", "12.0.12", "10.0.1"]
+        $versions = json_decode($response, true);
+        natsort($versions);
+        $versions = array_reverse($versions);
+        preg_match("/\d.\d.\d/", $this->domjudgeVersion, $matches);
+        $extractedLocalVersionString = $matches[0];
+        if ($this->config->get('check_new_version', false) === UpdateStrategy::Strategy_incremental) {
+            /* Steer towards the nearest new patch release first
```
meisterT: I might be alone with this opinion, but I think we should tell users the most recent version instead.

Feel free to close

etc/db-config.yaml
vmcj: By accident,

webapp/src/Service/DOMJudgeService.php
vmcj: I've now added both strategies.

I can see a case for both strategies if we do enough of the work (and had the code already for both),

We should discuss this in person though as to me doing the patch releases is that extra time which I don't think we invest often enough. We change quite a lot between versions and we seem to forget to backport often so the patch releases are often very outdated to maintain and not often useful because the fixes are not being backported (and therefor a sort of false promise).

Kevinjil: I think that having both strategies is nice, if we actually do the bugfix backports. If we do not plan to more actively do these, then I agree that we should guide users towards the latest version.

webapp/src/Twig/TwigExtension.php
```diff
@@ -159,6 +159,7 @@ public function getGlobals(): array
             'doc_links'                     => $this->dj->getDocLinks(),
             'allow_registration'            => $selfRegistrationCategoriesCount !== 0,
             'enable_ranking'                => $this->config->get('enable_ranking'),
+            'new_version_available'         => $this->dj->checkNewVersion(),
```
Kevinjil: From what I see and tried, this will make a HTTP request to `versions.domjudge.org` on every page load. This seems very aggressive, and might even cause an accidental DDOS of our server when large contests are running. Should we store it with some cache time in the DB?

webapp/src/Service/DOMJudgeService.php
eldering: I think we should do backports/releases more indeed and I think offering both patch and major/minor release updates makes sense. Why can't we display both at the same time?

meisterT: Have we people asking for backport releases? They are a lot of work and I'm not sure whether it's worth the effort

webapp/src/Twig/TwigExtension.php
vmcj: Given where the server is running I doubt if that ever happens but we can handle that case. I don't prefer to store this in the database but we can use this:
https://symfony.com/doc/current/components/cache.html

meisterT: Alternatively just do it on the index page?

webapp/src/Service/DOMJudgeService.php
```diff
@@ -1699,4 +1709,98 @@ public function getAllowedLanguagesForContest(?Contest $contest) : array {
             ->getQuery()
             ->getResult();
     }
+
+    /**
+     * @return String[]|false
+     */
+    public function cacher_checkNewVersion(ItemInterface $item): array|false {
```
vmcj: I'll rename this one.

meisterT:
I think this was merged prematurely, as https://versions.domjudge.org doesn't exist yet.

vmcj:
> I think this was merged prematurely, as https://versions.domjudge.org doesn't exist yet.

When the site is not reachable it should just do nothing. So if you get problems in your local copy something is wrong.

PR #3057, opened 2025-08-29
Add attachments to problem endpoint in API

nickygerritsen:
This should have the same permissions as the statement.

I added the mimetype in the database so we can expose that in the API.


webapp/src/Entity/Problem.php
```diff
@@ -200,6 +200,14 @@ class Problem extends BaseApiEntity implements
     #[Serializer\Exclude]
     private ?FileWithName $statementForApi = null;
 
+    /**
+     * @var FileWithName[]
+     */
+    // This field gets filled by the contest problem visitor with an array of data transfer
```
vmcj, line 206: I don't understand this remark

vmcj (approved):

meisterT (approved):

PR #3066, opened 2025-09-02
Allow for ms precision in contest start time.

meisterT:
Also explain why we are stripping off .000.

### DIFF
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -749,9 +749,10 @@ class ExternalContestSourceService
                     [$feedTimezone->getName(), $ourTimezone->getName()]
                 );
             }
+            // We do use precise time with milliseconds if provided, but strip trailing .000 off to make it look nicer.
             $toCheck = [
                 'start_time_enabled' => true,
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -749,9 +749,10 @@ protected function validateAndUpdateContest(Event $event, EventData $data): void
                     [$feedTimezone->getName(), $ourTimezone->getName()]
                 );
             }
+            // We do use precise time with milliseconds if provided, but strip trailing .000 off to make it look nicer.
             $toCheck = [
                 'start_time_enabled' => true,
-                'start_time_string'  => $startTime->format('Y-m-d H:i:s ') . $timezoneToUse,
+                'start_time_string'  => preg_replace('/\.000$/', '', $startTime->format('Y-m-d H:i:s.v ') . $timezoneToUse),
```
eldering: Otherwise the regex won't match:
```suggestion
                'start_time_string'  => preg_replace('/\.000$/', '', $startTime->format('Y-m-d H:i:s.v') . ' ' . $timezoneToUse),
```

eldering (approved):

Issue #1164, opened 2021-08-24
Calculate load of judgedaemon better


nickygerritsen:
[Currently](https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Controller/Jury/JudgehostController.php#L100-L109) we calculate the load of a judgehost by looking at the time between starting a judgetask and getting data back from the judgehost. However, we assign mulitple judgetasks to a judgehost by default and the starttime of a judgetask is calculated when assigning. This means that load is not correct: if we would assign 5 judgetasks to a host at the same time and the runtimes of these runs are `a`, `b`, `c`, `d` and `e` and the compile time is `t`, we report load as `5a + 4b + 3c + 2d + e + 5t`. Better would be to report `a + b + c + d + e + t`. Another way to do it with the current data is to use the runtime of runs, but then we miss compile time.

@meisterT suggested that a judgehost is busy as soon as `fetch-work` returns anything and stops when a next `fetch-work` does not return anything anymore. We should think about how we calculate this.

nickygerritsen:
After discussing IRL we came to the conclusion that we are actually interested in the total load of the judgehosts together, not necessary the load of a single judgehost. We should see if we can use this information to better define what we want and can calculate.

meisterT:
I think this can be closed with the load and interactive timeline now

PR #3072, opened 2025-09-03
Add simple load numbers to judgedaemon overview page.

meisterT:
Part of https://github.com/DOMjudge/domjudge/issues/1164

-----

Example: 
image

### DIFF
--- a/webapp/src/Controller/Jury/JudgehostController.php
+++ b/webapp/src/Controller/Jury/JudgehostController.php
@@ -7,6 +7,7 @@ use App\Doctrine\DBAL\Types\JudgeTaskType;
 use App\Entity\Judgehost;
 use App\Entity\JudgeTask;
 use App\Entity\Judging;
+use App\Entity\JudgingRun;
 use App\Form\Type\JudgehostsType;
 use App\Service\ConfigurationService;
 use App\Service\DOMJudgeService;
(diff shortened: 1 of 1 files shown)


vmcj (commented):
Is this not enough so the 2 above can be fixed? And would it be any faster?

meisterT:
What does the 2 in your comment refer to?

webapp/src/Controller/Jury/JudgehostController.php
```diff
@@ -117,6 +119,44 @@ public function indexAction(Request $request): Response
...
+            $now = Utils::now();
+            $timings = $this->em->createQueryBuilder()
+                ->from(JudgeTask::class, 'jt')
+                ->join(JudgingRun::class, 'jr', 'WITH', 'jr.judgetask = jt')
+                ->join('jr.judging', 'j')
+                ->join('jt.judgehost', 'jh')
+                ->select('jr.endtime, jr.startTime')
+                ->andWhere('jt.judgehost = :judgehost')
+                ->andWhere('jr.startTime IS NOT NULL')
+                ->andWhere('jr.endtime IS NOT NULL')
```
eldering: I guess this way we're ignoring currently running judging runs. Could we also include those?

webapp/src/Controller/Jury/JudgehostController.php
```diff
@@ -117,6 +119,44 @@ public function indexAction(Request $request): Response
                 'value' => 'j' . $lastJobId['jobid'],
             ];
 
+            $now = Utils::now();
+            $timings = $this->em->createQueryBuilder()
```
eldering: Would it make sense to run this query once outside the for loop for all judgehosts?

webapp/src/Controller/Jury/JudgehostController.php
```diff
@@ -117,6 +119,44 @@ public function indexAction(Request $request): Response
...
+                    if ($start_time < $end_time) {
+                        $loads[$i] += ($end_time - $start_time);
+                    }
+                }
+            }
+            // Normalize to [0,1] range.
+            $loads[0] = min(1.0, $loads[0] / (60.0 * 1.0));
+            $loads[1] = min(1.0, $loads[1] / (60.0 * 5.0));
+            $loads[2] = min(1.0, $loads[2] / (60.0 * 15.0));
+            $loads[3] = min(1.0, $loads[3] / (60.0 * 60.0));
```
eldering: Maybe change above
```
          $loadMinutes = [60*1, 60*5, 60*15, 60*60];
```
and then reuse that data here again for deduplication?

webapp/src/Controller/Jury/JudgehostController.php
meisterT: Done.

webapp/src/Controller/Jury/JudgehostController.php
meisterT: Done.

webapp/src/Controller/Jury/JudgehostController.php
meisterT: No, we can't - we don't know what judging_runs were started when until we get the data back from the judgedaemons.

Issue #3009, opened 2025-06-17
Add clarification counter/preview to problemsoverview page


vmcj:
### Description of the enhancement request
The current `/team/problems` page has the statement for a problem but not a counter with the (possible) errata about that problem. As we already know which clarifications are about problems we could provide the errata next to the problem.

### The goal you want to achieve
Make it easier for beginning teams who haven't checked all clarifications. I'm not 100% sure if this is better but to me the statement & clarifications are semi-related.

### Expected behaviour
My first idea was to make:
- An extra button next to statement:
- Default grayed out
- Active when there are clarifications
- Linking to a modal with the same interface as on the `/team` page (so a preview of the text) or directly linking to a list of the `View clarification` (but than the full list) ordered from newest to oldest (and consider if top posting is correct here)
- Making the text of the button bold when a message is not read yet similar to what we have on the main page.

PR #3087, opened 2025-09-06
Add clarifications to the team problemset page.

meisterT:
Fixes https://github.com/DOMjudge/domjudge/issues/3009

------

Example:

image

------

Clicking on one opens the regular clarification modal dialog:

image


vmcj:
Im not 100% sure if I like this better than what I suggested in the issue. This feels out of place to me.

meisterT:
How do like something like this then?

image


-----

image

vmcj:
> How do like something like this then?
> image image

Yes but I was clearly wrong with the greyed out button. It doesn't add anything and just takes up space.

meisterT:
Alright, changed my commit to that effect. Please take a look.

meisterT:
(I need to fix the unit test, will do that after we agreed whether this is the right approach)

vmcj:
I assume `red` is unread & `blue` is read? I think this is good. I first considered bold for unread but that might look out of place.

So LGTM after fixing the unit tests.

PR #3088, opened 2025-09-06
Properly restrict clarification thread to team.

meisterT:
Previously, the following could happen:
- team A requests a clarification, potentially including details the jury doesn't want to broadcast
- the jury replies to everyone, removing the part from the question that they don't want to have broadcasted
- team B views the reply from the jury, sees the whole clarification thread

Found while working on https://github.com/DOMjudge/domjudge/pull/3087


meisterT:
> Isn't the idea that we don't have threads so why are we displaying this?

We do have threads, in case they are accessible.

Issue #2883, opened 2024-12-16
Rejudged solution not properly taken into account in the scoreboard


RaphaelMarinier:
### Description of the problem
With the following events for a given team on a given problem:
1. submission 1 gets TLE
2. submission 2 gets TLE
3. submission 3 gets AC
4. submission 1 gets rejudged and becomes AC

The scoreboard only takes into account submission 3 but not submission 1.

### Your environment
I don't know the version as I don't have access to the machine hosting Domjudge. I will report this later.

### Steps to reproduce

1. Team A does submission 1 on problem B, gets TLE.
2. Team A does submission 2 on problem B, gets TLE.
3. Team A does submission 3 on problem B, gets AC.
4. Judges decide to rejudge all submission of problem B. Submission 1 of that team becomes AC (and the verdict gets automatically applied).

### Expected behaviour

In the scoreboard, submission 1 of team A that finally became AC should be properly taken into account:
- there should be no time penalty for the team on that problem.
- the time taken into account for the team solving the problem should be the time at which submission 1 was done.

### Actual behaviour

In the final scoreboard, we observe that team A got the time penalty for 2 non-AC submissions, and that the time of submission 3 is taken into account.

### Additional info
Note that this was discussed on slack:
https://domjudge-org.slack.com/archives/CHHURFUM7/p1733778384368249
Refreshing the scoreboard cache led to submission 1 being taken into account properly.

meisterT:
Thanks for filing the issue, I missed it earlier. We should fix this.

Some debugging info:
- we do call the `calculateScoreRow` function in the right place after setting the new judging to valid and all other's to invalid
- we do query only for valid judgings, see https://github.com/DOMjudge/domjudge/blob/95df1de7bce91f43600cfcb8760ea85762936614/webapp/src/Service/ScoreboardService.php#L240
- we do receive more than one "valid" judging for the relevant submission:

<img width="728" height="805" alt="Image" src="https://github.com/user-attachments/assets/bf63cd25-4976-48c0-9e68-2e90bb4d71b2" />

But when you expand the judging, you will see it is not valid: 

<img width="853" height="1140" alt="Image" src="https://github.com/user-attachments/assets/b59128c6-b6c8-481c-9527-cf1e59c0ec4c" />

Then we get the first one here: https://github.com/DOMjudge/domjudge/blob/95df1de7bce91f43600cfcb8760ea85762936614/webapp%2Fsrc%2FService%2FScoreboardService.php#L267

[...]

PR #3089, opened 2025-09-06
Fix scoreboard update after rejudging.

meisterT:
Previous buggy flow was something like:
- fetch all submissions that for this team/contest/problem combination
- that are valid
- and have a valid judging

The idea of https://github.com/DOMjudge/domjudge/blob/95df1de7bce91f43600cfcb8760ea85762936614/webapp/src/Service/ScoreboardService.php#L240 was that this would filter the OneToMany collection down to only valid judgings. This is also what happens on the database side.

However, on the doctrine side, the entity manager keeps a cache, and if the submission is already in the cache with the full collection of valid and invalid judgings, it will only merged with new data from the database query, no data (i.e. no judgings) will magically disappear.

In order to fix this, we explicitly filter on the PHP side. An alternative would be to add a new `OneToMany` mapping with a `where` on it and use that.

Fixes https://github.com/DOMjudge/domjudge/issues/2883

### DIFF
--- a/webapp/src/Entity/Submission.php
+++ b/webapp/src/Entity/Submission.php
@@ -359,6 +359,16 @@ class Submission extends BaseApiEntity implements
         return $this->judgings;
     }
 
+    public function getValidJudging(): ?Judging
+    {
+        foreach ($this->judgings as $j) {
+            if ($j->getValid()) {
+                return $j;
+            }
+        }
+        return null;
+    }
+
     public function setLanguage(?Language $language = null): Submission
     {
         $this->language = $language;
@@ -535,6 +545,16 @@ class Submission extends BaseApiEntity implements
     public function getExternalJudgements(): Collection
     {
         return $this->external_judgements;
+    }
+
+    public function getValidExternalJudgement(): ?ExternalJudgement
+    {
+        foreach ($this->external_judgements as $ej) {
+            if ($ej->getValid()) {
+                return $ej;
+            }
+        }
+        return null;
     }
 
     public function setFileForApi(?FileWithName $fileForApi = null): Submission
--- a/webapp/src/Service/ScoreboardService.php
+++ b/webapp/src/Service/ScoreboardService.php
@@ -262,9 +262,9 @@ class ScoreboardService
         foreach ($submissions as $submission) {
             /** @var Judging|ExternalJudgement|null $judging */
             if ($useExternalJudgements) {
-                $judging = $submission->getExternalJudgements()->first() ?: null;
+                $judging = $submission->getValidExternalJudgement();
             } else {
-                $judging = $submission->getJudgings()->first() ?: null;
+                $judging = $submission->getValidJudging();
             }
 
             // three things will happen in the loop in this order:


vmcj:
Why is it so hard (or expensive) to fix this in the DB side? I miss that comment currently.

meisterT:
> In order to fix this, we explicitly filter on the PHP side. An alternative would be to add a new `OneToMany` mapping with a `where` on it and use that.

I am not aware of any fix on the DB side. We could decide to switch off doctrine or disable any entity caching which would cost performance.

In case you are refering to what I wrote in the commit description:
> In order to fix this, we explicitly filter on the PHP side. An alternative would be to add a new OneToMany mapping with a where on it and use that.

The two options are relatively equal. I picked filtering on the PHP side since 99% of submissions have 1 judging, so it is very cheap to do so. The other solution would not change the DB but allow us to to do an implicit DB query but it would require a migration.

webapp/src/Entity/Submission.php
```diff
@@ -359,6 +359,16 @@ public function getJudgings(): Collection
...
     }
 
+    public function getValidJudging(): ?Judging
+    {
+        foreach ($this->judgings as $j) {
+            if ($j->getValid()) {
+                return $j;
+            }
+        }
+        return null;
```
nickygerritsen, line 369:
```suggestion
        return $this->judgings->findFirst(fn(Judging $j) => $j->getValid());
```
maybe?

nickygerritsen (approved):

vmcj:
> The two options are relatively equal. I picked filtering on the PHP side since 99% of submissions have 1 judging, so it is very cheap to do so. The other solution would not change the DB but allow us to to do an implicit DB query but it would require a migration.

I assume the current code will get merged but just for the discussion. I think doing this with the migration is better. We use Doctrine to handle the translation between database and what we get in PHP and we found one location where this is an issue. By doing this with the migration I think we force the solution to other (future) locations, by doing it in the PHP we have more code to read because of something which Doctrine is doing which is easily removed which would re-introduce the bug. We wouldn't change a OneToMany that easily so when we touch that code again we will most likely pickup this change in the `git blame`. A weaker argument is also that I prefer to send the right query already to the database purely to make sure we have a higher change of getting the expected performance by giving the database all the information so the query plan gets correctly optimized.

`tl;dr` I prefer to fix Doctrine bugs/features in the database model to get consistent results. I'm afraid that we would remove the current fix if someone isn't fully aware of what was fixed here.

meisterT:
@vmcj note that in both cases (with this change and the migration), you have a PHP function `getValidJudging()` that you need to call in all places where you want a valid judging. Also in both cases there would still be the function `getJudgings()` which returns all of the judgings. Then on the caller side you need to decide what function you need / want on a case by case basis, regardless of the implementation of that function. I might be missing something, but I don't see which class of bugs gets prevented.

vmcj:
> @vmcj note that in both cases (with this change and the migration), you have a PHP function `getValidJudging()` that you need to call in all places where you want a valid judging. Also in both cases there would still be the function `getJudgings()` which returns all of the judgings. Then on the caller side you need to decide what function you need / want on a case by case basis, regardless of the implementation of that function. I might be missing something, but I don't see which class of bugs gets prevented.

Ok, I still don't understand but I'll assume you know what you're doing.

PR #3090, opened 2025-09-07
Bump timeout on GitHub to 30mins while preparing the chroot takes lon…

meisterT:
…ger than usual.


meisterT:
https://github.com/DOMjudge/domjudge/actions/runs/17460184433/job/49582860429?pr=3074 is an example where the whole integration tests pass in 11m and setting up chroot takes 88s. Now that step alone takes sometimes more than 20mins.

PR #3091, opened 2025-09-07
Record submission source in database.

meisterT:
The motivation for this is that we can use the data to fix the various errors we have seen across the interface when shadowing one by one. But it can also be useful for other reasons, e.g. being able to count how many submissions were received via the API vs web UI.


meisterT:
(marked as draft since it requires a little more work)

nickygerritsen (approved):

vmcj (approved):
I assume displaying it somewhere is left for another PR.

PR #3093, opened 2025-09-08
Fix cgroup checks for containers

Kevinjil:
There does not have to be a `slice` string in containers. For example, when running the container in podman:
```
domjudge@5ad0debee049:/$ cat /proc/self/cgroup
0::/libpod_parent/libpod-5ad0debee049e150d851d81fa3e8184ee20725412e5702c7ffe3aae7e5173207
```

Fixes starting issue introduced in b2720396d7f2c9e696002703c3a160a2fb895995

### DIFF
--- a/judge/create_cgroups.in
+++ b/judge/create_cgroups.in
@@ -34,8 +34,8 @@ You can try using cgroup V1 by adding systemd.unified_cgroup_hierarchy=0 to the 
     if ! echo "+cpuset" >> /sys/fs/cgroup/cgroup.subtree_control; then
         cgroup_error_and_usage "Error: Cannot add +cpuset to cgroup.subtree_control; check kernel params."
     fi
-    if ! grep -q "slice" /proc/self/cgroup; then
-        cgroup_error_and_usage "Error: Cgroups not configured properly, missing systemd slice under /proc/self/cgroup. If running under docker, make sure to set cgroupns=host."
+    if grep -q ":/$" /proc/self/cgroup; then
+        cgroup_error_and_usage "Error: Cgroups not configured properly, missing cgroup hierarchy prefix under /proc/self/cgroup. If running in a container, make sure to set cgroupns=host."
     fi
 
 else # Trying cgroup V1:


Kevinjil:
Note: container ran with this command:
```
sudo podman --cgroup-manager cgroupfs run --cgroupns host --rm -it --entrypoint /bin/bash docker.io/domjudge/domjudge-contributor
```

meisterT:
We added this check specifically for a problem that @nickygerritsen had.

When not properly set he saw
`0::/`

So one option would be to test that it's not just this but there is more after the `/`.

WDYT?

Kevinjil:
I am not sure that we should enforce a path prefix in the cgroup hierarchy. When running in a systemd slice it will exist, and when running in a container it might exist depending on your deployment. There might be people out there using a different setup where there is no path prefix.

Let me dig a little deeper.

Kevinjil:
@meisterT After digging some deeper, I do not think it makes sense to enforce a prefix of the initiating process in the cgroup hierarchy.

vmcj:
> @meisterT After digging some deeper, I do not think it makes sense to enforce a prefix of the initiating process in the cgroup hierarchy.

If this is a revert I prefer to see the old commit also for the discussion.

Just to be sure, you can run a working judgedaemon in your podman container?

meisterT:
@Kevinjil have you found a platform where we would see `0::/` but it is correctly set up and works?

Kevinjil:
So far, from what I've noticed, is that you can perfectly fine run processes with a cgroup line of `0::/`.

I've been playing around with a non-systemd distro to see how they behave. I picked Alpine, which uses OpenRC. When you open a shell, you are actually in the `0::/` cgroup. But, as soon as you launch it on boot using the init system, it assigns you a cgroups service is running.

As time is limited, I assume that most init systems will assign a cgroup with prefix if you run a service when cgroups are enabled. I'll update the PR.

P.S. something to discuss during NWERC: we might want to either update our software requirements in the manual or change some scripts, as there are quite some hidden assumptions.

PR #3094, opened 2025-09-08
Switch bucket size for problem analysis

Kevinjil:
The graph has less space on the page, so the buckets got too narrow.

Old graph:
image

New graph:
image


Kevinjil:
Kind of related note: I checked the team analysis page, which still uses the old graph. IMO, it does not make sense to change that graph to this bucket-style graph, as in most contest a single team only submits rarely.

PR #3095, opened 2025-09-10
Change diff mode selection to radio button

Kevinjil:
Visually, this is more in line with the rest of the design. It also clears some vertical space which can be used by the monaco editor.

Old:
image
image

New:
image
image

### DIFF
--- a/webapp/templates/jury/partials/submission_diff.html.twig
+++ b/webapp/templates/jury/partials/submission_diff.html.twig
@@ -57,10 +57,10 @@
                 {% endif %}
                 <div id="diffselect-{{ diff_id }}" class="btn-group">
                     <a href="#" role="button" class="btn btn-secondary btn-sm pe-none" aria-disabled="true"><i class="fas fa-code-branch"></i></a>
-                    <input type="radio" class="btn-check" name="{{ diff_id }}-mode" id="{{ diff_id }}-side-by-side" value="side-by-side" autocomplete="off">
-                    <label class="btn btn-outline-secondary btn-sm" for="{{ diff_id }}-side-by-side">side-by-side</label>
-                    <input type="radio" class="btn-check" name="{{ diff_id }}-mode" id="{{ diff_id }}-inline" value="inline" autocomplete="off">
-                    <label class="btn btn-outline-secondary btn-sm" for="{{ diff_id }}-inline">inline</label>
+                    {%- for mode_id, mode in diff_modes %}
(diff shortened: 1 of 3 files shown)


webapp/templates/jury/partials/submission_diff.html.twig
```diff
@@ -54,8 +55,15 @@
                         <i class="fas fa-pencil-alt"></i> Edit
                     </a>
                 {% endif %}
+                <div id="diffselect-{{ diff_id }}" class="btn-group">
+                    <a href="#" role="button" class="btn btn-secondary btn-sm pe-none" aria-disabled="true"><i class="fas fa-code-branch"></i></a>
+                    <input type="radio" class="btn-check" name="{{ diff_id }}-mode" id="{{ diff_id }}-side-by-side" value="side-by-side" autocomplete="off">
+                    <label class="btn btn-outline-secondary btn-sm" for="{{ diff_id }}-side-by-side">side-by-side</label>
+                    <input type="radio" class="btn-check" name="{{ diff_id }}-mode" id="{{ diff_id }}-inline" value="inline" autocomplete="off">
+                    <label class="btn btn-outline-secondary btn-sm" for="{{ diff_id }}-inline">inline</label>
```
meisterT: perhaps "inline diff"?

nickygerritsen (approved):

webapp/templates/jury/partials/submission_diff.html.twig
Kevinjil: @meisterT Hmm, I think that either both `side-by-side` and `inline` should have the diff suffix then or none. Other option might be to add it next to the icon:
image

meisterT: I don't have a strong opinion, so pick whatever you think is best

eldering:
Not sure how easy it would be, but would be nice to have a third option "diff off" to just show the plain source.

And even better would be to store this in a cookie, I think. But feel free to let either be for another PR.

Kevinjil:
@eldering This is the best I could easily do. The commit message contains the reasoning why we do it this way. This also adds context that this is about diffs, resolving @meisterT comment.
image

Kevinjil:
@eldering The mode was already stored in local storage. I think it makes sense to keep it there and not move it to cookies, as there is no need to communicate it to the webserver.

meisterT:
LGTM, :ship: it

eldering (approved):

PR #3096, opened 2025-09-13
Make sure the webserver can reach the installation directory

vmcj:
My current maintainer-install is on a very non standard location and didn't work. I think some of the ACLs were implied by standard FHS but it's better to make them explicit IMO.

I wonder why the ` $(CURDIR)/etc/domserver-static.php"` never broke as it's clearly needed but hasn't been allowed if I understand the `git log` correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -250,6 +250,9 @@ inplace-install-l:
 	@echo "    - Give the webserver access to things it needs"
 	@echo "        setfacl    -m   u:$(WEBSERVER_GROUP):r    $(CURDIR)/etc/dbpasswords.secret"
 	@echo "        setfacl    -m   u:$(WEBSERVER_GROUP):r    $(CURDIR)/etc/symfony_app.secret"
+	@echo "        setfacl    -m   u:$(WEBSERVER_GROUP):r    $(CURDIR)/etc/domserver-static.php"
+	@echo "        setfacl -R -m d:u:$(WEBSERVER_GROUP):rx   $(CURDIR)/webapp"
+	@echo "        setfacl -R -m   u:$(WEBSERVER_GROUP):rx   $(CURDIR)/webapp"
 	@echo "        setfacl -R -m d:u:$(WEBSERVER_GROUP):rwx  $(CURDIR)/webapp/var"
 	@echo "        setfacl -R -m   u:$(WEBSERVER_GROUP):rwx  $(CURDIR)/webapp/var"
 	@echo "        setfacl -R -m d:m::rwx          $(CURDIR)/webapp/var"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -313,7 +319,12 @@ inplace-postinstall-nginx: inplace-postinstall-permissions
 	fi; \
 	service="systemctl restart $$service"; \
 	ln="ln -sf $(CURDIR)/etc/domjudge-fpm.conf $$phppool/domjudge-fpm.conf"; \
-	echo $$ln; echo $$service; $$ln; $$service
+	echo $$ln; echo $$service; $$ln; $$service;
+	
+	while [ `pwd` != "/" ]; do \
```
meisterT: Do we have to do a similar thing for apache?

vmcj: Now we have, I moved the check.

Makefile
```diff
@@ -250,13 +250,18 @@ inplace-install-l:
...
+	@echo "        setfacl -R -m d:u:$(WEBSERVER_GROUP):rx   $(CURDIR)/webapp"
+	@echo "        setfacl -R -m   u:$(WEBSERVER_GROUP):rx   $(CURDIR)/webapp"
 	@echo "        setfacl -R -m d:u:$(WEBSERVER_GROUP):rwx  $(CURDIR)/webapp/var"
 	@echo "        setfacl -R -m   u:$(WEBSERVER_GROUP):rwx  $(CURDIR)/webapp/var"
 	@echo "        setfacl -R -m d:m::rwx          $(CURDIR)/webapp/var"
 	@echo "        setfacl -R -m   m::rwx          $(CURDIR)/webapp/var"
 	@echo "        # Also make sure you keep access"
 	@echo "        setfacl -R -m d:u:$(DOMJUDGE_USER):rwx  $(CURDIR)/webapp/var"
 	@echo "        setfacl -R -m   u:$(DOMJUDGE_USER):rwx  $(CURDIR)/webapp/var"
+  @echo "        And manually make sure the webserver has traversal access to: $(CURDIR)
```
eldering: Indentation is off here.

vmcj: Fixed, and directly also fixed the missing quotes.

Makefile
```diff
@@ -288,6 +297,10 @@ inplace-postinstall-permissions:
 		chcon -R -t httpd_sys_rw_content_t $(CURDIR)/webapp/public/images; \
 		chcon    -t httpd_exec_t $(CURDIR)/lib/alert; \
 	fi
+	while [ `pwd` != "/" ]; do \
```
Kevinjil: Is there a reason why we need to list contents up to the root? I'd assume only starting from the installation path should be enough. I would prefer not changing permissions on system-defined directories.

eldering: Some programs are executed using the full absolute path. If any component of that path is not traversable (`x` permission) then it will fail.
But I do agree that we should be careful modifying stuff above the installation directory, so I'd say at least a warning if changing that would be in order.

vmcj: > Is there a reason why we need to list contents up to the root? I'd assume only starting from the installation path should be enough. I would prefer not changing permissions on system-defined directories.

I needed it on my setup where `o` has no `x` bit set on any directory. We wouldn't touch system-defined directories.. we only add an additional `ACL` which allows the webserver to follow the path.

The main problem is that doing the `su` to the `www-data` to check if it has access is quite hard and might often not allowed. So we either need to make this explicit like we have here or would need to come up with a nice test which might do interactive things which is not allowed for that account.

vmcj: I'll take the `setfacl` out which goes out of the installation path. We can discuss it in another PR.

PR #3098, opened 2025-09-14
Restructure documentation to prefer nginx

vmcj:
This is highly opiniated but we have deployed with nginx for years and most new users will use nginx over apache2 currently (although they might even already use other webservers).

Most of our enhancements are added to the nginx config and not always added to the apache2 config also.

This should make support slightly easier in Slack as new users should be steered into a configuration we're more familiar with.


nickygerritsen (approved):

eldering (approved):
No objections, but I didn't fully check the PR itself (whether all documentation was updated).

Kevinjil (approved):

vmcj:
> No objections, but I didn't fully check the PR itself (whether all documentation was updated).

I did a grep for `nginx` and those were the logical places for such documentation. We can do others if we find them.

PR #3099, opened 2025-09-15
MathJax demo

vmcj:
Purely a PoC, only tested/demonstrated at the jury clarification pages. I assume it works for the team pages but that's only because of the shared code.

I didn't test for MathML yet, and the other options for the MathJax object. We would need a couple of example clarifications to test with and do some visual inspection on the result.

I think the fonts can also be version controlled, but GitHub doesn't have a release in the MathJax organization which seems to ship those.


meisterT:
> I assume it works for the team pages but that's only because of the shared code.

Weirdly enough not for the modal:

image

Also it takes a few seconds to render correctly.

> I think the fonts can also be version controlled, but GitHub doesn't have a release in the MathJax organization which seems to ship those.

Yeah would be nice if would also come from packagist.

meisterT:
I tested a lot of things with the help of Gemini, the cross product is the only thing that didn't work:

```
    $$\vec{a} \times \vec{b} = \begin{vmatrix} \mathbf{i} & \mathbf{j} & \mathbf{k} \\ a_x & a_y & a_z \\ b_x & b_y & b_z \end{vmatrix}$$
```

image


Should look like this:
image


I also tested in a mathjax playground that this works in general. Not sure why it doesn't work here and whether it is important to fix.

meisterT:
Or rather like this:

image

vmcj:
> > I assume it works for the team pages but that's only because of the shared code.
> 
> Weirdly enough not for the modal:
> image
> 

I think I need to apply the MathJax code to the new element from the modal, I assume those elements don't exist in the DOM when MathJax is loaded. Similar to what we need for clarifications previews.

> Also it takes a few seconds to render correctly.

I think that's because of the rendering being done on clientside. It might be possible to do this serverside if we use nodejs but I don't think MathJax has a PHP equivalent.

> Yeah would be nice if would also come from packagist.

I'll keep on looking, but might be for a future PR if this works before the weekend.

vmcj:
> I tested a lot of things with the help of Gemini, the cross product is the only thing that didn't work:
> 
> ```
>     $$\vec{a} \times \vec{b} = \begin{vmatrix} \mathbf{i} & \mathbf{j} & \mathbf{k} \\ a_x & a_y & a_z \\ b_x & b_y & b_z \end{vmatrix}$$
> ```

So when you add that part to the menu it does work so it's special to the clarifications. I would almost say we make a separate issue for it if the other issues are fixed.

vmcj:
> I tested a lot of things with the help of Gemini, the cross product is the only thing that didn't work:
> 
> ```
>     $$\vec{a} \times \vec{b} = \begin{vmatrix} \mathbf{i} & \mathbf{j} & \mathbf{k} \\ a_x & a_y & a_z \\ b_x & b_y & b_z \end{vmatrix}$$
> ```

So when you add that part to the menu it does work so it's special to the clarifications. I would almost say we make a separate issue for it if the other issues are fixed. 

> > I tested a lot of things with the help of Gemini, the cross product is the only thing that didn't work:
> > ```
> >     $$\vec{a} \times \vec{b} = \begin{vmatrix} \mathbf{i} & \mathbf{j} & \mathbf{k} \\ a_x & a_y & a_z \\ b_x & b_y & b_z \end{vmatrix}$$
> > ```
> 
> So when you add that part to the menu it does work so it's special to the clarifications. I would almost say we make a separate issue for it if the other issues are fixed.

I found the problem, the `\\` gets replaced by `\` by the default `markdown_to_html` filter.

vmcj:
The packages on packagist doesn't have the fonts so we need to store those locally. I now installed the 2 packages with `npm`. So we have 2 possible solutions:

- Use `npm` in the development scripts next to `composer`, the advantage is that we can version control those
- Check the needed files in and not burden ourselves with another package manager.

If it was only for those 2 `npm` packages I would vote for option 2, but we have some other JS libraries which we might want to also manage with `npm`.

webapp/src/Twig/TwigExtension.php
```diff
@@ -1406,4 +1408,35 @@ public function extensionToMime(string $extension): string
     {
         return DOMJudgeService::EXTENSION_TO_MIMETYPE[$extension];
     }
+
+    public function domjudgeMarkdownToHTML(string $markdown): string
+    {
+        $latexFound = [];
+        while(true) {
```
vmcj:
```suggestion
        while (true) {
```

webapp/src/Twig/TwigExtension.php
```diff
@@ -1406,4 +1408,35 @@ public function extensionToMime(string $extension): string
...
+            $start = strpos($markdown,'$$');
```
vmcj:
```suggestion
            $start = strpos($markdown, '$$');
```

webapp/src/Twig/TwigExtension.php
```diff
@@ -1406,4 +1408,35 @@ public function extensionToMime(string $extension): string
...
+            $markdown = $newMarkdown;
+        }
+
+        /** @var MarkdownRuntime $runtime */
+        $runtime = $this->twig->getRuntime(MarkdownRuntime::class);
+        $markdown = (string)$runtime->convert($markdown);
+
+        $new = '';
+        foreach ($latexFound as $inlineLatex) {
+            $replacedStart = strpos($markdown,'$LaTeX$');
```
vmcj:
```suggestion
            $replacedStart = strpos($markdown, '$LaTeX$');
```

webapp/templates/jury/clarification.html.twig
```diff
@@ -18,7 +18,16 @@
             </div>
         {% endif %}
 
+
+
```
vmcj:
```suggestion
```

vmcj:
Most of the last commits are actually to make yarn work so will get rebased away when the other PR is merged.

vmcj:
We could also render everything serverside, but in that case we can also use https://github.com/KaTeX/KaTeX or even have an LaTeX installation and get an image out of that.

webapp/src/EventListener/AddContentSecurityPolicyListener.php
```diff
@@ -17,7 +17,10 @@ public function __invoke(ResponseEvent $event): void
         // the profiler requires 'unsafe-eval' for script-src 'self'.
         $response = $event->getResponse();
         $cspExtra = $this->profiler ? "'unsafe-eval'" : "";
-        $csp = "default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' $cspExtra; img-src 'self' data:; worker-src 'self' blob:; font-src 'self' data:;";
+        $csp  = "default-src 'self';";
+        $csp .= "font-src 'self' data:; img-src 'self' data:;";
+        $csp .= "style-src 'self' 'unsafe-inline'; worker-src 'self' blob:";
+        $csp .= "script-src 'self' 'unsafe-inline' $cspExtra;";
```
nickygerritsen: This is an unrelated change and seems to only change formatting. If we do this, lets put each -src on a separate line (but preferably in a separate PR)

webapp/src/Twig/TwigExtension.php
```diff
@@ -124,6 +125,7 @@ public function getFilters(): array
             new TwigFilter('medalType', $this->awards->medalType(...)),
             new TwigFilter('numTableActions', $this->numTableActions(...)),
             new TwigFilter('extensionToMime', $this->extensionToMime(...)),
+            new TwigFilter('domjudge_markdown_to_html', $this->domjudgeMarkdownToHTML(...), ['is_safe' => ['html']]),
```
nickygerritsen:
```suggestion
            new TwigFilter('domjudgeMarkdownToHtml', $this->domjudgeMarkdownToHTML(...), ['is_safe' => ['html']]),
```
? at least the others are cased this way

webapp/src/Twig/TwigExtension.php
```diff
@@ -1406,4 +1408,35 @@ public function extensionToMime(string $extension): string
     {
         return DOMJudgeService::EXTENSION_TO_MIMETYPE[$extension];
     }
+
+    public function domjudgeMarkdownToHTML(string $markdown): string
+    {
+        $latexFound = [];
+        while (true) {
```
nickygerritsen: It's not very clear to me what you are doing here? Are you making sure that we aren't markdowning any of the LaTex code? If so, maybe at least explain that we are doing that.

webapp/src/EventListener/AddContentSecurityPolicyListener.php
vmcj: It's from the merge conflict I got, will fix it with rebasing.

webapp/src/Twig/TwigExtension.php
vmcj: I'll add a comment.

meisterT:
> > I tested a lot of things with the help of Gemini, the cross product is the only thing that didn't work:
> > ```
> >     $$\vec{a} \times \vec{b} = \begin{vmatrix} \mathbf{i} & \mathbf{j} & \mathbf{k} \\ a_x & a_y & a_z \\ b_x & b_y & b_z \end{vmatrix}$$
> > ```
> 
> So when you add that part to the menu it does work so it's special to the clarifications. I would almost say we make a separate issue for it if the other issues are fixed.
> 
> > > I tested a lot of things with the help of Gemini, the cross product is the only thing that didn't work:
> > > ```
> > >     $$\vec{a} \times \vec{b} = \begin{vmatrix} \mathbf{i} & \mathbf{j} & \mathbf{k} \\ a_x & a_y & a_z \\ b_x & b_y & b_z \end{vmatrix}$$
> > > ```
> > 
> > 
> > So when you add that part to the menu it does work so it's special to the clarifications. I would almost say we make a separate issue for it if the other issues are fixed.
> 
> I found the problem, the `\\` gets replaced by `\` by the default `markdown_to_html` filter.

Interestingly enough, only the preview is broken. The actual clarification looks correct:

image

.github/jobs/baseinstall.sh
```diff
@@ -30,7 +30,6 @@ section_end
 
 section_start "Install JS frontend dependencies"
 cd webapp
-apt-get update; apt-get install -y yarnpkg
```
meisterT: Why did this change?

webapp/src/Twig/TwigExtension.php
```diff
@@ -1406,4 +1408,39 @@ public function extensionToMime(string $extension): string
...
     }
+
+    /**
+     * Extract all LaTeX code from the given string, sanitize the markdown and
+     * inject the original LaTeX code back so MathJax can render it.
+     */
+    public function domjudgeMarkdownToHTML(string $markdown): string
+    {
+        $latexFound = [];
+        while (true) {
```
meisterT: I wonder whether we should rather use a regular expression matching here, e.g. something like:

```
$markdown = preg_replace_callback(
        '/(\$\$[\s\S]*?\$\$)/',
        function (array $matches) {
            // Store and replace matches
            $this->latexFound[] = $matches[1];
            return '$LaTeX$';
        },
        $markdown
    );
```

webapp/src/Twig/TwigExtension.php
```diff
@@ -1406,4 +1408,39 @@ public function extensionToMime(string $extension): string
...
+            $newMarkdown .= '$LaTeX$';
+            $newMarkdown .= substr($markdown, $end+$start+4);
+            $markdown = $newMarkdown;
+        }
+
+        /** @var MarkdownRuntime $runtime */
+        $runtime = $this->twig->getRuntime(MarkdownRuntime::class);
+        $markdown = (string)$runtime->convert($markdown);
+
+        $new = '';
```
meisterT: Then this loop can be replaced with this single line:

```
$markdown = str_replace('$LaTeX$', fn() => array_shift($this->latexFound), $markdown);
```

webapp/templates/jury/export/clarifications.html.twig
```diff
@@ -45,7 +45,11 @@
             <div class="page-break"></div>
             <h1>{{ title }}</h1>
         {% endif %}
-        <h2>{{ queues[queue] }}</h2>
+        {% if queues[queue] is defined %}
```
meisterT, line 48: nit: seems unrelated to adding JAX, so let's extract into separate PR?

webapp/templates/jury/menu.html.twig
```diff
@@ -7,6 +7,8 @@
 
     {% set current_route = app.request.attributes.get('_route') %}
 
+    <span>Euler's identity: $( e^{i\pi} + 1 = 0 )$</span>
```
meisterT: just leaving a comment here so that we don't forget to remove it

Kevinjil:
Overall, looks good.


> Also it takes a few seconds to render correctly.

I could not reproduce this. Where did it take long for you as it was instant for me in Firefox.


Do we want to think of some clever way to handle truncating within a MathJax expression:
image

.github/jobs/baseinstall.sh
vmcj: Because yarnpkg is now added to the image.

meisterT: :+1: , let's separate this from this PR?

vmcj:
> Overall, looks good.
> 
> > Also it takes a few seconds to render correctly.
> 
> I could not reproduce this. Where did it take long for you as it was instant for me in Firefox.

I might have changed where the JS applies so it gets faster as it doesn't need to parse the whole file.
> 
> Do we want to think of some clever way to handle truncating within a MathJax expression: image

Might be smart yes. But I rather first see if we think this is useful, it still feels very complicated to me.

webapp/src/Twig/TwigExtension.php
eldering: This will break if you add a literal `$LaTeX$` string.

vmcj:
> Do we want to think of some clever way to handle truncating within a MathJax expression:

Discuss it with @eldering.

meisterT (approved):

webapp/src/Twig/TwigExtension.php
```diff
@@ -1282,4 +1289,36 @@ public function extensionToMime(string $extension): string
...
+            },
+            $markdown
+        );
+
+        /** @var MarkdownRuntime $runtime */
+        $runtime = $this->twig->getRuntime(MarkdownRuntime::class);
+        $markdown = (string)$runtime->convert($markdown);
+
+        return preg_replace_callback(
+            '/'.$latexPlaceholder.'/',
```
nickygerritsen:
```suggestion
            "/$latexPlaceholder/",
```

webapp/src/Twig/TwigExtension.php
```diff
@@ -1282,4 +1289,36 @@ public function extensionToMime(string $extension): string
...
+            fn() => array_shift($this->latexFound), $markdown
```
nickygerritsen:
```suggestion
            fn(): string => array_shift($this->latexFound), $markdown
```

webapp/src/Twig/TwigExtension.php
```diff
@@ -1282,4 +1289,36 @@ public function extensionToMime(string $extension): string
...
+    public function domjudgeMarkdownToHTML(string $markdown): string
+    {
+        $latexPlaceholder = Uuid::uuid4()->toString();
+        while (str_contains($markdown, $latexPlaceholder)) {
+            $latexPlaceholder = Uuid::uuid4()->toString();
+        }
+
+        $markdown = preg_replace_callback(
+            '/(\$[\s\S]*?\$)/',
+            function (array $matches) use ($latexPlaceholder) {
```
nickygerritsen:
```suggestion
            function (array $matches): string use ($latexPlaceholder) {
```

PR #3109, opened 2025-09-20
Document setting timelimit strategy

vmcj:


doc/manual/config-basic.rst
```diff
@@ -140,4 +140,11 @@ it need not have started yet). You can verify whether the submissions
 gave the expected answer in the Judging Verifier, available from
 the jury index page.
 
+After this check change the evaluation to full in the configuration
```
meisterT: I would avoid recommending changing configuration. People forget to change it back. What we could say instead: rejudge the whole contest with a higher overshoot, as well as judge remaining testcases.

webapp/tests/E2E/Controller/ControllerRolesTraversalTest.php
```diff
@@ -227,10 +227,12 @@ public function visitWithNoContest(string $url, bool $dropdown): void
             try {
                 self::assertSelectorExists('a#navbarDropdownContests:contains("no contest")', "Failed at: " . $url);
             } catch (\Exception $e) {
-                $gitlabArtifacts = getenv('GITLABARTIFACTS');
-                if ($gitlabArtifacts != '') {
-                    $fileHandler = fopen(sprintf("%s/%s", $gitlabArtifacts, str_replace('/', '_s_', $url)), 'w');
-                    fwrite($fileHandler, $response->getContent());
+                if (getenv('CI')) { // We're running in GHA.
```
meisterT: This change seems unrelated to the rest of the PR?

vmcj: Yes and no, something broke so I needed it.

And after fixing it's indeed unrelated. Will take it out.

meisterT (approved):
needs some commit squashing, otherwise LGTM

PR #3110, opened 2025-09-21
Allow www-data to store the files from import-contest

vmcj:
On none standard locations the extended ACLs wouldn't allow this. I've now specified those explicit but I think giving access to the whole images dir would be easier and more future proof as alternative. It would give the web server access to change the DOMjudge logo though which it never needs to do (and would be an indication of broad access).


meisterT:
The changes look fine, but I wonder: how comes that we didn't need this before, e.g. my maintainer install is working fine without it. I don't remember doing these steps manually, but perhaps I forgot.

vmcj:
> The changes look fine, but I wonder: how comes that we didn't need this before, e.g. my maintainer install is working fine without it. I don't remember doing these steps manually, but perhaps I forgot.

I think your git checkout is on a much better location than mine is, you most likely use `/home/user/git/domjudge` mine lives in `/nonFHS/git/domjudge` so some defaults are not there at all.

And we've had some issues with those ACLs not being set for WF setups. I think this is part of that problem, every OS can set those defaults different.

meisterT (approved):

Kevinjil (approved):

PR #3111, opened 2025-09-21
#2307 Multi-pass: copy feedback dir between passes (but remove nextpass.in)

mpsijm:
This commit has been used in NWERC 2024 (d7cc716216ec0261864e02d844cfd0207264616f), as well as GCPC 2025 and BAPC 2025.
It follows the latest release-candidate specification of multi-pass validation: https://www.kattis.com/problem-package-format/spec/2025-09.html#multi-pass-validation (the mirror at https://icpc.io/problem-package-format/spec/2025-09.html should be deployed "soon™").
Copying the feedback directory to the next pass is an implementation of "All other files inside the feedback directory are guaranteed to persist between passes." Additionally, "Note that the `nextpass.in` will be removed before the next pass."

See also https://github.com/Kattis/problem-package-format/issues/446#issuecomment-3246509869


judge/judgedaemon.main.php
```diff
@@ -1451,6 +1451,11 @@ function judge(array $judgeTask): bool
         $passdir = $testcasedir . '/' . $passCnt;
         mkdir($passdir, 0755, true);
 
+        if ($passCnt > 1) {
+            system('cp -R ' . $passdir . '/../' . ($passCnt - 1) . '/feedback ' . $passdir . '/');
```
meisterT: let's add some quoting here

judge/judgedaemon.main.php
```diff
@@ -1451,6 +1451,11 @@ function judge(array $judgeTask): bool
...
+            system('rm ' . $passdir . '/feedback/nextpass.in');
```
meisterT: ... and here

judge/judgedaemon.main.php
```diff
@@ -1451,6 +1451,11 @@ function judge(array $judgeTask): bool
         $passdir = $testcasedir . '/' . $passCnt;
         mkdir($passdir, 0755, true);
 
+        if ($passCnt > 1) {
```
meisterT, line 1458: Add a comment explaining the intent here?

judge/judgedaemon.main.php
mpsijm: Good one! I've found `dj_escapeshellargs` in surrounding `system()` calls, so I've used that :slightly_smiling_face:

judge/judgedaemon.main.php
mpsijm: Good idea, done :smile:

PR #3112, opened 2025-09-22
Fix incorrect link generation for testcase audit logs

zapotocnylubos:
The stored ID in the audit log is the problem ID, not a testcase ID

So searching for a testcase by a problem ID results in an incorrect link to a completely different problem.

Storing the problematic audit log happens here https://github.com/DOMjudge/domjudge/blob/af926294f1a5c054dd170a373fe538d47012b0e8/webapp/src/Controller/Jury/ProblemController.php#L634

I also saw a generic audit log mechanism in BaseController
https://github.com/DOMjudge/domjudge/blob/af926294f1a5c054dd170a373fe538d47012b0e8/webapp/src/Controller/BaseController.php#L154

I am wondering if this is a correct-intended behavior. On one side, I see performance/logical reasons why to store only the problem ID, but keeping consistency in mind, it would also seem logical to store the testcase ID and look up the problem ID as it was originally coded.

Should we change the 

```php
$this->dj->auditlog('testcase', $probId
```

in all **3 locations** to pass a testcase ID instead? Or use the change in the link generation I provided in this MR?

- https://github.com/DOMjudge/domjudge/blob/af926294f1a5c054dd170a373fe538d47012b0e8/webapp/src/Controller/Jury/ProblemController.php#L634
- https://github.com/DOMjudge/domjudge/blob/af926294f1a5c054dd170a373fe538d47012b0e8/webapp/src/Controller/Jury/ProblemController.php#L736
- https://github.com/DOMjudge/domjudge/blob/af926294f1a5c054dd170a373fe538d47012b0e8/webapp/src/Controller/Jury/ProblemController.php#L863

### DIFF
--- a/webapp/src/Controller/Jury/AuditLogController.php
+++ b/webapp/src/Controller/Jury/AuditLogController.php
@@ -155,11 +155,8 @@ class AuditLogController extends AbstractController
                 }
                 return $this->generateUrl('jury_user', ['userId' => $id]);
             case 'testcase':
-                $testcase = $this->em->getRepository(Testcase::class)->find($id);
-                if ($testcase && $testcase->getProblem()) {
-                    return $this->generateUrl('jury_problem_testcases', ['probId' => $testcase->getProblem()->getProbid()]);
-                }
-                break;
+                // For testcase audit logs, the ID is actually the problem ID
+                return $this->generateUrl('jury_problem_testcases', ['probId' => $id]);
         }
         return null;
     }


vmcj:
> I am wondering if this is a correct-intended behavior. On one side, I see performance/logical reasons why to store only the problem ID, but keeping consistency in mind, it would also seem logical to store the testcase ID and look up the problem ID as it was originally coded.
> 

Are you willing to do some testing for this? I think the call shouldn't be that heavy so sounds reasonable to log all relevant information.

zapotocnylubos:
Sure, I can try to do some load testing and measure it.
The main problem is that it's now inconsistent.
The controller stores the problem ID, and the auditlog table uses that problem ID as it is a testcase ID (which it is not).
Only one way should be chosen.

1. Store the testcase ID and the lookup problem ID when rendering the link
2. Store problem ID, don't do a lookup (in this MR)

meisterT:
Since we link to the problem (or all of its testcases) anyway (and the text mentions the test case ID, so all info is there), I think the PR is the right thing to do.

meisterT (approved):

PR #3113, opened 2025-09-22
Add MathML sanitizer

vmcj:
As alternative for MathJax, this PR fixes removing MathML tags.

I verified with: https://support.freedomscientific.com/SurfsUp/MathML-Samples.html
More elements can be found at: https://developer.mozilla.org/en-US/docs/Web/MathML/Reference/Element

All seem to be reasonable safe, e.q. no XSS unless we allow for onload,href etc. we still remove the <body> tag etc.

For now we only allow for the tags and most of the time no attributes as those are probably too advanced for a contest setting. The only reason some are added is for demonstration and to make the examples work.

mathml

### DIFF
--- a/webapp/config/packages/html_sanitizer.yaml
+++ b/webapp/config/packages/html_sanitizer.yaml
@@ -4,3 +4,43 @@ framework:
             app.clarification_sanitizer:
                 allow_safe_elements: true
                 allow_relative_medias: true
+                allow_elements:
+                    msup: []
+                    mrow: []
+                    mfrac: []
+                    msqrt: []
+                    mfenced:
+                        - open
+                        - close
+                    mtable: []
+                    mtd: []
+                    mtr: []
+                    msubsup: []
+                    semantics: []
+                    mroot: []
+                    menclose:
+                        - notation
+                    mlabeledtr: []
+                    mstyle:
+                        - mathcolor
+                        - fontfamily
+                        - displaystyle
+                    math:
+                        - display
+                        - class
+                    mi:
+                        - class
+                    mn:
+                        - class
+                    mo:
+                        - class
+                    annotation-xml:
+                        - encoding
+                    apply: []
+                    bvar: []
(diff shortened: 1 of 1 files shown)


meisterT:
Looking at the source code of these examples, picking a random one the syntax looks quite complex. I don't think teams will want to deal with it (without internet!) during a contest:
```
<math xmlns='http://www.w3.org/1998/Math/MathML' display='block'>
<mrow>
  <mrow>
    <msup>
      <mi>x</mi>
      <mn>2</mn>
    </msup>
    <mo>+</mo>
    <mrow>
      <mn>4</mn>
      <mo>&InvisibleTimes;</mo>
      <mi>x</mi>
    </mrow>
    <mo>+</mo>
    <mn>4</mn>
  </mrow>
  <mo>=</mo>
  <mn>0</mn>
</mrow>
</math>
```

vmcj:
> Looking at the source code of these examples, picking a random one the syntax looks quite complex. I don't think teams will want to deal with it (without internet!) during a contest:
> 
> ```
> <math xmlns='http://www.w3.org/1998/Math/MathML' display='block'>
> <mrow>
>   <mrow>
>     <msup>
>       <mi>x</mi>
>       <mn>2</mn>
>     </msup>
>     <mo>+</mo>
>     <mrow>
>       <mn>4</mn>
>       <mo>&InvisibleTimes;</mo>
>       <mi>x</mi>
>     </mrow>
>     <mo>+</mo>
>     <mn>4</mn>
>   </mrow>
>   <mo>=</mo>
>   <mn>0</mn>
> </mrow>
> </math>
> ```

Yes, this would most likely be more beneficial to jury members who do have internet. I just played around with this as it would remove most of the issues in the MathJax PR and although not perfect would be relatively harmless and easier to maintain.

eldering:
Agreed with @meisterT : this is too complex to use and I think doesn't really add much value.

PR #3114, opened 2025-09-23
Restrict clarifications to teams in the contest

vmcj:
This is helpful for when:
You have a shared DOMjudge instance with 2 different contests
The contests are not open to all teams but limited by categories (or directly assigned).
When you're a jury member in one of the contests you would still see all teams (union of both contests) in the `send clarification` page.

I do wonder if this is duplicated somewhere, and if we shouldn't let a dedicated Service handle the teams. The other reasonable place is the Contest itself but as that's an Entity it doesn't have an easy way of exposing which teams are in the contest when the contest is open to all teams (as you would need to query the database).
If there would be a TeamService we can change all queries to there and make sure we always get the consistent set without having to do this logic everywhere. For example I first tried this with the `Contest->getTeams()` which would return nothing in normal cases (but I would say is reasonable without this background knowledge).

### DIFF
--- a/webapp/src/Service/DOMJudgeService.php
+++ b/webapp/src/Service/DOMJudgeService.php
@@ -1699,4 +1699,20 @@ class DOMJudgeService
             ->getQuery()
             ->getResult();
     }
+
+    /** @return Team[] */
+    public function getTeamsForContest(?Contest $contest) : array {
+        if ($contest && !$contest->isOpenToAllTeams()) {
+            $contestTeams = $contest->getTeams()->toArray();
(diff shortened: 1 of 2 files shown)


webapp/src/Service/DOMJudgeService.php
```diff
@@ -1699,4 +1699,21 @@ public function getAllowedLanguagesForContest(?Contest $contest) : array {
...
+            }
+            return $contestTeams;
+        }
+        return $this->em->createQueryBuilder()
+            ->select('l')
+            ->from(Team::class, 'l')
+            ->getQuery()
+            ->getResult();
+    }
+
```
```suggestion
```

Issue #3032, opened 2025-07-06
Clarification page breaks for removed problems


ntm97:
### Description of the problem
Someone accidentally added a wrong problem to the GCPC extended contest.
A team noticed it and sent a clarification request, upon which the problem was deleted from the contest and replaced with the correct one.
However, now the clarifications page (`/jury/clarifications`) returns 500 Internal Server Error:

```
An exception has been thrown during the rendering of a template ("App\Twig\TwigExtension::problemBadge(): Argument #1 ($problem) must be of type App\Entity\ContestProblem, null given, called in /home/domjudge/muensterjudge/webapp/var/cache/prod/twig/9d/9d57c54119ebe7ee7c0a288abc88b039.php on line 210").
```

### Your environment
- DOMjudge version: `b32fc5c8c5a84160e6e26203228fbbe1ec8444e9` with `028995f9c00e7897ec863283986ef995661e38b9` cherry-picked on top of that
- Operating system / Linux distribution and version: Debian GNU/Linux 12 (bookworm), 6.1.0-27-amd64
- Webserver: nginx/1.22.1

### Steps to reproduce
1. Request clarification for a specific problem
1. Remove that problem from the contest
1. Clarification page fails to render

### Expected behaviour
The clarification page should still be able to display clarifications for all existing problems.
Not sure what should happen with the clarifications related to the removed problem.
Maybe they could be displayed with an empty subject or maybe they should just be deleted when the problem is removed.

### Actual behaviour
The clarification page fails to render, making it impossible for the jury to see and react to new clarification requests.

### Any other information that you want to share?

ntm97:
I applied this as a workaround for now:

```diff
diff --git a/webapp/templates/jury/partials/clarification_list.html.twig b/webapp/templates/jury/partials/clarification_list.html.twig
index 33ffcbcf3..2d7f6c667 100644
--- a/webapp/templates/jury/partials/clarification_list.html.twig
+++ b/webapp/templates/jury/partials/clarification_list.html.twig
@@ -71,7 +71,11 @@
 
             <td><a href="{{ link }}">
                     {%- if clarification.problem -%}
-                        problem {{ clarification.contestProblem | problemBadge -}}
+                        {%- if clarification.contestProblem -%}
+                            problem {{ clarification.contestProblem | problemBadge -}}
+                        {%- else -%}
+                            removed problem {{ clarification.problem.probid }}
+                        {%- endif -%}
                     {%- elseif clarification.category -%}
[...]

PR #3117, opened 2025-09-24
Fix clarifications for unlinked problems

vmcj:
Closes: https://github.com/DOMjudge/domjudge/issues/3032

I'm not 100% sure if we even do the right thing. If someone removes a problem from a contest would they expect the clarifications to stay. 

This does remove some problems from the API which we keep in the UI, if we want to fix that we would need to remove the `problem` key from the clarification and do this check in code. I think that's worse as it's very hard to forget and that would create referential integrity issues.

### DIFF
--- a/webapp/src/Controller/API/ClarificationController.php
+++ b/webapp/src/Controller/API/ClarificationController.php
@@ -277,8 +277,10 @@ class ClarificationController extends AbstractRestController
             ->leftJoin('clar.sender', 's')
             ->leftJoin('clar.recipient', 'r')
             ->leftJoin('clar.problem', 'p')
+            ->innerJoin('c.problems', 'cp')
             ->select('clar, c, r, reply, p')
             ->andWhere('clar.contest = :cid')
+            ->andWhere('clar.problem IS NULL OR clar.problem = cp.problem')
             ->setParameter('cid', $this->getContestId($request))
             ->orderBy('clar.clarid');
 


meisterT:
I think I would prefer to delete clarifications that are associated with problems when the corresponding problem gets deleted.

Kevinjil:
> I think I would prefer to delete clarifications that are associated with problems when the corresponding problem gets deleted.

I was thinking back about the discussion we had in Baku about teams possibly sharing other info in the clars (which could be exposed by quoting) of a problem and for that reason I think it makes sense to keep them.

meisterT:
You mean write about two different problems or select the incorrect one?

Kevinjil:
I mostly meant writing about stuff that relates to different / more problems and/or is slightly derailed of the thread about the problem and contains some other info.

meisterT:
I think unlinked data is confusing to teams IMO.

In case a team uses the wrong problem, the jury can change it before deleting the problem.

In case a team replies to a thread of problem A with a question about B, the jury can change it before deleting a problem. 

In case a team is asking about multiple problems in one question, and  the jury decides to reply meaningfully to both questions, they should reply twice, one for each problem and set the subject accordingly for each reply respectively.

vmcj:
> I think I would prefer to delete clarifications that are associated with problems when the corresponding problem gets deleted.

I agree, and couldn't come up with a good reasoning except there are currently no real warnings when you unlink. So pressing it by accident would remove the clarifications. I think the last one who touched that code is @tuupke he might know the reasoning.

Kevinjil:
> I think unlinked data is confusing to teams IMO.

@meisterT I can live with that reasoning ;). The warning suggested by @vmcj would be nice-to-have in that case.

vmcj:
I've opened an issue as reminder to discuss this (well... it's more a reminder to myself to implement what we need).

So this can be merged.

meisterT (approved):
Approved.

I think after the next release, we should revert the change and replace it with changing the cascade to `DELETE` here after thinking through all the consequences: https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Entity/Clarification.php#L90

Issue #2855, opened 2024-11-24
Cannot delete external judgement type


eldering:
### Description of the problem
Cannot remove external judgement types in the admin configuration UI.

### Your environment
Current `main` branch using WF Astana database dump.

### Steps to reproduce
1. Go to http://localhost/domjudge/jury/config?diffs=%5B%5D#Externalsystems
2. Add an external judgement type
3. See that it exists at http://localhost/domjudge/api/contests/wf48_finals/judgement-types
4. Try to delete it again.

### Expected behaviour
The judgement type is gone.

### Actual behaviour
It's still there, both in the config UI and API.

eldering:
Same issue as #2472: only the last one cannot be deleted.

Issue #2472, opened 2024-04-17
Cannot delete last clarification queue 


eldering:
### Description of the problem

When you try to delete on the configuration page the last clarification queue(s), then that fails with the message "No changes made, no changes saved."

Deleting any number of clarification queue(s) works fine as long as there is at least one left after deleting.

### Your environment
Current `main` branch.

nickygerritsen:
This would (also) be easier with a proper Symfony form.

PR #3118, opened 2025-09-25
Fix storing empty config arrays

vmcj:
Closes: https://github.com/DOMjudge/domjudge/issues/2855
Closes: https://github.com/DOMjudge/domjudge/issues/2472

@nickygerritsen do you want to keep the last one open as you wanted to use Symfony Forms there?


nickygerritsen (approved):

nickygerritsen:
No let's close it. We can do it later

PR #3119, opened 2025-09-25
Only delete problem zip after detecting problem directory in import-contest

eldering:
Otherwise, if you only have the zip, it gets deleted and then the script fails, leaving you without the data of that problem. In my case, I didn't notice, and ran `import-contest` again after unzipping all problem archives, which then failed because the first problem was missing.


vmcj:
In case we don't have the problem directory but do have the zip I think we can just upload that one?

PR #3120, opened 2025-09-25
Bump to debootstrap for latest Debian stable (Trixie)

eldering:


eldering:
Also pushed this to the `8.3` branch assuming no objections.

nickygerritsen (approved):

PR #3121, opened 2025-09-25
Store failing page for Crawler test

vmcj:
As the E2E/Crawler is very generic it's hard to pinpoint what page is failing and why. We stored the failing page(s) with GitLab CI but this was lost after the migration.

### DIFF
--- a/webapp/tests/E2E/Controller/ControllerRolesTraversalTest.php
+++ b/webapp/tests/E2E/Controller/ControllerRolesTraversalTest.php
@@ -227,10 +227,12 @@ class ControllerRolesTraversalTest extends BaseTestCase
             try {
                 self::assertSelectorExists('a#navbarDropdownContests:contains("no contest")', "Failed at: " . $url);
             } catch (\Exception $e) {
-                $gitlabArtifacts = getenv('GITLABARTIFACTS');
-                if ($gitlabArtifacts != '') {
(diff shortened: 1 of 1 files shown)


webapp/tests/E2E/Controller/ControllerRolesTraversalTest.php
```diff
@@ -227,10 +227,12 @@ public function visitWithNoContest(string $url, bool $dropdown): void
...
                 self::assertSelectorExists('a#navbarDropdownContests:contains("no contest")', "Failed at: " . $url);
             } catch (\Exception $e) {
-                $gitlabArtifacts = getenv('GITLABARTIFACTS');
-                if ($gitlabArtifacts != '') {
-                    $fileHandler = fopen(sprintf("%s/%s", $gitlabArtifacts, str_replace('/', '_s_', $url)), 'w');
-                    fwrite($fileHandler, $response->getContent());
+                if (getenv('CI')) { // We're running in GHA.
+                    $ciArtifacts = getenv('ARTIFACTS');
+                    if ($ciArtifacts != '') {
+                        $fileHandler = fopen(sprintf("%s/%s", $ciArtifacts, str_replace('/', '_s_', $url)), 'w');
```
Kevinjil, line 233: I'm not sure if this is a perfect sanitizer, but at first glance it sounds safe?

vmcj: It worked for years and it's only for debugging in CI. In case it breaks we can easily fix it with the new insights.

PR #3123, opened 2025-09-27
Do not expose submission data for hidden teams on public scoreboard

nickygerritsen:


nickygerritsen:
We should add a test for this at some point probably.

meisterT (approved):

PR #3124, opened 2025-09-27
Version control javascript libraries with NPM

vmcj:
NPM is another packagemanager for JS. So we use composer first and when we can't find the package there we check with npm.

This change does require more changes for maintainer-install & any checkout.

If we do this it makes introducing MathJax a lot easier and we can now also version control other javascript libraries.


nickygerritsen:
If we do this, I'd suggest going for npm (or yarn) fully for JS dependencies

vmcj:
> If we do this, I'd suggest going for npm (or yarn) fully for JS dependencies

So also remove JS dependencies in `composer` and add them in `npm`? I'm fine with either of the 2 but for adaptation doing it in `composer` is an easier learningpath.

I somehow never used `yarn` but if that's the better of the 2 but I have no experience with that translation path.

I also think we should only do this in the next release so we have some time to discuss.

nickygerritsen:
> > If we do this, I'd suggest going for npm (or yarn) fully for JS dependencies
> 
> So also remove JS dependencies in `composer` and add them in `npm`? I'm fine with either of the 2 but for adaptation doing it in `composer` is an easier learningpath.
> 
> I somehow never used `yarn` but if that's the better of the 2 but I have no experience with that translation path.
> 
> I also think we should only do this in the next release so we have some time to discuss.

I think it doesn't make sense to have 2 package managers for frontend dependencies, that's why I suggest using npm (or yarn, which can read the same package.json file) for all.

vmcj:
The other packages are harder to either find (dataTables) or don't work 1-on-1 (font-awesome). I'll look at those in another PR. Reviewing is easiest commit by commit and by just skipping the monaco one as that's the one touching most files.

vmcj:
> We need to update the docs for yarn as well
I think I already mention yarn in the docs? So what is missing?

> and later update the Docker images.
Yes, and ansible will also need it.

nickygerritsen:
> > We need to update the docs for yarn as well
> > I think I already mention yarn in the docs? So what is missing?
> 
> > and later update the Docker images.
> > Yes, and ansible will also need it.

sorry i missed it. it's missing for fedora though

doc/manual/develop.rst
```diff
@@ -41,7 +41,7 @@ already listed under
 
   sudo apt install autoconf automake bats \
     python3-sphinx python3-sphinx-rtd-theme fontconfig python3-yaml \
-    latexmk texlive-latex-recommended texlive-latex-extra tex-gyre
+    latexmk yarnpkg texlive-latex-recommended texlive-latex-extra tex-gyre
 
 For Fedora use::
```
vmcj, line 46:
```suggestion
For Fedora use::

  sudo dnf install git autoconf automake bats \
    sphinx-build python3-sphinx_rtd_theme latexmk yarnpkg texlive-cmap \
    texlive-metafont texlive-tex-gyre texlive-fncychap texlive-wrapfig \
    texlive-capt-of texlive-framed texlive-upquote texlive-needspace \
    texlive-tabulary texlive-parskip texlive-oberdiek texlive-makeindex \
    texlive-ellipse texlive-pict2e texlive-collection-fontsextra
    ```

Comment (vmcj) [reply doc/manual/develop.rst:46]: Did not work as good as I wanted... will fix later.

File: webapp/Makefile [hunk 2]
```diff
@@ -12,15 +12,15 @@ SUBDIRS = config
 domserver: composer-dump-autoload
 
 # Install PHP dependencies
-composer-dependencies:
+composer-dependencies: node-dependencies
```
eldering: This looks weird. Why not make it a single target `dependencies` that does both or a phony target that depends on both. Now we have a weird ordering from composer to node. Unless there's a specific reason for this?

vmcj: I've added the extra dependency

webapp/Makefile
```diff
@@ -11,6 +11,8 @@ SUBDIRS = config
 
 domserver: composer-dump-autoload
 
+dependencies: composer-dependencies node-dependencies
```
eldering, line 14: These should (all, also `composer-dependencies-dev` below) be added as phony dependencies under `.PHONY` to make it clear that no actual target with this name gets created.

eldering: Hmmm, I see that `composer-dependencies{,-dev}` are actually recursive targets from the root Makefile, so that should then be renamed there too.

vmcj: I think I already renamed the `composer-dependencies{,-dev}` to `dependencies{,-dev}` in all files (so also the root one). I've now added the `.PHONY` targets in the `webapp/Makefile`. I think that's what you meant?

Makefile
```diff
@@ -385,5 +385,4 @@ clean-autoconf:
 .PHONY: $(addsuffix -create-dirs,domserver judgehost docs) check-root \
         $(addprefix inplace-,conf conf-common install uninstall) \
         $(addprefix maintainer-,conf install) clean-autoconf config distdocs \
-        composer-dependencies composer-dependencies-dev \
-        coverity-conf coverity-build
+        dependencies dependencies-dev coverity-conf coverity-build
```
eldering, line 388: Adding the phony targets here is not strictly necessary, as recursive targets are already marked phony in https://github.com/DOMjudge/domjudge/blob/main/Makefile.global#L94, but this doesn't hurt.

PR #3125, opened 2025-09-28
Only show diff between submissions in the same contest

vmcj:
We get this case for jury imports where the same team DOMjudge uploads in multiple contests (dress + finals) and the first "finals" submission would show a diff with the last jury submission of the dress contest.

We could also have this if a team is in 2 contests at the same time but although in theory possible this shouldn't happen in practice.


nickygerritsen (approved):

meisterT:
> We get this case for jury imports where the same team DOMjudge uploads in multiple contests (dress + finals) and the first "finals" submission would show a diff with the last jury submission of the dress contest.

Huh, did you test with the same problem in both contests?

vmcj:
> > We get this case for jury imports where the same team DOMjudge uploads in multiple contests (dress + finals) and the first "finals" submission would show a diff with the last jury submission of the dress contest.
> 
> Huh, did you test with the same problem in both contests?

In my test db I have the same problem in multiple contests yes. Which makes this even more of an edge case.

meisterT:
It's a good change anyway, just wanted to understand what caused this

Issue #3115, opened 2025-09-24
Submission source missing


cubercsl:
**Description of the problem**

Related to #3031.

If a team submits the same code in two submissions, this page will display nothing.

While this rarely happens, my contest dumps show it happening quite often towards the end of the contest.

<img width="1802" height="402" alt="Image" src="https://github.com/user-attachments/assets/e3a2c973-6c92-47d8-929f-320c94ded50f" />

<img width="1802" height="402" alt="Image" src="https://github.com/user-attachments/assets/02fb5d30-3a08-48e6-861d-c0f1831cdbac" />

In demoweb, since all the code is removed, this problem can be reproduced.

meisterT:
I was hoping https://github.com/DOMjudge/domjudge/commit/3957fcd9e97a4e1d4b0271179841e82c40a01671 recently fixed it, cc @brinky

cubercsl:
> I was hoping [3957fcd](https://github.com/DOMjudge/domjudge/commit/3957fcd9e97a4e1d4b0271179841e82c40a01671) recently fixed it, cc [@brinky](https://github.com/brinky)

Yes, this commit does fix part of the issues (at least it shows the download and edit buttons). I can only view the source code by clicking the edit or download buttons, but it still doesn't make sense.

Similarly, if jury edit and resubmit the same code (i.e. no diff with original submission), will produce similar results.

![Image](https://github.com/user-attachments/assets/e3d2a6f4-7f77-418e-8aad-9f1258a76339)

Kevinjil:
I would say this is not a bug, but currently intended behavior: we are on a page to show the diff and there is none. I came across it when working on #3058, and have some ideas to improve the UX/UI.

cubercsl:
#3126 may fix this.

PR #3129, opened 2025-09-29
Always display all files (even unchanged) for a submission.

meisterT:
Fixes https://github.com/DOMjudge/domjudge/issues/3115


meisterT:
Alternative to https://github.com/DOMjudge/domjudge/pull/3126/files

meisterT:
What do you think about this then? https://github.com/DOMjudge/domjudge/pull/3130

Less `if`s.

PR #3130, opened 2025-09-29
Always display all files (even unchanged) for a submission.

meisterT:
Fixes https://github.com/DOMjudge/domjudge/issues/3115


webapp/templates/jury/partials/submission_diff.html.twig
```diff
@@ -18,31 +18,38 @@
         </tr>
     </table>
 {% endif %}
+{% set is_first_tab = true %}
```
nickygerritsen: This line is obsolete

Issue #3058, opened 2025-08-31
Double scrollbar when viewing submission source


eldering:
When viewing the source (and diff) of a submission, both the whole page and the source frame have a scrollbar. This is really annoying and we should make sure that only one of these has scrollbar by either

1. make the source frame as high as required to show the full source without a scrollbar
2. make the source frame small enough to render on the page without requiring a scrollbar

I think the first option is best as the second would break when you resize the browser window (or you'd have to dynamically scale the source code frame too.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/32f2a64f-e035-43be-9584-957773259ab8" />

eldering:
The scrollbars are not really visible in the screenshot, but were definitely there.

Kevinjil:
I pushed option 1 to https://github.com/Kevinjil/domjudge/tree/adaptive-editor-height. However, I do really not like this solution, as this is not supported by monaco at all and all the nice features do no longer work properly. At this point, if I cannot get a smooth implementation of 2, I'd prefer keeping double scroll bars.

Kevinjil:
In the source edit page, option 2 is not feasible as there are resubmission details below the source editor.

vmcj:
> In the source edit page, option 2 is not feasible as there are resubmission details below the source editor.

Can you send a screenshot? Maybe we can re-order that part and put it behind a toggle if we don't need it that often.

Kevinjil:
<img width="2558" height="604" alt="Image" src="https://github.com/user-attachments/assets/95451b7f-b1cf-437b-9a05-3e04bacc2af3" />

A bigger issue is probably the multiple editors, as in #3103:
<img width="2553" height="1266" alt="image" src="https://github.com/user-attachments/assets/2fd909fd-a9db-4ce0-96cb-35d2a5903698" />

Kevinjil:
> img
> 
> A bigger issue is probably the multiple editors, as in [#3103](https://github.com/DOMjudge/domjudge/pull/3103): img

Just dropping the idea quickly without working it out, as a reminder that at some point I will have to look into this: instead of showing 1 or more diff editors, I think it might make more sense to show the source of the submission, and optionally select a previous submission to diff against.

PR #3132, opened 2025-10-03
Switch diff viewers to single viewer

Kevinjil:
Render the diff viewers for a submission as if it was a single IDE, in
which you can select the submission to diff against. We move the button
interface out of the tabs and create functions to update the state of
individual editors.

**This PR is pushed already s.t. we can have a discussion whether or not we want this, before spending even more time polishing it.**

Main goals:
- Increase the information density for working efficiency.
- Clear as much vertical space, in preparation for #3058.
- Improve UX (although this is subjective)

Notes:
- No longer uses diff logic on the back-end: Monaco has to compute it anyway.
- Replace the very space-consuming changed files table(s) with icons in the source tabs.
- Synchronizes diff mode across all source tabs in the diff viewer.

Screen recording:
dj - Recording 2025-10-03 at 08 25 23
_dark mode is not included in the PR ;)_

TODO:
- [ ] Store and restore diff tag preference to/from local storage.
- [ ] Add tab panes for deleted source files.
- [ ] Discuss: show source code instead of diff to empty file for new/removed files?
- [ ] Handle single-file submission case with renamed file.
- [ ] Remove some now dead code from `TwigUtils`
- [ ] Polish some code
- [ ] Rebase onto `main`


meisterT:
Nice, I think this looks very promising!

>  Discuss: show source code instead of diff to empty file for new/removed files?

Indeed it looks weird, but I would still prefer the source code in the side by side view on half of the page / the correct side (left for deleted files, right for added files) if possible.

How hard is it to remember to the diff preference? For example: I personally prefer to always show a diff; if the submission has been resubmitted, then the diff to the original, otherwise the diff to the previous submission if exists.

Kevinjil:
> How hard is it to remember to the diff preference? For example: I personally prefer to always show a diff; if the submission has been resubmitted, then the diff to the original, otherwise the diff to the previous submission if exists.

@meisterT My idea is currently to use the tagged diff submission and store / restore those with similar logic to how the diff mode (`side-by-side` vs `inline`) is stored. In this PR, all selectable diff submissions have a tag (either being the `previous` or the `original` submission). So if a user selects a submission to diff against or no submission, the tag (or `no-diff`) is stored in local storage. When a submission source page is opened (same or other), the diff viewer is initialized with the last stored tag.

I also experimented a little bit with loading all previous submissions of the team for a problem, because teams sometimes might be working on 2 versions of code for a problem simultaneously and you might want to diff against the non-previous one. The UI scales perfectly with that (assuming a "normal" contests). In such a case, the "untagged" diff submissions will not be stored, but I will not include loading these in this PR and prefer to discuss such cases later.

Kevinjil:
> Indeed it looks weird, but I would still prefer the source code in the side by side view on half of the page / the correct side (left for deleted files, right for added files) if possible.

That is currently what is implemented for new files. Just have to replicate it for deleted files.

RagnarGrootKoerkamp:
Looks pretty nice! Just noticed that maybe the title isn't updated correctly after changing the file that is diffed against?

Kevinjil:
> Just noticed that maybe the title isn't updated correctly after changing the file that is diffed against?

@RagnarGrootKoerkamp Good catch! I've added it to the TODO list. Would you consider it helpful to keep it in the title, as the same information is already below it in the selection box? The only reason I would keep the duplicate information is that I'm not sure for a nice UX s.t. we can link to the diffed submission without the title.

RagnarGrootKoerkamp:
I'd keep it in the title probably just for easy reference. But it probably doesn't matter much.

Kevinjil:
For reference, screenshot of file renaming:
image

Kevinjil:
The current TODO's are polishing (where deemed necessary), including the note from CodeSniffer about a double `;;`. I think that the fundamental changes and the logic itself are ready for review.

webapp/public/js/domjudge.js
```diff
@@ -138,6 +138,20 @@ function setDiffMode(value)
     localStorage.setItem('domjudge_editor_diff_mode', value);
 }
 
+function getDiffTag()
+{
+    let diffTag = localStorage.getItem('domjudge_editor_diff_tag');
+    if (diffTag === undefined) {
```
meisterT: According to https://developer.mozilla.org/en-US/docs/Web/API/Storage/getItem, this returns `null` not `undefined` on missing keys.

webapp/public/js/domjudge.js
```diff
@@ -127,7 +127,7 @@ function applyEditorTheme(theme = undefined, isExternal = false)
 function getDiffMode()
 {
     let diffMode = localStorage.getItem('domjudge_editor_diff_mode');
-    if (diffMode === undefined) {
+    if (diffMode === null) {
```
vmcj: Can we use the ?? here?

webapp/public/js/domjudge.js
```diff
@@ -138,6 +138,20 @@ function setDiffMode(value)
     localStorage.setItem('domjudge_editor_diff_mode', value);
 }
 
+function getDiffTag()
+{
+    let diffTag = localStorage.getItem('domjudge_editor_diff_tag');
+    if (diffTag === null) {
```
vmcj: And here,

webapp/public/js/domjudge.js
meisterT: Good suggestion, I keep forgetting that `??` is a thing

webapp/public/js/domjudge.js
```diff
@@ -1290,3 +1296,267 @@ function initScoreboardSubmissions() {
...
+
+    const initialTag = getDiffTag();
+    const select = wrapper.find(".diff-select");
+    for (let i = 0; i < select[0].options.length; i++) {
+        if (select[0].options[i].dataset.tag == initialTag) {
+            select[0].selectedIndex = i;
+            break;
+        }
+    }
+    // Fall back to other tagged diff if preferred tag is not available for this submission.
```
vmcj, line 1323: Can you extend the comment to explain when this is not available?

webapp/public/js/domjudge.js
```diff
@@ -1290,3 +1296,267 @@ function initScoreboardSubmissions() {
...
+        });
+    };
+    editors[editorId].onDiffModeChange(updateMode);
+
+    const renamedFrom = (oldName) => {
+        let renamed = navItem.querySelector('.renamed');
+        let arrow = navItem.querySelector('.fa-arrow-right');
+        if (oldName === undefined) {
+            if (renamed) {
+                navItem.removeChild(renamed);
```
vmcj: can you refactor `renamed` to `renamedTag` (or something better) I read it as a boolean first and wanted to comment about the state here.

webapp/public/js/domjudge.js
```diff
@@ -1290,3 +1296,267 @@ function initScoreboardSubmissions() {
...
+        if (!(submitId in models)) {
+            setIcon('file-circle-plus');
+        } else if (lineChanges !== null && lineChanges.length > 0) {
+            setIcon('file-circle-exclamation');
+        } else {
+            setIcon('file-circle-check');
+        }
+    }
+    diffEditor.onDidUpdateDiff(updateIcon);
+}
```
vmcj: EOF missing.

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -900,30 +886,70 @@ public function sourceAction(
...
-            ->orderBy('file.ranknumber')
+            ->andWhere('file.submission in (:submissions)')
+            ->setParameter('submissions', $files_query)
+            ->orderBy('file.submission, file.ranknumber')
             ->getQuery()
             ->getResult();
 
-        $oldFileStats      = $oldFiles !== null ? $this->determineFileChanged($files, $oldFiles) : [];
-        $originalFileStats = $originalFiles !== null ? $this->determineFileChanged($files, $originalFiles) : [];
+        /** @var array<string, array<int, array{
```
vmcj, line 908: We don't care for the `array<int` part right? I think we only make it explicit if we set that key explicit.

I prefer to have those annotations as short as possible to keep them easier to read.

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -900,30 +886,70 @@ public function sourceAction(
...
+            $submitId = $f->getSubmission()->getSubmitid();
+            $files[$f->getFilename()] ??= [];
+            $files[$f->getFilename()][$submitId] = [
+                'rank' => $f->getRank(),
+                'filename' => $f->getFilename(),
+                'source'   => mb_check_encoding($f->getSourcecode(), 'UTF-8') ? $f->getSourcecode() : "Could not display binary file",
+            ];
+
+            // Keep track of the single filename within a submission for handling renaming.
+            $renames[$submitId] = array_key_exists($submitId, $renames) ? false : $f->getFilename();
```
vmcj, line 927: So here it would almost make sense to set the $renames as:
```@var array<int, (string|false)>```

because the id is important,

File: webapp/src/Twig/TwigExtension.php [hunk 9]
```diff
@@ -889,7 +892,7 @@ public function codeEditor(
 HTML;
         $rank   = $index;
         $id     = sprintf('editor%s', $rank);
-        $code   = htmlspecialchars($code);
+        $source = mb_check_encoding($code, 'UTF-8') ? $code : "Could not display binary file";
```
vmcj:
```suggestion
        $source = mb_check_encoding($code, 'UTF-8') ? $code : "Could not display file as UTF-8, is it binary?";
```

I wonder if anyone would ever upload something as non `UTF-8` but it would make it easier for them to see why this fails.

webapp/src/Twig/TwigExtension.php
```diff
@@ -940,86 +942,50 @@ public function getMonacoModel(SubmissionFile $file): string
...
 JS,
-            $this->twig->getRuntime(EscaperRuntime::class)->escape($file->getSourcecode(), 'js'),
+            $this->serializer->serialize($source, 'json'),
             $file->getSubmitfileid(),
             $file->getFilename(),
         );
     }
 
-    public function showDiff(string $id, SubmissionFile $newFile, SubmissionFile $oldFile): string
+    /** @param array<int, array{
```
vmcj, line 960: Same remark as before,

webapp/templates/jury/partials/submission_diff.html.twig
```diff
@@ -1,112 +1,66 @@
...
+<div class="tab-content source-tab">
+    {%- for name, name_files in files %}
+        {% set diff_id = "diff-" ~ name %}
+        <div class="tab-pane fade {{ extra_css_classes }}" id="{{ diff_id }}-tab" role="tabpanel">
+            {{ showDiff(editor_id, diff_id, submission.submitid, name, name_files) }}
         </div>
         {% set extra_css_classes = "" %}
     {%- endfor %}
 </div>
+</div>
```
vmcj, line 66: A next PR should fix the indentation IMO

webapp/templates/jury/submission_source.html.twig
```diff
@@ -25,66 +25,15 @@
...
-            <a href="{{ path('jury_submission', {submitId: originalSubmission.submitid}) }}">
-                s{{ originalSubmission.submitid }}
-            </a>
-        </h2>
-
-        {%- include 'jury/partials/submission_diff.html.twig' with {oldSubmission: originalSubmission, oldFiles: originalFiles, oldFileStats: originalFileStats} %}
-    {%- endif %}
-
+    {%- include 'jury/partials/submission_diff.html.twig' with {editor_id: editor_id} %}
 {% endblock %}
```
vmcj: EOF missing

vmcj (approved):
Some nits, probably a couple pre-existing.

webapp/src/Controller/Jury/SubmissionController.php
Kevinjil: We do care about the `array<int` part. `files` is explicitly a map `filename -> (submitid -> data)`.

webapp/src/Controller/Jury/SubmissionController.php
Kevinjil: Good point

webapp/src/Twig/TwigExtension.php
Kevinjil: Same as above: it actually is a `submitid -> data` map.

PR #3133, opened 2025-10-03
Add multi-delete option for problems.

meisterT:
Part of https://github.com/DOMjudge/domjudge/issues/229

----

If this gets merged, I can extend this to other entities.

### DIFF
--- a/webapp/templates/jury/delete_modal.html.twig
+++ b/webapp/templates/jury/delete_modal.html.twig
@@ -1,6 +1,12 @@
 {% extends "partials/modal.html.twig" %}
 
-{% block title %}Delete {{ type }} {{ primaryKey }} - "{{ description }}"{% endblock %}
+{% block title %}
+    {% if count > 1 %}
+        Delete {{ count }} {{ type }}s
+    {% else %}
+        Delete {{ type }} {{ primaryKey }} - "{{ description }}"
+    {% endif %}
+{% endblock %}
 
 {% block content %}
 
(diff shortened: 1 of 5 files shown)


vmcj:
Do you have a screenshot of how it looks?

nickygerritsen:
If we do this for more entities we should probably generalise it a bit

meisterT:
image


image

meisterT:
> If we do this for more entities we should probably generalise it a bit

Happy to generalize once we decided this is roughly how we want multi-delete to work. (I would also merge this first, and then generalize.)

meisterT:
Here is a potential generalization demonstrated for the user entity: https://github.com/DOMjudge/domjudge/commit/749efc93b627776f28a7fb6442941ba670ba8b9d

nickygerritsen:
Looks good

webapp/templates/jury/delete_modal.html.twig
```diff
@@ -9,7 +15,16 @@
             <strong>Error: {{ messages.0 }}</strong>
         </div>
     {% else %}
-        <p>You're about to delete {{ type }} {{ primaryKey }} "{{ description }}".</p>
+        {% if count > 1 %}
+            <p>You're about to delete the following {{ count }} {{ type }}s:</p>
```
eldering, line 19: Nitpick: the plural doesn't work for "category", not sure that's important enough to fix.

meisterT: indeed, I will see how to solve it when I get to team categories

PR #3135, opened 2025-10-03
Clean up team index page.

meisterT:
Changes:
- the striped tables don't provide a lot of additional value (anymore?), and I think the gray bites with table header - perhaps do the same changes later on other pages
- The bold-ing of the submissions that have not been clicked is unnecessary - there is no real additional content that we want teams to acknowledge. We probably should use the unread column in the database for notifications instead
- Remove the from/to columns for the clarification side and also remove the word "problem" in the subject column. from/to makes sense for jury interface but not so much for teams.

----

before:
image

after:
image


nickygerritsen:
I like it

nickygerritsen (approved):

vmcj:
I agree with removal of the `from`. The `to` adds info IMO. If it's worth the extra space which we can now give to the message preview? Probably not.

I think the bolding is a clear indicator for people to check this so I don't see a reason to remove it.

Removal of the `gray` makes it harder to read to me.

meisterT:
> I agree with removal of the `from`. The `to` adds info IMO.

My reasoning was:
- most clarifications that teams see are for everyone, so clarifications that are only for this team are very rare
- why would teams care whether other teams also see the response from the jury or they are the only ones? If they care they can see it after clicking on it
- I still added some indications for the ones that are for this team only in the summary column - see the first in the example image

> If it's worth the extra space which we can now give to the message preview? Probably not.

Do you mean increase the length of the message summary? We can certainly play around with that.

 
> I think the bolding is a clear indicator for people to check this so I don't see a reason to remove it.

What does it indicate? The bold currently only goes away after clicking on the submission.

I checked at finals and out of all the teams I checked (10 or so) none clicked on their submissions to remove the bold. Why would they? The modal dialog is basically only relevant if they want to see the compiler errors or (if shown) the result to samples. In all other cases, there is no additional info given.
 
> Removal of the `gray` makes it harder to read to me.

For me, it makes it easier to read without the gray, and also more modern, but we can just see what the majority thinks and decide based on that.

vmcj:
> > I agree with removal of the `from`. The `to` adds info IMO.
> 
> My reasoning was:
> 
>     * most clarifications that teams see are for everyone, so clarifications that are only for this team are very rare
> 
>     * why would teams care whether other teams also see the response from the jury or they are the only ones? If they care they can see it after clicking on it
> 
>     * I still added some indications for the ones that are for this team only in the summary column - see the first in the example image
> 
Can we add a `team only` badge there instead of what you did? This way we know the spacing consistently. I didn't see it before but that would be enough of an indication so that change is fine by me. Alternative would be `Reply to you` but not the teamname as it is now quite long.

> > If it's worth the extra space which we can now give to the message preview? Probably not.
> 
> Do you mean increase the length of the message summary? We can certainly play around with that.
> 
That's what I meant yes. If we can make more screen space we should use it to show more clarification.

> > I think the bolding is a clear indicator for people to check this so I don't see a reason to remove it.
> 
I assumed you removed the bold for clarifications? But you seem to suggest that submissions could be bold if that's the case I never used it and don't care for it so fine to remove it.

> 
> > Removal of the `gray` makes it harder to read to me.
> 
> For me, it makes it easier to read without the gray, and also more modern, but we can just see what the majority thinks and decide based on that.

Ok, let's just do it.

meisterT:
> Can we add a `team only` badge there instead of what you did? This way we know the spacing consistently. I didn't see it before but that would be enough of an indication so that change is fine by me. Alternative would be `Reply to you` but not the teamname as it is now quite long.

Something like this? It shows "Reply to you" when hovering over the reply icon: 
image
 
> > > If it's worth the extra space which we can now give to the message preview? Probably not.
> > 
> > 
> > Do you mean increase the length of the message summary? We can certainly play around with that.
> 
> That's what I meant yes. If we can make more screen space we should use it to show more clarification.

Sounds good, I will increase it from currently 80 chars to 120 chars - done in the screenshot above.

 
> > > I think the bolding is a clear indicator for people to check this so I don't see a reason to remove it.
> 
> I assumed you removed the bold for clarifications? But you seem to suggest that submissions could be bold if that's the case I never used it and don't care for it so fine to remove it.

Ah, it was a misleading screenshot that I had initially, see the one above for one with an unread clarification. I do think that this is valuable.

webapp/templates/team/partials/clarification_list.html.twig
```diff
@@ -64,7 +41,15 @@
...
                     <a data-ajax-modal data-ajax-modal-after="markSeen" href="{{ link }}">
-                        {{ clarification.summary | markdown_to_html | sanitize_html('app.clarification_sanitizer') }}
+                        {%- if clarification.sender is null and clarification.recipient is not null %}
+                            <i class="fa-solid fa-reply" title="Reply to you"></i>&nbsp;
+                        {%- endif %}
+                        <i>{{ clarification.summary
+                            | markdown_to_html
+                            | replace({'<p>': '', '</p>': ''})
+                            | sanitize_html('app.clarification_sanitizer')
+                            | raw
```
vmcj, line 57: Why do we make this raw?

vmcj (approved):

webapp/templates/team/partials/clarification_list.html.twig
meisterT: because the last filter already sanitizes it, so no double sanitation required

eldering:
What about keeping a column `to` that has value either `all` or `you`? I think that's very short and clear.

I also think that the striped grey was slightly clearer than without it.

Maybe @thijskh can have a look too?

meisterT:
> What about keeping a column `to` that has value either `all` or `you`? I think that's very short and clear.

Done, that's indeed a good space saving and clear idea:
image
 
> I also think that the striped grey was slightly clearer than without it.

Note that full rows still get highlighted if you hover over them.

meisterT:
Merging - we can always incrementally improve from here.

PR #3136, opened 2025-10-03
Give back unfinished judgings in repeated rejudgings.

### LINKED ISSUE #3069: Judging runs shown in pending state after doing a repeated rejudging

Judging runs stay showing as pending, while they should be marked as unevaluated.

### Your environment

Current `main`, actually reproduced on `wfbaku` branch.

### Steps to reproduce

1. Submit a program that fails on all test cases and have lazy evaluation enabled.
2. Rejudge it with a number of times to repeat, e.g. 5 times.
3. Let it get rejudged.
4. Observe that for all rejudgings *except* the last one, there are judging runs that show as pending, while they should just be marked as unevaluated (with a `?`) due to lazy evaluation.

Image

meisterT:
Fixes #3069.

### DIFF
--- a/webapp/src/Controller/API/JudgehostController.php
+++ b/webapp/src/Controller/API/JudgehostController.php
@@ -175,7 +175,7 @@ class JudgehostController extends AbstractFOSRestController
             ->andWhere('jh.hostname = :hostname')
             ->andWhere('j.judgingid = jt.jobid')
             ->andWhere('jr.runresult IS NULL')
-            ->andWhere('j.valid = 1 OR r.valid = 1')
+            ->andWhere('j.valid = 1 OR j.rejudging IS NOT NULL')
             ->andWhere('j.result != :compiler_error')
             ->setParameter('hostname', $hostname)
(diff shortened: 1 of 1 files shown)


meisterT:
Good question, I reworded the commit, let me know if this answers everything.

vmcj:
The third paragraph doesn't read nice in the commit.
`Previously, we did this for incorrectly assume that active rejudgings
are marked as valid`

meisterT:
Thanks, fixed.

eldering:
From the commit message now:

> judgedaemons check in after start-up and after failed judgetask, see
> https://github.com/DOMjudge/domjudge/blob/2222a867f99275cbdb8e81d4681e47522a29669f/judge/judgedaemon.main.php#L984

so I understand that the `judge` function returns true only when the submission was correct, so this should always mark these unfiinished judging runs as done? Makes sense, but I can't say I 100% follow the details in code.

But it's fine, no objections to merging.

meisterT:
It returns true when it needs more work. I will see how to add some comments in judgedaemon to make this more clear.

PR #3137, opened 2025-10-04
Add problem status to the team problem set page.

meisterT:
Example:
image

### DIFF
--- a/webapp/src/Service/DOMJudgeService.php
+++ b/webapp/src/Service/DOMJudgeService.php
@@ -20,6 +20,7 @@ use App\Entity\Language;
 use App\Entity\Problem;
 use App\Entity\ProblemAttachment;
 use App\Entity\QueueTask;
+use App\Entity\ScoreCache;
 use App\Entity\Rejudging;
 use App\Entity\Submission;
 use App\Entity\SubmissionSource;
@@ -1123,6 +1124,7 @@ class DOMJudgeService
         $problems = [];
         $samples = [];
         $clars = [];
+        $problemStatuses = [];
         if ($contest && ($forJury || $contest->getFreezeData()->started())) {
             $problems = $this->em->createQueryBuilder()
                 ->from(ContestProblem::class, 'cp')
@@ -1136,6 +1138,26 @@ class DOMJudgeService
                 ->addOrderBy('cp.shortname')
                 ->getQuery()
                 ->getResult();
+
+            if ($teamId) {
+                $scoreCache = $this->em->createQueryBuilder()
+                    ->from(ScoreCache::class, 's')
+                    ->select('s')
+                    ->andWhere('s.contest = :contest')
+                    ->andWhere('s.team = :team')
+                    ->setParameter('contest', $contest)
+                    ->setParameter('team', $this->getTeam($teamId))
+                    ->getQuery()
+                    ->getResult();
+
+                foreach ($scoreCache as $scoreRow) {
+                    $problemStatuses[$scoreRow->getProblem()->getProbid()] = [
+                        'is_correct' => $scoreRow->getIsCorrectRestricted(),
+                        'submissions' => $scoreRow->getSubmissionsRestricted(),
+                        'solvetime' => $scoreRow->getSolvetimeRestricted(),
+                    ];
+                }
+            }
 
             $samplesData = $this->em->createQueryBuilder()
                 ->from(ContestProblem::class, 'cp')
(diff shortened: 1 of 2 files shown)


nickygerritsen (approved):
I like it

vmcj (commented):
It feels out of place in that location,

I think the score doesn't add that much extra information for correct solutions, and if we remove it we could gray out (I would pick green) the problems and have the same effect?

Instead of the score we could in the future add the new rank in case the team solves the problem.

meisterT:
> It feels out of place in that location,

Happy to try out other locations, which ones should we try out?

 
> I think the score doesn't add that much extra information for correct solutions, and if we remove it we could gray out (I would pick green) the problems and have the same effect?

You mean color the background of the card, I actually tried that first and it looked horrible with the green submit button and the green-ish boxes.

 
> Instead of the score we could in the future add the new rank in case the team solves the problem.

I don't like it so much because it requires (A) explanation to the team, and (B) is largely independent of the problem to solve (modulo penalty time).

meisterT:
> > I think the score doesn't add that much extra information for correct solutions, and if we remove it we could gray out (I would pick green) the problems and have the same effect?
> 
> You mean color the background of the card, I actually tried that first and it looked horrible with the green submit button and the green-ish boxes.

See this screenshot for one example shade that I tried, other's didn't work better in my experience:
image

eldering (approved):

meisterT:
> > It feels out of place in that location,
> 
> Happy to try out other locations, which ones should we try out?

Or something like this?
image

Kevinjil:
I like this card the most:
image
Perhaps we can align the icon with the problem badge?

meisterT:
I find the green background with the other greens in the box quite ugly :-)

I can see how easy it is to align things

meisterT:
@Kevinjil one way to solve the alignment problem would be moving the problem badge to the left: 
image

What are everyone's thoughts on the options?

eldering:
> one way to solve the alignment problem would be moving the problem badge to the left:
> What are everyone's thoughts on the options?

I don't like it so much: this way the problem badge seems less pronounced and more easily confused with a green/red (un)solved badge. And I think the problem badges should be the most obvious discernable thing here.

vmcj:
I didn't check how it looked but the reason a team visits this page after the contest start is to pick a next problem. Could we split the page in 2?
Unsolved problems:
[Card C like you have it initially] [Card E] [Card G]
[Card H]
Solved problems:
[Card B] [Card C] [Card D]
[Card F] [Card I]

I suspect that the new badge is less out of place in that case (for me).

nickygerritsen:
I think people also use the page to download samples or testing tools, if you don't provide them.

I'm not entirely sure if not having the problems in contest order is the way to go.

If people indeed visit this page mainly to pick the next problem, then we should sort them by number of correct attempts by other teams. I'm not suggesting we do that though.

vmcj:
> I think people also use the page to download samples or testing tools, if you don't provide them.
> 
> I'm not entirely sure if not having the problems in contest order is the way to go.
> 
> If people indeed visit this page mainly to pick the next problem, then we should sort them by number of correct attempts by other teams. I'm not suggesting we do that though.

What I propose is contest order, I just propose to do it twice. In case you re-visit the page you would want to see the problem's you haven't solved yet. Another sorting option with sorted by Solved/Tried or even #solved would be an option but what I'm suggesting is to move stuff out of view which is most likely not relevant anymore (similar for disabled languages, non-activated contests etc.).

nickygerritsen:
I do really wonder if that fixes the issue with it not looking good with Tobi's addition though. And it will move stuff around still, even though it's in order.

But I might be wrong and maybe that is exactly what people want.

meisterT:
I think it is worth a try. There is also another option that came to mind: gray out the whole card if a problem has been solved. Do not add any indication to the attempted problems. I will try out both and add screenshots. And then we should probably ask in Slack or so to vote for options :-)

eldering:
A side comment: I'd try changing to the color of the submit button from green to something a bit less outstanding: I think the color green creates a "conflict" with the green solved badge.

eldering:
And what do the two rows of boxes actually mean? The fact that I don't know makes me wonder if we could remove those and declutter the page a bit.

nickygerritsen:
> And what do the two rows of boxes actually mean? The fact that I don't know makes me wonder if we could remove those and declutter the page a bit.

Hovering over them does give you a hint:
image

I do think it's useful, but maybe we can not show it by default or something?

meisterT:
I think the boxes are super useful and if you hover over them once you understand the concept. We should document them in the team manual though.

> but maybe we can not show it by default or something

I think that's worse than dropping them. We should be careful with adding options, especially to the team interface.

PR #3138, opened 2025-10-04
Add multi-delete for users.

meisterT:
Also generalize multi-delete code where it makes sense.

More progress towards #229.

### DIFF
--- a/webapp/public/js/multi-delete.js
+++ b/webapp/public/js/multi-delete.js
@@ -1,0 +1,43 @@
+function initializeMultiDelete(options) {
+    var $deleteButton = $(options.buttonSelector);
+    var checkboxClass = options.checkboxClass;
+    var deleteUrl = options.deleteUrl;
+
+    function toggleDeleteButton() {
+        var checkedCount = $('.' + checkboxClass + ':checked').length;
+        $deleteButton.prop('disabled', checkedCount === 0);
+    }
+
+    $(document).on('change', '.' + checkboxClass, function() {
+        var table = $(this).closest('table');
+        var $tableCheckboxes = table.find('.' + checkboxClass);
+        var $selectAllInTable = table.find('.select-all');
+        $selectAllInTable.prop('checked', $tableCheckboxes.length > 0 && $tableCheckboxes.length === $tableCheckboxes.filter(':checked').length);
+        toggleDeleteButton();
+    });
+
+    $(document).on('change', '.select-all', function() {
+        var table = $(this).closest('table');
+        table.find('.' + checkboxClass).prop('checked', $(this).is(':checked'));
+        toggleDeleteButton();
+    });
+
+    toggleDeleteButton();
+
+    $deleteButton.on('click', function () {
+        var ids = $('.' + checkboxClass + ':checked').map(function () {
+            return 'ids[]=' + $(this).val();
+        }).get();
+        if (ids.length === 0) return;
+
+        var url = deleteUrl + '?' + ids.join('&');
+
+        var $tempLink = $('<a>', {
+            'href': url,
+            'data-ajax-modal': ''
+        }).hide().appendTo('body');
+
+        $tempLink.trigger('click');
+        $tempLink.remove();
+    });
+}
--- a/webapp/src/Controller/BaseController.php
+++ b/webapp/src/Controller/BaseController.php
@@ -458,6 +458,34 @@ abstract class BaseController extends AbstractController
      * @param array<string, array<array{'target': string, 'targetColumn': string, 'type': string}>> $relations
      * @return string[]
      */
+    protected function deleteMultiple(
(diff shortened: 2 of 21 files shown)


meisterT:
Now also including teams, team affiliation, and team categories.

webapp/templates/jury/delete_modal.html.twig
```diff
@@ -2,7 +2,7 @@
 
 {% block title %}
     {% if count > 1 %}
-        Delete {{ count }} {{ type }}s
+        Delete {{ count }} {% if type|slice(-1) == 'y' %}{{ type|slice(0, -1) }}ies{% else %}{{ type }}s{% endif %}
```
meisterT, line 5: @eldering this is the fix for pluralizing `team category`

webapp/src/Controller/Jury/TeamAffiliationController.php
```diff
@@ -59,6 +60,13 @@ public function indexAction(
             'name' => ['title' => 'name', 'sort' => true, 'default_sort' => true],
         ];
 
+        if ($this->isGranted('ROLE_ADMIN')) {
```
nickygerritsen: Should we maybe make a generalized method of this in the base class? Also for the one below? So that we don't have this repeated html everywhere, if we ever want to change styling or something.

nickygerritsen (approved):

webapp/src/Controller/Jury/TeamAffiliationController.php
meisterT: Done, please check the last two commits.

webapp/public/js/multi-delete.js
```diff
@@ -0,0 +1,43 @@
+function initializeMultiDelete(options) {
```
vmcj, line 1: What is the reason to not add this to domjudge.js? If it's because this is only relevant for admins I prefer to name this domjudge-admin.js instead of extra js files.

webapp/src/Controller/BaseController.php
```diff
@@ -454,6 +454,40 @@ protected function deleteEntities(
...
+
+        if (empty($entities)) {
+            $this->addFlash('warning', $warningMessage);
+            return $this->redirectToRoute($redirectRoute);
+        }
+
+        return $this->deleteEntities($request, $entities, $this->generateUrl($redirectRoute));
+    }
+
+
```
vmcj:
```suggestion
```

vmcj:
I think we should add unit tests for this,

webapp/public/js/multi-delete.js
meisterT: For multiple reasons:
- Splitting up into multiple files makes it clear what the purpose is and keeps things readable
- Only load it on pages where it is necessary in order to not regress performance

If you have a strong opinion, I will merge it with the other files or rename it, but I personally think we should rather go in the opposite direction and split large files up.

meisterT:
> I think we should add unit tests for this,

Do we have good examples for unit tests where individual entities get deleted?

webapp/public/js/multi-delete.js
nickygerritsen: Note that big frontend javascript frameworks also put a lot in one file, but then does only activate the javascript applicable for whatever page/element you are on.

Since we don't have a JS framework for now, I think splitting is better. If we ever do something like Vue, it will be fixed then

meisterT:
@vmcj I have added a unit test for the problems entity in the last commit, PTAL.

webapp/templates/jury/problems.html.twig
```diff
@@ -40,49 +38,13 @@
 
 {% block extrafooter %}
     {{ parent() }}
+    <script src="{{ asset('js/multi-delete.js') }}"></script>
```
vmcj, line 41: Why can't we add this in the head?

meisterT: Could be in `extrahead` I assume (didn't test yet), but having it here makes it close to the `initializeMultiDelete` call, which needs to happen after the DOM has been inserted.

Is there a benefit of moving it from `extrafooter` to `extrahead`?

webapp/public/js/multi-delete.js
vmcj: Consensus seems to be to split so fine by me.

And I might have been fine with splitting, but then this should have been one PR where we split. Now its just deviating from the codestyle of other files, or introducing multiple changes in the same PR which makes reviewing/discussing harder.

webapp/templates/jury/problems.html.twig
vmcj: I prefer to see all loaded scripts in the head, so consistency.

But feel free to leave it there. I'll check in the future if I care enough about it to move it myself.

PR #3146, opened 2025-10-07
Fix codestyle errors

nickygerritsen:


webapp/src/Service/DOMJudgeService.php
```diff
@@ -1801,8 +1801,7 @@ public function checkNewVersion(): string|false {
                     }
                 }
             }
-        }
-        elseif ($this->config->get('check_new_version', false) === UpdateStrategy::Strategy_major_release) {
+        } elseif ($this->config->get('check_new_version', false) === UpdateStrategy::Strategy_major_release) {
```
```suggestion
        } elseif (
            $this->config->get('check_new_version', false) === UpdateStrategy::Strategy_major_release
        ) {
```

I do think this is only slightly better. I would keep the original line.

nickygerritsen: For me the line doesn't feel too long but if others think it does, we should come up with a way to make it shorter. I really don';t think the original, which saves 2 characters but has the `}` weirdly on its own line, makes it better.

eldering: If these enums are always written like `UpdateStrategy::Strategy_major_release`, then why does it contain "strategy" twice? Removing that would fix this case and in general produce shorter lines.

webapp/src/Entity/Contest.php
```diff
@@ -1280,7 +1280,9 @@ public function validate(ExecutionContextInterface $context): void
     {
         foreach (['Activate', 'Deactivate', 'Start', 'End', 'Freeze', 'Unfreeze'] as $timeString) {
             $tmpValue = $this->{'get' . $timeString . 'timeString'}();
-            if ($tmpValue === null) continue;
+            if ($tmpValue === null) {
+                continue;
+            }
```
eldering: I think that short statements like this are shorter and clearer as is.

webapp/src/Service/DOMJudgeService.php
nickygerritsen: I think that makes sense and is indeed beter to read

webapp/src/Utils/Utils.php
```diff
@@ -1082,10 +1082,10 @@ public static function extendMaxExecutionTime(int $minimumMaxExecutionTime): voi
     /**
      * Call ob_flush() unless the top-level output buffer does not allow it.
      */
-    public static function ob_flush_if_possible(): bool
+    public static function ob_flush_if_possible(): bool // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
```
eldering, line 1085: I mentioned this one in slack, but I see it's already annotated. I wrote it like this on purpose snice it wraps `ob_flush`.

webapp/src/Service/DOMJudgeService.php
vmcj: Alternative is to shorten UpdateStrategy to US for this inclusion. I think we shouldn't change this though.

vmcj (approved):

PR #3154, opened 2025-10-12
Simplify runpipe code including some minor stylistic improvements.

meisterT:

### DIFF
--- a/judge/runpipe.cc
+++ b/judge/runpipe.cc
@@ -42,7 +42,7 @@
 // Additionally, we set up a signal handler for SIGUSR1 which runguard will use
 // to indicate when it killed its child process due to a time limit.
 //
-// If the proxy is not enabled (i.e. no traffic capturing), the pipes are setup
+// If the proxy is not enabled (i.e. no traffic capturing), the pipes are set up
 // like this:
 //
 //   #0            #1
@@ -52,7 +52,7 @@
 // SIGUSR1 -----------------> epoll
 //
 //
-// If the proxy is enabled the pipes are setup like this:
+// If the proxy is enabled the pipes are set up like this:
 //
 //   #0            proxy           #1
 // stdout  ----->  epoll  -----> stdin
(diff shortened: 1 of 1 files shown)


judge/runpipe.cc
```diff
@@ -451,7 +449,8 @@ struct state_t {
 
     progname = argv[0];
     int opt = -1;
-    while ((opt = getopt_long(argc, argv, "+o:M:vh", long_opts, NULL)) != -1) {
+    while ((opt = getopt_long(argc, argv, "+o:M:vh", long_opts, nullptr)) !=
+           -1) {
```
eldering: Nitpick: I'd prefer to not break the line here.

judge/runpipe.cc
```diff
@@ -689,9 +663,8 @@ struct state_t {
     };
 
     for (size_t i = 0; i < processes.size(); i++) {
-      size_t j = (i + 1) % N_PROC;
-      // Setup the communication #i -> #j (optionally with an proxy in
-      // between).
+      size_t j = (i + 1) % 2;
```
eldering: Given that `processes.size() == 2` now, this could also be simplified to
```suggestion
      size_t j = 1 - i;
```

judge/runpipe.cc
```diff
@@ -844,7 +813,8 @@ struct state_t {
         int time_millis = time % 1000;
         char direction = from.index == 0 ? ']' : '[';
         char eofbuf[128];
-        sprintf(eofbuf, "[%3d.%03ds/%ld]%c", time_sec, time_millis, 0L, direction);
+        sprintf(eofbuf, "[%3d.%03ds/%ld]%c", time_sec, time_millis, 0L,
+                direction);
```
eldering: Nitpick: Is this line too long? I'd prefer not to break it.

judge/runpipe.cc
```diff
@@ -914,26 +885,25 @@ struct state_t {
...
         // A process wrote in one of the pipes to the proxy.
-        for (size_t i = 0; i < processes.size(); i++) {
-          auto &from = processes[i];
+        for (size_t j = 0; j < processes.size(); j++) {
+          auto &from = processes[j];
           if (fd != from.process_to_proxy) {
             continue;
           }
-          auto &to = processes[(i + 1) % processes.size()];
+          auto &to = processes[(j + 1) % processes.size()];
```
eldering: Maybe also `1-j`.

judge/runpipe.cc
meisterT: Too aggressive `clang-format`, reverted.

judge/runpipe.cc
meisterT: same as above, reverted

PR #3156, opened 2025-10-16
judgedaemon: Replace `system` and other command executions.

meisterT:
Introduce a safer and more readable variant that executes, logs and checks return values centrally.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -293,6 +293,41 @@ function read_judgehostlog(int $numLines = 20) : string
     return trim(ob_get_clean());
 }
 
+define('DONT_CARE', new class {});
+/**
+ * Execute a shell command given an array of strings forming the command.
+ * The command and all its arguments are automatically escaped.
+ *
+ * @param array $command_parts         The command and its arguments (e.g., ['ls', '-l', '/tmp/']).
+ * @param mixed $retval                The (integer) variable to store the command's exit code.
+ * @param bool  $log_nonzero_exitcode  Whether non-zero exit codes should be logged.
+ *
+ * @return bool Returns true on success (exit code 0), false otherwise.
+ */
+function run_command_safe(array $command_parts, &$retval = DONT_CARE, $log_nonzero_exitcode = true): bool
+{
+    if (empty($command_parts)) {
+        logmsg(LOG_WARNING, "Need at least the command that should be called.");
+        $retval = -1;
+        return false;
+    }
+
+    $command = implode(' ', array_map('dj_escapeshellarg', $command_parts));
+
+    logmsg(LOG_DEBUG, "Executing command: $command");
+    system($command, $retval_local);
+    if ($retval !== DONT_CARE) $retval = $retval_local;
+
+    if ($retval_local !== 0) {
+        if ($log_nonzero_exitcode) {
+            logmsg(LOG_WARNING, "Command failed with exit code $retval_local: $command");
+        }
+        return false;
+    }
+
+    return true;
+}
+
 // Fetches a new executable from database if not cached already, and runs build script to compile executable.
 // Returns an array with
 // - absolute path to run script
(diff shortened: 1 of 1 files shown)


judge/judgedaemon.main.php
```diff
@@ -290,6 +290,39 @@ function read_judgehostlog(int $numLines = 20) : string
...
+ * Execute a shell command given an array of strings forming the command.
+ * The command and all its arguments are automatically escaped.
+ *
+ * @param array    $command_parts         The command and its arguments (e.g., ['ls', '-l', '/tmp/']).
+ * @param int|null $retval                The variable to store the command's exit code.
+ * @param bool     $log_nonzero_exitcode  Whether non-zero exit codes should be logged.
+ *
+ * @return bool Returns true on success (exit code 0), false otherwise.
+ */
+function run_command_safe(array $command_parts, ?int &$retval = null, $log_nonzero_exitcode = true): bool
```
eldering: When I look at most call sites, then `$retval` is not used afterwards, so it would be a bit simpler to be able to not pass it and then only within this function a local `$retval` variable is used.

meisterT: How do imagine that working with the code snippet where we use the `$retval` like this?

```
 if (!isset($EXITCODES[$retval])) {
            alert('error');
            error("Unknown exitcode ($retval) from testcase_run.sh for s$judgeTask[submitid]");
        }
        $result = $EXITCODES[$retval];
```

judge/judgedaemon.main.php
```diff
@@ -290,6 +290,39 @@ function read_judgehostlog(int $numLines = 20) : string
...
+    if (empty($command_parts)) {
+        logmsg(LOG_WARNING, "Need at least the command that should be called.");
+        $retval = -1;
+        return false;
+    }
+
+    $command = implode(' ', array_map('dj_escapeshellarg', $command_parts));
+
+    logmsg(LOG_DEBUG, "Executing command: $command");
+    system($command, $retval);
```
eldering:
```suggestion
    system($command, $retval_local);
    if ( $retval!==null ) $retval = $retval_local;
```

judge/judgedaemon.main.php
eldering: Inside `run_command_safe` I'd use a separate variable `$retval_local` and then assign that to `$retval` if it is passed, see below. Then you'd be able to call it as
```php
run_command_safe(['echo', 'foobar']);
```
when you don't care about the exit code, right?

judge/judgedaemon.main.php
meisterT: This doesn't work directly because we cannot distinguish the different cases, replaced with a sentinel

judge/judgedaemon.main.php
meisterT: Got rid of `$retval` in the function call in more places, let me know if there is one more that I missed

judge/judgedaemon.main.php
```diff
@@ -293,6 +293,41 @@ function read_judgehostlog(int $numLines = 20) : string
...
+ * Execute a shell command given an array of strings forming the command.
+ * The command and all its arguments are automatically escaped.
+ *
+ * @param array $command_parts         The command and its arguments (e.g., ['ls', '-l', '/tmp/']).
+ * @param mixed $retval                The (integer) variable to store the command's exit code.
+ * @param bool  $log_nonzero_exitcode  Whether non-zero exit codes should be logged.
+ *
+ * @return bool Returns true on success (exit code 0), false otherwise.
+ */
+function run_command_safe(array $command_parts, &$retval = DONT_CARE, $log_nonzero_exitcode = true): bool
```
nickygerritsen, line 307: Why not
```suggestion
function run_command_safe(array $command_parts, ?int &$retval = null, $log_nonzero_exitcode = true): bool
```

and just always assing to retval? What's the benefit of the don't care thingy?

meisterT: This is what I had before but then you cannot distinguish between not passing anything or passing a "freshly created" variable anymore. See also https://github.com/DOMjudge/domjudge/pull/3156#discussion_r2463869568

nickygerritsen: I don't think I understand: why do we want to distinguish? But if you thought about it, I'm fine with it of course.

meisterT: I thought it did break for compile errors, but when I just tried again to verify, it doesn't break. Probably I was holding it wrong before.

judge/judgedaemon.main.php
eldering: I didn't realize you couldn't just use `null`, but if this works, LGTM!

Issue #3148, opened 2025-10-11
Possible Scoreboard Refresh Memory Leak


gozzarda:
### Description of the problem
Apparent memory leak observed when Jury scoreboard left running.
Mayb be a factor that this contest has 300+ teams and is multiple weeks long.
Seen as high as 10.8 GB memory usage and a whole CPU core.

### Your environment
- DOMjudge v8.3.1
- Pop!_OS 24.04
- Chromium 138.0.7204.157

### Steps to reproduce
1. Set up a new contest
  - In my case, there are several contests, all weeks long, all with the same hundreds of teams, in case this is relevant.
2. Log in as a Jury user and open the scoreboard
3. Leave the scoreboard open for a significant period of time (overnight, for me)
  - I do not know if users submitting during this time is necessary to trigger the issue, but I don't think so.
4. Observe high CPU and memory usage.

### Expected behaviour
Reasonable memory usage.

### Actual behaviour
Unreasonable memory usage (10+ GB)

### Any other information that you want to share?
I had a little time to do some very quick profiling, and there seems to be a periodic (when the scoreboard refreshes?) leak of a bunch of `<tr>`s from the scoreboard. As observed on a fresh page it was a few kB of detached elements directly but about 5 MB indirect (I think, I forget the exact numbers, sorry).
It would take about 2000 such leaks to account for 10+ GB, which is about 16-18 hours with a scoreboard refresh every 30 seconds.

meisterT:
Thanks for reporting! I have a potential fix in https://github.com/DOMjudge/domjudge/pull/3157 - can you try it out in your setup, whether it helps?

PR #3157, opened 2025-10-17
Attempt to fix memory leak in AJAX refresh.

meisterT:
As reported in https://github.com/DOMjudge/domjudge/issues/3148

### DIFF
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -588,7 +588,7 @@ function processAjaxResponse(jqXHR, data) {
         if ($refreshTarget.data('ajax-refresh-before')) {
             window[$refreshTarget.data('ajax-refresh-before')]();
         }
-        $refreshTarget.html($data);
+        $refreshTarget.empty().html($data);
         if ($refreshTarget.data('ajax-refresh-after')) {
             window[$refreshTarget.data('ajax-refresh-after')]();
         }


nickygerritsen:
Haven't tested if it actually fixes the issue but I don't disagree with the code. Would be nice to wait for @gozzarda

meisterT:
@gozzarda, did you have time to test whether this helps?

gozzarda:
> @gozzarda, did you have time to test whether this helps?

No, I've not, and am unlikely to soon, sorry. The reason I encountered this is that I (once a year) run a long-running "contest" for a class of hundreds of students. I did not have time to deploy this patch while it was running this year, and if I run another similar one it won't be until next year.
I did try to test this hackily by live-patching it client-side, but my couple of attempts were foiled by my session then expiring. This leads to an extra reproduction step, I think, though I didn't get to test it: You need to be occasionally doing things in other tabs to stop your session from expiring while keeping the scoreboard open for a long period, maybe?

If I get an opportunity to test it I will check in, but I can't guarantee if or when that will be.

Kevinjil:
I'm not sure why this would solve the memory leak. The `.html(...)` replaces all children with the provided data. This effectively includes the behavior of the `.empty()` call.

meisterT:
I initially thought empty does more, see https://api.jquery.com/empty and search for memory leak, but html does have the same effect indeed.

Closing this PR, we need to think again, and perhaps repro

PR #3159, opened 2025-10-24
Sync database schema with entities

nickygerritsen:
This is extracted from #3043 (and the submission source change is added).

### DIFF
--- a/webapp/migrations/Version20250620082406.php
+++ b/webapp/migrations/Version20250620082406.php
@@ -1,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20250620082406 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Change comments to reflect entities';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE judging CHANGE max_runtime_for_verdict max_runtime_for_verdict NUMERIC(32, 9) UNSIGNED DEFAULT NULL COMMENT \'The maximum runtime for all runs that resulted in the verdict\'');
+        $this->addSql('ALTER TABLE problem CHANGE types types INT NOT NULL COMMENT \'Bitmask of problem types, default is pass-fail.\'');
+    }
+
+    public function down(Schema $schema): void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE problem CHANGE types types INT NOT NULL COMMENT \'Bitset of problem types, default is pass-fail.\'');
+        $this->addSql('ALTER TABLE judging CHANGE max_runtime_for_verdict max_runtime_for_verdict NUMERIC(32, 9) UNSIGNED DEFAULT NULL COMMENT \'The maximum run time for all runs that resulted in the verdict\'');
+    }
+
+    public function isTransactional(): bool
+    {
+        return false;
+    }
+}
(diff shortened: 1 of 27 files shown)


webapp/src/Entity/RankCache.php
```diff
@@ -19,8 +19,8 @@
 #[ORM\Index(columns: ['cid', 'points_public', 'totaltime_public', 'totalruntime_public'], name: 'order_public')]
 #[ORM\Index(columns: ['cid'], name: 'cid')]
 #[ORM\Index(columns: ['teamid'], name: 'teamid')]
-#[ORM\Index(columns: ['sort_key_public'], name: 'sortKeyPublic')]
-#[ORM\Index(columns: ['sort_key_restricted'], name: 'sortKeyRestricted')]
+#[ORM\Index(columns: ['sort_key_public'], name: 'sortKeyPublic', options: ['lengths' => [768]])]
```
meisterT: where does 768 come from?

nickygerritsen: We have this in more places. I believe this is a mysql limit on keys

meisterT (approved):

webapp/migrations/Version20250620082406.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    public function getDescription(): string
+    {
+        return 'Change comments to reflect entities';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE judging CHANGE max_runtime_for_verdict max_runtime_for_verdict NUMERIC(32, 9) UNSIGNED DEFAULT NULL COMMENT \'The maximum runtime for all runs that resulted in the verdict\'');
```
vmcj, line 23: It took me a while to spot the difference.. If we're already doing this and touching the database, should it also be `max_run_time_for_verdict` as now the comment and column name imply to different things. Because now we imply the largest environment with the column and the longest run for the judging. I think it's important to make sure the columns imply the right thing in case someone ever wants to rate on the smallest resulting binary for example or more serious, if we change this let's be consistent.

webapp/src/Entity/ExternalSourceWarning.php
```diff
@@ -81,12 +81,12 @@ public function setLastEventId(?string $lastEventId): ExternalSourceWarning
         return $this;
     }
 
-    public function getLastTime(): float
+    public function getLastTime(): string|float
```
vmcj: Why can't we accept a string and convert it to a float (or the other way around?) This feels like something where we would need to repeat the code to handle conversion with always the same implementation?

webapp/src/Entity/RankCache.php
vmcj: Can 
```suggestion
PHP_CONSTANT MYSQL_KEY_LIMIT 768
#[ORM\Index(columns: ['sort_key_public'], name: 'sortKeyPublic', options: ['lengths' => [MYSQL_KEY_LIMIT]])]
```

I suspect this code doesn't work but I assume you get what I'm suggesting.

webapp/migrations/Version20250620082406.php
nickygerritsen: Are you suggesting adding a `_`? Since we call it runtime not run time everywhere, so I'm not sure that's better.

webapp/src/Entity/RankCache.php
nickygerritsen: Good idea, done.

webapp/src/Entity/ExternalSourceWarning.php
nickygerritsen: I made them all `?float` now and PHPUnit is still happy. I like it.

webapp/migrations/Version20250620082406.php
vmcj: Well why do you change the comment? I assumed to fix one of those mistakes.

nickygerritsen: The comment changed from `run time` to `runtime`, to be in sync with what is in the entity file.

eldering (approved):

PR #3160, opened 2025-10-24
Pin a VNU release instead of picking latest.

meisterT:
Looks like they had a release today and that might be the reason for our test failures.

### DIFF
--- a/.github/jobs/webstandard.sh
+++ b/.github/jobs/webstandard.sh
@@ -117,7 +117,7 @@ if [ "$TEST" = "w3cval" ]; then
 
     section_start "Install testsuite"
     cd "$DIR"
-    wget https://github.com/validator/validator/releases/latest/download/vnu.linux.zip
+    wget https://github.com/validator/validator/releases/download/20.6.30/vnu.linux.zip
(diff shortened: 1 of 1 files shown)


vmcj:
I rather not pin as they seemed to not release often so it's more of a hassle to check manually

meisterT:
IMO, pinning versions is best practice to avoid errors like this one.

But I will not fight for it if you dislike pinning. However, let's be pragmatic and merge this PR and then separately fix whatever changed at HEAD so that other PRs can be merged in the meantime.

vmcj:
> IMO, pinning versions is best practice to avoid errors like this one.
> 
> But I will not fight for it if you dislike pinning. However, let's be pragmatic and merge this PR and then separately fix whatever changed at HEAD so that other PRs can be merged in the meantime.

I think for components which upgrade often pinning is fine as I tend to check those often and upgrade them when needed. IRC this component does not often so I rather accept a day like today where we need to fix CI instead of having the check often with no new release. Especially for VNU I hope they finally implemented the current "accepted" spec for some elements so we can remove a couple of warnings as it tends to be very conservative.

PR #3164, opened 2025-10-25
When no mapping to CLICS role exists, map to first role

### LINKED ISSUE #3163: Account type can be null, which is not allowed by ccs specification
I got this event on BAPC 2025 contest. It contradicts [specififcation](https://ccs-specs.icpc.io/draft/contest_api#accounts), as type is not allowed to be null
```
{"token":"52184","id":"balloon","type":"accounts","data":{"last_login_time":null,"last_api_login_time":null,"first_login_time":null,"team":null,"team_id":null,"roles":["balloon"],"type":null,"userid":78,"id":"balloon","username":"balloon","name":"Balloon Account","email":null,"last_ip":null,"ip":null,"enabled":true},"time":"2025-10-25T08:43:17.992+02:00"}
```

nickygerritsen:
Fixes #3163

### DIFF
--- a/webapp/src/Entity/User.php
+++ b/webapp/src/Entity/User.php
@@ -451,7 +451,7 @@ class User extends BaseApiEntity implements
             }
         }
 
-        return null;
+        return $this->getRoleList()[0];
     }
 
     /**
--- a/webapp/tests/Unit/Controller/API/AccountBaseTestCase.php
+++ b/webapp/tests/Unit/Controller/API/AccountBaseTestCase.php
@@ -26,8 +26,8 @@ abstract class AccountBaseTestCase extends BaseTestCase
         [['type' => 'api_writer'], ['type' => 'admin', 'roles' => ['api_writer']]],
         [['type' => 'api_reader'], ['type' => 'admin', 'roles' => ['api_reader']]],
         [['type' => 'api_source_reader'], ['type' => 'judge', 'roles' => ['api_source_reader']]],
-        [['type' => 'balloon'], ['roles' => ['balloon'], 'type' => null]],
-        [['type' => 'clarification_rw'], ['roles' => ['clarification_rw'], 'type' => null]],
+        [['type' => 'balloon'], ['roles' => ['balloon']]],
+        [['type' => 'clarification_rw'], ['roles' => ['clarification_rw']]],
         [['type' => 'cds'], ['roles' => ['api_source_reader', 'api_reader', 'api_writer'], 'type' => 'admin']],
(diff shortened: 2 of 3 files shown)


webapp/src/Entity/User.php
```diff
@@ -451,7 +451,7 @@ public function getType(): ?string
             }
         }
 
-        return null;
+        return $this->getRoleList()[0];
```
meisterT, line 454: Is this list guaranteed to be non empty?

webapp/tests/Unit/Controller/API/AccountBaseTestCase.php
```diff
@@ -26,16 +26,16 @@ abstract class AccountBaseTestCase extends BaseTestCase
...
+        [['type' => 'clarification_rw'], ['roles' => ['clarification_rw']]],
         [['type' => 'cds'], ['roles' => ['api_source_reader', 'api_reader', 'api_writer'], 'type' => 'admin']],
         [['username' => 'cds', 'type' => 'admin'], ['roles' => ['api_source_reader', 'api_reader', 'api_writer'], 'type' => 'admin']],
         [['username' => 'cds', 'type' => 'jury'], ['roles' => ['api_source_reader', 'api_reader', 'api_writer'], 'type' => 'admin']],
     ];
 
     protected static array $accountAddCombinationsWithoutFile = [
         [['username' => 'cds', 'roles' => ['admin']], ['roles' => ['api_source_reader', 'api_reader', 'api_writer'], 'type' => 'admin']],
-        [['username' => 'another_judgehost', 'roles' => ['judgehost']], ['type' => null]],
+        [['username' => 'another_judgehost', 'roles' => ['judgehost']], ['type' => 'judgehost']],
```
vmcj, line 38:
```suggestion
        [['username' => 'another_judgehost', 'roles' => ['judgehost']], ['type' => 'dj_judgehost']],
```

webapp/tests/Unit/Controller/API/AccountControllerTest.php
```diff
@@ -22,7 +22,7 @@ class AccountControllerTest extends AccountBaseTestCase
             "id"       => "judgehost",
             "username" => "judgehost",
             "team_id"  => null,
-            "type"     => null,
+            "type"     => "judgehost",
```
vmcj, line 25:
```suggestion
            "type"     => "dj_judgehost",
```

This is so specific to DOMjudge that prefixing is better in case it ever ends up in another system...

vmcj (commented):
Should we expose those accounts even?

nickygerritsen:
Closing in favor of #3165

PR #3165, opened 2025-10-25
Return 'other' if no standard account type is found in API

eldering:
Also rename variables and comment to clarify that the listed types are not the only ones allowed, just the ones suggested/ mentioned in the CLICS specification (as examples).

Closes: #3163

### DIFF
--- a/webapp/src/Entity/User.php
+++ b/webapp/src/Entity/User.php
@@ -434,24 +434,24 @@ class User extends BaseApiEntity implements
     #[Serializer\VirtualProperty]
     #[Serializer\SerializedName('type')]
     #[Serializer\Type('string')]
-    public function getType(): ?string
+    public function getType(): string
     {
-        // Types allowed by the CCS Specs Contest API in order of most permissions to least.
+        // Types suggested by the CCS Specs Contest API in order of most permissions to least.
         // Either key=>value where key is the DOMjudge role and value is the API account type or
         // only value, where both the DOMjudge role and API type are the same.
-        $allowedTypes = ['admin', 'api_writer' => 'admin', 'api_reader' => 'admin',
-                         'jury' => 'judge', 'api_source_reader' => 'judge',
-                         'team'];
-        foreach ($allowedTypes as $role => $allowedType) {
+        $mappedTypes = ['admin', 'api_writer' => 'admin', 'api_reader' => 'admin',
+                        'jury' => 'judge', 'api_source_reader' => 'judge',
+                        'team'];
+        foreach ($mappedTypes as $role => $mappedType) {
             if (is_numeric($role)) {
-                $role = $allowedType;
+                $role = $mappedType;
             }
             if (in_array($role, $this->getRoleList())) {
-                return $allowedType;
+                return $mappedType;
             }
         }
 
-        return null;
+        return 'other';
     }
 
     /**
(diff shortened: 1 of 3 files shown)


nickygerritsen:
I like this better than mine 😄 . We should fix the test though but you can see what I did in my PR.

webapp/src/Entity/User.php
```diff
@@ -436,22 +436,21 @@ public function getRoleList(): array
...
-        // Types allowed by the CCS Specs Contest API in order of most permissions to least.
+        // Types suggested by the CCS Specs Contest API in order of most permissions to least.
         // Either key=>value where key is the DOMjudge role and value is the API account type or
         // only value, where both the DOMjudge role and API type are the same.
-        $allowedTypes = ['admin', 'api_writer' => 'admin', 'api_reader' => 'admin',
-                         'jury' => 'judge', 'api_source_reader' => 'judge',
-                         'team'];
-        foreach ($allowedTypes as $role => $allowedType) {
+        $mappedTypes = ['admin', 'api_writer' => 'admin', 'api_reader' => 'admin',
+                        'jury' => 'judge', 'api_source_reader' => 'judge', 'team'];
```
vmcj:
```suggestion
                        'jury' => 'judge', 'api_source_reader' => 'judge',
                        'team'];
```
The newline was there with a reason, to show all things mapped to admin, all things to judge and all things to teams.

webapp/src/Entity/User.php
```diff
@@ -436,22 +436,21 @@ public function getRoleList(): array
...
+                $role = $mappedType;
             }
             if (in_array($role, $this->getRoleList())) {
-                return $allowedType;
+                return $mappedType;
             }
         }
 
-        return null;
+        return 'other';
```
vmcj, line 454:
```suggestion
        return 'domjudge_internal';
```
So other systems reading it know who to contact if they want to do anything with the account.

webapp/src/Entity/User.php
eldering: fixed

webapp/src/Entity/User.php
eldering: I'm not strongly against this, but why is this better than just `other`? If you're querying the DOMjudge API, you know the source already. Do we want to distinguish our nonstandard accounts from possibly another system's nonstandard accounts?x

eldering:
> I like this better than mine 😄 . We should fix the test though but you can see what I did in my PR.

Done

webapp/src/Entity/User.php
vmcj: The CDS might know the source, but things behind the CDS don't, if you make contest exports it's more convenient etc.

nickygerritsen: I have a slight preference for `other` but am also fine with `domjudge_internal`.

nickygerritsen (approved):

webapp/src/Entity/User.php
eldering: I also prefer `other`. I prefer a generic term that can potentially be added to the spec than something DOMjudge specific, as I don't see a case where you'd not know or easily figure out this came from DOMjudge.

webapp/src/Entity/User.php
```diff
@@ -436,22 +436,22 @@ public function getRoleList(): array
     #[Serializer\Type('string')]
     public function getType(): ?string
```
meisterT: We can remove the `?` in this line now, correct?

eldering: Yes, done.

PR #3169, opened 2025-10-26
C++-ify lib.misc.

meisterT:
Best reviewed commit by commit

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -4,7 +4,7 @@ endif
 
 include $(TOPDIR)/Makefile.global
 
-OBJECTS = lib.error$(OBJEXT) lib.misc$(OBJEXT)
+OBJECTS = $(addsuffix $(OBJEXT),lib.error lib.misc)
 
 build: $(OBJECTS)
 
--- a/lib/lib.misc.cc
+++ b/lib/lib.misc.cc
@@ -7,91 +7,87 @@
 
 #include "config.h"
 
-#include <stdlib.h>
+#include <cstdlib>
 #include <unistd.h>
-#include <string.h>
-#include <stdarg.h>
-#include <signal.h>
+#include <cstring>
+#include <cstdarg>
+#include <csignal>
 #include <sys/wait.h>
 #include <fcntl.h>
+#include <vector>
+#include <iostream>
 
 #include "lib.misc.h"
 #include "lib.error.h"
 
 /* Array indices for input/output file descriptors as used by pipe() */
-#define PIPE_IN  1
-#define PIPE_OUT 0
+constexpr int PIPE_IN = 1;
+constexpr int PIPE_OUT = 0;
 
 const int def_stdio_fd[3] = { STDIN_FILENO, STDOUT_FILENO, STDERR_FILENO };
 
-int execute(const char *cmd, const char **args, int nargs, int stdio_fd[3], int err2out)
+int execute(const std::string& cmd, const std::vector<std::string>& args,
+            std::array<int, 3>& stdio_fd, bool err2out)
 {
-	pid_t pid, child_pid;
-	int redirect;
-	int status;
-	int pipe_fd[3][2];
-	char **argv;
-	int i, dir;
-
-	if ( (argv=(char **) malloc((nargs+2)*sizeof(char *)))==NULL ) return -1;
-
 	if ( err2out ) stdio_fd[2] = FDREDIR_NONE;
 
-	redirect = ( stdio_fd[0]!=FDREDIR_NONE ||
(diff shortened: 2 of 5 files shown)


lib/Makefile
```diff
@@ -4,11 +4,12 @@ endif
 
 include $(TOPDIR)/Makefile.global
 
-OBJECTS = $(addsuffix $(OBJEXT),lib.error lib.misc)
+OBJECTS = lib.error$(OBJEXT) lib.misc$(OBJEXT)
```
eldering: Any reason for this change? I think their result is identical and I'd somewhat prefer the original.

Makefile.global
```diff
@@ -62,12 +62,11 @@ endef
 # Library objects required in multiple places:
 LIBSBASE   = $(addprefix $(TOPDIR)/lib/,lib.error lib.misc)
 LIBHEADERS = $(addsuffix .h,$(LIBSBASE))
-LIBSOURCES = $(addsuffix .c,$(LIBSBASE))
 LIBOBJECTS = $(addsuffix $(OBJEXT),$(LIBSBASE))
 CFLAGS   += -I$(TOPDIR)/lib -I$(TOPDIR)/etc
 CXXFLAGS += -I$(TOPDIR)/lib -I$(TOPDIR)/etc
 
-$(LIBOBJECTS): %$(OBJEXT): %.c %.h
+$(LIBOBJECTS):
```
eldering, line 69: Did you test that these changes work and trigger a rebuild if you change either the source or the header? I'm assuming you're relying on implicit make rules?

lib/lib.misc.c
```diff
@@ -7,91 +7,83 @@
...
-	             stdio_fd[1]!=FDREDIR_NONE ||
-	             stdio_fd[2]!=FDREDIR_NONE );
+	const bool redirect = ( stdio_fd[0]!=FDREDIR_NONE ||
+	                        stdio_fd[1]!=FDREDIR_NONE ||
+	                        stdio_fd[2]!=FDREDIR_NONE );
 
-	/* Build the complete argument list for execvp.
-	 * We can const-cast the pointers, since execvp is guaranteed
-	 * not to modify these (or the data pointed to).
-	 */
```
eldering, line 68: Maybe keep the comment part on const-casting, especially since the original data from the caller is passed in?

lib/lib.misc.cc
```diff
@@ -115,21 +109,15 @@ int execute(const char *cmd, const char **args, int nargs, int stdio_fd[3], int
...
-are welcome to redistribute it under certain conditions.  See the GNU\n\
-General Public Licence for details.\n", prog, vers);
+	std::cout << prog << " -- part of DOMjudge version " << vers << std::endl
+	          << "Written by the DOMjudge developers" << std::endl << std::endl
+	          << "DOMjudge comes with ABSOLUTELY NO WARRANTY.  This is free software, and you" << std::endl
+	          << "are welcome to redistribute it under certain conditions.  See the GNU" << std::endl
+	          << "General Public Licence for details." << std::endl;
 	exit(0);
-}
+}
```
eldering: Nitpick: keep the final newline

lib/lib.misc.h
```diff
@@ -17,23 +17,22 @@ extern "C" {
...
- * int stdio_fd[3]  File descriptors for stdin, stdout and stderr respectively.
- *                    Each can separately be set to one of the following:
- *                      FDREDIR_NONE - don't do redirection
- *                      FDREDIR_PIPE - connect to pipe and set value to file
- *                                     descriptor of other end of the pipe
- *                      fd >= 0      - make this a duplicate of <fd>
- * int err2out      Set non-zero to redirect command stderr to stdout. When set
- *                    the redirection of stderr by stdio_fd[2] is ignored.
+ * const std::string& cmd                command to be executed (PATH is searched)
+ * const std::vector<std::string>& args  vector of arguments to command
```
eldering: Suggestion: maybe leave out the typing (as it can be seen from the function declaration) and it would save a lot of space and ease reading.

lib/Makefile
meisterT: reverted

lib/lib.misc.c
meisterT: Added back

lib/lib.misc.cc
meisterT: Done

lib/lib.misc.h
meisterT: Done

Makefile.global
meisterT: Forgot to reply here: yes, I tested this and it works.

PR #3173, opened 2025-10-26
Use string|float for database decimals

nickygerritsen:
This reverts part of 761f3238de3dc21c3da6b0af07c9fb1e1742f2a3 since Doctrine stroes decimals as strings to not loose precision.

When we upgrade to PHP 8.4, we can use a differnt type to store them as `BCMath\Number` objects instead, which seems to be the better way to do this.

### DIFF
--- a/webapp/src/Controller/API/JudgehostController.php
+++ b/webapp/src/Controller/API/JudgehostController.php
@@ -644,8 +644,8 @@ class JudgehostController extends AbstractFOSRestController
         }
 
         $runResult    = $request->request->get('runresult');
-        $startTime    = (float)$request->request->get('start_time');
-        $endTime      = (float)$request->request->get('end_time');
+        $startTime    = $request->request->get('start_time');
+        $endTime      = $request->request->get('end_time');
         $runTime      = $request->request->get('runtime');
         $outputRun    = $request->request->get('output_run');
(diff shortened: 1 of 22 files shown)


vmcj:
Is there a reason to keep the `float` in this case? Why can't we just pick 2 of the 3 options?

nickygerritsen:
> Is there a reason to keep the `float` in this case? Why can't we just pick 2 of the 3 options?

See slack. I tried this and it is very messy to do so since we need to change many places where these functions are called. Like 100+.

I'd rather tackle that once we can use the new Number class

nickygerritsen:
> You have a typo in the commit message.

I think I'm blind. What's the typo?

meisterT:
stroes -> stores

nickygerritsen:
Lol on the mobile app I didn't even see I had more than one line. Fixed.

vmcj:
> > You have a typo in the commit message.
> 
> I think I'm blind. What's the typo?

differnt

But I see you already merged because of another typo.

nickygerritsen:
> > > You have a typo in the commit message.
> > 
> > 
> > I think I'm blind. What's the typo?
> 
> differnt
> 
> But I see you already merged because of another typo.

🫠

Issue #2159, opened 2023-09-25
Problem import message shows on wrong page


eldering:
### Description of the problem
While importing problems into a new contest (from problem zip files), the message of problem id 108 showed up on the result page for problem 107, together with the message of problem 107 itself.

### Your environment
DOMjudge version 8.0 running on CentOS 8

### Steps to reproduce
- Import a new contest from `contest.yaml`
- Open multiple tabs for each problem after importing them from `problems.yaml`
- On each tab, go to "edit" and import the problem from its zip file, into the newly created contest

This only happened once out of 12-13 problems imported, so it seems not consistently reproducible.

### Expected behaviour
Each message shows on the page of that problem itself.

nickygerritsen:
It shows on the first page that is rendered, so this is a synchronization thing. Unless we use a separate error block for each problem (of which I don’t know the downsides) I don’t think there is much we can do

eldering:
Ok, indeed: now it shows on the submissions page:
![image](https://github.com/DOMjudge/domjudge/assets/4692022/f02dfaea-ca62-4c71-831d-938b980ba766)

I'd say it would be good to fix this. It looks really strange to see this message on a page it really doesn't belong.

nickygerritsen:
For this to be fixed we need to:
* investigate which flash messages we use
* Make sure you can only end up on one page after showing a flash message
* Make each flash message unique

Seems doable (search for “flash” in all controllers). Just a lot of work.

Anyone interested is welcome to create a PR, otherwise I can take a look but probably only after finals since we have plenty stuff to do that seems more important.

eldering:
Sure. I agree this is not very important, but would be good to fix. Maybe at a hackathon or so.

PR #3174, opened 2025-10-26
Handle pre-contest clarifications

vmcj:
I looked into if we can let both API & UI get the data via a service (to remove having to handle everything twice) but that would require a lot of rewrites which need a bit of further thought.

The first 2 commits are unrelated, I'll open another PR for those but they broke my local build.

### DIFF
--- a/webapp/src/Controller/API/ClarificationController.php
+++ b/webapp/src/Controller/API/ClarificationController.php
@@ -8,7 +8,6 @@ use App\Entity\Contest;
 use App\Entity\ContestProblem;
 use App\Entity\Team;
 use App\Utils\Utils;
-use DateTime;
 use Doctrine\ORM\NonUniqueResultException;
 use Doctrine\ORM\QueryBuilder;
 use Exception;


meisterT:
I would only do the warning, but include implications for the API in the message.

vmcj:
> I would only do the warning, but include implications for the API in the message.

So drop all other commits? I personally think keeping the UI/API uniform is an improvement.

webapp/src/Controller/Jury/ClarificationController.php
```diff
@@ -34,13 +34,30 @@ public function __construct(
         protected readonly EventLogService $eventLogService
     ) {}
 
+    private function warnClarificationBeforeContestStart(): void
+    {
+        $cc = $this->dj->getCurrentContest();
+        $message = "Generic clarifications are visible before contest start.";
+        if ($cc && $cc->getStartTime() > Utils::now()) {
+            $this->addFlash('warning', $message);
```
meisterT, line 42: Does this flash message show on the current page visit or the next one?

webapp/src/Controller/Jury/ClarificationController.php
```diff
@@ -34,13 +34,30 @@ public function __construct(
         protected readonly EventLogService $eventLogService
     ) {}
 
+    private function warnClarificationBeforeContestStart(): void
+    {
+        $cc = $this->dj->getCurrentContest();
+        $message = "Generic clarifications are visible before contest start.";
```
meisterT, line 40: not just generic, right?

webapp/src/Controller/Jury/ClarificationController.php
```diff
@@ -239,6 +257,7 @@ public function composeClarificationAction(
         #[MapQueryParameter]
         ?string $teamto = null,
     ): Response {
+        $this->warnClarificationBeforeContestStart();
```
meisterT, line 260: in addition, I was thinking whether we should add a modal dialog on submission warning of the implications

webapp/templates/team/partials/index_content.html.twig
```diff
@@ -6,6 +6,10 @@
     <h2 id="contestnotstarted" class="text-center">
         Contest {{ contest | printContestStart }}
     </h2>
+    {% if clarifications is not empty %}
```
meisterT: Thinking a little bit more: when do we send out "static events" in the event feed? Do we do that on state changes? Then hiding them in the API is perhaps not too much code complexity, but one would have to implement it to really see.

webapp/templates/team/partials/clarifications_team.html.twig
meisterT, line 1: I would rather reject clar requests before contest start through the API

webapp/src/Twig/TwigExtension.php
```diff
@@ -205,13 +205,16 @@ public function printelapsedminutes(float $start, float $end): string
...
-    public function printtime(string|float|null $datetime, ?string $format = null, ?Contest $contest = null): string
+    public function printtime(string|float|null $datetime, ?string $format = null, ?Contest $contest = null, bool $squash = false): string
     {
         if ($datetime === null) {
             $datetime = Utils::now();
         }
         if ($contest !== null && $this->config->get('show_relative_time')) {
             $relativeTime = $contest->getContestTime((float)$datetime);
+            if ($relativeTime < 0 && $squash) {
+                return "Before contest";
```
meisterT: wonder if this makes sense in general and not just here?

meisterT:
> > I would only do the warning, but include implications for the API in the message.
> 
> So drop all other commits? I personally think keeping the UI/API uniform is an improvement.

I misunderstood what was happening, and left now individual comments on the code

webapp/templates/team/partials/clarifications_team.html.twig
vmcj: So instead of letting teams ask them via UI (because they can via API) disallow both via UI & API?

I think I'm fine with that, I want the jury to still be able to send them via API because of imports and suspected we didn't want the extra branch. I did not really like having to add this to the UI but found consistency important.

webapp/src/Twig/TwigExtension.php
vmcj: This should just display it in the table, I didn't find other locations yet (also didn't really search) but in general I agree. Probably best to move this commit out and do this in a separate PR to find the other locations.

webapp/src/Controller/Jury/ClarificationController.php
vmcj: I don't like this for 2 reasons... modals are more work and in case you know what you're doing this feels a little patronizing. You have more experience with judges so if you think it's smarter I can take a look how much work it is.

webapp/src/Controller/Jury/ClarificationController.php
vmcj: You're loading the request for the page, in gathering all data for the request you add the flash. When we render the twig template there is a function which collects all flash messages and displays them.

So the answer would be "The next rendered twig template" which if you only have 1 tab open is the current page (there is an issue open where it displays on the wrong page).

If you mean that we would only provide the flash when the clarification already happened, no, not that I'm aware of.

webapp/src/Controller/Jury/ClarificationController.php
vmcj: In the later commits I don't disclose clarifications about problems, I know you guys had opinions on this but I think we should try the referential integrity. If we don't like the extra code path for maintainability we can discuss it now and change this message.

webapp/templates/team/partials/index_content.html.twig
vmcj: I don't think I understand.

webapp/src/Controller/Jury/ClarificationController.php
meisterT: Thanks, at one point we probably want to fix https://github.com/DOMjudge/domjudge/issues/2159 more generically

webapp/src/Controller/Jury/ClarificationController.php
meisterT: I think this PR does many things at once, some of which can be merged easily and maybe should be lifted to their own PRs so that they can be merged already.

webapp/src/Controller/Jury/ClarificationController.php
meisterT: This a very edge case with some consequences, so yes, I would prefer to put into the face of the jury. Under normal operation, nobody should see the modal dialog.

webapp/templates/team/partials/index_content.html.twig
meisterT: Probably best described in person - @nickygerritsen, is my understanding correct that static events are created on state changes?

webapp/templates/team/partials/clarifications_team.html.twig
meisterT: Yes, there is no point in asking for clarification for a contest where no data is supposed to be known yet to teams.

>  I want the jury to still be able to send them via API because of imports

Do we have a way to import clarifications through problem upload or what do you mean by "imports" here?

webapp/templates/team/partials/index_content.html.twig
nickygerritsen: Clarifications are not static events. But we send them out [here](https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Service/EventLogService.php#L371-L371), so when a dependency is missing basically.

meisterT: hmmm, needs more thought

webapp/templates/team/partials/clarifications_team.html.twig
vmcj: > Yes, there is no point in asking for clarification for a contest where no data is supposed to be known yet to teams.

I think `can I go to the toilet` is always an important one. But I'm fine with either way.. either the extra branch in the UI or the extra branch in the API.

> 
> Do we have a way to import clarifications through problem upload or what do you mean by "imports" here?

I'm not sure if we have it yet, but I think it's something we should have. In BAPC prelims (multisite over different days) the jury now updates the PDF. But I can imagine a year where we would just also upload the clars directly in case as they don't have the time.

webapp/src/Controller/Jury/ClarificationController.php
vmcj: > This a very edge case with some consequences, so yes, I would prefer to put into the face of the jury. Under normal operation, nobody should see the modal dialog.

Sounds reasonable, I'll spend the time on it.

webapp/src/Controller/Jury/ClarificationController.php
vmcj: > I think this PR does many things at once, some of which can be merged easily and maybe should be lifted to their own PRs so that they can be merged already.

Which ones would you approve? I really like to close this PR soon as getting merge conflicts here is risky.

webapp/src/Controller/API/ClarificationController.php
```diff
@@ -213,6 +214,17 @@ public function addAction(
...
+            $startTime = $contest->getStarttimeString();
+            $startTimeTimeZone = DateTime::createFromFormat('Y-m-d h:i:s e', $startTime)->getTimezone();
+            $now = DateTime::createFromFormat('U.u', (string) $time);
+            $now->setTimezone($startTimeTimeZone); // We can't the timezone in createFromFormat as it always picks UTC.
+            throw new BadRequestHttpException(
+                "Sending a clarification before the contest can disclose restricted information, "
+                . "provide an explicit time when this clarification should be visible. "
+                . "For the start of this contest: " . $contest->getStarttimeString() . ", "
+                . "for the current time: " . $now->format('Y-m-d H:i:s e') . "."
+            );
```
vmcj:
```suggestion
```

Already added with another commit.

webapp/src/Controller/API/ClarificationController.php
```diff
@@ -8,6 +8,7 @@
 use App\Entity\ContestProblem;
 use App\Entity\Team;
 use App\Utils\Utils;
+use DateTime;
```
vmcj:
```suggestion
```

PR #3178, opened 2025-11-01
Display all languages in the contest in teaminterface

vmcj:
Also limit the languages in case the problem has different languages.

image

I don't really like the `[All]` as it is not really true but `rest` is also not that good. Found this when playing around with something and needed the info so added it.

@nickygerritsen suggested this as alternative:
image
Or this:
image

### DIFF
--- a/webapp/src/Controller/Team/LanguageController.php
+++ b/webapp/src/Controller/Team/LanguageController.php
@@ -40,7 +40,7 @@ class LanguageController extends BaseController
      * @param array<string, array{'problems': ContestProblem[], 'language': Language}> $languages
      * @return array<string, array{'problems': ContestProblem[], 'language': Language}>
      */
-    private function addLanguage(array $languages, Language $language, ContestProblem $problem): array
+    private function addLanguage(array $languages, Language $language, ?ContestProblem $problem = null): array
     {
         $langId = $language->getLangid();
         if (!isset($languages[$langId])) {
@@ -65,23 +65,30 @@ class LanguageController extends BaseController
         foreach ($this->dj->getAllowedLanguagesForContest($currentContest) as $language) {
             $allLanguages->add($language);
         }
-        // Add the problem specific languages
-        foreach ($this->dj->getCurrentContest()->getProblems() as $problem) {
-            $problemLanguages = new Set($problem->getProblem()->getLanguages());
-            if ($problemLanguages->isEmpty()) {
-                continue;
-            }
-            if (!Utils::equalSets($allLanguages, $problemLanguages)) {
-                if (!$allLanguages->isEmpty()) {
-                    $limited = true;
+        if ($currentContest) {
+            // Add the problem specific languages
+            $currentContestProblems = $currentContest?->getProblems() ?? [];
+            foreach ($currentContestProblems as $problem) {
+                $problemLanguages = new Set($problem->getProblem()->getLanguages());
+                if ($problemLanguages->isEmpty()) {
+                    continue;
                 }
-                $allLanguages->merge($problem->getProblem()->getLanguages());
+                if (!Utils::equalSets($allLanguages, $problemLanguages)) {
+                    if (!$allLanguages->isEmpty()) {
+                        $limited = true;
+                    }
+                    $allLanguages->merge($problem->getProblem()->getLanguages());
+                }
             }
-        }
-        foreach ($this->dj->getCurrentContest()->getProblems() as $problem) {
-            $problemLanguages = count($problem->getProblem()->getLanguages()) ? new Set($problem->getProblem()->getLanguages()) : $allLanguages;
-            foreach ($problemLanguages as $language) {
-                $languages = $this->addLanguage($languages, $language, $problem);
+            foreach ($currentContestProblems as $problem) {
+                $problemLanguages = count($problem->getProblem()->getLanguages()) ? new Set($problem->getProblem()->getLanguages()) : $allLanguages;
+                foreach ($problemLanguages as $language) {
+                    $languages = $this->addLanguage($languages, $language, $problem);
+                }
+            }
+        } else {
+            foreach ($allLanguages as $language) {
+                $languages = $this->addLanguage($languages, $language);
             }
         }
         return $this->render('team/languages.html.twig', ['languages' => $languages, 'limited' => $limited]);
--- a/webapp/src/Controller/Team/LanguageController.php
+++ b/webapp/src/Controller/Team/LanguageController.php
@@ -67,7 +67,7 @@ class LanguageController extends BaseController
         }
         if ($currentContest) {
             // Add the problem specific languages
-            $currentContestProblems = $currentContest?->getProblems() ?? [];
+            $currentContestProblems = $currentContest->getProblems();
             foreach ($currentContestProblems as $problem) {
                 $problemLanguages = new Set($problem->getProblem()->getLanguages());
                 if ($problemLanguages->isEmpty()) {
--- a/webapp/src/Controller/Team/LanguageController.php
+++ b/webapp/src/Controller/Team/LanguageController.php
@@ -3,10 +3,14 @@
 namespace App\Controller\Team;
 
 use App\Controller\BaseController;
+use App\Entity\ContestProblem;
+use App\Entity\Language;
 use App\Service\ConfigurationService;
 use App\Service\DOMJudgeService;
 use App\Service\EventLogService;
+use App\Utils\Utils;
 use Doctrine\ORM\EntityManagerInterface;
+use Ds\Set;
 use Symfony\Component\ExpressionLanguage\Expression;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
@@ -32,6 +36,20 @@ class LanguageController extends BaseController
         parent::__construct($em, $eventLogService, $dj, $kernel);
     }
 
+    /**
+     * @param array<string, array{'problems': ContestProblem[], 'language': Language}> $languages
+     * @return array<string, array{'problems': ContestProblem[], 'language': Language}>
+     */
+    private function addLanguage(array $languages, Language $language, ContestProblem $problem): array
+    {
+        $langId = $language->getLangid();
+        if (!isset($languages[$langId])) {
+            $languages[$langId] = ['problems' => [], 'language' => $language];
+        }
+        $languages[$langId]['problems'][] = $problem;
+        return $languages;
+    }
+
     #[Route(path: '', name: 'team_languages')]
     public function languagesAction(): Response
     {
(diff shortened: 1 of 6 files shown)


nickygerritsen:
I like the idea, but I would not do `All`, but:

* Not display any badge anywhere if we have no problem specific languages. (you already do this I think)
* Display exact problewms otherwise. So Instead of `All` display a badge for each problem that does not exclude it.

vmcj:
> I like the idea, but I would not do `All`, but:
> 
>     * Not display any badge anywhere if we have no problem specific languages. (you already do this I think)
Yes

> 
>     * Display exact problewms otherwise. So Instead of `All` display a badge for each problem that does not exclude it.
I didn't try that but thought it would be overcrowded, but I can try it.

meisterT:
I like adding this info in general, wonder if we need something special for too many problems (e.g. > 15) like online judges where this different languages per problem are more likely

vmcj:
> I like adding this info in general,

So also if there are no specific languages per problem? I think I disagree

> wonder if we need something special for too many problems (e.g. > 15) like online judges where this different languages per problem are more likely

Let me play around with such a situation.

meisterT:
> > I like adding this info in general,
> 
> So also if there are no specific languages per problem? I think I disagree


I think we are on the same page, I just worded it confusingly. I meant, I think the change is an improvement, but there might be the edge case that I pointed out

vmcj:
image

I think it's crowded but not much worse than with displaying all the problems. Labels are terrible mostly as I generated the import file.

I also tried this:
image
which works under the condition that there are more than 25 problems, and picks the shortest list of either the problems allowed or not allowed. This only makes the problem happens less but in case you have 52 problems with a language being allowed for 26 problems (and not for the other 26) it would still be crowded.

I'm aware of the issue with `C` not having a problem, in case we want the option I described above I can fix it.

The last alternative (which I think is the worst):
image

nickygerritsen:
I don't like the 'not for' variant, since it seems to draw you towards a colored box while the color is exactly what you don't want to know.

vmcj:
> I don't like the 'not for' variant, since it seems to draw you towards a colored box while the color is exactly what you don't want to know.

I agree, but I also don't like the alternatives.

Basically we can:
- The `not for` variant only if you have more than 25 items,
- Make it scrollable (see the last screenshot with `C#`),
- Display all of them, no limitations.

vmcj:
@meisterT I prefer to get this merged at one point, we can think about the edge case some more later.

webapp/src/Controller/Team/LanguageController.php
```diff
@@ -33,16 +34,73 @@ public function __construct(
         parent::__construct($em, $eventLogService, $dj, $kernel);
     }
 
+    /**
+     * @param Language[] $languages
+     * @return Language[]
```
meisterT: This is not the return type that we actually return, is it?

webapp/src/Controller/Team/LanguageController.php
```diff
@@ -33,16 +34,73 @@ public function __construct(
...
             throw new BadRequestHttpException("You are not allowed to view this page.");
         }
+        $languages = [];
         $currentContest = $this->dj->getCurrentContest();
-        /** @var Language[] $languages */
-        $languages = $this->dj->getAllowedLanguagesForContest($currentContest);
-        return $this->render('team/languages.html.twig', ['languages' => $languages]);
+        $limited = false;
+        $allLanguages = [];
+        foreach($this->dj->getAllowedLanguagesForContest($currentContest) as $language) {
```
meisterT: nit: phpcs error

webapp/src/Controller/Team/LanguageController.php
meisterT: I assume it is something complex like
```
array<string, array{problems: ContestProblem[], limitedProblems: ContestProblem[], language: Language}>
```

webapp/src/Controller/Team/LanguageController.php
```diff
@@ -33,16 +34,73 @@ public function __construct(
         parent::__construct($em, $eventLogService, $dj, $kernel);
     }
 
+    /**
+     * @param Language[] $languages
+     * @return Language[]
+     */
+    private function addLanguage(array $languages, Language $language, ContestProblem $problem, bool $inverted = false): array
+    {
+        $langId = $language->getName();
```
meisterT: is it intentional here to use `->getName()` and not `->getLandId()`?

vmcj:
Extreme case
image

Normal case:
image

I think I like the 2nd placement the most, so left aligned and no scroll but it would make it more crowded for the online judge scenario

meisterT:
Can you please do the screenshot of the full page (or at least full box for one language)?

vmcj:
image

Not 100% sure if this is an improvement but it would provide all needed information instead of forcing the team to think about this themselves.

webapp/src/Controller/Team/LanguageController.php
vmcj: No,

webapp/src/Controller/Team/LanguageController.php
vmcj: Yes, no idea why I didn't document that properly, fixed now.

vmcj:
> Can you please do the screenshot of the full page (or at least full box for one language)?

image
image

webapp/src/Utils/Utils.php
```diff
@@ -1147,4 +1148,13 @@ public static function numericToBcMath(float|int|string $value, int $scale = 9):
...
         return bcadd($value, '0', $scale);
     }
+
+    /**
+     * @template V
+     * @param Set<V> $a
+     * @param Set<V> $b
+     */
+    public static function equalSets(Set $a, Set $b): bool {
+        return $a->diff($b)->isEmpty() && $b->diff($a)->isEmpty();
```
meisterT: consider using this instead:
```
return $a->xor($b)->isEmpty();
```

webapp/src/Controller/Team/LanguageController.php
```diff
@@ -40,7 +58,32 @@ public function languagesAction(): Response
...
-        return $this->render('team/languages.html.twig', ['languages' => $languages]);
+        $languages = [];
+        $limited = false;
+        $allLanguages = new Set();
+        // Get all languages specific to the contest, if those are empty all globally enabled languages
+        foreach ($this->dj->getAllowedLanguagesForContest($currentContest) as $language) {
+            $allLanguages->add($language);
+        }
+        // Add the problem specific languages
+        foreach ($this->dj->getCurrentContest()->getProblems() as $problem) {
```
meisterT: `getCurrentContest` can return null, can't it? Also consider using the local var `$currentContest` that you create above

meisterT (commented):
we have something like "collect all languages including problem-specific ones" in a couple places, consider putting some helper method in DOMjudgeService

webapp/src/Utils/Utils.php
vmcj: Would we still need the helper method or can it be taken out?

vmcj:
> we have something like "collect all languages including problem-specific ones" in a couple places, consider putting some helper method in DOMjudgeService

I tried, but from the 3 places 2 need extra information to return the right format. So I'm not sure if it will make the code better.

webapp/src/Service/DOMJudgeService.php
```diff
@@ -1177,6 +1178,20 @@ public function getTwigDataForProblemsAction(
...
+            $limited = false;
+            foreach ($problems as $problem) {
+                if (count($problem->getProblem()->getLanguages())) {
+                    $limited = true;
+                }
+            }
+            foreach ($problems as $problem) {
+                if (!count($problem->getProblem()->getLanguages()) && $limited) {
+                    foreach ($this->getAllowedLanguagesForContest($contest) as $language) {
+                        $problem->getProblem()->addLanguage($language);
```
meisterT, line 1190: I am not particularly fond of mutating the problem's internal state (calling `addLanguage`) in a method which is called `getSomething`.

webapp/src/Service/DOMJudgeService.php
```diff
@@ -1177,6 +1178,20 @@ public function getTwigDataForProblemsAction(
                 ->getQuery()
                 ->getResult();
 
+            $limited = false;
+            foreach ($problems as $problem) {
+                if (count($problem->getProblem()->getLanguages())) {
+                    $limited = true;
```
meisterT, line 1184: Is this `$limited` dead code?

webapp/src/Service/DOMJudgeService.php
```diff
@@ -1847,9 +1862,25 @@ public function shadowMode(): bool
...
+                foreach ($contest->getProblems() as $problem) {
+                    $problemLanguages = new Set($problem->getProblem()->getLanguages());
+                    if ($problemLanguages->isEmpty()) {
+                        $limited = false;
+                    } else {
+                        $allLanguages = $allLanguages->merge($problemLanguages);
+                    }
+                }
+                if ($limited) {
+                    $allLanguages = $allLanguages->merge($languages);
```
meisterT, line 1881: Where is `$allLanguages` used?

meisterT: Looks like the whole `if ($alsoProblemSpecific)` branch is dead code?

webapp/src/Controller/Team/LanguageController.php
```diff
@@ -32,6 +36,20 @@ public function __construct(
...
+     * @param array<string, array{'problems': ContestProblem[], 'language': Language}> $languages
+     * @return array<string, array{'problems': ContestProblem[], 'language': Language}>
+     */
+    private function addLanguage(array $languages, Language $language, ?ContestProblem $problem = null): array
+    {
+        $langId = $language->getLangid();
+        if (!isset($languages[$langId])) {
+            $languages[$langId] = ['problems' => [], 'language' => $language];
+        }
+        $languages[$langId]['problems'][] = $problem;
```
meisterT, line 49: Is that what we want if `$problem` is `null`?

webapp/templates/team/languages.html.twig
```diff
@@ -7,19 +7,46 @@
...
                                 <h2 class="lang-name h4">{{ lang.name }}</h2>
-                                <div class="d-flex gap-2 mt-2">
+                                <div class="d-flex gap-2 mt-2 extensions-team-interface">
                                     {% for ext in lang.extensions %}
                                         <span class="extension-badge">.{{ ext }}</span>
                                     {% endfor %}
+                                    {% if limited %}
+                                        <div style="margin-left: auto; float: right; text-align: right;">
+                                            {% for problem in problems %}
+                                                {{ problem | problemBadge }}
```
meisterT, line 25: Not sure which of these three you want to keep. Leaving the comment just to make sure we remember to pick one option.

PR #3180, opened 2025-11-02
Check floats according to 2025-09

mzuenni:
The `2025-09` spec defines what a floating point is. This PR changes the `default_compare.cc` accordingly.

### DIFF
--- a/sql/files/defaultdata/compare/compare.cc
+++ b/sql/files/defaultdata/compare/compare.cc
@@ -1,10 +1,14 @@
 // default_validator from kattis problemtools package
 // licensed under MIT license
 //
-// modified: float comparison
+// modified:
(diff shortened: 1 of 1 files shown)


meisterT:
@mzuenni I have just created some basic tests in https://github.com/DOMjudge/domjudge/pull/3183 - I think it makes sense to merge that one first and then add more tests that do not pass today but pass after your changes.

Also consider splitting up the formatting-only changes into a separate commit

mzuenni:
@meisterT sounds good. Some inspiration for tests: https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/main/test/default_output_validator/default_output_validator.yaml (with these changes all those tests should be passed)

PR #3181, opened 2025-11-02
Cache the results of getAssetFiles

cubercsl:
Fix the disk backlog during scoreboard pictures rendering.

When running a contest with 500+ teams, this implementation consumes a large amount of CPU and causes disk backlog issues. However, it seems some unit test failed after my fix.

In our running contest now, the actual mitigation is to turn off the scoreboard org logo display

cc @moesoha. If he proposes a better implementation in a few days, I will close this PR.


nickygerritsen:
Maybe it's even better to cache the results of `getAssetFiles`? We can easily add a class property that stores the result. What do you think? If so, we need to also implement the `Symfony\Contracts\Service\ResetInterface` and reset it for when it is used inside a long running process.

moesoha:
@nickygerritsen Yes, you are correct. I have already drafted a patch with cache earlier this day, and we are going to have it tested before push here.

nickygerritsen:
> @nickygerritsen Yes, you are correct. I have already drafted a patch with cache earlier this day, and we are going to have it tested before push here.

Great, thanks!

nickygerritsen:
@cubercsl is this still draft? If not, can you rebase onto main and mark it as a non draft?

PR #3182, opened 2025-11-02
Replace async reporting back of results with guzzle.

meisterT:
Previously, we were in an ugly way calling the judgedaemon again.

A later step can be to replace all curl calls with guzzle.


meisterT:
Needs some other approach

PR #3185, opened 2025-11-05
Use alternative github download URL

vmcj:
The other ended with a 404. The question is how they determine the `latest` tag and if they keep on repushing that one.


vmcj:
The 404 seems to have fixed itself.

PR #3186, opened 2025-11-08
Don't display future clarifications to teams

vmcj:
Only time where this can ever happen is a replay of a contest but change is trivial enough.

### DIFF
--- a/webapp/src/Controller/Team/MiscController.php
+++ b/webapp/src/Controller/Team/MiscController.php
@@ -101,13 +101,16 @@ class MiscController extends BaseController
                 ->andWhere('c.contest = :contest')
                 ->andWhere('c.sender IS NULL')
                 ->andWhere('c.recipient = :team OR c.recipient IS NULL')
+                ->andWhere('c.submittime <= :time')
                 ->setParameter('contest', $contest)
                 ->setParameter('team', $team)
+                ->setparameter('time', time())
                 ->addOrderBy('c.submittime', 'DESC')
                 ->addOrderBy('c.clarid', 'DESC');
             if ($contest->getStartTimeObject()->getTimestamp() > time()) {
                 $qb->andWhere('c.problem IS NULL');
             }
+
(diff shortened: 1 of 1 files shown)


Kevinjil:
Could this have an impact on multi-site contests / setups where people submit on a shadowing DJ instance and the server time is not well enough in sync? In such cases, do we want to give the clarification asap, or on the same time within the contest according to the server?

vmcj:
> Could this have an impact on multi-site contests / setups where people submit on a shadowing DJ instance and the server time is not well enough in sync?

I don't think we should consider cases where people misconfigure something. Either the spec is to broad here and should not allow clarifications in the future (so we should change any time which is future to now)  or this is a valid usecase where I would say we should consider replays and do what I propose.

In case you submit to different instances something should always be in sync, because otherwise teams get information on different times. I think what you propose can be better fixed with having a `rel_time` per clarification but not by assuming the absolute time is incorrect.

> In such cases, do we want to give the clarification asap, or on the same time within the contest according to the server?

I think I don't understand what you propose. But if we really want to consider misalignment's of time between backend systems (and having a backendsystem is already in the 20% cases..) which are in the 20%s again we should also allow submitting after contest is over and getting info before the contest starts as I think you could make this case for all of those.

Kevinjil (approved):

webapp/src/Controller/Team/MiscController.php
```diff
@@ -105,9 +104,14 @@ public function homeAction(Request $request): Response
                 ->setParameter('contest', $contest)
                 ->setParameter('team', $team)
                 ->addOrderBy('c.submittime', 'DESC')
-                ->addOrderBy('c.clarid', 'DESC')
-                ->getQuery()
-                ->getResult();
+                ->addOrderBy('c.clarid', 'DESC');
+            if (!$this->dj->checkrole('jury')) {
```
nickygerritsen: Again I would not make a distinction for jury here.

PR #3187, opened 2025-11-08
Handle team clarification edge cases

vmcj:
Couple of commits from: https://github.com/DOMjudge/domjudge/pull/3174 which are expected to already get approved to make discussion easier.

### DIFF
--- a/webapp/src/Controller/Team/MiscController.php
+++ b/webapp/src/Controller/Team/MiscController.php
@@ -5,13 +5,13 @@ namespace App\Controller\Team;
 use App\Controller\BaseController;
 use App\DataTransferObject\SubmissionRestriction;
 use App\Entity\Clarification;
-use App\Entity\Language;
 use App\Form\Type\PrintType;
 use App\Service\ConfigurationService;
 use App\Service\DOMJudgeService;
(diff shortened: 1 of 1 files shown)


webapp/src/Controller/Team/MiscController.php
```diff
@@ -105,9 +104,12 @@ public function homeAction(Request $request): Response
                 ->setParameter('contest', $contest)
                 ->setParameter('team', $team)
                 ->addOrderBy('c.submittime', 'DESC')
-                ->addOrderBy('c.clarid', 'DESC')
-                ->getQuery()
-                ->getResult();
+                ->addOrderBy('c.clarid', 'DESC');
+            if (!$this->dj->checkrole('jury') && $contest->getStartTimeObject()->getTimestamp() > time()) {
```
nickygerritsen: Are we sure we want this first part? Even if you are jury, do we want to show these clarifications on the team pages? I think this might confuse people.

vmcj: I was thinking the same thing, and the thing I think I wanted the most is to as a jury member be able to "restrict" the view to what a team would see or get all the information.

I'll make it uniform for now.

PR #3188, opened 2025-11-08
Use the correct monaco package

Kevinjil:
The `monaco` package is some indepdantly maintained build and not the original monaco editor. It also seemed incredibly outdated (9y) and not API compatible with the current `monaco` release and broke the editors. Switch to `monaco-editor` instead.

### DIFF
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -5,7 +5,7 @@
     "d3": "3.5.17",
     "file-saver": "^2.0.5",
     "jquery-debounce-throttle": "^1.0.6-rc.0",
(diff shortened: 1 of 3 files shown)


Kevinjil:
For reference: [monaco-editor](https://www.npmjs.com/package/monaco-editor) vs [monaco](https://www.npmjs.com/package/monaco)

Kevinjil:
@vmcj E.g. missing the `setLanguage` function on https://www.domjudge.org/demoweb/jury/submissions/9675/source.

PR #3189, opened 2025-11-09
Fix Monaco-editor font rendering

vmcj:
Fixes a CSP violation as monaco-editor loads a font over data:.

@Kevinjil can you check if you get the same warning without it?


Kevinjil (approved):
I did not notice, but you are correct. I also see this error, which is caused when loading Monaco but I have not figured out which image is missing, if any.
```
The page’s settings blocked the loading of a resource (img-src) at blob:http://localhost:12345/6fb207db-4f78-4ff5-889b-74f17e82641f because it violates the following directive: “img-src 'self' data:”
```

PR #3192, opened 2025-11-09
Don't display clarification time from before/after contest

vmcj:
It looks ugly in some cases and less professional in others where printing happened much earlier. The exact time is not relevant and could lead to discussions with teams.

Originally this was only for the team interface but it does add information for the jury. It's more important to know that something happened before the contest (or after) and the interpretation of that time as knowing when it exactly happened.

### DIFF
--- a/webapp/src/Twig/TwigExtension.php
+++ b/webapp/src/Twig/TwigExtension.php
@@ -204,7 +204,6 @@ class TwigExtension extends AbstractExtension implements GlobalsInterface
 
     /**
      * Print a time formatted as specified. The format is according to date().
-     * @param Contest|null $contest If given, print time relative to that contest start.
      * @param bool $maskOutsideContest When true and contest is given replace time with before/after.
      */
     public function printtime(string|float|null $datetime, ?string $format = null, ?Contest $contest = null, bool $maskOutsideContest = true): string


vmcj:
> I wonder if we have any cases where this would look ugly but can't think of any.
> 
> I like the idea but would wait for other opinions before merging.

I looked at the pages and I think it works in most locations, I disabled it for one. The question is if we do like this for `jury` pages or if we prefer the time there, I prefer this but others might want the time.

Fine with waiting for some extra opinions.

nickygerritsen:
> I disabled it for one. 

I don't see any changes to .twig files?

vmcj:
> > I disabled it for one.
> 
> I don't see any changes to .twig files?

I was talking about https://github.com/DOMjudge/domjudge/blob/b65759827ef25b4839b1186268b6f243562aeb93/webapp/templates/partials/problem_list.html.twig#L108 and I think I removed it as it should already work. I've now added it back just to be sure.

webapp/src/Twig/TwigExtension.php
```diff
@@ -205,13 +205,20 @@ public function printelapsedminutes(float $start, float $end): string
...
     {
         if ($datetime === null) {
             $datetime = Utils::now();
         }
         if ($contest !== null && $this->config->get('show_relative_time')) {
             $relativeTime = $contest->getContestTime((float)$datetime);
+            if ($relativeTime < 0 && $squash) {
+                return "Before contest";
+            }
+            if ($relativeTime > $contest->getContestTime($contest->getEndtime())) {
```
eldering: I'd say this also requires `$squash`.

webapp/src/Twig/TwigExtension.php
```diff
@@ -205,13 +205,20 @@ public function printelapsedminutes(float $start, float $end): string
      * Print a time formatted as specified. The format is according to date().
      * @param Contest|null $contest If given, print time relative to that contest start.
```
meisterT: nit: update params

webapp/src/Twig/TwigExtension.php
```diff
@@ -205,13 +205,20 @@ public function printelapsedminutes(float $start, float $end): string
...
      */
-    public function printtime(string|float|null $datetime, ?string $format = null, ?Contest $contest = null): string
+    public function printtime(string|float|null $datetime, ?string $format = null, ?Contest $contest = null, bool $squash = true): string
```
meisterT: nit: I am not sure the `$squash` name is intuitive, perhaps something like `$checkBounds` or `$maskOutsideContest`?

webapp/src/Twig/TwigExtension.php
vmcj: I probably misunderstand you but for now I'll assume you meant because of the new parameter added.

I don't think we should add basic parameters but only use a `@param` in case we really need to explain that parameter which I would say is actually an anti-pattern or in case of `arrays` where we need to show the type of the array.

webapp/src/Twig/TwigExtension.php
```diff
@@ -204,14 +204,24 @@ public function printelapsedminutes(float $start, float $end): string
     /**
      * Print a time formatted as specified. The format is according to date().
      * @param Contest|null $contest If given, print time relative to that contest start.
+     * @param bool $maskOutsideContest, When true and contest is given replace time with before/after.
```
nickygerritsen:
```suggestion
     * @param bool $maskOutsideContest When true and contest is given replace time with before/after.
```

And maybe remove the contest param since it's trivial what it does

PR #3193, opened 2025-11-10
Allow teams to download all samples/attachments

vmcj:
We already allow this over the API so use the same URL.

image

This doesn't check if there are actually samples as I think it will in 99% of the cases and in case there are none someone would just get a 404.

### DIFF
--- a/webapp/src/Service/DOMJudgeService.php
+++ b/webapp/src/Service/DOMJudgeService.php
@@ -884,8 +884,62 @@ class DOMJudgeService
         return Utils::streamAsBinaryFile($zipFileContent, $outputFilename, 'zip');
     }
 
+    public function checkIfSamplesZipForContest(Contest $contest): bool
+    {
+        // Note, we reload the contest with the problems and attachments, to reduce the number of queries
+        // We do not load the testcases here since addSamplesToZip loads them
+        $contestQuery = $this->em->createQueryBuilder()
+            ->from(Contest::class, 'c')
+            ->innerJoin('c.problems', 'cp')
+            ->innerJoin('cp.problem', 'p')
+            ->select('c', 'cp', 'p')
+            ->andWhere('c.cid = :cid')
+            ->andWhere('cp.allowSubmit = 1')
+            ->setParameter('cid', $contest->getCid());
+
+        /** @var Contest $contest */
+        $contest = $contestQuery->getQuery()->getOneOrNullResult();
+        if ($contest->getContestProblemsetType()) {
+            return true;
+        }
+
+        /** @var Problem[] $nonInteractiveProblems */
+        $nonInteractiveProblems = (clone $contestQuery)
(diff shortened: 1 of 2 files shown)


nickygerritsen:
I would prefer adding a check if it exists, I don't think it's very nice to add buttons that don't always work. Even if it is very rare.

webapp/src/Service/DOMJudgeService.php
```diff
@@ -896,30 +900,42 @@ public function getSamplesZipForContest(Contest $contest): StreamedResponse
             ->getQuery()
             ->getSingleResult();
 
-        $zip = new ZipArchive();
-        if (!($tempFilename = tempnam($this->getDomjudgeTmpDir(), "export-"))) {
-            throw new ServiceUnavailableHttpException(null, 'Could not create temporary file.');
-        }
+        $zip = null;
```
nickygerritsen: I get what you're doing, but I think just doing a simple query to count the number of sample testcases across all problems in the contest in `checkIfSamplesZipForContest` is cleaner than this (and more performant)

vmcj: Yes, but in case you have no samples but do have attachments you would not create the zip. But I wonder if this will ever return false, why would there be a contest with neither of:
- samples,
- problem statements
- attachments

All of those are possible, but having neither of those files really feels strange.

nickygerritsen: The zip with all samples contains only samples or also attachment? In the latter case the button has a weird name maybe, not sure.

But still I would not mis-use a method like this.

eldering: > All of those are possible, but having neither of those files really feels strange.

When everything is provided on the team machine and/or as hardcopy problem statements. I don't think that's a far-fetched case.

webapp/src/Service/DOMJudgeService.php
```diff
@@ -879,8 +879,62 @@ public function getSamplesZipStreamedResponse(ContestProblem $contestProblem): S
...
+            ->setParameter('cid', $contest->getCid());
+
+        /** @var Contest $contest */
+        $contest = $contestQuery->getQuery()->getOneOrNullResult();
+        if ($contest->getContestProblemsetType()) {
+            return true;
+        }
+
+        /** @var Problem[] $nonInteractiveProblems */
+        $nonInteractiveProblems = $contestQuery
```
nickygerritsen:
```suggestion
        $nonInteractiveProblems = (clone $contestQuery)
```

I think. Otherwise I think it modifies the existing query object and the below queries have all joins. If so, all should be `clone`d

nickygerritsen (approved):

webapp/src/Service/DOMJudgeService.php
vmcj: I verified and you're right, added it the `clone` to the queries.

PR #3194, opened 2025-11-11
Reorder solved problems

vmcj:
I've added 2 solutions, I think the 2nd is less intuitive but when you get it looks better (only 1 button). I think if we do this it makes https://github.com/DOMjudge/domjudge/pull/3137 also easier as we have to handle 1 case less and have more space per card to display the stats @meisterT tried to add.

Hidden:
image

Opened (both solutions)
image
image


meisterT:
I am against reordering / collapsing / hiding problems and would prefer one of the solutions from #3137 (not sure which one is the one that I like most yet).

vmcj:
> I am against reordering / collapsing / hiding problems and would prefer one of the solutions from #3137 (not sure which one is the one that I like most yet).

Ok, I'll close this one than.

Issue #3048, opened 2025-08-03
import-contest should not ask contest-id if there is only 1


vmcj:
If you import in 2 different passes we don't know the contest_id anymore. But if there is only 1 we should be able to guess which contest we want to import into.

PR #3195, opened 2025-11-13
Hint during 2nd import of contests with the existing contest id(s)

vmcj:
Closes: https://github.com/DOMjudge/domjudge/issues/3048

Originally I wanted to just pick the one contest in case there is only one as you typically enter this situation by fixing a failed import. But as I'm not sure if we should choose for the user even if it works in 99% of the cases the 2nd is maybe better.


nickygerritsen:
I agree that this is better

vmcj:
> I agree that this is better

I provided 2 options, I'll assume you want the 2nd.

nickygerritsen:
Yes, thisPR is what I meant

Issue #2162, opened 2023-09-27
Build script for executables should not use chroot


eldering:
### Description of the problem
Currently, executables are built using https://github.com/DOMjudge/domjudge/blob/main/judge/build_executable.sh. This runs inside the chroot, but the executable that gets built is run outside the chroot. This leads to incompatibility problems, see for example below with the glibc version, and causes internal errors.

### Your environment
DOMjudge version 8.0 on CentOS 8 with chroot Debian Bookworm.

### Steps to reproduce
Install the system, try to judge something.

### Actual behaviour
Then get this internal error:
```
[Sep 27 07:27:31.221] judgedaemon[1421]: Building executable in /home/domjudge/domjudge/output/judgings/domjudge-201-28/endpoint-default/executable/run/3/8f6a40aab30f3a816e80bb1b76fc0d6c, under 'build/'
[Sep 27 07:27:31.439] judgedaemon[1421]:   💾 Fetching new executable 'compare/1' with hash '80822e09ea714b3238620203f466caff'.
[Sep 27 07:27:31.439] judgedaemon[1421]: API request GET judgehosts/get_files/compare/1
[Sep 27 07:27:31.600] judgedaemon[1421]: Building executable in /home/domjudge/domjudge/output/judgings/domjudge-201-28/endpoint-default/executable/compare/1/80822e09ea714b3238620203f466caff, under 'build/'
[Sep 27 09:27:32.197] testcase_run.sh[7439]: starting '/home/domjudge/domjudge/lib/judge/testcase_run.sh', PID = 7439
[Sep 27 09:27:32.200] testcase_run.sh[7439]: arguments: '/home/domjudge/domjudge/output/judgings/domjudge-201-28/endpoint-default/testcase/2055/7f930ff4f8b7fd3735aed2a13bce3380_652820dfe1f4660f8ad233ba09ad6eb3.in' '/home/domjudge/domjudge/output/judgings/domjudge-201-28/endpoint-default/testcase/2055/7f930ff4f8b7fd3735aed2a13bce3380_652820dfe1f4660f8ad233ba09ad6eb3.out' '1:6' '/home/domjudge/domjudge/output/judgings/domjudge-201-28/endpoint-default/1833/3764/testcase02055'
[...]

PR #3196, opened 2025-11-13
Use static link for compare scripts

vmcj:
In case you have different GLIBC versions inside and outside of the chroot we would get errors. As dynamic linking saves a bit of space we shouldn't always do this but as we only do this for compare scripts and not the submissions this is quite a small diskoverhead.

We already have an issue to fix this in a proper way but as I encountered this both for alpine and in my laptop install (which is quite outdated with 22.04) I do think this is a good temporary fix.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -463,10 +463,10 @@ function fetch_executable_internal(
                 }
                 switch ($execlang) {
                     case 'c':
-                        $buildscript .= "gcc -Wall -O2 -std=gnu11 $source -o run -lm\n";
+                        $buildscript .= "gcc -Wall -O2 -static -std=gnu11 $source -o run -lm\n";
                         break;
                     case 'cpp':
-                        $buildscript .= "g++ -Wall -O2 -std=gnu++20 $source -o run\n";
+                        $buildscript .= "g++ -Wall -O2 -static -std=gnu++20 $source -o run\n";
                         break;
                     case 'java':
                         $buildscript .= "javac -cp . -d . $source\n";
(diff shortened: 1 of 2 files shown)


meisterT:
I have to think through this, just leaving a reference to https://github.com/DOMjudge/domjudge/issues/2162 here in case someone is interested in the larger issue

eldering:
I tried this when running into it at our BAPC preliminary and it wasn't working. I don't remember the exact reason anymore though. One possible reason: IIRC not all Linux distributions provide static link libraries.

meisterT:
Jaap, what distro did you use?

eldering:
> Jaap, what distro did you use?

Host distro CentOS 9, chroot using Debian Bookworm (oldstable) I think, but could have been Trixie (stable).

eldering:
> Host distro CentOS 9, chroot using Debian Bookworm (oldstable) I think, but could have been Trixie (stable).

FYI, I'm now referring  to this year's setup. In https://github.com/DOMjudge/domjudge/issues/2162 that was an older setup with CentOS 8 in 2023.

Actually, this year, I think I fell back to Bookworm because that had a compatible GLIBC, while Trixie didn't and I didn't get that to work, including with using static compilation.

eldering:
Just for reference, these are the diffs from my attempts to make it work. I need to dive a bit deeper to find out what actually failed for what reasons.

https://pastebin.com/jMeC8CEg
https://pastebin.com/waXRApA6

vmcj:
So I checked an on CentOS9 we also have other problems. It might be this becomes easier to fix when people switch to `clang` as there static building is possible due to other libraries.

PR #3197, opened 2025-11-13
Move jury submission source edit details to modal

Kevinjil:
Previously, we always rendered the submission details (problem, language, entry point) form below the code editor. This prevents us from scaling the editor window to use all available height (towards #3058).

We move this form to a modal, triggered by a submit button on the top right of the IDE. This uses the same design as the proposed changes for the diff editor (#3132).

Screenshot:
image

### DIFF
--- a/webapp/src/Controller/Jury/SubmissionController.php
+++ b/webapp/src/Controller/Jury/SubmissionController.php
@@ -979,8 +979,7 @@ class SubmissionController extends BaseController
             ->add('entry_point', TextType::class, [
                 'label' => 'Optional entry point',
                 'required' => false,
-            ])
-            ->add('submit', SubmitType::class);
+            ]);
 
         foreach ($files as $file) {
             $formBuilder->add('source' . $file->getRank(), TextareaType::class);
@@ -1039,12 +1038,17 @@ class SubmissionController extends BaseController
             return $this->redirectToRoute('jury_submission', ['submitId' => $submittedSubmission->getSubmitid()]);
         }
 
-        return $this->render('jury/submission_edit_source.html.twig', [
+        $twigData = [
             'submission' => $submission,
             'files' => $files,
             'form' => $form,
             'selected' => $rank,
-        ]);
+        ];
+        if ($request->isXmlHttpRequest()) {
+            return $this->render('jury/submission_edit_source_modal.html.twig', $twigData);
+        } else {
+            return $this->render('jury/submission_edit_source.html.twig', $twigData);
+        }
     }
 
     /**
--- a/webapp/templates/jury/submission_edit_source.html.twig
+++ b/webapp/templates/jury/submission_edit_source.html.twig
@@ -11,9 +11,7 @@
         source files
     </h1>
 
-    {{ form_start(form) }}
-
-    <ul class="nav nav-tabs source-tab-nav">
+    <ul class="nav nav-tabs source-tab-nav align-items-end">
         {%- for file in files %}
 
             <li class="nav-item">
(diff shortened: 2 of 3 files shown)


Kevinjil:
~@nickygerritsen Can I easily put the submit button of the form next to the close button?~
Edit: found it

webapp/templates/jury/submission_edit_source_modal.html.twig
```diff
@@ -0,0 +1,24 @@
...
+                </h5>
+                <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+            </div>
+            {{ form_start(form) }}
+            <div class="modal-body">
+                {{ form_widget(form) }}
+            </div>
+            <div class="modal-footer">
+                <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Cancel</button>
+                {# {{ form_row(form.submit, { 'label': 'Submit me' }) }} #}
```
vmcj:
```suggestion
```

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -1039,12 +1038,17 @@ public function editSourceAction(Request $request, Submission $submission, #[Map
...
             'submission' => $submission,
             'files' => $files,
             'form' => $form,
             'selected' => $rank,
-        ]);
+        ];
+        if ($request->isXmlHttpRequest()) {
+            return $this->render('jury/submission_edit_source_modal.html.twig', $twigData);
+        } else {
+            return $this->render('jury/submission_edit_source.html.twig', $twigData);
```
vmcj, line 1050: I assume you tested if it all still works when you do this from another tab/window?

webapp/templates/jury/submission_edit_source.html.twig
```diff
@@ -22,28 +20,38 @@
...
+                <i class="fas fa-cloud-upload-alt"></i> Submit
+            </a>
+        </li>
     </ul>
+
+    <script>
+    function transferSourceToModal(elem) {
+        const editors = monaco.editor.getEditors();
+        const idx = {{ files | keys | json_encode }}
+        for (let i = 0; i < editors.length; ++i) {
```
vmcj:
```suggestion
        for (let i = 0; i < editors.length; i++) {
```
If you really need the other form an explanation is needed, but I think `i++` is more used.

webapp/templates/jury/submission_edit_source_modal.html.twig
```diff
@@ -0,0 +1,24 @@
+<div class="modal fade" tabindex="-1" role="dialog" aria-hidden="true">
+    <div class="modal-dialog modal-lg" role="document">
+        <div class="modal-content">
+            <div class="modal-header">
+                <h5 class="modal-title">
+                    Submission details
+                </h5>
+                <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+            </div>
+            {{ form_start(form) }}
```
vmcj, line 10: I assume we still can't use the `form(form))` instead of splitting it out but did you check?

vmcj (approved):
I assume you checked that everything is still prefilled such as the problem & language?

webapp/templates/jury/submission_edit_source_modal.html.twig
nickygerritsen: @Kevinjil if you re-add the submit button in the form type and then do
```twig
{{ form_widget(form.submit) }}
```

I think that works. the label can be set in the type as well

Kevinjil: > @Kevinjil if you re-add the submit button in the form type and then do
> 
> ```
> {{ form_widget(form.submit) }}
> ```
> 
> I think that works. the label can be set in the type as well

I tried getting this to work, but it failed as rendering the form widget will include this button, and I need to place it next to the cancel button.

webapp/src/Controller/Jury/SubmissionController.php
Kevinjil: I'm not sure what you mean with this. The XML route is the request to render the modal, the other (normal) route is the one that renders the page which opens the modal when clicking the new submit button.

webapp/templates/jury/submission_edit_source_modal.html.twig
Kevinjil: The submit button needs to be within the form.

webapp/templates/jury/submission_edit_source_modal.html.twig
nickygerritsen: Form_end needs to be after the button. That should fix it

Kevinjil: I could be doing something wrong, but from my experiments this does not work, as the form widget will include the submit button, and the rendering of the submit button afterwards throws an exception because it is already rendered. Putting the submit button before the form_widget would work, but is not the desired result.

I could call all form widget rendering individually by name, but I think that is more error prone for future changes than including a custom submit button.

nickygerritsen: > I could call all form widget rendering individually by name, but I think that is more error prone for future changes than including a custom submit button.

This is what we do in at least some places. But I'm also fine having a custom button.

webapp/src/Controller/Jury/SubmissionController.php
vmcj: Ok, normally the modal page & the normal page do the same thing.

PR #3198, opened 2025-11-13
Dynamically stretch the code editor height

### LINKED ISSUE #3058: Double scrollbar when viewing submission source

Kevinjil:
Depends on #3132 + #3197 to clear vertical space for better UX.
Fixes #3058

### DIFF
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -1290,3 +1290,11 @@ function initScoreboardSubmissions() {
         });
     });
 }
+
+function resizeEditors() {
+    // This function assumes that there is at most one tabbed editor in the screen.
+    const activeEditor = document.querySelector(".tab-pane.active .editor");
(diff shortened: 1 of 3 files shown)


nickygerritsen (approved):

vmcj:
Why doesn't this work with 100vh? I think that should also be dynamic.

Kevinjil:
> Why doesn't this work with 100vh? I think that should also be dynamic.

100vh always references to the full viewport height. This would result in a scrollbar as some of the height is occupied by e.g. the header and footer of the page.

Kevinjil:
Closing in favor of #3206

PR #3201, opened 2025-11-16
Teach coverity that pipe initializes its argument

eldering:


eldering:
Too much fuss to get this to work, closing and simply ignoring the coverity issue.

PR #3202, opened 2025-11-16
Improve judge daemon code health and readability

meisterT:


vmcj:
I find this very hard to review, I think I like the general idea but can't really check where the lines moved to.

meisterT:
> I find this very hard to review, I think I like the general idea but can't really check where the lines moved to.

Yes, as often with bigger refactorings, the GitHub review UI is meh. My recommendation is to check things out locally and review commit by commit by opening the old and new version of the file side by side, e.g. in vim.

judge/judgedaemon.main.php
```diff
@@ -317,57 +327,18 @@ public function run(): void
...
+                $this->checkDiskSpace($workdirpath);
             }
 
             // Request open judge tasks to be executed.
             // Any errors will be treated as non-fatal: we will just keep on retrying in this loop.
-            $row = [];
-            $judging = $this->request('judgehosts/fetch-work', 'POST', ['hostname' => $this->myhost], false);
+            $row = $this->fetchWork();
+
             // If $judging is null, an error occurred; we marked the endpoint already as errorred above.
```
eldering: Also rename `$judging` in the comment.

judge/judgedaemon.main.php
```diff
@@ -314,6 +280,12 @@ private function loop(): void
             if (function_exists('pcntl_signal_dispatch')) {
                 pcntl_signal_dispatch();
             }
+            if (function_exists('pcntl_waitpid')) {
+                // Reap any finished child processes.
+                while (pcntl_waitpid(-1, $status, WNOHANG) > 0) {
+                    // Do nothing.
```
eldering: I'd generate a warning at least if the return value is -1, when an error happened.

judge/judgedaemon.main.php
```diff
@@ -647,7 +619,7 @@ private function judging_directory(string $workdirpath, array $judgeTask): strin
             . $judgeTask['jobid'];
     }
 
-    private function read_credentials(): void
+    private function readCredentials(): void
```
eldering, line 637: Nitpick: I'd be consistent and leave this as snake case as is.

judge/judgedaemon.main.php
meisterT: Done

judge/judgedaemon.main.php
meisterT: Done

judge/judgedaemon.main.php
meisterT: I did the opposite and remove all snake case from this file (as camel case is the best practice, and phpcs complains)

PR #3203, opened 2025-11-17
Fall back to external ID when exporting results

nickygerritsen:
As reported by @mkfuron on Slack.


mkfuron:
I manually applied these changes and I was successful in exporting the results.  

The prompt resolution to this issue is appreciated!

PR #3204, opened 2025-11-17
PoC - Generic Tasks implementation

vmcj:
This should be a minimal version to allow to run specific general tasks on the judgehosts. When we agree on this format the next step is to use such a script when a judgehosts registers and use that output to compare against other judgehosts and disable it if it differs from the others. That way someone can have a notion of how similar judgehosts should be without having to check manual.

I've tried to explain my reasoning in the commits to make reviewing and changes easier, this is probably not how we want it yet but a good start to discuss how this should work. So basically 2 questions:
- Is this something we want?
- Is this the way how we want it?

There are some empty commits for remarks I already had myself but left unanswered to not waste time on this route if we don't want it.

### DIFF
--- a/sql/files/defaultdata/judgehost_info/run
+++ b/sql/files/defaultdata/judgehost_info/run
@@ -10,14 +10,10 @@ output_file="$1"
 # Posted by fjarlq, modified by community. See post 'Timeline' for change history
 # Retrieved 2025-11-15, License - CC BY-SA 4.0
 
-if [ "${SHELL#*"bash"}" != "${SHELL}" ]; then
-    exec &>> "${output_file}"
-else
-    # We can't use &>> as it's not POSIX,
-    # this does introduce a racecondition
-    exec 1>> "${output_file}"
-    exec 2>> "${output_file}"
-fi
+# We can't use &>> as it's not POSIX,
+# this does introduce a racecondition
+exec 1>> "${output_file}"
+exec 2>> "${output_file}"
 
 # Generic linux/distro information
 uname -a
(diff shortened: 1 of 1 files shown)


meisterT:
I think something like this in general can be useful, didn't look at the code yet.

We need to think through the current UI/UX though:
- The new table looks out of place and becomes ugly if you have too many judgehosts.
- I haven't found where in the UI a scheduled generic task is reported,
- or where the result is reported
- It's unclear at first glance whether clicking on a possible script (on the page of an individual judgehost) would schedule it or let you view it
- the add new executable page doesn't allow adding a "generic executable"
- "Used as generic_task script" on the executable page doesn't really make sense (it is not associated with problems)

Perhaps we can find a whiteboard at KIT next week and talk through how it should look like?

vmcj:
Ui is indeed just a mockup, if we want it the next step was displaying the result with a good diff viewer.

vmcj:
> I think something like this in general can be useful, didn't look at the code yet.
> 
> We need to think through the current UI/UX though:
> 
>     * The new table looks out of place and becomes ugly if you have too many judgehosts.

I think it needs a proper bootstrap table, but this was just for the mockup.

> 
>     * I haven't found where in the UI a scheduled generic task is reported,

I think this is almost instant so not sure if that is really needed.
> 
>     * or where the result is reported

There is no page yet, if we like this I think I want it with a diff viewer as this makes most sense for scripts to compare output over different judgehosts (a startup script for example, or something which is ran before judgement starts) and to give a possibility to mark some output as desired and disable the judgehost when it gets another output.

> 
>     * It's unclear at first glance whether clicking on a possible script (on the page of an individual judgehost) would schedule it or let you view it

It works the same as on the overview page, probably having the judgehost link as link to the script is better and have a button to trigger the run.
> 
>     * the add new executable page doesn't allow adding a "generic executable"
> 
>     * "Used as generic_task script" on the executable page doesn't really make sense (it is not associated with problems)
> 
But having that empty is also strange IMO.

> 
> Perhaps we can find a whiteboard at KIT next week and talk through how it should look like?

I would have liked to get parts merged before NWERC to test with it at NWERC but lets discuss onsite.

PR #3206, opened 2025-11-22
Dynamically stretch the code editor height using flexboxes

Kevinjil:
Alternative to #3198 which relies on flexbox layouts instead of computing the height ourselves, which makes it more generally applicable and probably prevents dozens of edge-cases and accessibility problems. The commit messages contain the rationale.

### DIFF
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -1588,3 +1588,15 @@ function initDiffEditorTab(editorId, diffId, submissionId, models) {
     }
     diffEditor.onDidUpdateDiff(updateIcon);
 }
+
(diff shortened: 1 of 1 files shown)


vmcj:
I like this more over the other PR.

Kevinjil:
@vmcj This force-push is a rebase and re-applying the `flex-pass` on the correct elements.

vmcj:
@Kevinjil can this be merged?

Kevinjil:
Yes, I kept it unmerged as it causes conflicts with the external id branch and this one is probably easier to rebase. Given that that one will take some more time, let's merge it anyway.

Issue #2019, opened 2023-05-07
Import of problem zip slow with many submission files


eldering:
Importing `example_problems/hello` as a zip takes about 10 seconds, and almost all of that time seems spent importing the submissions.

I think the problem is the double for loop over the files in the zip at https://github.com/DOMjudge/domjudge/blob/01b754ab46bd7195cc412313f559ccb7ed03f86c/webapp/src/Service/ImportProblemService.php#L635-L659, which may become slow when there's in the case of `hello` 223 files/directories in the zip file.

meisterT:
I tried to investigate a bit, the double loop doesn't seem to be an issue. Also we only entering the inner loop for multi-file submissions - for the `hello` problem we only have a few of these. Commenting that code made it barely faster.

meisterT:
Elaborating here  a little bit more: the slow part is that we create all judging runs and judge tasks (one per test case and judging).

We discussed in Luxor that it we could have a special queuetask to do that instead per judging and let the judgedaemon create the actual judging_run and judge tasks for this submission.

If we do so, also rejudgings should be a *lot* faster.

meisterT:
I looked into this again, and it gets a *lot* faster (before 21s, after 4s) if I disable the event log here:
https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Service/SubmissionService.php#L637-L641

Slow part in there is the internal API request (goes up from 4s to 5s if I do the rest but not the internal API request): https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Service/EventLogService.php#L319

The interesting part seems to be that each of these requests seems to take longer, they take around 5-20ms for the first few requests and can go up to 200ms and more per request:
![image](https://github.com/user-attachments/assets/2895e553-997a-44d1-9ce9-8af97b56f1eb)

Now I first thought that this increases with number of submissions in the database, but that doesn't seem to be true. Perhaps there is some other inefficiency in the stacked subrequests that grows with the number of subrequests within one parent request?

nickygerritsen:
Did you run this in prod mode? In dev mode queries are logged and the more queries you do the slower this gets.

meisterT:
No, I have used dev mode and also enabled the profiler, otherwise I would be blind. But that's a sensible explanation, I'll try later with prod mode and see how overall timings change.

nickygerritsen:
[...]

PR #3207, opened 2025-11-22
Micro optimization to speed up problem import.

meisterT:
On my machine, this does speed up a problem with hundreds of test cases and hundreds of submissions by 4%.

Part of https://github.com/DOMjudge/domjudge/issues/2019

### DIFF
--- a/webapp/src/Service/ImportProblemService.php
+++ b/webapp/src/Service/ImportProblemService.php
@@ -68,6 +68,11 @@ class ImportProblemService
             $messages[$type] = [];
         }
 
+        $zipEntries = [];
+        for ($j = 0; $j < $zip->numFiles; $j++) {
+            $zipEntries[$j] = $zip->getNameIndex($j);
+        }
+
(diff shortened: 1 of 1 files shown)


webapp/src/Service/ImportProblemService.php
```diff
@@ -904,11 +905,10 @@ public function importProblemFromRequest(Request $request, ?int $contestId = nul
     /**
      * @param array{danger?: string[], info?: string[]} $messages
      */
-    private function searchAndAddValidator(ZipArchive $zip, ?array &$messages, string $externalId, string $validationMode, ?Problem $problem): bool
+    private function searchAndAddValidator(ZipArchive $zip, array $zipEntries, ?array &$messages, string $externalId, string $validationMode, ?Problem $problem): bool
```
nickygerritsen, line 910: We should update the docblock with the type of this array

nickygerritsen (approved):

PR #3212, opened 2025-11-23
Check dependencies for released versions

vmcj:


.github/dependabot.yml
```diff
@@ -44,6 +44,15 @@ updates:
     allow:
         - dependency-type: "all"
     target-branch: "8.2"
+  - package-ecosystem: "composer"
+    directory: "/"
```
Kevinjil: Some of your PRs use `/`, whilst others use `/webapp`
https://github.com/DOMjudge/domjudge/pull/3218/files#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R4

vmcj: I think we can now move to `/webapp` but only thought of that today.

Kevinjil (approved):

PR #3213, opened 2025-11-23
Fix doc acls for inplace installs

vmcj:
Needs another `.gitignore` entry and I'm not sure if we need to set an ACL to the symlink but this does work on my setup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -262,6 +262,7 @@ inplace-install-l:
 	@echo "        setfacl    -m   u:$(WEBSERVER_GROUP):x    $(CURDIR)/doc/manual/build"
 	@echo "        setfacl -R -m d:u:$(WEBSERVER_GROUP):rx   $(CURDIR)/doc/manual/build/html"
 	@echo "        setfacl -R -m   u:$(WEBSERVER_GROUP):rx   $(CURDIR)/doc/manual/build/html"
+	@echo "        setfacl -R -m   u:$(WEBSERVER_GROUP):r    $(CURDIR)/doc/manual/build/domjudge-team-manual.pdf"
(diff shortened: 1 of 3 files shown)


Kevinjil:
What does the `Needs another .gitignore entry` refer to, as no `.gitignore` are committed?

vmcj:
> What does the `Needs another .gitignore entry` refer to, as no `.gitignore` are committed?

That I still need to add that.

Makefile
```diff
@@ -292,6 +301,15 @@ inplace-postinstall-permissions:
 	setfacl    -m   u:$(WEBSERVER_GROUP):r    $(CURDIR)/etc/verdicts.php
 	setfacl -R -m d:u:$(WEBSERVER_GROUP):rx   $(CURDIR)/webapp
 	setfacl -R -m   u:$(WEBSERVER_GROUP):rx   $(CURDIR)/webapp
+	setfacl    -m d:u:$(WEBSERVER_GROUP):x    $(CURDIR)/doc
+	setfacl    -m   u:$(WEBSERVER_GROUP):x    $(CURDIR)/doc
+	setfacl    -m d:u:$(WEBSERVER_GROUP):x    $(CURDIR)/doc/manual
+	setfacl    -m   u:$(WEBSERVER_GROUP):x    $(CURDIR)/doc/manual
+	setfacl    -m d:u:$(WEBSERVER_GROUP):x    $(CURDIR)/doc/manual/build
+	setfacl    -m   u:$(WEBSERVER_GROUP):x    $(CURDIR)/doc/manual/build
```
eldering, line 309: These should not be needed since the one below is recursive.

PR #3214, opened 2025-11-23
Document how to upgrade the dependencies

vmcj:


doc/manual/develop.rst
```diff
@@ -187,9 +187,17 @@ Loading development fixture data
 To debug failing Unit tests the fixtures can be loaded with:
 ``./webapp/bin/console domjudge:load-development-data SampleSubmissionsFixture`` in the current database.
 
+Updating and upgrading dependencies
+-----------------------------------
+To upgrade the PHP dependencies run:
+``cd ./webapp; composer update``.
```
meisterT: Is there also a way to update a single dependency (and perhaps its transitive deps) instead of everything?

nickygerritsen: I think we should add both indeed (`composer update package/name` and `composer update -W package/name`)

vmcj: Please check the text now. I've now also added `composer require` as that is probably the more important command for new contributors.

nickygerritsen (approved):

PR #3221, opened 2025-11-25
Reencode submission source in UTF-8 when displaying in monaco

### LINKED ISSUE #3217: 500 Internal Error when jury view code with non-UTF-8 encoding code.
Jury can not view code with non-UTF-8 encoding code (and diff with it)

`The string to escape is not a valid UTF-8 string in "jury/partials/submission_diff.html.twig" at line 72.`

### Your environment

- DOMjudge version 9.0 or later (with Monaco Editor)
- Debian 13
- Nginx

### Steps to reproduce
1. Team or jury submit a code with non-UTF-8 encoded source code. (This may happen in some local contest, some team may type some non-ascii character in their source code)
2. Jury view the code.

### Expected behaviour
It should be shown either in the correct encoding or force display in UTF-8 encoding.
At the very least, it shouldn't be a 500 error; the code can't be displayed at all.

### Actual behaviour
Jury can view the code.

I have repro this in demoweb now. See https://www.domjudge.org/demoweb/jury/submissions/9677/source 
It is also happened in our last weekend ICPC Regional Contest (only one jury test submission with Chinese GBK-encoded comment in his code and submit from his Windows laptop), and I have a db-dump.
If more information is needed, I can provide it via slack DM.

Kevinjil:
Try to reencode the submission source if it currently is not valid in UTF-8. Add some extra safeguard by forcing another sanity check reencode from UTF-8 to UTF-8, which filters out binary blobs.

**Before:**
image

**After:**
image

Blobs still filtered out:
image

Fixes #3217

### DIFF
--- a/webapp/src/Service/CheckConfigService.php
+++ b/webapp/src/Service/CheckConfigService.php
@@ -52,6 +52,7 @@ class CheckConfigService
             'php_version' => $this->checkPhpVersion(),
             'php_extensions' => $this->checkPhpExtensions(),
             'php_settings' => $this->checkPhpSettings(),
+            'mbstring_settings' => $this->checkMbstringSettings(),
             'mysql_settings' => $this->checkMysqlSettings(),
         ];
 
@@ -192,6 +193,23 @@ class CheckConfigService
         $this->stopwatch->stop(__FUNCTION__);
         return new ConfigCheckItem(
             caption: 'PHP settings',
+            result: $result,
+            desc: $desc
+        );
+    }
+
+    public function checkMbstringSettings(): ConfigCheckItem
+    {
+        $this->stopwatch->start(__FUNCTION__);
+
+        $desc = "\nFor submission source code displaying, the detect order should at least contain all expected submission encodings:\n";
+        $desc .= sprintf("  - `mbstring.detect_order` should at least contain `UTF-8` (now set to `%s`).\n", implode(', ', mb_detect_order()));
+        $desc .= sprintf("  - Supported encodings: `%s`.\n", implode(', ', mb_list_encodings()));
+        $result = in_array('UTF-8', mb_detect_order()) ? 'O' : 'W';
+
+        $this->stopwatch->stop(__FUNCTION__);
+        return new ConfigCheckItem(
+            caption: 'PHP multibyte string settings',
             result: $result,
             desc: $desc
         );


meisterT (approved):

webapp/src/Utils/Utils.php
```diff
@@ -1053,6 +1053,27 @@ public static function jsonEncode(mixed $data): string
...
+     *                can be reencoded correctly, an error message otherwise.
+     */
+    public static function reencodeUtf8(string $source): string {
+        $encoding = mb_detect_encoding($source, mb_list_encodings(), strict: true);
+        if ($encoding !== false) {
+            if ($encoding !== 'UTF-8') {
+                $source = mb_convert_encoding($source, 'UTF-8', $encoding);
+            }
+            // Some binary files are strictly valid in an encoding but fail to reencode correctly.
+            // A successive strict call to mb_detect_encoding still sais it is validly encoded, but rendering the string fails.
```
vmcj:
```suggestion
            // A successive strict call to mb_detect_encoding still says it is validly encoded, but rendering the string fails.
```

vmcj (changes requested):
Let's discuss this in person, this might be a proper fix but I'm not sure if we should change submissions for participants (out of principle)

So mostly blocking merging now.

meisterT:
This is not changing submissions. It is changing the display of submissions. The submissions it changes would not be displayed previously, see bug and screenshot

meisterT:
This is not changing submissions. It is changing the display of submissions. The submissions it changes would not be displayed previously, see bug and screenshot

Kevinjil:
> Let's discuss this in person

As we'll meet this week anyway, I don't see a reason not to discuss in person so I'll wait with merging.

vmcj:
> This is not changing submissions. It is changing the display of submissions. The submissions it changes would not be displayed previously, see bug and screenshot

Wrongly read this as that we do this on import.

Kevinjil:
I'll postpone merging this until after a small in-person discussion about https://github.com/DOMjudge/domjudge/issues/3217#issuecomment-3576512283

webapp/src/Utils/Utils.php
```diff
@@ -1053,6 +1053,27 @@ public static function jsonEncode(mixed $data): string
...
+     * @return string The reencoding if the source file is strictly valid in any of the supported encodings and
+     *                can be reencoded correctly, an error message otherwise.
+     */
+    public static function reencodeUtf8(string $source): string {
+        $encoding = mb_detect_encoding($source, mb_list_encodings(), strict: true);
+        if ($encoding !== false) {
+            if ($encoding !== 'UTF-8') {
+                $source = mb_convert_encoding($source, 'UTF-8', $encoding);
+            }
+            // Some binary files are strictly valid in an encoding but fail to reencode correctly.
```
vmcj:
```suggestion
            // Some binary files are strictly valid in an encoding but fail to re-encode correctly.
```

webapp/src/Service/CheckConfigService.php
```diff
@@ -189,6 +189,14 @@ public function checkPhpSettings(): ConfigCheckItem
                     (isset($sizes[$var]) ? Utils::printsize($sizes[$var]) : "unlimited"));
         }
 
+        $desc .= "\nFor submission encoding detection, the detect order should atcontain all expected submission encodings:\n";
```
vmcj: Wat wil je hier zeggen?

webapp/src/Service/CheckConfigService.php
```diff
@@ -189,6 +189,12 @@ public function checkPhpSettings(): ConfigCheckItem
                     (isset($sizes[$var]) ? Utils::printsize($sizes[$var]) : "unlimited"));
         }
 
+        $desc .= "\nFor submission encoding detection, the detect order should at least contain all expected submission encodings:\n";
```
vmcj: Less true but maybe more useful,

```suggestion
        $desc .= "\nFor submission source code displaying, the detect order should at least contain all expected submission encodings:\n";
```

webapp/src/Utils/Utils.php
```diff
@@ -1053,6 +1053,26 @@ public static function jsonEncode(mixed $data): string
         return json_encode($data, JSON_PRESERVE_ZERO_FRACTION | JSON_UNESCAPED_SLASHES | JSON_THROW_ON_ERROR);
     }
 
+    /**
+     * @return string The re-encoding if the source file is strictly valid in any of the supported encodings and
+     *                can be re-encoding correctly, an error message otherwise.
```
vmcj:
```suggestion
     * @return string The re-encoding if the source file is strictly valid in any of the supported encodings and
     *                can be re-encoding correctly, an error message otherwise.
```
Line might be too long? Also I think that return type is simple enough to not document.

webapp/src/Utils/Utils.php
```diff
@@ -1053,6 +1053,26 @@ public static function jsonEncode(mixed $data): string
...
+            $encoded = $encoding !== 'UTF-8' ? mb_convert_encoding($source, 'UTF-8', $encoding) : $source;
+            // Some binary files are strictly valid in an encoding but fail to re-encoding correctly.
+            // A successive strict call to mb_detect_encoding still says it is validly encoded, but rendering the string fails.
+            // This will filter these files.
+            $sanity = mb_convert_encoding($encoded, $encoding, 'UTF-8');
+            if ($source === $sanity) {
+                return $encoded;
+            }
+        }
+        return "Could not display file as UTF-8.\nCheck the supported PHP mbstring encodings if you did not expect this file to be binary.";
```
vmcj:
```suggestion
        return "Could not display file as UTF-8.\nCheck the supported PHP mbstring encodings (/jury/page/where/I/check/this) if you did not expect this file to be binary.";
```

Kevinjil: Agreed not adding this link IRL

PR #3227, opened 2025-11-27
Display clarifications when queue has been removed

vmcj:


webapp/templates/jury/export/clarifications.html.twig
```diff
@@ -45,7 +45,11 @@
             <div class="page-break"></div>
             <h1>{{ title }}</h1>
         {% endif %}
-        <h2>{{ queues[queue] }}</h2>
+        {% if queues[queue] is defined %}
+            <h2>{{ queues[queue] }}</h2>
+        {%  else %}
+            <h2><i>Unknown queue: {{ queue }}</h2>
```
meisterT: `<i>` tag never closes

PR #3229, opened 2025-11-27
Don't hardcode adminer DB name

nickygerritsen:
While we are at it, update it as well


vmcj:
In the future we should find a way to directly get this with composer (https://packagist.org/packages/vrana/adminer).

nickygerritsen:
> In the future we should find a way to directly get this with composer ([packagist.org/packages/vrana/adminer](https://packagist.org/packages/vrana/adminer)).

I tried but it's not really made for that it seems.

nickygerritsen:
@vmcj his idea: require the composer version and on post install / update cmd compile it.

PR #3232, opened 2025-11-27
Do expose contests before start in the API

### LINKED ISSUE #3166: Active, not started contests not exposed in the API
Since e2eaca5abe782878f2201b4cae615eae0a42f4d9 we don't expose contests that are active but not started yet anymore in the API.

We did this because for some entities we don't want to expose them before contest start, but for the contest itself we do.

We had some discussion and proposals in Slack but no resolution yet, so I'm creating an issue so we can track it
.

nickygerritsen:
Fixes #3166.

Also return 404 for problemset before contest start instead of 403.

### DIFF
--- a/webapp/tests/Unit/Controller/API/ContestControllerTest.php
+++ b/webapp/tests/Unit/Controller/API/ContestControllerTest.php
@@ -2,6 +2,7 @@
 
 namespace App\Tests\Unit\Controller\API;
 
+use App\DataFixtures\Test\DemoPreStartContestFixture;
 use App\Entity\Contest;
 
 class ContestControllerTest extends BaseTestCase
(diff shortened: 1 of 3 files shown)


webapp/tests/Unit/Controller/API/ContestControllerTest.php
```diff
@@ -26,4 +27,48 @@ class ContestControllerTest extends BaseTestCase
...
+        // Find the demo contest in the response
+        $foundContest = null;
+        foreach ($objects as $contest) {
+            if ($contest['shortname'] === 'demo') {
+                $foundContest = $contest;
+                break;
+            }
+        }
+
+        self::assertNotNull($foundContest, 'Demo contest should be visible after activation but before start');
```
meisterT:
```suggestion
        self::assertNotNull($foundContest, 'Demo contest should be visible after activation even before start');
```

meisterT (approved):

Issue #3024, opened 2025-07-06
Use external IDs in web interface everywhere


nickygerritsen:
We discussed that it makes sense to not expose internal IDs anymore. For this, we need to use the external IDs in URLs and tables and not display internal IDs anymore.

This also means we need to find a way to make `j/k` keyboard shortcuts work. My current idea is to add some data to the head of a page about what the next and previous submission IDs are. This would also fix that you can currently go beyond the last ID.

nickygerritsen:
Some things discussed on Slack:

* We will not display internal IDs anywhere anymore.
* We will use the external ID in URLs and display it in the UI. We will call it ID in the UI, not external ID, since it is the de facto unique ID.
* The external ID will be a unique column in the database.
  * Note that for submissions and clarifications this needs to be unique within a contest.
* Note that judgements and runs are special: they will have our IDs shown, since we have a separate table for external judgements and runs.
* We will change the primary key of the language table to be an auto increment integer, rather than the string we have now.

vmcj:
>     * Note that judgements and runs are special: they will have our IDs shown, since we have a separate table for external judgements and runs.
> 
Isn't our ID the external also? And why shouldn't we merge the 2 tables (in the future)?

nickygerritsen:
> > ```
> > * Note that judgements and runs are special: they will have our IDs shown, since we have a separate table for external judgements and runs.
> > ```
> 
> Isn't our ID the external also?

It is not because we have two tables.

> And why shouldn't we merge the 2 tables (in the future)?

As far as I remember we choose this for the reason that having it in one table makes it very complex: imagine we are in shadow mode and we have 10 local submissions and 500 remote submissions of which we already judged 400 ourself. Then what should the `/judgements` endpoint return? With 2 separate tables we made the decision to only return external judgements. We could do the same if we merge the tables, but that means we pre-create all runs as soon as the external system sends them to us and we need to match them when we create a run ourself. Or we have double data (one row for the external run, one for our own), but then having two tables seems better.

nickygerritsen:
Working on this in https://github.com/nickygerritsen/domjudge/tree/external-id-in-ui. First change is making langid integer.

vmcj:
> > > ```
> > > * Note that judgements and runs are special: they will have our IDs shown, since we have a separate table for external judgements and runs.
> > > ```
> > 
> > 
> > Isn't our ID the external also?
> 
> It is not because we have two tables.
> 
[...]

PR #3237, opened 2025-11-28
Fixes judgehost down notification

tuupke:
The route requires the judgehostid as route parameter, not the hostname.

Currently clicking the notification results in the screen shown below. Arguably this should be a 4xx error, but then it should be fixed within all route specifications by appending `<\d+>` to the parameter name. e.g.:

```
#[Route(path: '/{judgehostid}', methods: ['GET'], name: 'jury_judgehost')]
to 
#[Route(path: '/{judgehostid<\d+>}', methods: ['GET'], name: 'jury_judgehost')]
```
image

### DIFF
--- a/webapp/src/Controller/Jury/JudgehostController.php
+++ b/webapp/src/Controller/Jury/JudgehostController.php
@@ -230,7 +230,7 @@ class JudgehostController extends BaseController
     /**
      * @throws NonUniqueResultException
      */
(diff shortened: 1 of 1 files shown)


tuupke:
@nickygerritsen is changing things in #3024 to use external ids (strings) everywhere but here. So will add that change here too.

PR #3238, opened 2025-11-28
Return 404 if not yet access to submission data

Kevinjil:
Note that we do this check before verifying the existence of team or problem, s.t. we do not leak any information.

Fixes Sentry issue 7070995611

### DIFF
--- a/webapp/src/Controller/PublicController.php
+++ b/webapp/src/Controller/PublicController.php
@@ -330,6 +330,11 @@ class PublicController extends BaseController
         }
 
         $scoreboard = $this->scoreboardService->getScoreboard($contest);
+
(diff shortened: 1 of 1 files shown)


webapp/src/Controller/PublicController.php
```diff
@@ -330,6 +330,11 @@ public function submissionsDataAction(Request $request, ?string $teamId, ?string
         }
 
         $scoreboard = $this->scoreboardService->getScoreboard($contest);
+
+        if (!$scoreboard) {
```
tuupke: Why check for any falsy value?

Kevinjil: Good question

PR #3240, opened 2025-11-28
Symfony 7.4 upgrade

nickygerritsen:


.github/jobs/composer_setup.sh
```diff
@@ -8,6 +8,7 @@ section_start "Configure PHP"
 PHPVERSION=$(php -r 'echo PHP_MAJOR_VERSION.".".PHP_MINOR_VERSION."\n";')
 export PHPVERSION
 echo "$PHPVERSION" | tee -a "$ARTIFACTS"/phpversion.txt
+sudo apt-get -y remove "php$PHPVERSION-redis"
```
vmcj: Probably purge it in this case?

.github/jobs/composer_setup.sh
```diff
@@ -8,6 +8,7 @@ section_start "Configure PHP"
 PHPVERSION=$(php -r 'echo PHP_MAJOR_VERSION.".".PHP_MINOR_VERSION."\n";')
 export PHPVERSION
 echo "$PHPVERSION" | tee -a "$ARTIFACTS"/phpversion.txt
+sudo apt-get -y purge "php$PHPVERSION-redis"
```
meisterT: add a comment why this is required

.github/jobs/detect_dump.sh
```diff
@@ -8,6 +8,7 @@ OUT=$(find ./ -name ".git*" -type d -prune -o \
               -name "bundles" -prune -o \
               -name "cache" -type d -prune -o \
               -name "ace" -type d -prune -o \
+              -path "./webapp/config/reference.php" -prune -o \
```
meisterT, line 11: What is this and why are we doing this?

webapp/phpstan.dist.neon
```diff
@@ -2,9 +2,8 @@ parameters:
     level: 6
     paths:
         - src
+        - migrations
         - tests
-    excludePaths:
-        - src/Utils/Adminer.php
```
meisterT, line 7: Why are you deleting this?

webapp/src/Service/ImportProblemService.php
```diff
@@ -468,7 +468,7 @@ public function importZippedProblem(
             }
             if ($numCases > 0) {
                 $messages['info'][] = sprintf("Added/updated %d %s testcase(s): {%s}.{in,ans}",
-                    $numCases, $type, join(',', $testcaseNames));
+                    $numCases, $type, implode(',', $testcaseNames));
```
meisterT: `join` is the much better name for the function, is it going away?

.github/jobs/baseinstall.sh
```diff
@@ -4,7 +4,7 @@
 
 export version="$1"
 db=${2:-install}
-phpversion="${3:-8.1}"
+phpversion="${3:-8.2}"
```
meisterT, line 7: Perhaps move this commit (and I think there were other similar changes in other commits) to a separate PR?

webapp/src/DependencyInjection/Compiler/DoctrineNativeLazyObjectsPass.php
```diff
@@ -0,0 +1,30 @@
...
+namespace App\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class DoctrineNativeLazyObjectsPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        if (PHP_VERSION_ID < 80400) {
```
meisterT, line 13: Add a comment?

.github/jobs/detect_dump.sh
nickygerritsen: See https://symfony.com/blog/new-in-symfony-7-4-better-php-configuration, it is an auto generated file for PHPstan/IDE basically to get code completion in the new format Symfony is moving towards (but we can't use yet). It will be generated everytime and they strongly advice to NOT put it in gitignore.

webapp/phpstan.dist.neon
nickygerritsen: I will add a comment in the new place, but basically PHPStan got very angry with how we did it before since it can't load adminer.

webapp/src/Service/ImportProblemService.php
nickygerritsen: https://www.php.net/join not really, will remove this rule.

.github/jobs/baseinstall.sh
nickygerritsen: SGTM

nickygerritsen: See #3280

.github/jobs/composer_setup.sh
vmcj: I think the new CI image doesn't have this anymore.

webapp/templates/jury/external_contest_manage.html.twig
```diff
@@ -41,7 +41,7 @@
             var typeChanged = function () {
                 var ccsApiFields = $('[data-type-ccs-api]');
                 switch (type.val()) {
-                    case '{{ constant('\App\\Entity\\ExternalContestSource::TYPE_CCS_API') }}':
+                    case '{{ constant('\\App\\Entity\\ExternalContestSource::TYPE_CCS_API') }}':
```
vmcj, line 44:
```suggestion
                    case '{{ constant('\\App\\Entity\\ExternalContestSource::TYPE_CCS_API') }}':
```
Can we use an `use` here?

webapp/src/Utils/Adminer.php
```diff
@@ -1,80 +0,0 @@
-<?php declare(strict_types=1);
```
vmcj, line 1: This shows as deleted in the GH UI, is it intended to be removed?

.github/jobs/composer_setup.sh
nickygerritsen: I'll remove it

webapp/resources/functions.php
```diff
@@ -1,10 +1,86 @@
...
+            {
+                return "DOMjudge DB Admin";
+            }
+
+            public function database(): string
+            {
+                return $this->getDatabaseCredentials()['db'];
+            }
+
+            public function databases($flush = true): array
```
vmcj: Why don't we need annotations for PHPstan here?

webapp/src/Twig/TwigGlobalsExtension.php
```diff
@@ -0,0 +1,99 @@
...
+        protected readonly TokenStorageInterface $tokenStorage,
+        protected readonly AuthorizationCheckerInterface $authorizationChecker,
+        protected readonly RouterInterface $router,
+        protected readonly SerializerInterface $serializer,
+        #[Autowire('%kernel.project_dir%')]
+        protected readonly string $projectDir,
+        protected array $renderedSources = []
+    ) {}
+
+    public function getGlobals(): array
```
vmcj, line 44: Why don't we need the annotation here?

webapp/src/Doctrine/DBAL/Types/TinyIntType.php
```diff
@@ -33,13 +28,13 @@ public function convertToPHPValue($value, AbstractPlatform $platform): ?int
...
     }
 
-    public function getBindingType(): int
+    public function getBindingType(): ParameterType
     {
         return ParameterType::INTEGER;
     }
 
-    public function requiresSQLCommentHint(AbstractPlatform $platform): bool
+    public function getMappedDatabaseTypes(AbstractPlatform $platform): array
```
vmcj, line 36: Can we annotate this?

webapp/config/packages/web_profiler.yaml
```diff
@@ -14,4 +14,6 @@ when@test:
         intercept_redirects: false
 
     framework:
-        profiler: { collect: false }
+        profiler:
+            collect: false
+            collect_serializer_data: true
```
vmcj, line 21: Why do we collect this but not the other thing? Can you add a comment for it?

webapp/templates/jury/external_contest_manage.html.twig
nickygerritsen: We can't use use in Twig.

webapp/src/Utils/Adminer.php
nickygerritsen: Tobi asked the same and the answer is both yes and answered in `functions.php`.

webapp/resources/functions.php
nickygerritsen: Because we ignore the functions file. I can try to enable it though.

webapp/src/Twig/TwigGlobalsExtension.php
nickygerritsen: I have no clue.

webapp/config/packages/web_profiler.yaml
nickygerritsen: I can add a comment, which boils down to:
```
not collecting serialized data is deprecated, so we enable it.
```

.github/jobs/composer_setup.sh
nickygerritsen: I did now remove it, which broke CI. I assume you'll look into it.

webapp/resources/functions.php
nickygerritsen: Not ignoring this file gives many many many errors about not found classes and functions. I will add comments but not check the file after.

nickygerritsen: Found a nice way, will update

webapp/src/Twig/TwigGlobalsExtension.php
nickygerritsen: Because the parent already has it and it knows.

webapp/src/Doctrine/DBAL/Types/TinyIntType.php
nickygerritsen: We can but the parent already has the annotation

.github/jobs/composer_setup.sh
nickygerritsen: I found a better way with `--ingore-platform-reqs`

webapp/resources/functions.php
```diff
@@ -1,10 +1,87 @@
...
+use function Adminer\lang;
 
 if (!function_exists('adminer_object')) {
-    function adminer_object()
+    function adminer_object(): Adminer
     {
-        return new Adminer();
+        // Use anonymous class to extend Adminer with our own settings.
+        // We can't use a normal class since the base class only exists when
+        // the adminer code is include in JuryMiscController.
```
vmcj:
```suggestion
        // the adminer code is included in JuryMiscController.
```

webapp/resources/functions.php
```diff
@@ -1,10 +1,87 @@
...
+
+            public function database(): string
+            {
+                return $this->getDatabaseCredentials()['db'];
+            }
+
+            /**
+             * @return list<string>
+             */
+            public function databases(bool $flush = true): array
```
vmcj, line 29: We don't use this parameter, I assume we have a reason for that but need to overwrite the function from the default file? If so we should leave a comment about it.

webapp/resources/functions.php
```diff
@@ -1,10 +1,87 @@
...
+             * @return array{string, string, string}
+             */
+            public function credentials(): array
+            {
+                ['host' => $host, 'user' => $user, 'pass' => $pass] = $this->getDatabaseCredentials();
+
+                return [$host, $user, $pass];
+            }
+
+            public function login(string $login, string $password): bool
```
vmcj, line 47: Same for here,

vmcj (approved):
Some small comments, but best way of testing this is just merging it IMO

vmcj:
Your PR fails on a test, normally we would increase the deprecation counters (which I think is valid) but before you do that you should look at:
```
   6x: Since symfony/framework-bundle 7.3: Not setting the "framework.profiler.collect_serializer_data" config option to "true" is deprecated.
    2x in DocsPageControllerTest::setUp from App\Tests\Unit\Controller\Team
    2x in DocsPageControllerTest::tearDown from App\Tests\Unit\Controller\Team
    1x in TeamAffiliationControllerTest::provideAddCorrectEntities from App\Tests\Unit\Controller\Jury
    1x in QueuetaskIntegrationTest::setUp from App\Tests\Unit\Integration
```
as I think we had a discussion about this that you do this so it is probably not picked up correctly.

nickygerritsen:
> Your PR fails on a test, normally we would increase the deprecation counters (which I think is valid) but before you do that you should look at:
> 
> ```
>    6x: Since symfony/framework-bundle 7.3: Not setting the "framework.profiler.collect_serializer_data" config option to "true" is deprecated.
>     2x in DocsPageControllerTest::setUp from App\Tests\Unit\Controller\Team
>     2x in DocsPageControllerTest::tearDown from App\Tests\Unit\Controller\Team
>     1x in TeamAffiliationControllerTest::provideAddCorrectEntities from App\Tests\Unit\Controller\Jury
>     1x in QueuetaskIntegrationTest::setUp from App\Tests\Unit\Integration
> ```
> 
> as I think we had a discussion about this that you do this so it is probably not picked up correctly.

yeah my setting went away, now it's back

PR #3248, opened 2025-11-29
Make error messages in the submit client stand out more prominently.

meisterT:
Before:

image

------

After with bright background:

image

------

After with dark background:

image

### DIFF
--- a/submit/submit
+++ b/submit/submit
@@ -60,7 +60,7 @@ def warn_user(msg: str) -> None:
 
 
 def usage(msg: str) -> NoReturn:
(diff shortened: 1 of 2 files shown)


nickygerritsen (approved):

vmcj:
Follow up should be telling which problems are there.. so if the label is not `A` but `Alphanumerical` you still know what to pass.

meisterT:
> Follow up should be telling which problems are there.. so if the label is not `A` but `Alphanumerical` you still know what to pass.

The recommended `--help` does that

PR #3254, opened 2025-11-29
Diff selection using keyboard shortcuts

Kevinjil:


webapp/public/js/domjudge.js
```diff
@@ -1010,7 +1010,7 @@ function initializeKeyboardShortcuts() {
                 parts[parts.length - 1] += '?' + params[1];
             }
             window.location = parts.join('/');
-        } else if (!ignore && (key === 's' || key === 't' || key === 'p' || key === 'j' || key === 'c')) {
+        } else if (!ignore && (key === 'd' || key === 's' || key === 't' || key === 'p' || key === 'j' || key === 'c')) {
```
meisterT: Perhaps it's time to make this more readable by doing something like this?

```
const validKeys = ['c', 'd', 'j', 'p', 's', 't'];

if (!ignore && validKeys.includes(key)) {
...
```

nickygerritsen: My AI made, for external IDs, this beauty:

```
if (/^[a-zA-Z0-9_.-]$/.test(e.key)) {
```

We an do something similar likeL
```
if (/^[cdjpst]$/.test(e.key)) {
```

meisterT (approved):

PR #3263, opened 2025-11-30
Use external ID in UI everywhere instead of internal database ID

### LINKED ISSUE #3024: Use external IDs in web interface everywhere

nickygerritsen:
Fixes #3024


judge/judgedaemon.main.php
```diff
@@ -624,7 +624,7 @@ private function checkDiskSpace(string $workdirpath): void
 
     private function judgingDirectory(string $workdirpath, array $judgeTask): string
     {
-        if (filter_var($judgeTask['submitid'], FILTER_VALIDATE_INT) === false ||
+        if (filter_var($judgeTask['submitid'], FILTER_VALIDATE_REGEXP, ['options' => ['regexp' => '/^[a-zA-Z0-9_.-]+$/']]) === false ||
```
meisterT: should we define this somewhere more generally?

meisterT: Also, what do submission URLs look like after this change?

webapp/migrations/Version20251024122021.php
```diff
@@ -0,0 +1,141 @@
...
+        $this->addSql('ALTER TABLE contestlanguage DROP FOREIGN KEY FK_ADCB43232271845');
+
+        $this->addSql('ALTER TABLE language ADD langid_str VARCHAR(32) NOT NULL AFTER langid');
+        $this->addSql('ALTER TABLE contestlanguage ADD langid_str VARCHAR(32) NOT NULL AFTER langid');
+        $this->addSql('ALTER TABLE problemlanguage ADD langid_str VARCHAR(32) NOT NULL AFTER langid');
+        $this->addSql('ALTER TABLE submission ADD langid_str VARCHAR(32) DEFAULT NULL AFTER langid');
+        $this->addSql('ALTER TABLE version ADD langid_str VARCHAR(32) DEFAULT NULL AFTER langid');
+
+        $this->addSql("UPDATE language SET langid_str = CASE externalid
+            WHEN 'ada' THEN 'adb'
```
meisterT, line 93: huh, ada -> adb? Is that what you want?

webapp/src/Controller/Jury/ClarificationController.php
```diff
@@ -194,14 +205,14 @@ public function viewAction(Request $request, int $id): Response
...
+                    $concernssubject = $contest->getExternalid() . "|" . $clar->getProblem()->getExternalid();
                 } else {
                     // Very special case, this problem is unlinked.
                     $concernssubject = "";
                 }
-                $data['subjectlink'] = $this->generateUrl('jury_problem', ['probId' => $clar->getProblem()->getProbid()]);
+                $data['subjectlink'] = $this->generateUrl('jury_problem', ['probId' => $clar->getProblem()->getExternalid()]);
             } elseif ($clar->getCategory()) {
-                $concernssubject = $contest->getCid() . "-" . $clar->getCategory();
+                $concernssubject = $contest->getExternalid() . "#" . $clar->getCategory();
```
meisterT: why not also use `|`?

webapp/src/Controller/Jury/TeamAffiliationController.php
```diff
@@ -208,20 +208,20 @@ public function deleteAction(Request $request, int $affilId): Response
     }
 
     #[IsGranted('ROLE_ADMIN')]
-    #[Route(path: '/delete-multiple', name: 'jury_team_affiliation_delete_multiple', methods: ['GET', 'POST'])]
+    #[Route(path: '/delete-multiple', name: 'jury_team_affiliation_delete_multiple', methods: ['GET', 'POST'], priority: 1)]
```
meisterT, line 211: what does `priority: 1` mean/do here?

webapp/src/Controller/BaseController.php
```diff
@@ -649,4 +665,120 @@ protected function processAddFormForExternalIdEntity(
...
+     *
+     * @param class-string                $entityClass     Entity class to query
+     * @param mixed                       $currentIdValue  Current value of the ID field
+     * @param string                      $idField         Field to return as the ID (e.g., 'externalid', 'submitid')
+     * @param array<string, 'ASC'|'DESC'> $orderBy         Sort criteria as field => direction (e.g., ['e.submittime' => 'ASC', 'e.submitid' => 'ASC'])
+     * @param bool                        $filterOnContest Whether to filter results by current contests
+     *
+     * @return array{previous: string|int|null, next: string|int|null}
+     */
+    protected function getPreviousAndNextObjectIds(
```
meisterT, line 680: Is it easy to add a few unit tests for this method?

webapp/templates/jury/analysis/contest_overview.html.twig
```diff
@@ -113,7 +113,7 @@ $(function() {
             </thead>
             <tbody>
             {% for p in problems %}
-              {% set id=p.probid %}
+              {% set id=p.externalId %}
```
meisterT, line 116: looking at all these places, should we just use `p.id` or `p.probid` to mean the external id and `p.internalId` to mean the probid? Could be done separately though

ChangeLog
meisterT, line 1: This might deserve a changelog entry :-)

judge/judgedaemon.main.php
nickygerritsen: What are you suggesting for generally defining this? The judgedaemon doesn't have access to the Symfony code.

As for urls, it's https://domjudge/jury/submissions/[id] still, but the [id] can be a string (although if we generate submission ID's they will actually be numbers, only when external systems generate them they will be strings).

webapp/migrations/Version20251024122021.php
nickygerritsen: Yes, we used `adb` as langid for ADA files, but the CLICS spec uses `ada`.

webapp/templates/jury/analysis/contest_overview.html.twig
nickygerritsen: I wonder if that is better. It will have less changes in the twigs but then have other changes instead. Not sure I actually like it, but we can try it separately indeed.

webapp/src/Controller/Jury/ClarificationController.php
nickygerritsen: Because then you can't have a problem with the same ID as a clarification category

webapp/src/Controller/Jury/TeamAffiliationController.php
nickygerritsen: It makes it that `jury/affiliations/delete-multiple` goes to this route and not suggests you want to view the affiliation called `delete-multiple`. In the past we reordered the methods to have this come earlier than the view action, but I like this way more.

webapp/src/Controller/BaseController.php
nickygerritsen: I will try, I think that makes sense.

nickygerritsen: Done

webapp/templates/jury/analysis/contest_overview.html.twig
vmcj: I would prefer to not do this for DJ10 yet as this change is still new and @nickygerritsen might have missed one.

In my mental model `p.id` is the DB column and making it explicit that we filter on possible external input (or that we expose this to the outside) makes sense in the code to me.

judge/judgedaemon.main.php
```diff
@@ -1284,7 +1284,7 @@ private function compile(
 
             // Revoke readablity for domjudge-run user to this workdir.
             chmod($workdir, 0700);
-            logmsg(LOG_NOTICE, "Judging s$judgeTask[submitid], task $judgeTask[judgetaskid]: compile error");
+            logmsg(LOG_NOTICE, "Judging $judgeTask[submitid], task $judgeTask[judgetaskid]: compile error");
```
vmcj: Are you sure the `s` should be gone? There can be multiple judgings for 1 submitid right? So maybe this should be an `j` actually? Or do we now prefix the externalid with the type so `s`?

nickygerritsen: Since external IDs can be anything, I think we should not prepend an `s`. However, this indeed should say `Submission` I think, not `Judging`.

webapp/tests/Unit/Controller/Jury/UserControllerTest.php
```diff
@@ -28,7 +28,7 @@ class UserControllerTest extends JuryControllerTestCase
                                                            'plainPassword' => 'plainpassword',
                                                            'ipAddress'     => '10.0.0.0',
                                                            'enabled'       => '1',
-                                                           'team'          => '0',
+                                                           'team'          => '',
```
vmcj, line 31: Why don´t we test assigning to a team anymore?

webapp/tests/Unit/Controller/Jury/TeamControllerTest.php
```diff
@@ -18,17 +18,17 @@ class TeamControllerTest extends JuryControllerTestCase
...
     protected static array   $deleteEntities           = ['DOMjudge','Example teamname'];
     protected static string  $deleteEntityIdentifier   = 'name';
-    protected static string  $getIDFunc                = 'getTeamid';
+    protected static string  $getIDFunc                = 'getExternalid';
     protected static string  $className                = Team::class;
     protected static array   $DOM_elements             = ['h1' => ['Teams']];
     protected static string  $addForm                  = 'team[';
     protected static array   $addEntitiesShown         = ['icpcid', 'label', 'displayName', 'location'];
-    protected static array   $overviewSingleNotShown   = ['addUserForTeam'];
+    protected static array   $overviewSingleNotShown   = ['addUserForTeam', 'category'];
```
vmcj: Why is this suddenly needed? There is a selection for `category` for teams right?

webapp/tests/Unit/Controller/Jury/LanguagesControllerTest.php
```diff
@@ -34,53 +33,33 @@ class LanguagesControllerTest extends JuryControllerTestCase
                                                            'compileExecutable' => 'java_javac',
                                                            'extensions' => ['1' => 'extension'],
                                                            'filterCompilerFiles' => '1'],
-                                                          ['langid' => 'ext',
-                                                           'externalid' => 'diffext',
+                                                          ['externalid' => 'diffext',
```
vmcj: We don´t need this test anymore. as we test this case L39

webapp/tests/Unit/Controller/Jury/LanguagesControllerTest.php
```diff
@@ -34,53 +33,33 @@ class LanguagesControllerTest extends JuryControllerTestCase
...
                                                            'name' => 'External'],
-                                                          ['langid' => 'lang123_.-',
-                                                           'name' => 'langid_expected_chars'],
-                                                          ['langid' => 'external_expected_chars',
-                                                           'externalid' => 'ext123_.-'],
-                                                          ['langid' => 'name_special_chars',
-                                                           'name' => '🕑ড|{}()*'],
-                                                          ['langid' => 'entry',
-                                                           'requireEntryPoint' => '1',
+                                                          ['name' => 'langid_expected_chars'],
```
vmcj: When I wrote this I updated the `langid` each time to make debugging easier. Alternative is to change `langid` for `externalid` now IMO.

webapp/src/Controller/Jury/ClarificationController.php
```diff
@@ -230,10 +241,15 @@ public function viewAction(Request $request, int $id): Response
...
+        $parameters['previousNext'] = $this->getPreviousAndNextObjectIds(
+            Clarification::class,
+            $clarification->getExternalid(),
+        );
+
         return $this->render('jury/clarification.html.twig', $parameters);
     }
 
-    #[Route(path: '/send', name: 'jury_clarification_new')]
+    #[Route(path: '/send', name: 'jury_clarification_new', priority: 1)]
```
meisterT, line 252: I probably asked before, but what is this `priority: 1` for?

webapp/migrations/Version20251024122021.php
meisterT, line 1: did you test with say the NWERC database?

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -441,8 +441,9 @@ public function updateJudgingAction(
 
                     $judgingId = $judging->getJudgingid();
                     $contestId = $judging->getSubmission()->getContest()->getCid();
-                    $this->dj->auditlog('judging', $judgingId, 'judged',
-                                        'compiler-error', $judgehost->getHostname(), $contestId);
+                    $contestExernalid = $judging->getSubmission()->getContest()->getExternalid();
```
meisterT: nit: typo, missing a *t*

judge/judgedaemon.main.php
vmcj: Reading it again I think it should state "Judging"  as its judging the `Submission`.

judge/judgedaemon.main.php
vmcj: Can´t we have it in `lib.misc.php` so we can share it with the judgedaemon?

judge/judgedaemon.main.php
```diff
@@ -1565,7 +1565,7 @@ private function runTestcase(
             // What does the exitcode mean?
             if (!isset($this->EXITCODES[$retval])) {
                 alert('error');
-                error("Unknown exitcode ($retval) from testcase_run.sh for s$judgeTask[submitid]");
+                error("Unknown exitcode ($retval) from testcase_run.sh for $judgeTask[submitid]");
```
```suggestion
                error("Unknown exitcode ($retval) from testcase_run.sh for s:$judgeTask[submitid]");
```

webapp/migrations/Version20251024122021.php
```diff
@@ -0,0 +1,141 @@
...
+        // - First we drop the foreign keys from tables referencing langid. We also drop compound
+        // - Then we add a temporary integer langid column to all these tables, since we can't
+        //   change the langid column itself, because MySQL will try to convert the strings to
+        //   integers and fail. We set the langid column in the language table to auto increment,
+        //   so it's filled by MySQL.
+        // - Now we copy the langid_int values to other tables.
+        // - Then we drop the old langid columns and drop any (compound) primary keys that use it.
+        // - Next, we rename the langid_int columns back to langid.
+        // - Finally we add back all primary and foreign keys.
+
```
vmcj, line 23:
```suggestion
        // We need to do some juggling to get this to work:
        // - First we drop the foreign keys from tables referencing langid. We also drop compound
        ... SQL code
        // - Then we add a temporary integer langid column to all these tables, since we can't
        //   change the langid column itself, because MySQL will try to convert the strings to
        //   integers and fail. We set the langid column in the language table to auto increment,
        //   so it's filled by MySQL.
        ... SQL code
        // - Now we copy the langid_int values to other tables.
        // - Then we drop the old langid columns and drop any (compound) primary keys that use it.
        // - Next, we rename the langid_int columns back to langid.
        // - Finally we add back all primary and foreign keys.

```

This is very hard to verify if you did the right thing. I'll believe you it works but hope you ran this on an old contest and it worked?

webapp/migrations/Version20251024122021.php
```diff
@@ -0,0 +1,141 @@
...
+        $this->addSql('ALTER TABLE language ADD langid_int INT UNSIGNED AUTO_INCREMENT UNIQUE AFTER langid');
+        $this->addSql('ALTER TABLE contestlanguage ADD langid_int INT UNSIGNED NOT NULL AFTER langid');
+        $this->addSql('ALTER TABLE problemlanguage ADD langid_int INT UNSIGNED NOT NULL AFTER langid');
+        $this->addSql('ALTER TABLE submission ADD langid_int INT UNSIGNED DEFAULT NULL AFTER langid');
+        $this->addSql('ALTER TABLE version ADD langid_int INT UNSIGNED DEFAULT NULL AFTER langid');
+
+        $this->addSql('UPDATE contestlanguage c JOIN language l ON c.langid = l.langid SET c.langid_int = l.langid_int');
+        $this->addSql('UPDATE problemlanguage p JOIN language l ON p.langid = l.langid SET p.langid_int = l.langid_int');
+        $this->addSql('UPDATE submission s JOIN language l ON s.langid = l.langid SET s.langid_int = l.langid_int WHERE s.langid IS NOT NULL');
```
vmcj: Why can we have submissions with no language?

webapp/migrations/Version20251024122021.php
```diff
@@ -0,0 +1,141 @@
...
+        $this->addSql('ALTER TABLE contestlanguage CHANGE langid_int langid INT UNSIGNED NOT NULL COMMENT \'Language ID\'');
+        $this->addSql('ALTER TABLE problemlanguage CHANGE langid_int langid INT UNSIGNED NOT NULL COMMENT \'Language ID\'');
+        $this->addSql('ALTER TABLE submission CHANGE langid_int langid INT UNSIGNED DEFAULT NULL COMMENT \'Language ID\'');
+        $this->addSql('ALTER TABLE version CHANGE langid_int langid INT UNSIGNED DEFAULT NULL COMMENT \'Language ID\'');
+
+        $this->addSql('ALTER TABLE language ADD PRIMARY KEY (langid), DROP INDEX langid_int');
+        $this->addSql('ALTER TABLE contestlanguage ADD PRIMARY KEY (cid, langid)');
+        $this->addSql('ALTER TABLE problemlanguage ADD PRIMARY KEY (probid, langid)');
+
+        $this->addSql('ALTER TABLE submission ADD INDEX langid (langid), ADD CONSTRAINT submission_ibfk_4 FOREIGN KEY (langid) REFERENCES language (langid) ON DELETE CASCADE');
```
vmcj, line 69: Just a general note, I checked and indeed when we remove a language we remove the submissions but I wonder if we should actually do that. I think we should set it to `NULL` and provide an vacuum script so in case someone deletes with SQL they still have a way to recover.

webapp/templates/jury/clarification.html.twig
```diff
@@ -94,12 +91,12 @@
...
-                                    <form action="{{ path('jury_clarification_change_queue', {'clarId': clar.clarid }) }}" class="d-flex" method="post">
+                                <span class="clarification-queue-form" data-current-selected-queue="{{ clar.queueid }}" data-clarification-id="{{ clar.externalid }}">
+                                    <form action="{{ path('jury_clarification_change_queue', {'clarId': clar.externalid }) }}" class="d-flex" method="post">
                                         <select name="queue" class="queue form-select w-auto form-select-sm" id="qselect">
                                             <option value="unassigned">Unassigned issues</option>
                                             {% for qid,queue in queues %}
-                                                <option value="{{ qid }}"{% if clar.queueid == qid %} selected{% endif %}>{{ queue }}</option>
+                                                <option value="{{ qid }}"{% if clar.queue == qid %} selected{% endif %}>{{ queue }}</option>
```
meisterT: is `queue` the name here or the ID? In other words, does this work correctly?

webapp/src/Controller/BaseController.php
```diff
@@ -649,4 +665,120 @@ protected function processAddFormForExternalIdEntity(
...
+                    $parameters[$paramName] = $fieldValues[$field];
+                }
+
+                // Add the comparison for this column
+                $field = $fields[$i];
+                $direction = $directions[$i];
+                // For "previous": if ASC, we want < ; if DESC, we want >
+                // For "next": if ASC, we want > ; if DESC, we want <
+                $compOp = ($operator === 'previous')
+                    ? ($direction === 'ASC' ? '<' : '>')
```
vmcj, line 728: I assume PHPStan will warn us in case we ever feed `$orderBy = ['something' => 'asc']` so as lowercase instead of what you check now?

webapp/src/Controller/Jury/LanguageController.php
```diff
@@ -302,7 +303,7 @@ public function editAction(Request $request, string $langId): Response
             }
             $this->saveEntity($language, $language->getLangid(), false);
             if ($language->getAllowJudge()) {
-                $this->dj->unblockJudgeTasksForLanguage($langId);
+                $this->dj->unblockJudgeTasksForLanguage($language->getLangid());
             }
             return $this->redirectToRoute('jury_language', ['langId' => $language->getLangid()]);
```
meisterT: does this redirect still work (if it doesn't, it's likely that other redirects also need updating)

webapp/src/Controller/API/ProblemController.php
```diff
@@ -281,7 +281,7 @@ public function unlinkProblemAction(Request $request, string $id): Response
         }
 
         $this->em->remove($contestProblem);
-        $id = [$contestProblem->getCid(), $contestProblem->getProbid()];
+        $id = [$contestProblem->getExternalId(), $contestProblem->getExternalId()];
```
meisterT: This looks incorrect, because it is now twice the same ID (previously contest id and problem id)

webapp/src/Controller/API/ProblemController.php
```diff
@@ -354,7 +354,7 @@ public function linkProblemAction(
         $this->em->persist($contestProblem);
         $this->em->flush();
 
-        $fullId = [$contestProblem->getCid(), $contestProblem->getProbid()];
+        $fullId = [$contestProblem->getExternalId(), $contestProblem->getExternalId()];
```
meisterT: same comment as above

webapp/src/Service/DOMJudgeService.php
meisterT: this line has not been updated, does it need to be updated?

webapp/src/Controller/Jury/ClarificationController.php
```diff
@@ -43,12 +48,12 @@ public function indexAction(
...
         if ($contest = $this->dj->getCurrentContest()) {
-            $contestIds = [$contest->getCid()];
+            $contestIds = [$contest->getExternalid()];
         } else {
             $contestIds = array_keys($this->dj->getCurrentContests());
-            // cid -1 will never happen, but otherwise the array is empty and that is not supported.
+            // cid @ will never happen, but otherwise the array is empty and that is not supported.
             if (empty($contestIds)) {
-                $contestIds = [-1];
+                $contestIds = ['@'];
```
vmcj:
```suggestion
                $contestIds = [''];
```

In case `CLICS` ever decides that identifiers can start with `@` and we forget to change this code?

judge/judgedaemon.main.php
nickygerritsen: How about `Judging submission ...`?

webapp/src/Form/Type/RejudgingType.php
meisterT, line 203: do we need to add choice_value here as well as above?

meisterT, line 168: choice_value as above?

webapp/src/Controller/Jury/TeamController.php
```diff
@@ -292,12 +297,17 @@ public function viewAction(
                     'judgehost' => 'judgehost',
                     default => throw new BadRequestHttpException(sprintf('Restriction on %s not allowed.', $key)),
                 };
-                $restrictions->$key = is_numeric($value) ? (int)$value : $value;
+                $restrictionValue = match ($key) {
+                    'problemId' => $this->em->getRepository(Problem::class)->findByExternalId($value)->getProbid(),
```
meisterT: Can this be `null`? Also the line below?

webapp/src/Controller/Jury/ClarificationController.php
vmcj, line 252:
```suggestion
    #[Route(path: '/_send', name: 'jury_clarification_new', priority: 1)]
```

Our externalids might make sense, but what will the ones be when we shadow?

webapp/src/Controller/Jury/UserController.php
meisterT: does this need to be updated?

webapp/migrations/Version20251026092516.php
```diff
@@ -0,0 +1,66 @@
...
+        // Note: this migration only works for entities that still exist. For others, it will not work but there is nothing we can do
+        $this->addSql('ALTER TABLE auditlog CHANGE cid cid VARCHAR(255) DEFAULT NULL COMMENT \'External contest ID associated to this entry\', CHANGE dataid dataid VARCHAR(64) DEFAULT NULL COMMENT \'(External) identifier in reference table\'');
+        $this->addSql('UPDATE auditlog INNER JOIN contest ON CAST(contest.cid AS CHAR) = auditlog.cid SET auditlog.cid = contest.externalid');
+        foreach (static::ENTITIES as $table => $fields) {
+            if (!is_array($fields)) {
+                $fields = [$fields];
+            }
+            foreach ($fields as $field) {
+                $this->addSql(sprintf(
+                    'UPDATE auditlog INNER JOIN %s ON CAST(%s.%s AS CHAR) = auditlog.dataid AND auditlog.datatype = "%s" SET auditlog.dataid = %s.externalid',
```
meisterT, line 41: is this `INNER` join problematic? should it be a regular join?

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -842,15 +838,15 @@ public function sourceAction(
         }
 
         /** @var array{
-         *      submitid: int,
+         *      submitid: string,
          *      tag?: string
          * } otherSubmissions
```
meisterT: nit: `$otherSubmissions`

webapp/src/Controller/Jury/ClarificationController.php
```diff
@@ -294,38 +310,44 @@ public function toggleAnsweredAction(Request $request, int $clarId): Response
...
         if (!$clarification) {
-            throw new NotFoundHttpException(sprintf('Clarification with ID %d not found', $clarId));
+            throw new NotFoundHttpException(sprintf('Clarification with ID %s not found', $clarId));
         }
 
         $subject = $request->request->get('subject');
-        [$cid, $probid] = explode('-', $subject);
+        $problemId = null;
+        $category = null;
+        if (str_contains($subject, '#')) {
```
vmcj:
```suggestion
        const CLARIFICATION_CONSTANT = '#'
        const PROBLEM_CONSTANT = '#'
        if (str_contains($subject, CLARIFICATION_CONSTANT)) {
```
We need to know this in a couple of places and as those are based on not being valid and not on clearly connected to either `problem` or `clarification` (I think) it makes sense to use a couple more chars and keep it consistent.

webapp/src/Controller/Jury/ClarificationController.php
vmcj: that this is picked by symfony before the route `/{clarid}`

So instead of 404: no clar with name "send", you would first enter this route.

webapp/src/Controller/Jury/RejudgingController.php
```diff
@@ -742,7 +748,11 @@ public function createAction(Request $request): Response
                 ->from(Judging::class, 'j')
                 ->leftJoin('j.submission', 's')
                 ->leftJoin('s.rejudging', 'r')
+                ->leftJoin('s.contest', 'c')
+                ->leftJoin('s.language', 'l')
+                ->leftJoin('s.problem', 'p')
                 ->leftJoin('s.team', 't')
+                ->leftJoin('s.user', 'u')
```
vmcj, line 755: You join but don´t select those, didn´t @meisterT  have such a PR where you also need to select these to make sure that you get them in the results?

webapp/src/Controller/Jury/SubmissionController.php
```diff
@@ -240,11 +240,17 @@ public function viewAction(
 
         // If judging ID is not set but rejudging ID is, try to deduce the judging ID from the database.
         if (!isset($judgingId) && isset($rejudgingId)) {
-            $judging = $this->em->getRepository(Judging::class)
-                ->findOneBy([
-                                'submission' => $submitId,
```
vmcj, line 245: why can´t we search for `externalid` here? Instead of doing the query.

webapp/tests/Unit/Controller/Jury/UserControllerTest.php
nickygerritsen: We never did. `0` meant `no team`. now empty string means the same.

webapp/tests/Unit/Controller/Jury/LanguagesControllerTest.php
nickygerritsen: Good idea, done

webapp/tests/Unit/Controller/Jury/TeamControllerTest.php
nickygerritsen: Uh we don't need this. Not sure why it's here.

webapp/src/Controller/Jury/ClarificationController.php
nickygerritsen: Exactly, this makes this route more important than the view route with `.*` basically. Another option is to put this route 'physically' above the other route, but I dislike that more.

webapp/migrations/Version20251024122021.php
nickygerritsen: Well no since you only uploaded that one yesterday. But I just did and the migration works.

webapp/templates/jury/clarification.html.twig
nickygerritsen: I think I was drunk, this doesn't work

webapp/src/Controller/Jury/LanguageController.php
nickygerritsen: Good catch, it needs updating. This was the only one that was actually broke but updated some more to use the controller argument, so it's cleaner.

webapp/src/Service/DOMJudgeService.php
nickygerritsen: Yep

webapp/src/Controller/Jury/TeamController.php
nickygerritsen: Clearly it can.

webapp/src/Controller/Jury/TeamAffiliationController.php
vmcj: This is the 2nd time Tobi asks this, so it´s worth discussing this somewhere if we like this more. I do for this PR but I might like it more when debugging if I don´t have to read the full file to find why a certain route takes priority (or I just need to get used to this).

webapp/migrations/Version20251026092516.php
nickygerritsen: Do you mean a LEFT JOIN? AFAIK (and AI agrees) INNER JOIN and normal JOIN are the same.
If entities don't exist anymore, we can't do much. But that doesn't really matter, since then you only can't click the link, but you can't anyway.

webapp/src/Controller/Jury/TeamAffiliationController.php
nickygerritsen:
```
$ webapp/bin/console debug:router | grep /jury/clarifications/send
  jury_clarification_new                                 ANY        /jury/clarifications/send   
```

#protip

judge/judgedaemon.main.php
nickygerritsen: I prefer writing `submission` then and I just did that.

webapp/src/Controller/Jury/TeamAffiliationController.php
vmcj: I think my argument still holds, but let's discuss if we like this at the next hackathon... I'm fine with it for now as it makes reviewing so much easier.

webapp/migrations/Version20251024122021.php
nickygerritsen: Fixed and yes I did. Took me quite a while.

webapp/migrations/Version20251024122021.php
nickygerritsen: If we want this, let's do it in a separate PR (after merging this one, pleaseeeee)

webapp/src/Controller/BaseController.php
nickygerritsen: It will.

webapp/src/Controller/Jury/ClarificationController.php
nickygerritsen: If we want this, we should discuss this, since then we also need to modify `/add` to be `/_add`. In that case I would vote to instead change all `/[x]`  URLs to be `/view/[x]`.

webapp/src/Controller/PublicController.php
```diff
@@ -143,7 +143,7 @@ protected function getContestFromRequest(?string $contestId = null): ?Contest
             } else {
                 // Find the contest with the given ID.
                 foreach ($this->dj->getCurrentContests(onlyPublic: true) as $possibleContest) {
-                    if ($possibleContest->getCid() == $contestId || $possibleContest->getExternalid() == $contestId) {
+                    if ($possibleContest->getExternalid() == $contestId) {
```
vmcj:
```suggestion
                    if ($possibleContest->getExternalid() === $contestId) {
```

nickygerritsen:
> We now have a lot of foreign keys still going to the internal identifier, wouldn´t the code be much cleaner if we point to the externalid?
> 
> I think it makes debugging also a lot easier as we see the value in the database which we also see in the UI.
> 
> I know it's a pain to change now, but probably it's even worse if we do this later.

I don't think I'm a fan for a couple of reasons:

* Some external IDs are only unique in a contest, so we would need a compound foreign key.
* Not all tables have an external ID, so we would not be consistent with foreign keys.
* Minor, but we would need a huge migration to do all this, similar to the language one.
* To me it feels like database IDs are used for... the database. If we want to use external IDs we should drop database IDs fully instead.

webapp/src/Controller/PublicController.php
```diff
@@ -158,28 +158,27 @@ protected function getContestFromRequest(?string $contestId = null): ?Contest
...
-    public function teamAction(Request $request, int $teamId): Response
+    #[Route(path: '/team/{teamId}', name: 'public_team')]
+    public function teamAction(Request $request, string $teamId): Response
     {
         /** @var Team|null $team */
         $team = $this->em->createQueryBuilder()
                          ->from(Team::class, 't')
                          ->innerJoin('t.categories', 'tc')
                          ->select('t, tc')
-                         ->andWhere('tc.visible = 1')
```
vmcj, line 181: Removing this makes the function slightly slower. Now you remove hidden teams with PHP code.

webapp/src/Controller/Jury/SubmissionController.php
nickygerritsen: Huh how? The judging table doesn't have an external ID, especially not the submission one.

webapp/migrations/Version20251024122021.php
vmcj: That's what I meant with general note, but it makes sense to note it down IMO.

webapp/src/Controller/Jury/ClarificationController.php
vmcj: I think what you propose is better. It also removes the `priority = 1` issue.

judge/judgedaemon.main.php
nickygerritsen: We don't include that in the webapp code currently.

judge/judgedaemon.main.php
nickygerritsen: Well no, it is actually done judging already. I don't think it should read that?

webapp/migrations/Version20251024122021.php
nickygerritsen: We can't.

webapp/src/Controller/Jury/ClarificationController.php
nickygerritsen: It does, but it does update a lot of URLs, which I didn't want to do. If people think this is better, we can.

webapp/src/Controller/Jury/RejudgingController.php
nickygerritsen: Only if we use them, but it doesn't hurt so I'll add it.

webapp/src/Controller/PublicController.php
nickygerritsen: Yeah I thought we needed the hidden ones but we don't. Re-adding.

webapp/src/Controller/Jury/RejudgingController.php
vmcj: But if we don´t use them we don´t need the joins, if I understand the code correctly you add the joins but never use them.

webapp/migrations/Version20251024122021.php
vmcj: So this is not needed right:
`WHERE s.langid IS NOT NULL`

webapp/src/Controller/Jury/RejudgingController.php
nickygerritsen: We use them in tablemap to do queries so we do need the joins. Do you want me to drop them from the select again?

webapp/migrations/Version20251024122021.php
nickygerritsen: Yes missed them there.

webapp/src/Controller/Jury/ClarificationController.php
vmcj: Let's do it in another (possible) PR.

webapp/src/Controller/Jury/RejudgingController.php
vmcj: No, I thought you added joins which we don´t need. If we need them I rather have we add them in the `select` to prevent that issue.

webapp/src/Controller/Team/ClarificationController.php
```diff
@@ -127,9 +127,9 @@ public function viewAction(Request $request, int $clarId): Response
         $formData = [];
         if ($clarification) {
             if ($clarification->getProblem()) {
-                $formData['subject'] = sprintf('%d-%d', $clarification->getContest()->getCid(), $clarification->getProblem()->getProbid());
+                $formData['subject'] = sprintf('%s|%s', $clarification->getContest()->getExternalid(), $clarification->getProblem()->getExternalid());
```
vmcj: Constant

webapp/src/Controller/Team/ClarificationController.php
```diff
@@ -127,9 +127,9 @@ public function viewAction(Request $request, int $clarId): Response
...
             } else {
-                $formData['subject'] = sprintf('%d-%s', $clarification->getContest()->getCid(), $clarification->getCategory());
+                $formData['subject'] = sprintf('%s#%s', $clarification->getContest()->getExternalid(), $clarification->getCategory());
```
vmcj: Constant, might be good to really `grep` for this.

webapp/src/Controller/Team/ClarificationController.php
```diff
@@ -227,15 +227,15 @@ private function newClarificationHelper(
...
-        // First part will always be the contest ID, as Symfony will validate this.
-        [, $problemId] = explode('-', $formData['subject']);
+        // First part will always be the contest external ID, as Symfony will validate this.
         $problem = null;
         $category = null;
         $queue = null;
-        if (!ctype_digit($problemId)) {
-            $category = $problemId;
+        if (str_contains($formData['subject'], '#')) {
+            [, $category] = explode('#', $formData['subject']);
```
vmcj: Constant

webapp/src/Controller/Team/ClarificationController.php
```diff
@@ -227,15 +227,15 @@ private function newClarificationHelper(
...
         } else {
-            $problem = $this->em->getRepository(Problem::class)->find($problemId);
+            [, $problemId] = explode('|', $formData['subject']);
```
vmcj: You might guess it, but can you use a constant here?

webapp/src/Controller/Team/ScoreboardController.php
```diff
@@ -71,12 +71,11 @@ public function teamAction(Request $request, int $teamId): Response
                          ->from(Team::class, 't')
                          ->innerJoin('t.categories', 'tc')
                          ->select('t, tc')
-                         ->andWhere('tc.visible = 1')
```
vmcj, line 74: Same remark.

webapp/src/Controller/PublicController.php
```diff
@@ -387,7 +384,7 @@ public function submissionsDataAction(Request $request, ?string $teamId, ?string
             }
             $submissionData[$teamKey][$problemKey][] = [
                 'time' => $this->twigExtension->printtime($submission->getSubmittime(), contest: $contest),
-                'language' => $submission->getLanguageId(),
+                'language' => $submission->getLanguage()->getName(),
```
vmcj, line 387: I think Language should have an `getExternalId` which internally just calls the `getName()||return $name`, reading this now makes me wonder if this code is wrong and I'll will always have that hunch.

webapp/src/DataFixtures/DefaultData/TeamCategoryFixture.php
```diff
@@ -33,8 +28,6 @@ public function load(ObjectManager $manager): void
                     ->setExternalid($item[4]);
                 $manager->persist($category);
                 $manager->flush();
-            } else {
```
vmcj, line 36: If people load our default data we still want to have this message I think?

webapp/src/Controller/Team/ClarificationController.php
nickygerritsen: I now did

webapp/src/Controller/PublicController.php
nickygerritsen: Language *has* a `getExternalid` but this is a table rendered to end users (the public). I don't want to show them a language ID, I want them to show `C++`.

webapp/src/DataFixtures/DefaultData/TeamCategoryFixture.php
nickygerritsen: I couldn't get it to work because we used a loader that didn't understand services in the tests, but I had a good idea last night and just implemented it.

webapp/templates/jury/entity_id_badge.html.twig
```diff
@@ -1,9 +1,9 @@
-<span class="badge text-bg-secondary id-badge" data-bs-toggle="tooltip" data-bs-placement="top" title="{% if label is defined and label | length %}Label: {{ label }}, {% endif %}{% if externalId is not null %}External ID: {{ externalId }}, {% endif %}Internal ID: {{ idPrefix }}{{ id }}">
+<span class="badge text-bg-secondary id-badge" data-bs-toggle="tooltip" data-bs-placement="top" title="{% if label is defined and label | length %}Label: {{ label }}, {% endif %}{% if externalId is not null %}ID: {{ externalId }}{% else %}ID: {{ id }}{% endif %}">
```
vmcj:
```suggestion
<span class="badge text-bg-secondary id-badge" data-bs-toggle="tooltip" data-bs-placement="top" title="{% if label is defined and label | length %}Label: {{ label }}, {% endif %}ID: {% if externalId is not null %}{{ externalId }}{% else %}{{ id }}{% endif %}">
```
When do we not have the externalId?

webapp/tests/Unit/BaseTestCase.php
```diff
@@ -66,7 +66,11 @@ protected function loadFixtures(array $fixtures): void
             $this->fixtureExecutor = new ORMExecutor(static::getContainer()->get(EntityManagerInterface::class));
         }
 
-        $loader = new Loader();
+        // Use the Symfony fixture loader so the fixtures can be autowired as a service
+        $loader = static::getContainer()->get('doctrine.fixtures.loader');
+        // Use reflection on the base class to clear the enabled fixtures, so we can only add what we want
+        $loaderReflection = new \ReflectionClass(Loader::class);
```
vmcj: Can we use `use` here?

judge/judgedaemon.main.php
meisterT: It's the `Submission` that gets the compiler error, so the wording is fine

webapp/templates/jury/entity_id_badge.html.twig
nickygerritsen: Looking at the code I think we have it everywhere now.

judge/judgedaemon.main.php
nickygerritsen: Wrong thread but I know what you mean.

For this, I'm not sure we can easily include lib.misc.php in PHP, at least I don't know where. For now I'll add a constant to the judgedaemon as well.

PR #3268, opened 2025-11-30
Remove collection of emails for self-registered users

vmcj:
In all years that I've worked with DOMjudge I've never seen a good reason to collect the email separately. So I propose to remove the collection in the next major release.

We keep storing the emails in development dumps and although it doesn´t hurt that much it is a bit more privacy sensitive than most other fields IMHO.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -15,6 +15,7 @@ Version 10.0.0DEV
    shadow mode.
  - Use application/x-www-form-urlencoded instead of multipart/form-data for
    API requests without file uploads to make them work with PUT/PATCH requests.
+ - Removed email collection for selfregistered users.
 
 Version 9.0.0 - 5 October 2025
 ------------------------------
--- a/webapp/migrations/Version20190803123217.php
+++ b/webapp/migrations/Version20190803123217.php
@@ -591,7 +591,6 @@ CREATE TABLE `user` (
     `userid` int(4) unsigned NOT NULL AUTO_INCREMENT COMMENT 'Unique ID',
     `username` varchar(255) NOT NULL COMMENT 'User login name',
     `name` varchar(255) NOT NULL COMMENT 'Name',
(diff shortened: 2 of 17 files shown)


Kevinjil:
This is one of those features I have no idea if anyone in the world is using it. I can imagine use cases when self-signup enabled and, e.g. after contest, you'd like to sent out mail to whoever entered an email address.

meisterT:
Is entering an email address mandatory right now? If so, could me make it optional instead of dropping it?

There might be installations out there which suddenly lose data that may be important to them (because that's how they can communicate with their users).

vmcj:
> Is entering an email address mandatory right now? If so, could me make it optional instead of dropping it?

It's already optional, we still get them in our dumps as it's implied with such installation that you might get an confirmation email?
> 
> There might be installations out there which suddenly lose data that may be important to them (because that's how they can communicate with their users).

In the 2nd commit we don't drop the column if it exists, only on new installs the column is not created. That is the intent at least.

meisterT:
Thinking more about it, it might just be OK to delete the emails there if we are very clear in the Change Log that they will be dropped.

ChangeLog
```diff
@@ -15,6 +15,7 @@ Version 10.0.0DEV
    shadow mode.
  - Use application/x-www-form-urlencoded instead of multipart/form-data for
    API requests without file uploads to make them work with PUT/PATCH requests.
+ - Removed email collection for selfregistered users.
```
meisterT, line 18:
```suggestion
 - Removed email collection for selfregistered users. Note that this will drop the email address from existing installations.
```

vmcj:
> Made a tiny suggestion to the commit message - you probably want to drop the middle commit and then merge the two remaining?

I actually think keeping the addresses there for people who wanted them is okay? Mostly as I think most people don´t read the changelog.

PR #3273, opened 2025-12-01
Only show compile if the judging is done

vmcj:
Based on https://domjudge.sentry.io/issues/7076103482/?project=4510437648695296&query=is%3Aunresolved&referrer=issue-stream

### DIFF
--- a/webapp/src/Controller/Team/SubmissionController.php
+++ b/webapp/src/Controller/Team/SubmissionController.php
@@ -150,7 +150,7 @@ class SubmissionController extends BaseController
         }
 
         $runs = [];
-        if ($showSampleOutput && $judging && $judging->getResult() !== 'compiler-error') {
+        if ($showSampleOutput && $judging?->getResult() !== 'compiler-error') {
             $outputDisplayLimit    = (int)$this->config->get('output_display_limit');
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

eldering:
> Based on https://domjudge.sentry.io/issues/7076103482/?project=4510437648695296&query=is%3Aunresolved&referrer=issue-stream

It would be better to (also) include the error log explicitly in the commit message. That way you don't need to log in to sentry to understand why this was needed, but also, it will be retained in our git history, even if the logs are not available in sentry anymore for whatever reason.

vmcj:
> > Based on https://domjudge.sentry.io/issues/7076103482/?project=4510437648695296&query=is%3Aunresolved&referrer=issue-stream
> 
> It would be better to (also) include the error log explicitly in the commit message. That way you don't need to log in to sentry to understand why this was needed, but also, it will be retained in our git history, even if the logs are not available in sentry anymore for whatever reason.

I think there is nothing relevant in that issue to explain why we need this. Either this fix shouldn't be needed as we have for every submission a judgement, or this fix is trivial enough. Sentry only warned us it happened in practice, not really why it happened for this specific case.

PR #3274, opened 2025-12-01
Temporary ugly fix

vmcj:


meisterT:
Fix for what?

vmcj:
> Fix for what?

We have a test which fails and blocks CI, I wrongly assumed this was in the image. 
See: https://github.com/DOMjudge/domjudge/actions/runs/19817699529/job/56772588320?pr=3274 for one of those for example. 

I'll convert this one to draft for now, as I don't have the intention to merge it now, I just need CI to check if it works.

PR #3275, opened 2025-12-01
Fix ci phpcscompat upgrades

vmcj:


.github/jobs/fix_pipelinecomponents_image.sh
```diff
@@ -20,5 +20,5 @@ sed -i 's/"squizlabs\/php_codesniffer": "3.13.2"/"squizlabs\/php_codesniffer": "
 echo "After /app/composer.json:"
 cat /app/composer.json
 
-cd /app; composer update
+cd /app; composer -W update
```
meisterT: consider adding an explanatory comment before this line so that we know why we added `-W` (or why we need to run `composer update` at all)

vmcj: If I understand the problem I'll probably add a proper fix so we don't need this at all.

If/when this works I think I remove the test completely and let our CI container install this tool to (re)move the ugliness to the scripts repo.

PR #3277, opened 2025-12-01
Testing PHP 8.1 is unnecessary after we dropped support for it.

meisterT:


vmcj:
Didn't I already have a PR for this?

meisterT:
I thought that's already merged and this might have been missed but apparently not

PR #3279, opened 2025-12-01
Disable phpcs CI run while it is broken

meisterT:
We probably should not use the latest tag but some fixed version so that it doesn't break under us at random times


vmcj:
> We probably should not use the latest tag but some fixed version so that it doesn't break under us at random times

We needed this because the latest released version lacked something important IRC

vmcj:
@meisterT I propose we close this one.

PR #3265, opened 2025-11-30
Start development with the latest PHP version

vmcj:

PR #3280, opened 2025-12-02
Set PHP 8.2 as minimum, dropping 8.1

nickygerritsen:
Also add 8.5 testing


vmcj:
That PR is mostly already open,

meisterT:
@vmcj I only see https://github.com/DOMjudge/domjudge/pull/3265 which is not updating code, I am likely blind (or missing something).

nickygerritsen:
There is #3269 but that also does less.

PR #3283, opened 2025-12-06
Fix submit tests and make them less brittle

eldering:
Resolving of domjudge.example.edu now gives a different error.

### DIFF
--- a/submit/submit_online.bats
+++ b/submit/submit_online.bats
@@ -10,12 +10,14 @@ setup() {
     export SUBMITBASEURL="$(grep '^BASEURL' ../paths.mk | tr -s ' ' | cut -d ' ' -f 3)"
 }
 
-@test "contest via parameter overrides environment" {
-    run ./submit -c bestaatniet
-    assert_failure 1
-    assert_partial "ERROR: No (valid) contest specified"
+@test "baseurl can end in slash" {
+    run ./submit --url "$SUBMITBASEURL/" --contest demo --help
+    assert_success
+    assert_regex "A *- *Hello World"
+}
 
-    run ./submit --contest=bestaatookniet
+@test "contest via parameter overrides environment" {
+    run ./submit -c nonexistent
     assert_failure 1
     assert_partial "ERROR: No (valid) contest specified"
 }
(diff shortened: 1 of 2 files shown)


submit/submit_online.bats
```diff
@@ -10,12 +10,14 @@ setup() {
     export SUBMITBASEURL="$(grep '^BASEURL' ../paths.mk | tr -s ' ' | cut -d ' ' -f 3)"
 }
 
-@test "contest via parameter overrides environment" {
-    run ./submit -c bestaatniet
-    assert_failure 1
-    assert_partial "ERROR: No (valid) contest specified"
+@test "baseurl can end in slash" {
+    run ./submit --url "$SUBMITBASEURL/" --contest demo --help
```
vmcj, line 14: I don´t think we should trust on `--help` to always do an online lookup.

submit/submit_standalone.bats
```diff
@@ -24,17 +24,7 @@ setup() {
...
-    run ./submit -u https://domjudge3.example.edu
-    assert_failure 1
-    assert_regex "domjudge3.example.edu.*/api(/.*)?/contests.*\[Errno -2\] Name or service not known"
-}
-
-@test "baseurl can end in slash" {
-    run ./submit --url https://domjudge.example.edu/domjudge/
-    assert_failure 1
-    assert_regex "domjudge.example.edu.*/api(/.*)?/contests.*\[Errno -2\] Name or service not known"
+    assert_regex "domjudge\.example\.edu"
```
vmcj, line 27: If we don´t test for any thing specific in the error, we can just drop this. IMO `domjudge\.example\.edu` is very generic.

submit/submit_online.bats
eldering: That's the same as below.

submit/submit_standalone.bats
eldering: It's the URL that was given, so it tests that something with that URL is wrong without depending on the specific way it fails. I think that's a good balance between not being too brittle and still test something.

submit/submit_online.bats
meisterT: `--help` does always do a lookup, e.g. to include the problem list in its output

meisterT: `--help` does always do a lookup, e.g. to include the problem list in its output

PR #3286, opened 2025-12-07
Fold `testcase_run.sh` logic into judgedaemon.

meisterT:
The primary motivation is to significantly reduce the per-testcase overhead of judging a submission.

Historically, with fewer test cases, the separation was carrying its weight. However, with modern problems using many test cases or multiple passes, the overhead from dozens of forked simple programs (`cp`, `mv`, `chmod`, `grep` etc.) within the shell script has become a major performance bottleneck. Moving them into PHP effectively replaces them with simple system calls.

This change results in a substantial performance improvement:
* Roughly a 50% speed-up in total judging overhead per submission
* Example: A simple C++ solution to a multi-pass problem (NWERC 2025 practice) was reduced from ~70s to ~33s end-to-end.


meisterT:
No need to review yet, I am still testing and there are some follow-up changes required.

doc/manual/config-advanced.rst
```diff
@@ -223,7 +223,7 @@ To allow for problems that do not fit within the standard scheme of
 fixed input and/or output, DOMjudge has the possibility to change the
 way submissions are run and checked for correctness.
 
-The back end script ``testcase_run.sh`` that handles
+The judgedaemon that handles
```
vmcj: The line is now short, so we can remove some newlines etc.

example_problems/hello/submissions/time_limit_exceeded/stress-test-fork-setsid.c
```diff
@@ -3,7 +3,7 @@
  *
  * Without cgroups however, this will crash the judging daemon: it
  * forks processes and places these in a new session, such that
- * testcase_run cannot retrace and kill these. They are left running
+ * the judgedaemon cannot retrace and kill these. They are left running
```
vmcj, line 6: I wonder if we should still have this file (or the explanation, running without CGroups is not tested anymore nor advised).

judge/runguard.cc
```diff
@@ -266,7 +266,7 @@ void die(int errnum, std::format_string<Args...> fmt, Args&&... args)
 			/*
 			 * continue, there is not much we can do here.
 			 * In the worst case, this will trigger an error
-			 * in testcase_run.sh, as the runuser may still be
+			 * in the judgedaemon, as the runuser may still be
```
vmcj, line 269: I might not know the current flow but to me it feels that we can accept an error in the testcase_run.sh but if we get it in the judgedaemon this becomes more of an problem.

So what is the flow, and why is it okay to still ignore this error?

judge/judgedaemon.main.php
```diff
@@ -18,6 +18,19 @@
 
 define('DONT_CARE', new class {});
 
+enum Verdict
```
vmcj, line 22: PHPCS complains, I assume you ignored this with intent?

judge/judgedaemon.main.php
```diff
@@ -1472,7 +1908,7 @@ private function runTestcase(
         }
 
         if ($combined_run_compare) {
-            // set to empty string to signal the testcase_run script that the
+            // set to empty string to signal that the
             // run script also acts as compare script
```
```suggestion
            // set to empty string to signal that the run script also acts as compare script
```

judge/judgedaemon.main.php
meisterT: Yeah, I added an ignore statement now

judge/runguard.cc
meisterT: This is already checked at HEAD in the judgedaemon https://github.com/DOMjudge/domjudge/blob/main/judge/judgedaemon.main.php#L189

example_problems/hello/submissions/time_limit_exceeded/stress-test-fork-setsid.c
meisterT: Slightly adjusted the text.

judge/judgedaemon.main.php
```diff
@@ -1439,6 +1459,423 @@ private function compileAndRunSubmission(array $judgeTask, string $workdirpath):
...
+                return Verdict::INTERNAL_ERROR;
+            }
+            if (!$combined_run_compare && !is_executable($compare_runpath)) {
+                logmsg(LOG_WARNING, "Compare script '$compare_runpath' not found or not executable.");
+                return Verdict::INTERNAL_ERROR;
+            }
+
+            foreach ([
+                         'system.out',                   # Judging system output (info/debug/error)
+                         'program.out','program.err',    # Program output and stderr (for extra information)
```
nickygerritsen: Nit:
```suggestion
                         'program.out', 'program.err',    # Program output and stderr (for extra information)
```

judge/judgedaemon.main.php
```diff
@@ -1439,6 +1459,423 @@ private function compileAndRunSubmission(array $judgeTask, string $workdirpath):
...
+            if ($program_meta_ini['exitcode'] !== '0') {
+                appendToFile("system.out", "Non-zero exitcode " . $program_meta_ini['exitcode']);
+                return Verdict::RUN_ERROR;
+            }
+
+            // Check whether stdout is in the list of truncated output streams
+            // These could be either 'stdout', 'stderr,stdout', or 'stdout,stderr'
+            $outputTruncated = $program_meta_ini['output-truncated'] ?? '';
+            $truncatedStreams = explode(',', $outputTruncated);
+            if (in_array('stdout', $truncatedStreams, true)) {
```
nickygerritsen: Why not `str_contains`?

meisterT: Not that there will be ever a `foo_stdout_bar`, but it felt still cleaner that way. I don't feel strongly though, do you?

nickygerritsen: Nope, this is also fine. If only we had sets...

judge/judgedaemon.main.php
```diff
@@ -1439,6 +1459,423 @@ private function compileAndRunSubmission(array $judgeTask, string $workdirpath):
...
+                    "--",
+                    "$prefix/execdir/program",
+                ]
+            );
+
+            $this->runCommandSafe($run_args, $exitcode, log_nonzero_exitcode: false, stderr_target: "runguard.err");
+
+            if (CREATE_WRITABLE_TEMP_DIR) {
+                // Revoke access to the TMPDIR as security measure
+                if (!chown("$realWorkdir/write_tmp", posix_getpwuid(posix_geteuid())['uid'])) {
```
vmcj: Please verify but I think because of `chmod: 0777 + ownership` a submission can remove this dir so we should check if the folder still exists.

meisterT: Done.

PR #3288, opened 2025-12-07
Format judgedaemon code according to PSR-2.

meisterT:


judge/judgedaemon.main.php
```diff
@@ -851,7 +865,9 @@ private function runCommandSafe(array $command_parts, &$retval = DONT_CARE, $log
 
         logmsg(LOG_DEBUG, "Executing command: $command");
         system($command, $retval_local);
-        if ($retval !== DONT_CARE) $retval = $retval_local; // phpcs:ignore Generic.ControlStructures.InlineControlStructure.NotAllowed
+        if ($retval !== DONT_CARE) {
+            $retval = $retval_local;
+        } // phpcs:ignore Generic.ControlStructures.InlineControlStructure.NotAllowed
```
eldering, line 870: This was explicitly marked to not get reformatted, so I wonder what happened here.
I'd prefer the old way, but at least the comment should not be left like this.

PR #3300, opened 2025-12-16
Redetect entrypoint when switching language

nickygerritsen:
Adapted from SEERC.


tuupke (approved):

vmcj:
When did this fail?

nickygerritsen:
> When did this fail? 

We had two pythons (pypy and cpython), and switching removed the entry point. But i think if you switch languages it always makes sense to recalculate it

Kevinjil (approved):

PR #3301, opened 2025-12-16
Shadowing import improvements

nickygerritsen:
All these were needed at SEERC when shadowing from the merged CDS contest and I think all of them make sense.

### DIFF
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -766,7 +766,7 @@ class ExternalContestSourceService
 
         // Also compare the penalty time
         $penaltyTime = $data->penaltyTime;
-        if ($this->config->get('penalty_time') != $penaltyTime) {
+        if ($penaltyTime !== null && $this->config->get('penalty_time') != $penaltyTime) {
             $this->logger->warning(
                 'Penalty time does not match between feed (%d) and local (%d)',
                 [$penaltyTime, $this->config->get('penalty_time')]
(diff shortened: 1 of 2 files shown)


webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -766,7 +766,7 @@ protected function validateAndUpdateContest(Event $event, EventData $data): void
 
         // Also compare the penalty time
         $penaltyTime = $data->penaltyTime;
-        if ($this->config->get('penalty_time') != $penaltyTime) {
+        if ($penaltyTime !== null & $this->config->get('penalty_time') != $penaltyTime) {
```
tuupke: Bitwise-and becomes a truthy value and is the same in this case, but does not seem intentional.

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -766,7 +766,7 @@ protected function validateAndUpdateContest(Event $event, EventData $data): void
 
         // Also compare the penalty time
         $penaltyTime = $data->penaltyTime;
-        if ($this->config->get('penalty_time') != $penaltyTime) {
+        if ($penaltyTime !== null && $this->config->get('penalty_time') != $penaltyTime) {
```
vmcj, line 769: Either the spec is hard to read but isn't this one non-optional so shouldn´t we emit a warning?

nickygerritsen: Yeah a warning makes sense

PR #3302, opened 2025-12-16
Also show memory limit for samples on team submission page

nickygerritsen:
I'm open to discussing these changes, since they might not be preferred by everyone.

I am in favor of the first change for sure (since teams could measure this on their own machine anyway). The second might be more controversial.

### DIFF
--- a/etc/db-config.yaml
+++ b/etc/db-config.yaml
@@ -250,6 +250,13 @@
             default_value: false
             public: true
             description: Should teams be able to view a diff of their and the reference output on sample testcases?
+        -   name: show_all_testcase_verdicts
+            type: bool
+            default_value: false
+            public: true
+            description: |
+                Show verdicts for all testcases on the team interface.
+                If `show_limits_on_team_page` is also set, also display runtime and memory usage.
         -   name: show_too_late_result
             type: bool
             default_value: false
--- a/webapp/templates/team/partials/submission.html.twig
+++ b/webapp/templates/team/partials/submission.html.twig
@@ -78,83 +78,34 @@
             {% endif %}
         {% endif %}
 
-        {% if showSampleOutput and judging.result != 'compiler-error' %}
+        {% if (showSampleOutput or showAllTestcases) and judging.result != 'compiler-error' %}
             <hr/>
-            <h4 class="text-center">Run(s) on the provided sample data</h4>
+            <h4 class="text-center">
+                {% if not showAllTestcases %}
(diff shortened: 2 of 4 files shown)


webapp/templates/team/partials/submission.html.twig
```diff
@@ -99,10 +99,18 @@
                                         <th>Description</th>
                                         <td>{{ run.0.description(true) }}</td>
                                     </tr>
-                                    <tr>
-                                        <th>Runtime</th>
-                                        <td>{{ judgingRun.runtime }} sec</td>
-                                    </tr>
+                                    {% if showLimits %}
```
vmcj: I prefer to not have this option, and first come up with a good reason why this shouldn´t be public.

If there is a good reason we should also add tests for this to make sure that we don´t expose that information, for now I miss the change to also expose this in the API to keep those in line.

nickygerritsen: I'm fine always doing it but if we have that setting for limits turned of I think it's weird.
As for the API, I'm fine adding the memory limit there as well

meisterT:
Can you please add a before/after screenshot?

nickygerritsen:
Before if all sample tests have been run:
image

Before if not:
image

After with the default settings, when not all sample testcases are run:
image

After with showing limits off:
image

After with showing limits on AND showing results for all testcases on:
image
[...]
image

vmcj:
> After with showing limits on AND showing results for all testcases on: image

I wonder if a graph with 2 y-axis & the limit would be even better, as we also have in the jury interface. I can look at that myself in another PR.

etc/db-config.yaml
```diff
@@ -250,6 +250,13 @@
             default_value: false
             public: true
             description: Should teams be able to view a diff of their and the reference output on sample testcases?
+        -   name: show_all_testcase_verdicts
+            type: bool
+            default_value: false
+            public: true
+            description: |
+                Show verdicts for all testcases on the team interface.
+                If `show_limits_on_team_page` is also set, also display runtime and memory usage.
```
meisterT, line 259: I find the implication of exposing actual usage when selecting the option "show limits" (as in upper bounds) quite confusing.

nickygerritsen: What is your suggestion? Always showing time/memory limit for samples/testcases (like we do now) if you enable those options or adding yet another option to show these limits?

PR #3303, opened 2025-12-16
Allow to mark multiple balloons as done

nickygerritsen:
This adds 2 features to the balloon page:

* Checkboxes and a button to mark multiple balloons as done.
* A table at the bottom, sorted by location (or if not set, team external ID), that shows all the balloons the team should have gotten, to be able to do a check.

Both were requested by Marius during SEERC 2025 and I think both make sense.

### DIFF
--- a/webapp/src/Service/BalloonService.php
+++ b/webapp/src/Service/BalloonService.php
@@ -212,9 +212,9 @@ class BalloonService
         }
 
         uasort($teamSummary, function ($a, $b) {
-            $aKey = $a['location'] ?? $a['team']->getExternalId();
-            $bKey = $b['location'] ?? $b['team']->getExternalId();
-            return strnatcasecmp((string)$aKey, (string)$bKey);
+            $aKey = join('-', array_filter([$a['location'], $a['team']->getExternalId()]));
+            $bKey = join('-', array_filter([$b['location'], $b['team']->getExternalId()]));
+            return strnatcasecmp($aKey, $bKey);
         });
 
         return $teamSummary;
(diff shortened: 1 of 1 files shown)


webapp/src/Service/BalloonService.php
```diff
@@ -181,14 +181,64 @@ public function collectBalloonTable(Contest $contest, bool $todo = false): array
...
+                    'location' => $data['location'],
+                    'category' => $data['category'],
+                    'categoryid' => $data['categoryid'],
+                    'total' => $data['total'],
+                ];
+            }
+        }
+
+        uasort($teamSummary, function ($a, $b) {
+            $aKey = $a['location'] ?? $a['team']->getExternalId();
```
vmcj: For us `location` can be a string right? So those can be the same, maybe also append the externalId to it to make sure you can always sort?

webapp/src/Service/BalloonService.php
```diff
@@ -181,14 +181,64 @@ public function collectBalloonTable(Contest $contest, bool $todo = false): array
...
-        if (!$balloon) {
-            throw new NotFoundHttpException('balloon not found');
+        $balloons = $em->createQueryBuilder()
+            ->from(Balloon::class, 'b')
+            ->select('b')
+            ->andWhere('b.balloonid IN (:balloonIds)')
+            ->setParameter('balloonIds', $balloonId)
+            ->getQuery()
+            ->getResult();
+        if (empty($balloons)) {
```
vmcj:
```suggestion
        if (count($balloons) !== count((array)$balloonId)) {
```

vmcj: So we drop the message if any of the balloonIds is invalid (for whatever reason)...

webapp/src/Service/BalloonService.php
nickygerritsen: I like that idea

PR #3304, opened 2025-12-16
Use shadow mode judgements in shadow mode when displaying team submissions on scoreboard

nickygerritsen:


vmcj:
I think this should be a toggle, when we shadow I would think we should trust no data of the primary except for what is strictly needed (starttime + submissions). Or we should have this as a flag in the URL or something.

nickygerritsen:
Currently all UI uses data from the primary and this makes it consistent.
Having a flag is a good idea but i think it should be separate from this change

Issue #2778, opened 2024-11-04
Add an option of sending "encourage balloon"


bhscer:
### Description of the enhancement request

In ACM contests, it's common to stop distributing balloons during the scoreboard freeze. However, to create a more supportive and motivating environment, I propose adding "encouragement balloons" for teams with fewer solved problems. These balloons would offer a morale boost and enhance the overall competition experience, particularly for teams struggling to advance.

### The goal you want to achieve

During the freeze period, teams that have solved fewer problems than a set threshold (x) can still receive balloons for any new problems they solve, until they exceed the threshold.

### Expected behaviour

1. Step 1
Add options in the contest settings to enable "encouragement balloons" and to specify a threshold for the number of solved problems. Teams with fewer solved problems than this threshold will receive the encouragement balloons.
2. Step 2
In balloon list page, we add a special mark showing whether this balloon is an "encouragement balloon".

### Any other information that you want to share?

I apologize for my poor English. I hope the developers can understand what I am trying to express. Thank you for your understanding!

vmcj:
You can already do this when using the icpctools balloon utility as that tool has this builtin.

Before considering this I think this would need more people requesting this as it is a bit controversial within the community. So I'll label it as a `low prio` for now until we get more people requesting this to prevent against feature creep.

bhscer:
> You can already do this when using the icpctools balloon utility as that tool has this builtin.
> 
> Before considering this I think this would need more people requesting this as it is a bit controversial within the community. So I'll label it as a `low prio` for now until we get more people requesting this to prevent against feature creep.

Thank you for your quick response! Actually, I'm using the balloon API to read the balloons and import them into my system. I noticed the option "Show balloons post-freeze." Does this mean that it will collect balloons during the freeze period, allowing me to use the balloon API to read the balloons and decide whether to distribute them or not?

bhscer:
Sorry that I accidentally closed this issue....

cubercsl:
> > You can already do this when using the icpctools balloon utility as that tool has this builtin.
> > 
> > Before considering this I think this would need more people requesting this as it is a bit controversial within the community. So I'll label it as a `low prio` for now until we get more people requesting this to prevent against feature creep.
> 
[...]

PR #3305, opened 2025-12-16
Reworks balloon for sending balloons in freeze

tuupke:
Replaces `show_balloons_postfreeze` with `minimum_number_of_balloons` which allows for sending out balloons to teams even during the freeze while still supporting the old behavior if desired.

This is useful as a motivational tool for teams that have not yet solved 'enough' problems before entering the freeze. e.g. 2 problems before the freeze while the expected number of problems to solve before the freeze is 3. By setting this new value to 3 the team will receive a balloon as long as the problem they've solved has been solved before the freeze.

Kept on draft until I fix the documentation and do some minor cleanup.

### DIFF
--- a/webapp/src/DataFixtures/ExampleData/TeamCategoryFixture.php
+++ b/webapp/src/DataFixtures/ExampleData/TeamCategoryFixture.php
@@ -8,6 +8,7 @@ use Doctrine\Persistence\ObjectManager;
 class TeamCategoryFixture extends AbstractExampleDataFixture
 {
     final public const PARTICIPANTS_REFERENCE = 'participants';
+    final public const OBSERVERS_REFERENCE = 'observers';
 
     public function load(ObjectManager $manager): void
     {
@@ -41,5 +42,6 @@ class TeamCategoryFixture extends AbstractExampleDataFixture
         $manager->flush();
 
         $this->addReference(self::PARTICIPANTS_REFERENCE, $participants);
+        $this->addReference(self::OBSERVERS_REFERENCE, $observers);
     }
 }
--- a/webapp/src/DataFixtures/Test/BalloonNotificationsSettingsFixture.php
+++ b/webapp/src/DataFixtures/Test/BalloonNotificationsSettingsFixture.php
@@ -1,0 +1,187 @@
+<?php declare(strict_types=1);
+
+namespace App\DataFixtures\Test;
+
+use App\DataFixtures\ExampleData\TeamAffiliationFixture;
+use App\Entity\Balloon;
+use App\Entity\Contest;
+use App\Entity\ContestProblem;
+use App\Entity\Language;
+use App\Entity\Problem;
+use App\Entity\Submission;
+use App\Entity\Team;
+use App\Entity\TeamAffiliation;
+use App\Entity\TeamCategory;
+use App\Utils\Utils;
+use Doctrine\Persistence\ObjectManager;
+
+class BalloonNotificationsSettingsFixture extends AbstractTestDataFixture
+{
+    public function load(ObjectManager $manager): void
+    {
+        /** @var Contest $contest */
+        $contest = $manager->getRepository(Contest::class)->findOneBy(['shortname' => 'beforeUnfreeze']);
+
+        $affiliation = (new TeamAffiliation())
+            ->setExternalid('balloontester')
+            ->setShortname('BT')
+            ->setName('Balloon Tester')
+            ->setCountry('NLD');
+
+        $manager->persist($affiliation);
+
+        // Setup two teams, and two problems, one of these problems must be solved before the freeze,
+        // while the other is only solved after the freeze.
+
+        // Create three categories, two sharing the same short order.
+        $cat1 = (new TeamCategory())
+            ->setSortorder(1)
+            ->setName('balloon cat1')
+            ->setExternalid('balloon cat1');
+        $cat2 = (new TeamCategory())
+            ->setSortorder(1)
+            ->setName('balloon cat2')
+            ->setExternalid('balloon cat2');
+        $cat3 = (new TeamCategory())
+            ->setSortorder(2)
+            ->setName('balloon cat3')
+            ->setExternalid('balloon cat3');
+
+        $manager->persist($cat1);
+        $manager->persist($cat2);
+        $manager->persist($cat3);
+
+        $teamA = (new Team())
+            ->setExternalid('exteam1')
+            ->setIcpcid('exteam1')
+            ->setLabel('exteama1')
+            ->setName('Balloon team 1')
+            ->setAffiliation($affiliation)
+            ->addCategory($cat1);
+
+        $teamB = (new Team())
+            ->setExternalid('exteam2')
+            ->setIcpcid('exteam2')
+            ->setLabel('exteama2')
+            ->setName('Balloon team 2')
+            ->setAffiliation($affiliation)
+            ->addCategory($cat2);
+
+        $teamC = (new Team())
+            ->setExternalid('exteam3')
+            ->setIcpcid('exteam3')
+            ->setLabel('exteama3')
+            ->setName('Balloon team 3')
+            ->setAffiliation($affiliation)
+            ->addCategory($cat2);
+
+        // Add one team in entirely different sort-order
+        $teamD = (new Team())
+            ->setExternalid('exteam4')
+            ->setIcpcid('exteam4')
+            ->setLabel('exteama4')
+            ->setName('Balloon team 4')
+            ->setAffiliation($affiliation)
+            ->addCategory($cat3);
+
+        $manager->persist($teamA);
+        $manager->persist($teamB);
+        $manager->persist($teamC);
+        $manager->persist($teamD);
+
+        // Create three problems that will be solved pre-freeze and one problem that will only be solved post-freeze, for teams in $cat.
+        // The team in $cat2, will solve two problems before and two problems after the freeze.
+        $preFreezeA = (new Problem())->setName("BalloonProblemA");
+        $cpA = (new ContestProblem())
+            ->setShortname('BA')
+            ->setProblem($preFreezeA)
+            ->setColor("#000000")
+            ->setContest($contest);
+        $manager->persist($preFreezeA);
+        $manager->persist($cpA);
+
+        $preFreezeB = (new Problem())->setName("BalloonProblemB");
+        $cpB = (new ContestProblem())
+            ->setShortname('BB')
+            ->setProblem($preFreezeB)
+            ->setColor("#000000")
+            ->setContest($contest);
+        $manager->persist($preFreezeB);
+        $manager->persist($cpB);
+
+        $preFreezeC = (new Problem())->setName("BalloonProblemC");
+        $cpC = (new ContestProblem())
+            ->setShortname('BC')
+            ->setProblem($preFreezeC)
+            ->setColor("#000000")
+            ->setContest($contest);
+        $manager->persist($preFreezeC);
+        $manager->persist($cpC);
+
+        $postFreeze = (new Problem())->setName("BalloonProblemD");
+        $cpD = (new ContestProblem())
+            ->setShortname('BD')
+            ->setProblem($postFreeze)
+            ->setColor("#000000")
+            ->setContest($contest);
+        $manager->persist($postFreeze);
+        $manager->persist($cpD);
+
+        $language = $manager->getRepository(Language::class)->findByExternalId('cpp');
+
+        $freezeTime = $contest->getFreezetime();
+        $preFreezeSubmitTime = $freezeTime - 2;
+        $postFreezeSubmitTime = $freezeTime + 1;
+        $submissionData = [
+            // Team A submits all three 'pre-freeze problems', before the freeze
+            // team, submittime,           problem,     contest problem,
+            [$teamA, $preFreezeSubmitTime, $preFreezeA, $cpA],
+            [$teamA, $preFreezeSubmitTime, $preFreezeB, $cpB],
+            [$teamA, $preFreezeSubmitTime, $preFreezeC, $cpC],
+
+            // Team B submits the final problem during the freeze.
+            // team, submittime,            problem,     contest problem,
+            [$teamB, $postFreezeSubmitTime, $postFreeze, $cpD],
+
+            // All teams end up solving all problems.
+            // team, submittime,            problem,     contest problem,
+            [$teamA, $postFreezeSubmitTime, $postFreeze, $cpD],
+
+            [$teamB, $postFreezeSubmitTime, $preFreezeA, $cpA],
+            [$teamB, $postFreezeSubmitTime, $preFreezeB, $cpB],
+            [$teamB, $postFreezeSubmitTime, $preFreezeC, $cpC],
+
+            // Submit some problems at submit time to test boundary condition.
+            [$teamC, $postFreezeSubmitTime, $preFreezeA, $cpA],
+            [$teamC, $freezeTime, $preFreezeB, $cpB],
+            [$teamC, $freezeTime, $preFreezeC, $cpC],
+            [$teamC, $postFreezeSubmitTime, $postFreeze, $cpD],
+
+            [$teamD, $preFreezeSubmitTime, $preFreezeA, $cpA],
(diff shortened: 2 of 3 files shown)


tuupke:
Mhh come to think of it, we don't really have the full old behavior anymore. Because now problems that are only solved in the freeze will never receive a balloon regardless of the value.

What do you think, should the old config var be reinstated?

cubercsl:
Possibly related to #2778

tuupke:
> Possibly related to #2778

Yes thank you! I was looking for this exact issue but couldn't find it for some reason. These indeed serve the same purpose.

tuupke:
I'm quite sure the remarks phpstan makes are incorrect. Since otherwise it would not work at all while it does :upside_down_face:

nickygerritsen:
> I'm quite sure the remarks phpstan makes are incorrect. Since otherwise it would not work at all while it does 🙃

Probably the docblocks for the methods being called are wrong.

Use
```bash
cd webapp
vendor/bin/phpstan analyze --configuration phpstan.dist.neon
```

to check locally btw.

webapp/src/Service/BalloonService.php
```diff
@@ -144,41 +134,74 @@ public function collectBalloonTable(Contest $contest, bool $todo = false): array
             $balloon = $balloonsData[0];
             $done = $balloon->getDone();
 
-            if ($todo && $done) {
-                continue;
+            // TODO explain what this var does.
+            if (!key_exists($balloonsData['sortorder'], $balloonSummaryPerTeam)) {
+                $balloonSummaryPerTeam[$balloonsData['sortorder']] = [];
             }
```
Kevinjil:
```suggestion
            $balloonSummaryPerTeam[$balloonsData['sortorder']] ??= [];
```

Kevinjil: I think the `TODO` can just be removed as the variable name is descriptive enough.

Kevinjil (commented):
Just a general question: has something (e.g. at SEERC) influenced why we want this feature compared to the last time it was discussed IRL at the hackathon?

webapp/src/Service/BalloonService.php
tuupke: :facepalm: I renamed the variable but forgot to remove the comment, will fix.

tuupke:
> Just a general question: has something (e.g. at SEERC) influenced why we want this feature compared to the last time it was discussed IRL at the hackathon?

Partly. Some were just because we we're already looking at this code because of a bug (hence Nicky's PR) Some discussion and talks 'inspired' me on how to implement and that this is a nice feature. It was also the case that there were some teams that only solved some of the easy problems in the freeze. Because this is somewhat common -especially on the early rungs of the tournament ladder- it might be a nice motivational feature to add.

The replacing of the old config flag came after (incorrectly) thinking that this change offered at least the same functionality as before.

Tbh I cannot recall this discussion, but if we're opposed we can close the PR; I don't think it hurts anything while it does add a new functionality. When declining the pr we should cherry-pick 706147e5c425d74d69cff7c231f539f6be8cb6c1 since it DRYs up the retrieving of balloon updates.

eldering:
I'm not in favor of adding this because it adds extra complexity and weird exceptions. For example, what if a teams solves their first problem in the freeze, but it is an as yet unsolved problem? Highly unlikely maybe, but then that exposes crucial information during the freeze. There are likely similar edge-case issues, which I think is the reason we should not add this.

eldering:
I think a better way to solve this, is when `show_balloons_postfreeze` is enabled  to show clearly which balloons are post-freeze and how many problems the team already solved. Then it becomes an easy non-technical detail to choose which teams to hand a post-freeze balloon out to.

Kevinjil:
> I think a better way to solve this, is when `show_balloons_postfreeze` is enabled to show clearly which balloons are post-freeze and how many problems the team already solved. Then it becomes an easy non-technical detail to choose which teams to hand a post-freeze balloon out to.

So, this proposal basically shifts the balloon selection from a technical problem to an organizational problem. One could e.g. decide that such balloons are not handed out, unless the contest jury decides to grant it.

vmcj:
> > I think a better way to solve this, is when `show_balloons_postfreeze` is enabled to show clearly which balloons are post-freeze and how many problems the team already solved. Then it becomes an easy non-technical detail to choose which teams to hand a post-freeze balloon out to.
> 
> So, this proposal basically shifts the balloon selection from a technical problem to an organizational problem. One could e.g. decide that such balloons are not handed out, unless the contest jury decides to grant it.

I think the balloon runners are not always the most experienced with all the rules and edge cases on when something would disclose information. So in that case I would say we should also include that information (handing this out provides additional crucial information), and if you already have that information you also have the edge cases. So leaving the choice up to the balloon runners feels wrong to me.

tuupke:
> I think we shouldn't merge this without tests. I do agree it happens relatively often but some in the team had strong opinions on such a feature, because of the same opinion we don't test this ourselves at competitions so I think if we ship this we should test it in a different way (with Unit tests for example..)

Sure, I'll go ahead and start adding some tests. Good to have for balloons anyways. Will probably/hopefully also help address some of @eldering's concerns.

tuupke:
> I'm not in favor of adding this because it adds extra complexity and weird exceptions. For example, what if a teams solves their first problem in the freeze, but it is an as yet unsolved problem? Highly unlikely maybe, but then that exposes crucial information during the freeze. There are likely similar edge-case issues, which I think is the reason we should not add this.

This is exactly why these are only shown for problems that have been solved pre-freeze by teams in the same sortorder. So in your example the team would *not* get a balloon. The entire goal is to not leak any information that was not available pre-freeze while still being able and hand out balloons when desired. I've tried making this abundantly clear in the code and the doc -more in the first than in the second- any suggestions for how to improve?

Adding the testcases as suggested by @vmcj will help discover and prevent regressions on these possible edge-cases. I've given this quite some thought and could not come up with relevant edge-cases within the scope of the feature. I'm open to any and all edge-cases you can think of @eldering. Would be great to address those beforehand.

tuupke:
> > > I think a better way to solve this, is when `show_balloons_postfreeze` is enabled to show clearly which balloons are post-freeze and how many problems the team already solved. Then it becomes an easy non-technical detail to choose which teams to hand a post-freeze balloon out to.
> > 
> > 
> > So, this proposal basically shifts the balloon selection from a technical problem to an organizational problem. One could e.g. decide that such balloons are not handed out, unless the contest jury decides to grant it.
> 
> I think the balloon runners are not always the most experienced with all the rules and edge cases on when something would disclose information. So in that case I would say we should also include that information (handing this out provides additional crucial information), and if you already have that information you also have the edge cases. So leaving the choice up to the balloon runners feels wrong to me.

I completely agree! Solving this at an organizational level is not a good idea. The expectation that (balloon) volunteers are aware of enough intricacies and act accordingly does not hold. We've all organized and been to plenty of contests where this does not even hold for the pre-freeze situation let alone for things that happen post-freeze.

tuupke:
To add to the edge-case discussion. The following scenario would potentially leak information:

1. Problem `A` is easy to solve.
2. Team `Foo` -a good team- focuses on more difficult problems first.
3. Many teams submit correct solutions for `A`. (pre-freeze)
4. The scoreboard is frozen.
5. `Foo` submits a correct solution for `A`. 
6. Some information has leaked once `Foo` receives their balloon for `A`.

The major issue with this scenario is that it assumes the value of balloons to hand out post freeze is set to a number so high that this well performing team still gets a balloon. Any team with a sufficient number of accepted submissions will not receive any more balloons regardless of which problem they solve.

I'd like to stress that the behavior is disabled by default. Only after setting the configuration variable to a strictly positive value will balloon notifications appear for correct post-freeze submissions. There is thus less potential for information leaking than with the old implementation of handing out all balloons post-freeze. 

---
For completeness sake, I will be also be adding back the old behavior of handing out balloons to all accepted post-freeze submissions somehow. Not yet sure whether this is with the old or a new config flag.

webapp/src/Service/BalloonService.php
```diff
@@ -114,28 +112,21 @@ public function collectBalloonTable(Contest $contest, bool $todo = false): array
...
         $balloons = $query->getQuery()->getResult();
-        // Loop once over the results to get totals.
-        $TOTAL_BALLOONS = [];
-        foreach ($balloons as $balloonsData) {
-            if ($balloonsData['color'] === null) {
-                continue;
-            }
 
-            $stime = $balloonsData['submittime'];
+        $minumumNumberOfBalloons = (int)$this->config->get('minimum_number_of_balloons');
```
eldering: typo
```suggestion
        $minimumNumberOfBalloons = (int)$this->config->get('minimum_number_of_balloons');
```

webapp/src/Service/BalloonService.php
```diff
@@ -144,41 +135,72 @@ public function collectBalloonTable(Contest $contest, bool $todo = false): array
...
+            // Handing out balloons for problems that have not been solved pre-freeze poses a potential information leak, so these are always excluded.
+            // So to decide whether to skip showing a balloon:
+            //   1. Check whether the scoreboard has been frozen.
+            //   2. Check whether the team has exceeded minimum number of balloons.
+            //   3. Check whether the problem been solved pre-freeze.
+            $stime = $balloonsData['submittime'];
+            if (isset($freezetime) && $stime >= $freezetime) {
+                if (key_exists($balloonsData['teamid'], $relevantBalloonSummaries) &&
+                    count($relevantBalloonSummaries[$balloonsData['teamid']]) >= $minumumNumberOfBalloons) {
+                    continue;
```
eldering: What if `key_exists` returns false but `$minumumNumberOfBalloons == 0`?

eldering:
> This is exactly why these are only shown for problems that have been solved pre-freeze by teams in the same sortorder. So in your example the team would _not_ get a balloon. The entire goal is to not leak any information that was not available pre-freeze while still being able and hand out balloons when desired. I've tried making this abundantly clear in the code and the doc -more in the first than in the second- any suggestions for how to improve?

Ok, I overlooked that you handled that case.

However, I'm still not a fan as this opens up a can of worms of inconsistencies. Next, someone comes and complains that the balloons API endpoint is not in sync with submissions/judgings/scoreboard, and I really don't want to go there.
 
> Adding the testcases as suggested by @vmcj will help discover and prevent regressions on these possible edge-cases. I've given this quite some thought and could not come up with relevant edge-cases within the scope of the feature. I'm open to any and all edge-cases you can think of @eldering. Would be great to address those beforehand.

I can't think of any right now (didn't think too carefully), but this feels the same as the first to solve balloon in contest where we uncovered various bugs, and finally gave up on fixing it. And tests are not a silver bullet.

Anyways, if no-one else objects, then I'd say we must at least have a big disclaimer that we don't guarantee consistency between balloon and other data, so we can keep this completely self-contained inside balloon-only code.

Kevinjil:
> The entire goal is to not leak any information that was not available pre-freeze while still being able and hand out balloons when desired.

I kind of agree with @eldering that this feels like skating on thin ice. There are numerous situations, for now leaving out the likelyhood, that will still leak information. Consider for example:
1. Problem `A` looks difficult, but is trivial to solve.
2. Few teams `X_1` ... `X_n` with `0 < n << #teams` identify this pre-freeze and are handed a balloon.
3. During the freeze, some bottom team `Y_1` also sees the real difficulty and solves it, receiving a balloon.
4. The other `Y_2` ... `Y_m` teams also want this balloon and focus on it, flooding the contest floor with this balloon and definitely leaking  information.

This example is exaggerated to illustrate the potential leakage. One could define some `n >> 1` or `n = p * #teams` with `0 < p < 1` as a threshold for handing out balloons, limiting the risk but never completely taking it away. However, perhaps by Murphy's law, eventually we will get a situation where such feature leaks some (crucial) info.

How we should weigh this risk against the potential joy perceived by teams still being handed a balloon, is where I'm kind of undecided. Right now, I lean towards having a robust automated system and having the jury (NOT the balloon runners) explicitly decide on encouragement balloons.

webapp/src/Service/BalloonService.php
tuupke: Good catch, fixing.

nickygerritsen:
This whole feature is still behind a config option. If you don't want to leak this information, disable the option?

I do agree with @eldering that we should add a big fat warning when enabling the feature btw.

webapp/src/Service/BalloonService.php
tuupke: Yeah I had a better implementation at first but then phpstan prompted me that I needed to change something. Apparently it has retracted its old feedback so will be reverting back to that implementation.

Kevinjil:
> This whole feature is still behind a config option. If you don't want to leak this information, disable the option?
> 
> I do agree with @eldering that we should add a big fat warning when enabling the feature btw.

Fair point, a big warning (also on the config checker page) seems right here.

tuupke:
I totally agree that there is a potential for a data leak but since this behavior is (1) both disabled by default and (2) there is now a potential for (at least) less data leakage and inconsistencies compared to the old setting. I always found it surprising we did not have this disclaimer before so definitely a good idea and will add it for sure.

Just an observation but what I find kind of funny is that there are suggestions that this can be solved on an organizational level with the old (non-disclaimer, 'show every balloon during the freeze') option because people can reason about all these cases, while at the same time arguing that there are so many edge-cases that we need a disclaimer to warn them. (Which I again agree with)

Wrt to the (internal) inconsistencies. This is why it is *should* all be DRYed up by using the `BalloonService`. The `UpdateController` used to have it's own implementation and is now using the `BalloonService`. If anything this commit can (read, should) be cherry-picked since it definitely is an improvement on consistency. Inconsistencies between endpoints like balloons and scoreboard are expected since they are explicitly broken when using this feature.

--
Edit, already added a warning. But showing this all the time might be a bit much :sweat_smile: 
image

--
Edit 2, removed the flashed message on the main page. Now added (always shown) flashed messages on the configuration and balloons pages. Config check added.

webapp/src/Service/BalloonService.php
```diff
@@ -144,41 +136,71 @@ public function collectBalloonTable(Contest $contest, bool $todo = false): array
...
+            //   2. Check whether the team has exceeded minimum number of balloons.
+            //   3. Check whether the problem been solved pre-freeze.
+            $stime = $balloonsData['submittime'];
+            if ($ignorePreFreezeSolves === false && isset($freezetime) && $stime >= $freezetime) {
+                if (count($relevantBalloonSummaries[$balloonsData['teamid']] ?? []) >= $minimumNumberOfBalloons) {
+                    continue;
+                }
+
+                // Check if problem has been solved before the freeze by someone in the same sortorder to prevent information leak.
+                // The DOMjudge team (that commonly runs jury submissions) has so must be ignored.
```
Kevinjil: This sentence looks like it was rewritten partly at some point. What should it say, as I don't see how it handles the DOMjudge team as an edge case?

tuupke: Yeah this reads a bit weird. I think I wrote this in one go. Will rephrase.

etc/db-config.yaml
```diff
@@ -255,11 +255,16 @@
             default_value: false
             public: true
             description: Show results of TOO-LATE submissions in team interface?
-        -   name: show_balloons_postfreeze
+        -   name: minimum_number_of_balloons
+            type: int
+            default_value: 0
+            public: true
+            description: How many balloons to hand out ignoring freeze time. Only hands out balloons for problems solved pre-freeze.
```
vmcj:
```suggestion
            description: How many balloons to hand out ignoring freeze time. Only hands out balloons for problems solved pre-freeze by other teams.
```

```suggestion
            description: How many balloons to hand out ignoring freeze time. Limited on problems solved pre-freeze.
```

etc/db-config.yaml
```diff
@@ -255,11 +255,16 @@
...
+        -   name: any_balloon_postfreeze
```
vmcj, line 271: How is this option different than #Balloons>#Problems?

If the intent is to disclose all the information it's currently unclear.

tuupke: No it is not the same. 'Normal' behavior is to only hand out balloons for problems solved pre freeze. Assume no problem is solved pre-freeze, then setting `#Balloons>#Problems` still does not result in any balloon being handed out while in the freeze. This option ensures balloons are handed out in these kinds of situations

PR #3307, opened 2025-12-20
Use the default packaged chroot

vmcj:
This should speedup the integration test by quite a lot. We don´t change the `dj_make_chroot` that often and already have tests to build the chroot so we still do it, just not in this test anymore.

This was discussed at NWERC25.


vmcj:
> Perhaps squash them before merging.

:+1: but I'll also wait for @meisterT his approval.

PR #3308, opened 2025-12-21
PHP8.5 deprecations

vmcj:
@nickygerritsen not sure if you already did this in another PR?


nickygerritsen:
I did but we can close it. Note that that PR contains more, which you might want to either apply here or in #3269. See #3280.

PR #3314, opened 2025-12-27
Install and apply sane Rector rules

nickygerritsen:
I removed some rules we didn't like before, but I think this is a good set of rules to use.

The rector.php file can be used to rerun rector.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -758,14 +758,14 @@ class JudgeDaemon
                 }
             }
             if ($trial == BACKOFF_STEPS) {
-                $errstr = $errstr . " Retry limit reached.";
+                $errstr .= " Retry limit reached.";
             } else {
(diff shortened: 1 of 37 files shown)


vmcj:
Why is this better? https://github.com/DOMjudge/domjudge/pull/3314/files#diff-ebb0d9a14289d7ffedd4c5f1df222e3215630edd6e6b96aaeaab21dd6f61bcb3R975 maybe we should add rector as an CI job to detect such cases like we have for PHPCS, otherwise I think we will get both styles inconsistently.

vmcj (approved):

nickygerritsen:
> Why is this better? https://github.com/DOMjudge/domjudge/pull/3314/files#diff-ebb0d9a14289d7ffedd4c5f1df222e3215630edd6e6b96aaeaab21dd6f61bcb3R975 maybe we should add rector as an CI job to detect such cases like we have for PHPCS, otherwise I think we will get both styles inconsistently.

Way less typing without losing much legibility imho. But if others disagree we can drop this one

PR #3315, opened 2025-12-29
Allow QUIET=0 as temporary fix in installation scripts

vmcj:
I encountered this when changing our update_*.sh scripts. We use `QUIET=1` there, I replaced that with `QUIET=0` which would enter `0` in a couple of places where we don't really need that. The same would happen with `QUIET=2` which would be a strange but valid boolean.

Those fixes should make it hopefully easier to read and remove this specific issue. I couldn't find a nice way to do an or to remove the duplication so we end up with some (ugly) replication of the flags.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,7 +195,7 @@ paths.mk:
 maintainer-conf: inplace-conf-common dependencies-dev
 inplace-conf: inplace-conf-common dependencies
 inplace-conf-common: dist
-	./configure $(subst 1,-q,$(QUIET)) --prefix=$(CURDIR) \
+	./configure $(QUIET_FLAG) --prefix=$(CURDIR) \
 	            --with-domserver_root=$(CURDIR) \
 	            --with-judgehost_root=$(CURDIR) \
 	            --with-domjudge_docdir=$(CURDIR)/doc \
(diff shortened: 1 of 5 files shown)


eldering (commented):
I'm fine with adding `QUIET_FLAG{,_CAPITAL}`, but I think right now the difference (if any?) between specifying `QUIET=0` or nothing is unclear.
I'd rather not have that difference at all, or if really necessary clearly document why/where that's needed.

vmcj:
> I'm fine with adding `QUIET_FLAG{,_CAPITAL}`, but I think right now the difference (if any?) between specifying `QUIET=0` or nothing is unclear. I'd rather not have that difference at all, or if really necessary clearly document why/where that's needed.

I was debugging the demoweb instance, where we have QUIET=1 in the scripts. As not all files are versioncontrolled I rather change the QUIET=1 to QUIET=0 as that way I can easily check where I should fix the option again after I was done on the demoweb fixing.

So I propose that we have:

Verbose:
`QUIET=`
`QUIET=0`

Quiet:
`QUIET=1...`

Which I think all reflect normal boolean rules and feel intuitive to me.

PR #3316, opened 2025-12-29
Current summary of interesting DOMjudge changes

vmcj:


ChangeLog
```diff
@@ -1,10 +1,16 @@
 DOMjudge Programming Contest Judging System
 
 Version 10.0.0DEV
----------------------------
+-----------------
+ - Removed support for multiple endpoints from judgedaemon.
+ - Remove support for cgroup v1.
+ - Reformatted the judgedaemon which might break customizations.
```
nickygerritsen: I'm not sure if this is something we should add to the changelog

vmcj: I also wondered but it is really a major overhaul where I suspect people do have customizations.

eldering: I think we should not add things like this here, but just functional changes that may affect users. If someone has patches on top of our code, then they'll find out when rebasing them, and you can always read the commit history.

ChangeLog
```diff
@@ -1,12 +1,15 @@
 DOMjudge Programming Contest Judging System
 
 Version 10.0.0DEV
----------------------------
-
- - Allow teams to be in more than one category and add more category types.
+-----------------
+ - Removed support for multiple endpoints from judgedaemon.
+ - Remove support for cgroup v1.
```
meisterT: Nit: let's use `Removed` or `Remove` in both places.

vmcj: I would blame the person writing the commit messages ;)

Will change it.

PR #3317, opened 2026-01-02
Make database in sync again with Doctrine

nickygerritsen:
When we updated to Doctrine DBAL 4.x, some things needed changing:

- `float` is now stored as `double precision` in the database
- enum cases are now in alphabetical order
- The `DC2Type` comments aren't needed anymore
- We need to be more explicit about using floats when they are typehinted as `string|float`
- The custom TinyIntType needed some massaging

### DIFF
--- a/webapp/migrations/Version20260102094726.php
+++ b/webapp/migrations/Version20260102094726.php
@@ -1,0 +1,74 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20260102094726 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Synchronize database after doctrine/dbal 4.x upgrade';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE configuration CHANGE value value LONGTEXT NOT NULL COMMENT \'Content of the configuration variable (JSON encoded)\'');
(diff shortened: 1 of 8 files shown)


webapp/src/Entity/Problem.php
```diff
@@ -56,7 +56,7 @@ class Problem extends BaseApiEntity implements
 
     #[ORM\Column(options: [
         'comment' => 'Maximum run time (in seconds) for this problem',
-        'default' => 0,
+        'default' => '0',
```
vmcj, line 59: Why is this needed? It's still a number and it feels strange to me that the DBAL maintainers intended this change to be needed?

You're probably right, but a reference to the notes why this is needed is good in case we encounter another situation where this is needed.

webapp/migrations/Version20260102094726.php
```diff
@@ -0,0 +1,74 @@
...
+{
+    public function getDescription(): string
+    {
+        return 'Synchronize database after doctrine/dbal 4.x upgrade';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE configuration CHANGE value value LONGTEXT NOT NULL COMMENT \'Content of the configuration variable (JSON encoded)\'');
```
vmcj, line 23: It looks like this only removes the `(DC2Type...)` not sure if anything else was intended? What was the reason why we had it there and now don't need it anymore?

webapp/src/Entity/Problem.php
nickygerritsen: I couldn't find a direct reference, but I think it's because floats are stored as strings in PHP.

webapp/migrations/Version20260102094726.php
nickygerritsen: Doctrine 3 required DC2Type, Doctrine 4 doesn't.

webapp/src/Entity/Problem.php
vmcj: But we have DBAL to fix this abstraction for us. But as we only have this for columns in Entities it's easy enough to remember this.

PR #3318, opened 2026-01-02
Pin to specific release

vmcj:
They've upgraded twice in the last 12 months which breaks PR's with unrelated problems.


nickygerritsen (approved):

eldering:
It would be better to mention in the commit message *what* you're pinning.

vmcj:
@eldering please merge it if you think this is good enough. @nickygerritsen needs it for his PR to merge it.

PR #3319, opened 2026-01-03
Disallow opening .php directly with GET request

vmcj:
This only worked for files in the public directory which is at the moment only index.php. This shouldn't be a real problem but doing it like this shouldn't hurt in case someone misconfigures.

Closes: https://github.com/DOMjudge/domjudge-packaging/issues/215

### DIFF
--- a/etc/nginx-conf-inner.in
+++ b/etc/nginx-conf-inner.in
@@ -17,6 +17,8 @@ set $domjudgeRoot @domserver_webappdir@/public;
 set $prefix /domjudge;
 
 # Uncomment to run it out of the root of your system
+# # Disallow opening .php files directly
+# location ~ ^/+[^/]+\.php$ { return 404; }
 # location / {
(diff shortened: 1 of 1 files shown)


Kevinjil:
Would this conflict with #3263, as URLs ending with `.php` are allowed for entities with external ids?

vmcj:
> Would this conflict with #3263, as URLs ending with `.php` are allowed for entities with external ids?

Maybe, see my next commit which should be more restrictive in what it picks.

nickygerritsen (approved):
We could even hardcode it to index.php of we want since that's the only file in the public folder, right?

vmcj:
> We could even hardcode it to index.php of we want since that's the only file in the public folder, right?

It is now, but I wanted to try to fix it for when we also get others.

Considering that Symfony only needs `index.php` I wonder if that extra bit of "security" is needed.

PR #3320, opened 2026-01-03
Eagerly load users to avoid integrity violations on rejudgings page

meisterT:
When listing rejudgings, users were being lazy-loaded. In some scenarios (likely involving session-stored user objects), the resulting Doctrine proxy objects could end up in a "dirty" state with uninitialized properties (e.g. calling `->getName`).

Since `calculateTodo()` triggers a `flush()` during rejudging list rendering, Doctrine would attempt to sync these corrupted objects back to the database, resulting in a `Column 'username' cannot be null` error.

Eagerly joining `start_user` and `finish_user` ensures these entities are fully populated from the start, preventing the creation of corrupted proxies and subsequent invalid database updates.

This was likely caused by a doctrine update.

### DIFF
--- a/webapp/src/Controller/Jury/RejudgingController.php
+++ b/webapp/src/Controller/Jury/RejudgingController.php
@@ -66,9 +66,14 @@ class RejudgingController extends BaseController
     public function indexAction(): Response
     {
         $curContest = $this->dj->getCurrentContest();
+        // Note that we are eagerly selecting and joining on the users to avoid
+        // integrity violations, as doctrine proxy objects could be uninitialized
+        // but dirty and the following flush in calculateTodo would fail.
         $queryBuilder = $this->em->createQueryBuilder()
-            ->select('r')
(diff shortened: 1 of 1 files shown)


webapp/src/Controller/Jury/RejudgingController.php
```diff
@@ -67,8 +67,10 @@ public function indexAction(): Response
     {
         $curContest = $this->dj->getCurrentContest();
         $queryBuilder = $this->em->createQueryBuilder()
-            ->select('r')
-            ->from(Rejudging::class, 'r');
+            ->select('r', 'su', 'fu')
```
vmcj, line 73: does it work without selecting the `su` & `fu`? So only the joins,

meisterT: No.

Also added a code comment to explain why we have the select/join.

PR #3321, opened 2026-01-03
Support both localhost and % host part for mysql users

eldering:
This reverts d6a82b235e59699059ae3aa9629d7 to use localhost again, but also keeps the user with `%` host. MySQL seems to treat these as fully separate users and when connecting with a mysql client locally, the `%` host results in the error:
```
ERROR 1045 (28000): Access denied for user 'domjudge'@'localhost' (using password: YES)
```

Also remove a verbose print statement left over from debugging.


Kevinjil:
`'$domjudge_DBUSER'@'%'` should allow access from localhost, unless there is a more specific user denying it access in your DB setup. Sometimes there is a `''@localhost` user denying access which by the [defined sorting](https://dev.mysql.com/doc/refman/8.4/en/connection-access.html) takes priority.

What is in your users table?

PR #3322, opened 2026-01-04
W3C Webstandard fixes to use latest release again.

vmcj:
This will require a lot of pushing, testing this locally is difficult as it requires a couple of changes to the script.


vmcj:
Solved with other PRs.

PR #3328, opened 2026-01-09
Change shadow mode to a per contest flag 

### LINKED ISSUE #3272: Change shadow mode to a per contest flag
Now that we always use external IDs, it should be easy to make the shadow mode a per contest setting, so you can have a normal contest and a shadowed one in the same instance

nickygerritsen:
This also adds support for the thing suggested by @vmcj: we can now decide whether to use internal or external judgings/runs when in shadow mode. 

Fixes #3272


doc/manual/shadow.rst
```diff
@@ -45,26 +60,22 @@ team linked to your account, it will complain that it can't import jury
...
-create an external contest source. Select the contest to import into and enter
-the source you want to import from.
-
-Another alternative is to use these DOMjudge only keys in the `contest.yaml`::
+You can also configure shadow mode settings via the ``contest.yaml`` file using
+these DOMjudge-specific keys::
 
   shadow:
-    type: contest-api
+    useJudgements: true
```
vmcj: I think I prefer `use_judgements` (so snake case) as most of those are written as full lowercase (but you could also state all keys are now camelCase..)

vmcj (commented):
I think we need an ChangeLog entry.

nickygerritsen:
> I think we need an ChangeLog entry.

Done

doc/manual/shadow.rst
nickygerritsen: Agreed and done

meisterT: unresolving since it is still camelCase, also below

webapp/src/Controller/API/RunController.php
```diff
@@ -112,34 +116,50 @@ public function singleAction(Request $request, string $id): Response
...
             $queryBuilder
-                ->andWhere('jr.runid = :last_id')
+                ->andWhere($this->useExternalRuns ? 'jr.extrunid = :last_id' : 'jr.runid = :last_id')
                 ->setParameter('last_id', $request->query->get('last_id'));
         }
 
         if ($request->query->has('judging_id')) {
             $queryBuilder
-                ->andWhere('jr.judging = :judging_id')
+                ->andWhere($this->useExternalRuns ? 'jr.external_judgement = :judging_id' : 'jr.judging = :judging_id')
```
meisterT: does this need to be `extjudgementid` or something similar?

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -462,7 +462,7 @@ protected function importFromCcsApi(array $eventsToSkip, ?callable $progressRepo
      */
     protected function importFromContestArchive(array $eventsToSkip, ?callable $progressReporter = null): bool
     {
-        $file = fopen($this->source->getSource() . '/event-feed.ndjson', 'r');
+        $file = fopen($this->contest->getExternalSourceSource() . '/event-feed.ndjson', 'r');
```
meisterT, line 475: I know this is not new code, but I think we should check `$file` for nullness?

webapp/src/Entity/Contest.php
```diff
@@ -325,6 +325,57 @@ public function getScoreboardTypeString(): string
...
+    #[Serializer\Exclude]
+    private string|float|null $externalSourceLastPollTime = null;
+
+    #[ORM\Column(
+        type: 'smallint',
+        nullable: true,
+        options: ['comment' => 'Last HTTP code received by event feed reader', 'unsigned' => true]
+    )]
+    #[Serializer\Exclude]
+    private ?int $externalSourceLastHTTPCode = null;
```
meisterT: nit: `externalSourceLastHttpCode`?

webapp/src/Controller/API/RunController.php
nickygerritsen: yes it does. also for judgement=>submission

doc/manual/shadow.rst
nickygerritsen: thanks, fixed

webapp/src/Command/ImportEventFeedCommand.php
```diff
@@ -165,27 +159,24 @@ protected function loadSource(InputInterface $input, ?string $contestId = null):
...
-            ->join('ecs.contest', 'c')
-            ->andWhere('c.cid = :cid')
-            ->setParameter('cid', $contestId)
-            ->getQuery()
-            ->getOneOrNullResult();
-        if ($source === null) {
-            $this->style->error('Contest does not have an external contest configured yet');
+        $contest = $this->em->getRepository(Contest::class)->findByExternalId($contestId);
+        if ($contest === null) {
+            $this->style->error('Contest not found');
```
vmcj:
```suggestion
            $this->style->error('Contest not found.');
```

webapp/src/Command/ImportEventFeedCommand.php
```diff
@@ -165,27 +159,24 @@ protected function loadSource(InputInterface $input, ?string $contestId = null):
...
+            return false;
+        }
+        if (!$contest->isExternalSourceEnabled()) {
+            $this->style->error('Contest does not have shadow mode enabled');
```
vmcj:
```suggestion
            $this->style->error('Contest does not have shadow mode enabled.');
```

webapp/src/Controller/API/JudgementController.php
```diff
@@ -36,6 +38,8 @@ class JudgementController extends AbstractRestController implements QueryObjectT
      */
     protected readonly array $verdicts;
 
+    private ?bool $useExternalJudgements = null;
```
vmcj: Why can't we pick a default here?

webapp/src/Controller/API/JudgementController.php
```diff
@@ -156,12 +180,15 @@ protected function getQueryBuilder(Request $request): QueryBuilder
...
+        if ($this->useExternalJudgements) {
+            return 'j.externalid';
+        }
         return 'j.judgingid';
     }
 
     public function transformObject($judging): JudgingWrapper
     {
-        /** @var Judging $judging */
+        /** @var Judging|ExternalJudgement $judging */
```
vmcj: Don't both inherit from the same (abstract) class? Should they?

webapp/src/Controller/API/JudgementController.php
nickygerritsen: I'm fine with a default. We set it every time, but doesn't hurt.

webapp/src/Controller/API/JudgementController.php
nickygerritsen: They do, but it doesn't have any methods (base api entity or something). Having a base class makes sense though. Let's see if I can make that work.

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -156,7 +156,7 @@ public function getSourceContest(): Contest
             throw new LogicException('The contest source is not valid');
```
vmcj:
```suggestion
            throw new LogicException('The contest source is not valid.');
```

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -165,7 +165,7 @@ public function getSourceContestId(): int
...
```
vmcj:
```suggestion
            throw new LogicException('The contest source is not valid.');
```

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -304,14 +304,14 @@ public function import(bool $fromStart, array $eventsToSkip, ?callable $progress
...
 
         $this->loadPendingEvents();
-        return match ($this->source->getType()) {
-            ExternalContestSource::TYPE_CCS_API => $this->importFromCcsApi($eventsToSkip, $progressReporter),
-            ExternalContestSource::TYPE_CONTEST_PACKAGE => $this->importFromContestArchive($eventsToSkip, $progressReporter),
-            default => false,
+        return match ($this->contest->getExternalSourceType()) {
+            ExternalContestSourceType::CCS_API => $this->importFromCcsApi($eventsToSkip, $progressReporter),
+            ExternalContestSourceType::CONTEST_PACKAGE => $this->importFromContestArchive($eventsToSkip, $progressReporter),
+            null => throw new LogicException('External source type is not set'),
```
vmcj:
```suggestion
            null => throw new LogicException('External source type is not set.'),
```

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -541,36 +546,36 @@ protected function readEventsFromFile($filePointer, callable $callback): void
 
     protected function loadContest(): void
     {
-        if ($this->source === null) {
-            throw new LogicException('You need to call setSource() first');
+        if ($this->contest === null) {
+            throw new LogicException('You need to call setSourceContest() first');
```
vmcj:
```suggestion
            throw new LogicException('You need to call setSourceContest() first.');
```

webapp/src/Service/ImportExportService.php
```diff
@@ -353,17 +378,19 @@ public function importContestData(mixed $data, ?string &$errorMessage = null, ?s
...
-                $fieldFunc = 'set'.ucwords($field);
-                $fieldArgs = [$value];
-                if (method_exists($externalSource, $fieldFunc)) {
-                    $externalSource->$fieldFunc(...$fieldArgs);
+                // Map shadow fields to Contest setter methods
+                $fieldFunc = 'setExternalSource' . ucfirst($inflector->camelize($field));
+                if ($field === 'type') {
+                    // Type is now an enum
+                    $value = ExternalContestSourceType::from($value);
+                }
```
vmcj:
```suggestion
                if ($field === 'type') {
                    // Type is now an enum
                    $value = ExternalContestSourceType::from($value);
                }
                // Map shadow fields to Contest setter methods
                $fieldFunc = 'setExternalSource' . ucfirst($inflector->camelize($field));
```

webapp/src/Controller/API/JudgementController.php
vmcj: My remark is mostly to make PHPStan more powerful, less cases is better reasoning.

webapp/src/Controller/API/JudgementController.php
nickygerritsen: I made something work. For `getRuns` we need to ignore a PHPStan error about covariance but not much we can do about it I think.

webapp/src/Controller/API/RunController.php
```diff
@@ -112,34 +117,50 @@ public function singleAction(Request $request, string $id): Response
...
+                ->leftJoin('jr.testcase', 'tc')
+                ->leftJoin('j.submission', 's')
+                ->leftJoin('j.contest', 'c')
+                ->select('jr, j, tc, c')
+                ->andWhere('j.contest = :cid')
+                // With the new judgehost API we pre-create the judging_runs; only expose those who correspond to a real run
+                // on a judgehost.
+                ->andWhere('jr.endtime IS NOT NULL')
+                ->setParameter('cid', $contestId);
+        }
```
vmcj, line 140:
```suggestion
        $queryBuilder = $this->em->createQueryBuilder()
                ->leftJoin('jr.testcase', 'tc')
                ->leftJoin('j.submission', 's')
                ->leftJoin('jr.contest', 'c')
                ->select('jr, j, tc, c')
                ->andWhere('jr.contest = :cid')
                ->setParameter('cid', $contestId);
        if ($this->useExternalRuns) {
            $queryBuilder
                ->from(ExternalRun::class, 'jr')
                ->leftJoin('jr.external_judgement', 'j')
        } else {
            $queryBuilder
                ->from(JudgingRun::class, 'jr')
                ->leftJoin('jr.judging', 'j')
                // With the new judgehost API we pre-create the judging_runs; only expose those who correspond to a real run
                // on a judgehost.
                ->andWhere('jr.endtime IS NOT NULL');
        }
```

webapp/src/DataTransferObject/JudgingWrapper.php
```diff
@@ -2,15 +2,15 @@
 
 namespace App\DataTransferObject;
 
-use App\Entity\Judging;
+use App\Entity\AbstractJudgement;
 use JMS\Serializer\Annotation as Serializer;
 
-class JudgingWrapper
+readonly class JudgingWrapper
```
vmcj, line 8: PHPCS doesn't agree with this, I think we never care for that warning?

nickygerritsen: Seems it doesn't understand it indeed. Maybe it is time to remove that warning after all? Unless we can upgrade it or something.

vmcj: I don't see where you ignored it, and I think it's another one as what I PR'd last time.

But for this specific case I think I disagree with PSR that this would be worse, might be I don't understand the involved risk..

webapp/src/Controller/API/RunController.php
```diff
@@ -112,34 +117,45 @@ public function singleAction(Request $request, string $id): Response
...
+                ->leftJoin('jr.testcase', 'tc')
+                ->leftJoin('j.submission', 's')
+                ->leftJoin('jr.contest', 'c')
+                ->select('jr, j, tc, c')
+                ->andWhere('jr.contest = :cid')
+                ->setParameter('cid', $contestId);
+        if ($this->useExternalRuns) {
+            $queryBuilder
+                ->from(ExternalRun::class, 'jr')
+                ->leftJoin('jr.external_judgement', 'j')
```
vmcj, line 128:
```suggestion
                ->leftJoin('jr.external_judgement', 'j');
```

webapp/src/DataTransferObject/JudgingWrapper.php
nickygerritsen: I did not ignore it, but I'm saying maybe we should. I think readonly is better (the drawback is you can't extend the class).

vmcj: https://github.com/DOMjudge/domjudge/blob/main/.github/workflows/phpcodesniffer.yml#L17-L18

I think we already have the latest/greatest version... so either we need to use the dev branch or go for the ignore route. For now I would prefer the ignore route.

nickygerritsen: How did the ignore work? Feel free to suggest/push it to this branch if you want.

vmcj: I will do it globally tomorrow.

vmcj: https://github.com/squizlabs/PHP_CodeSniffer/pull/3728 It seems it should be in main. So in the future we can remove it.

webapp/src/Controller/API/RunController.php
vmcj: It seems there is not a nice way to remove this duplication.

webapp/src/DataTransferObject/JudgingWrapper.php
nickygerritsen: Nice!

webapp/src/Controller/API/RunController.php
nickygerritsen: I think I found a nice way, please check.

webapp/src/Entity/ExternalRun.php
```diff
@@ -45,21 +50,26 @@ class ExternalRun
         scale: 9,
         options: ['comment' => 'Time run ended', 'unsigned' => true]
     )]
+    #[Serializer\Exclude]
     private string|float $endtime;
 
     #[ORM\Column(options: ['comment' => 'Running time on this testcase'])]
+    #[Serializer\Exclude]
```
vmcj, line 57: ~~Why do we not share this information in the API?~~

We do in getApiRuntime()

webapp/src/Controller/API/RunController.php
vmcj: Yes, so simple and much better.

PR #3329, opened 2026-01-09
Add config option to rate limit submissions per time window.

meisterT:


etc/db-config.yaml
```diff
@@ -172,6 +172,11 @@
             default_value: true
             public: false
             description: Activate a judgehost when it registers for the first time.
+        -   name: submission_rate_limit
+            type: array_keyval
+            default_value: []
+            public: false
```
vmcj: Shouldn´t the rules be clear if someone uses this option?

webapp/src/Service/SubmissionService.php
```diff
@@ -605,6 +605,55 @@ public function submitSolution(
             $user = $this->em->getRepository(User::class)->find($user->getUserid());
         }
 
+        // Submission rate limiting
+        $rateLimits = $this->config->get('submission_rate_limit');
+        if (!empty($rateLimits) && !$this->dj->checkrole('jury')) {
```
vmcj:
```suggestion
        if (!(empty($rateLimits) || $this->dj->checkrole('jury'))) {
```

etc/db-config.yaml
meisterT: Yeah, good catch. Will fix tomorrow.

webapp/tests/Unit/Controller/API/SubmissionRateLimitTest.php
```diff
@@ -0,0 +1,179 @@
...
+        /** @var Contest $contest */
+        $contest = static::getContainer()->get(EntityManagerInterface::class)
+            ->getRepository(Contest::class)
+            ->findOneBy(['shortname' => 'demo']);
+        return $contest->getExternalid();
+    }
+
+    protected function setupTestUser(string $username, array $roles): void
+    {
+        $em = static::getContainer()->get(EntityManagerInterface::class);
```
vmcj: Why can´t we use a Fixture here?

etc/db-config.yaml
vmcj: Maybe even display it in the UI somewhere, either before you submit or in the team main page.

meisterT: New code (not completely done yet): image

webapp/src/Service/SubmissionService.php
meisterT: This is obsolete by the refactoring I did to reuse this on the team side

webapp/tests/Unit/Controller/API/SubmissionRateLimitTest.php
meisterT: Done

etc/db-config.yaml
vmcj: I wonder if I would want to know this ahead of time already as it might have an effect on your (theoretical) team strategy. In the future we would also need this in the submit client.

meisterT: If you set generous limits (as indicated in the description), there is no real need to know the limits ahead of time (for humans, not bots). If you set tight/silly limits, it is IMHO on contest organizers to communicate.

PR #3330, opened 2026-01-10
Add and use new option to stay on one filesystem.

meisterT:
While theoretically we should not leave filesystems behind, sometimes during testing new judging changes `/proc` was still mounted, causing `evict` to traverse into proc where it gets stuck in a loop causing weird failure modes.

### DIFF
--- a/judge/evict.cc
+++ b/judge/evict.cc
@@ -25,9 +25,11 @@ std::string_view progname;
 int be_verbose;
 int show_help;
 int show_version;
+int one_file_system;
 
 struct option const long_opts[] = {
 	{"verbose", no_argument,       NULL,         'v'},
+	{"one-file-system", no_argument, NULL,       'x'},
 	{"help",    no_argument,       &show_help,    1 },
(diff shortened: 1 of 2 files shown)


judge/evict.cc
```diff
@@ -62,33 +65,37 @@ void evict_directory(const std::string& dirname) {
...
+				if (be_verbose) logmsg(LOG_DEBUG, "Skipping {}: different filesystem", entry_path);
 				continue;
 			}
+
 			if (S_ISDIR(s.st_mode)) {
 				/* Recurse into subdirectories */
-				evict_directory(entry_path);
-			} else {
+				evict_directory(entry_path, root_dev);
+			} else if (S_ISREG(s.st_mode)) {
```
vmcj, line 82: Can you add the last `} else {` with a comment explaining what we do to non dirs/ non regular files? I assume we should never encounter those with `evict` but i think documenting what should happen makes it easier to understand what `evict` does.

meisterT: Done

Issue #1387, opened 2021-12-10
Import jury solutions when expected results fails.


vmcj:
As reported by @eldering:

When uploading jury solutions; Accept any submission (since it's just meta-data), but simply not recognize any expected result when parsing fails, so it's easier to detect that we didn't parse any @EXPECTED_RESULTS@ strings.

vmcj:
@eldering can this be closed? I think you fixed those issues?

eldering:
I didn't fix anything specifically related to this, and also am not sure it is fixed. So best to leave open, also to track the reference to that problem package issue.

PR #3337, opened 2026-01-17
Add support for partial scoring / optimization problems

meisterT:
See #2518

### DIFF
--- a/webapp/templates/jury/problem_testcases.html.twig
+++ b/webapp/templates/jury/problem_testcases.html.twig
@@ -39,149 +39,122 @@
             </tr>
             </thead>
             <tbody>
-            {% set lastLineage = [] %}
-            {% for rank, testcase in testcases %}
-                {% set lineage = [] %}
-                {% set g = testcase.testcaseGroup %}
-                {% for i in 0..10 %}
-                    {% if g %}
-                        {% set lineage = [g]|merge(lineage) %}
-                        {% set g = g.parent %}
-                    {% endif %}
-                {% endfor %}
-
-                {% set commonPrefixLength = 0 %}
-                {% set foundDiff = false %}
-                {% for i in 0..10 %}
-                    {% if not foundDiff %}
-                        {% if lastLineage[i] is defined and lineage[i] is defined and lastLineage[i].testcaseGroupId == lineage[i].testcaseGroupId %}
-                            {% set commonPrefixLength = i + 1 %}
-                        {% else %}
-                            {% set foundDiff = true %}
-                        {% endif %}
-                    {% endif %}
-                {% endfor %}
-
-                {% for i in 0..10 %}
-                    {% if i >= commonPrefixLength and i < lineage|length %}
-                        {% set group = lineage[i] %}
-                        {% set dispName = group.name | split('/') | last %}
-
-                        <tr class="table-secondary">
-                            <td colspan="{{ allowEdit ? 8 : 6 }}" style="padding-left: {{ (i * 1.5) + 0.5 }}rem;">
-                                <i class="fas {{ i == 0 ? 'fa-sitemap' : 'fa-folder-open' }} me-1"></i>
-                                <strong>{{ dispName }}</strong>
-                                <span class="badge ms-2 scoring-agg-pill scoring-agg-{{ group.aggregationType.value }}">{{ group.aggregationType.value | upper }}</span>
-                                {% if group.acceptScore is not null %}
-                                    <span class="badge ms-2 scoring-accept-badge">Accept score: {{ (group.acceptScore + 0) }}</span>
-                                {% endif %}
-                            </td>
-                        </tr>
-                    {% endif %}
-                {% endfor %}
-
-                {% if lineage|length == 0 and lastLineage|length > 0 %}
+            {% for row in rows %}
+                {% if row.type == 'group' %}
+                    {% set group = row.group %}
+                    {% set dispName = group.name | split('/') | last %}
+                    <tr class="table-secondary">
+                        <td colspan="{{ allowEdit ? 8 : 6 }}" style="padding-left: {{ (row.level * 1.5) + 0.5 }}rem;">
+                            <i class="fas {{ row.level == 0 ? 'fa-sitemap' : 'fa-folder-open' }} me-1"></i>
+                            <strong>{{ dispName }}</strong>
+                            <span class="badge ms-2 scoring-agg-pill scoring-agg-{{ group.aggregationType.value }}">{{ group.aggregationType.value | upper }}</span>
+                            {% if group.acceptScore is not null %}
+                                <span class="badge ms-2 scoring-accept-badge">Accept score: {{ (group.acceptScore + 0) }}</span>
+                            {% endif %}
+                        </td>
+                    </tr>
+                {% elseif row.type == 'no_group' %}
                     <tr class="table-secondary">
                         <td colspan="{{ allowEdit ? 8 : 6 }}">
                             <strong>No group</strong>
                         </td>
                     </tr>
-                {% endif %}
-
-                {% set lastLineage = lineage %}
-                <tr>
-                    {% if allowEdit %}
-                        <td rowspan="2">
-                            <a href="{{  path('jury_testcase_delete', {'testcaseId': testcase.testcaseid}) }}"
-                               title="delete testcase {{ testcase.testcaseid }}"
-                               onclick="confirm('Proceed with the deletion of testcase {{ testcase.testcaseid }}?');">
-                                <i class="fas fa-trash-alt"></i>
+                {% else %}
+                    {% set testcase = row.testcase %}
+                    <tr>
+                        {% if allowEdit %}
+                            <td rowspan="2">
+                                <a href="{{  path('jury_testcase_delete', {'testcaseId': testcase.testcaseid}) }}"
+                                   title="delete testcase {{ testcase.testcaseid }}"
+                                   onclick="confirm('Proceed with the deletion of testcase {{ testcase.testcaseid }}?');">
+                                    <i class="fas fa-trash-alt"></i>
+                                </a>
+                            </td>
+                        {% endif %}
+                        <td rowspan="2" class="testrank">
+                            {% if allowEdit %}
+                                <a href="{{ path('jury_problem_testcase_move', {'probId': problem.probid, 'rank': testcase.rank, 'direction': 'up'}) }}"
+                                   title="Move testcase up"><i class="fas fa-arrow-up"></i></a><br>
+                            {% endif %}
+                            {{ testcase.rank }}
+                            {% if allowEdit %}
+                                <br>
+                                <a href="{{ path('jury_problem_testcase_move', {'probId': problem.probid, 'rank': testcase.rank, 'direction': 'down'}) }}"
+                                   title="Move testcase down"><i class="fas fa-arrow-down"></i></a>
+                            {% endif %}
+                        </td>
+                        <td rowspan="2" class="testsample">
+                            <input type="checkbox" name="sample[{{ testcase.rank }}]" value="sample"
+                                   {% if testcase.sample %}checked{% endif %} />
+                        </td>
+                        <td class="filename">
+                            <a href="{{ path('jury_problem_testcase_fetch', {'probId': problem.probid, 'rank': testcase.rank, 'type': 'input'}) }}">
+                                {{ testcase.downloadName | split('/') | last }}.{{ extensionMapping.input }}
                             </a>
                         </td>
-                    {% endif %}
-                    <td rowspan="2" class="testrank">
+                        <td class="size">
+                            {{ row.inputSize | printSize }}
+                        </td>
+                        <td class="md5 small">
+                            {{ testcase.md5sumInput }}
+                        </td>
                         {% if allowEdit %}
-                            <a href="{{ path('jury_problem_testcase_move', {'probId': problem.probid, 'rank': testcase.rank, 'direction': 'up'}) }}"
-                               title="Move testcase up"><i class="fas fa-arrow-up"></i></a><br>
+                            <td><div class="input-group">
+                                <label class="form-control text-truncate" for="update_input_{{testcase.rank}}" id="update_input_label_{{testcase.rank}}">new input file</label>
+                                <input type="file" class="d-none custom-file-input" name="update_input[{{ testcase.rank }}]" id="update_input_{{testcase.rank}}">
+                            </div></td>
                         {% endif %}
-                        {{ testcase.rank }}
-                        {% if allowEdit %}
-                            <br>
-                            <a href="{{ path('jury_problem_testcase_move', {'probId': problem.probid, 'rank': testcase.rank, 'direction': 'down'}) }}"
-                               title="Move testcase down"><i class="fas fa-arrow-down"></i></a>
-                        {% endif %}
-                    </td>
-                    <td rowspan="2" class="testsample">
-                        <input type="checkbox" name="sample[{{ testcase.rank }}]" value="sample"
-                               {% if testcase.sample %}checked{% endif %} />
-                    </td>
-                    <td class="filename">
-                        <a href="{{ path('jury_problem_testcase_fetch', {'probId': problem.probid, 'rank': testcase.rank, 'type': 'input'}) }}">
-                            {{ testcase.downloadName | split('/') | last }}.{{ extensionMapping.input }}
-                        </a>
-                    </td>
-                    <td class="size">
-                        {{ testcaseData[testcase.rank].input_size | printSize }}
-                    </td>
-                    <td class="md5 small">
-                        {{ testcase.md5sumInput }}
-                    </td>
-                    {% if allowEdit %}
-                        <td><div class="input-group">
-                            <label class="form-control text-truncate" for="update_input_{{testcase.rank}}" id="update_input_label_{{testcase.rank}}">new input file</label>
-                            <input type="file" class="d-none custom-file-input" name="update_input[{{ testcase.rank }}]" id="update_input_{{testcase.rank}}">
-                        </div></td>
-                    {% endif %}
-                    <td class="testdesc">
-                        <span class="testcase-description">{{ testcase.description(true) | descriptionExpand }}</span>
-                        {% if allowEdit %}
-                            <textarea class="d-none form-control testcase-description-field" name="description[{{ testcase.rank }}]">{{ testcase.description(true) }}</textarea>
-                            <button type="button" class="float-end testcase-edit-description btn btn-secondary btn-sm">Edit</button>
-                        {% endif %}
-                    </td>
-                </tr>
-                <tr>
-                    <td class="filename">
-                        <a href="{{ path('jury_problem_testcase_fetch', {'probId': problem.probid, 'rank': testcase.rank, 'type': 'output'}) }}">
-                            {{ testcase.downloadName | split('/') | last }}.{{ extensionMapping.output }}
-                        </a>
-                    </td>
-                    <td class="size">
-                        {{ testcaseData[testcase.rank].output_size | printSize }}
-                    </td>
-                    <td class="md5 small">
-                        {{ testcase.md5sumOutput }}
-                    </td>
-                    {% if allowEdit %}
-                        <td><div class="input-group">
-                            <label class="form-control text-truncate" for="update_output_{{testcase.rank}}" id="update_output_label_{{testcase.rank}}">new output file</label>
-                            <input type="file" class="d-none custom-file-input" name="update_output[{{ testcase.rank }}]" id="update_output_{{testcase.rank}}">
-                        </div></td>
-                    {% endif %}
-                    <td class="testimage">
-                        {% if testcaseData[testcase.rank].image_size > 0 %}
-                            <a href="{{ path('jury_problem_testcase_fetch', {'probId': problem.probid, 'rank': testcase.rank, 'type': 'image'}) }}">
-                                <span class="filename">p{{ problem.probid }}.t{{ testcase.rank }}.{{ testcaseData[testcase.rank].image_type }}</span>
+                        <td class="testdesc">
+                            <span class="testcase-description">{{ testcase.description(true) | descriptionExpand }}</span>
+                            {% if allowEdit %}
+                                <textarea class="d-none form-control testcase-description-field" name="description[{{ testcase.rank }}]">{{ testcase.description(true) }}</textarea>
+                                <button type="button" class="float-end testcase-edit-description btn btn-secondary btn-sm">Edit</button>
+                            {% endif %}
+                        </td>
+                    </tr>
+                    <tr>
+                        <td class="filename">
+                            <a href="{{ path('jury_problem_testcase_fetch', {'probId': problem.probid, 'rank': testcase.rank, 'type': 'output'}) }}">
+                                {{ testcase.downloadName | split('/') | last }}.{{ extensionMapping.output }}
                             </a>
-                            {% if allowEdit %}
-                            &nbsp;
-                            <div class="input-group">
-                                <label class="form-control text-truncate" for="update_image_{{testcase.rank}}" id="update_imapge_label_{{testcase.rank}}">replace image</label>
-                                <input type="file" class="d-none custom-file-input" name="update_image[{{ testcase.rank }}]" id="update_image_{{testcase.rank}}" accept="image/*">
-                            </div>
-                            {% endif %}
-                        {% else %}
-                            {% if allowEdit %}
-                            <div class="input-group">
-                                <label class="form-control text-truncate" for="update_image_{{testcase.rank}}" id="update_imapge_label_{{testcase.rank}}">add an image</label>
-                                <input type="file" class="d-none custom-file-input" name="update_image[{{ testcase.rank }}]" id="update_image_{{testcase.rank}}" accept="image/*">
-                            </div>
-                            {% else %}
-                            <span class="nodata">No image</span>
-                            {% endif %}
+                        </td>
+                        <td class="size">
+                            {{ row.outputSize | printSize }}
+                        </td>
+                        <td class="md5 small">
+                            {{ testcase.md5sumOutput }}
+                        </td>
+                        {% if allowEdit %}
+                            <td><div class="input-group">
+                                <label class="form-control text-truncate" for="update_output_{{testcase.rank}}" id="update_output_label_{{testcase.rank}}">new output file</label>
+                                <input type="file" class="d-none custom-file-input" name="update_output[{{ testcase.rank }}]" id="update_output_{{testcase.rank}}">
+                            </div></td>
                         {% endif %}
-                    </td>
-                </tr>
+                        <td class="testimage">
+                            {% if row.imageSize > 0 %}
+                                <a href="{{ path('jury_problem_testcase_fetch', {'probId': problem.probid, 'rank': testcase.rank, 'type': 'image'}) }}">
+                                    <span class="filename">p{{ problem.probid }}.t{{ testcase.rank }}.{{ testcase.imageType }}</span>
+                                </a>
+                                {% if allowEdit %}
+                                &nbsp;
+                                <div class="input-group">
+                                    <label class="form-control text-truncate" for="update_image_{{testcase.rank}}" id="update_imapge_label_{{testcase.rank}}">replace image</label>
+                                    <input type="file" class="d-none custom-file-input" name="update_image[{{ testcase.rank }}]" id="update_image_{{testcase.rank}}" accept="image/*">
+                                </div>
+                                {% endif %}
+                            {% else %}
+                                {% if allowEdit %}
+                                <div class="input-group">
+                                    <label class="form-control text-truncate" for="update_image_{{testcase.rank}}" id="update_imapge_label_{{testcase.rank}}">add an image</label>
+                                    <input type="file" class="d-none custom-file-input" name="update_image[{{ testcase.rank }}]" id="update_image_{{testcase.rank}}" accept="image/*">
+                                </div>
+                                {% else %}
+                                <span class="nodata">No image</span>
+                                {% endif %}
+                            {% endif %}
+                        </td>
+                    </tr>
+                {% endif %}
             {% endfor %}
             </tbody>
         </table>
--- a/webapp/src/Controller/Jury/ProblemController.php
+++ b/webapp/src/Controller/Jury/ProblemController.php
@@ -4,6 +4,7 @@ namespace App\Controller\Jury;
 
 use App\Controller\BaseController;
 use App\DataTransferObject\SubmissionRestriction;
+use App\DataTransferObject\TestcaseViewRow;
 use App\Entity\Contest;
 use App\Entity\ContestProblem;
 use App\Entity\Judging;
@@ -576,6 +577,43 @@ class ProblemController extends BaseController
         /** @var Testcase[] $testcases */
         $testcases = array_map(fn($data) => $data[0], $testcaseData);
 
+        $rows        = [];
+        $lastLineage = [];
+        foreach ($testcaseData as $data) {
+            /** @var Testcase $testcase */
+            $testcase = $data[0];
+            $lineage  = $testcase->getTestcaseGroup() ? $testcase->getTestcaseGroup()->getLineage() : [];
+
+            $commonPrefixLength = 0;
+            $foundDiff          = false;
+            foreach ($lineage as $i => $group) {
+                if (!$foundDiff && isset($lastLineage[$i]) && $lastLineage[$i]->getTestcaseGroupId() === $group->getTestcaseGroupId()) {
+                    $commonPrefixLength = $i + 1;
+                } else {
+                    $foundDiff = true;
+                    $rows[]    = new TestcaseViewRow(
+                        type: TestcaseViewRow::TYPE_GROUP,
+                        group: $group,
+                        level: $i
+                    );
+                }
+            }
+
+            if (empty($lineage) && !empty($lastLineage)) {
+                $rows[] = new TestcaseViewRow(type: TestcaseViewRow::TYPE_NO_GROUP);
+            }
+
+            $rows[] = new TestcaseViewRow(
+                type: TestcaseViewRow::TYPE_TESTCASE,
+                testcase: $testcase,
+                inputSize: (int)$data['input_size'],
+                outputSize: (int)$data['output_size'],
+                imageSize: (int)$data['image_size'],
+            );
+
+            $lastLineage = $lineage;
+        }
+
         if ($request->isMethod('POST')) {
             if (!empty($lockedContests)) {
                 $this->addFlash('danger', 'Cannot edit problem / testcases, it belongs to locked contest(s) '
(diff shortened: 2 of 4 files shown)


meisterT:
Some examples of the breakdown:

image

----

image

----

image

meisterT:
Example scoreboard:
image

------------

I have a few ideas how this can be improved, but would  prefer to do it outside of this PR

meisterT:
Example test case overview:

image

.github/workflows/scoring_integration.yml
```diff
@@ -0,0 +1,162 @@
+name: Run scoring integration tests
```
vmcj: I think I prefer if this becomes a matrix job, I'm fine with doing the work as it makes reviewing of the differences a bit easier.

.github/jobs/baseinstall.sh
```diff
@@ -137,7 +137,11 @@ if [ "${db}" = "install" ]; then
 	    # Make sure admin has no team associated so we will not insert submissions during unit tests.
 	    mysql_log "UPDATE user SET teamid=null WHERE userid=1;" $DATABASE_NAME
     fi
-    /opt/domjudge/domserver/bin/dj_setup_database install-examples | tee -a "$ARTIFACTS/mysql.txt"
+    INSTALL_CMD="install-examples"
+    if [ "${INSTALL_SCORING:-}" = "1" ]; then
+        INSTALL_CMD="install-scoring-examples"
```
vmcj, line 142: Why can't we install both in the same command? If we add scoring we should add it for default installs for people to test with. We could install them against another contest? So when someone installs they can see the differences in API output etc (scoreboard_type).

configure.ac
```diff
@@ -179,6 +179,7 @@ if test "x$FHS_ENABLED" = xyes ; then
 	AC_SUBST(domserver_rundir,          $localstatedir/run/${PACKAGE_TARNAME})
 	AC_SUBST(domserver_tmpdir,          /tmp)
 	AC_SUBST(domserver_exampleprobdir,  $datadir/${PACKAGE_TARNAME}/example_problems)
+	AC_SUBST(domserver_scoringexampleprobdir,  $datadir/${PACKAGE_TARNAME}/example_scoring_problems)
```
vmcj, line 182: Whitespace should be aligned, or we should remove it everywhere.

example_scoring_problems/addone/submissions/partial/addone-tiny.py
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python3
+# Only handles small numbers (n < 100)
+# Artificially limited to demonstrate partial scoring
+# @EXPECTED_SCORE@: 10
+
+n = int(input())
+if n < 100:
```
vmcj, line 7:
```suggestion
if n > 100 and n <= 10**6:
```

This should be the case for `SCORE: 20`, as I think that is not tested and I think we should not stop evaluation if the first testcase group fails? Might be you already test this somewhere else ofcourse.

example_scoring_problems/fibonacci/data/secret/extreme/1.ans
```diff
@@ -0,0 +1 @@
+918091266
```
vmcj: Missing `EOL`

example_scoring_problems/fibonacci/data/secret/hard/1.ans
```diff
@@ -0,0 +1 @@
+210345902
```
vmcj: Missing `EOL`.

example_scoring_problems/fibonacci/data/secret/medium/1.ans
```diff
@@ -0,0 +1 @@
+134903163
```
vmcj: Missing `EOL`.

example_scoring_problems/knapsack/output_validators/knapsack_validator/validate.py
```diff
@@ -0,0 +1,117 @@
...
+        sys.exit(43)
+
+    # Calculate score (0-100 based on achieved value vs optimal)
+    if optimal_value == 0:
+        score = 100.0 if total_value == 0 else 0.0
+    else:
+        score = (total_value / optimal_value) * 100.0
+
+    # Cap at 100 in case team found better solution than judge
+    score = min(100.0, score)
```
vmcj, line 98: I think the spec allowes for better solutions to be found.

webapp/tests/Unit/Service/SubmissionServiceTest.php
```diff
@@ -44,4 +59,543 @@ public function provideRunResults(): Generator
...
+        $group = $this->createTestcaseGroup('group1', TestcaseAggregationType::SUM, acceptScore: '20');
+        $problem = new Problem();
+        $problem->setTimelimit(1)->setName('test');
+        $problem->setParentTestcaseGroup($group);
+
+        $tc1 = $this->createTestcase($problem, $group, 1);
+        $tc2 = $this->createTestcase($problem, $group, 2);
+
+        $judging = $this->createJudging();
+        // Individual testcase scores are 0, but group has accept_score 20
```
vmcj, line 395: Why should this work? The individual scores don't matter you just need correct in all testcases for this group? What if the score of the testcases summed is higher than the acceptScore?

webapp/tests/Unit/Service/SubmissionServiceTest.php
```diff
@@ -44,4 +59,543 @@ public function provideRunResults(): Generator
...
+
+        self::assertFalse($result);
+    }
+
+    /**
+     * Test parsing with both annotation types returns false.
+     */
+    public function testParseExpectedAnnotationMixedTags(): void
+    {
+        $source = "// @EXPECTED_RESULTS@: CORRECT\n// @EXPECTED_SCORE@: 60\nint main() {}";
```
vmcj, line 518: How hard is it to support: 
```
        $source = "// @EXPECTED_RESULTS@: CORRECT, RUN-ERROR\n// @EXPECTED_SCORE@: 60\nint main() {}";
```

I would expect something like that for different versions of Python for example where a solution can either sometimes fail or get the expected score.

webapp/tests/Unit/Integration/ScoringScoreboardIntegrationTest.php
```diff
@@ -0,0 +1,519 @@
...
+        $lang = $this->em->getRepository(Language::class)->findByExternalId('c');
+
+        // Team 0: Gets 30 points on problem 0 (groups 0 and 1 correct = 10+20)
+        $this->createScoringSubmission($lang, $this->problems[0], $this->teams[0], 10*60, [
+            'problem_0_tc_0' => ['correct', '10'],
+            'problem_0_tc_1' => ['correct', '20'],
+            'problem_0_tc_2' => ['wrong-answer', '0'],
+        ]);
+
+        // Team 1: Gets 60 points on problem 0 (all correct = 10+20+30)
```
vmcj, line 254: They should now also have the `FTS`? Do we still implement that?

webapp/tests/Unit/Integration/ScoringScoreboardIntegrationTest.php
```diff
@@ -0,0 +1,519 @@
...
+            'problem_0_tc_1' => ['correct', '20'],
+            'problem_0_tc_2' => ['wrong-answer', '0'],
+        ]);
+
+        // Team 0: Third submission gets 20 points (worse than second, should be ignored)
+        $this->createScoringSubmission($lang, $this->problems[0], $this->teams[0], 30*60, [
+            'problem_0_tc_0' => ['correct', '10'],
+            'problem_0_tc_1' => ['wrong-answer', '0'],
+            'problem_0_tc_2' => ['correct', '30'],
+        ], score: '40'); // Actually this would be 40, better
```
vmcj:
```suggestion
        // Team 0: Third submission gets 20 points (worse than second, should be ignored)
        $this->createScoringSubmission($lang, $this->problems[0], $this->teams[0], 30*60, [
            'problem_0_tc_0' => ['correct', '10'],
            'problem_0_tc_1' => ['wrong-answer', '0'],
            'problem_0_tc_2' => ['correct', '30'],
        ], score: '40'); // Actually this would be 40, better
```

I understand the comment on L300, I understand the comment at L305 but I don't understand why those are not 2 different pieces of code. I suspect L300 should still be implemented?

example_scoring_problems/knapsack/submissions/wrong_answer/knapsack_invalid.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python3
```
vmcj, line 1: Not related to this file,

For one of those problems we should also have solutions in Java/Kotlin, and maybe even in all languages we have so people can use the example-problems to test their setup.

example_scoring_problems/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+### B - Fibonacci Number (100 points total)
+Compute the n-th Fibonacci number.
+
+Test groups:
+- **basic** (25 points): n <= 20
+- **medium** (25 points): n <= 45
+- **hard** (25 points): n <= 90
+- **extreme** (25 points): n <= 10^6 (requires matrix exponentiation, modulo 10^9+7)
+
```
vmcj, line 24: Should problems `C` & `D` be mentioned here?

webapp/migrations/Version20250704181912.php
```diff
@@ -0,0 +1,66 @@
...
+        $this->addSql('ALTER TABLE problem DROP FOREIGN KEY FK_D7E7CCC8A090DCC7');
+        $this->addSql('DROP INDEX IDX_D7E7CCC8A090DCC7 ON problem');
+        $this->addSql('ALTER TABLE problem DROP parent_testcase_group_id');
+        $this->addSql('ALTER TABLE testcase_group DROP FOREIGN KEY FK_F02888FE727ACA70');
+        $this->addSql('DROP INDEX IDX_F02888FE727ACA70 ON testcase_group');
+        $this->addSql('ALTER TABLE testcase_group DROP parent_id');
+        $this->addSql('ALTER TABLE judging DROP score');
+        $this->addSql('ALTER TABLE judging_run DROP score');
+        $this->addSql('ALTER TABLE testcase DROP FOREIGN KEY FK_4C1E5C391FF421A3');
+        $this->addSql('DROP TABLE testcase_group');
```
vmcj: I would assume this should be the last item.

webapp/migrations/Version20250804192300.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE external_judgement ADD score NUMERIC(32, 9) DEFAULT \'0.000000000\' NOT NULL COMMENT \'Optional score for this run, e.g. for partial scoring\'');
```
vmcj, line 23: Is this in another migration by design? If so I would prefer to see the description.

webapp/src/Command/ImportEventFeedCommand.php
```diff
@@ -209,6 +209,17 @@ protected function compareContestId(): bool
             }
         }
 
+        if ($contest->getScoreboardType() !== $this->sourceService->getScoreboardType()) {
```
vmcj, line 212: The function should now be renamed. And probably the warning in L205 but that's something we should have done in the other PR.

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -654,15 +655,16 @@ public function addJudgingRunAction(
         $teamMessage  = $request->request->get('team_message');
         $metadata     = $request->request->get('metadata');
         $testcasedir  = $request->request->get('testcasedir');
-        $compareMeta = $request->request->get('compare_metadata');
+        $compareMeta  = $request->request->get('compare_metadata');
+        $score        = $request->request->get('score');
```
vmcj, line 660: Should we accept a score if the problem is not a scoring problem?

webapp/src/Controller/Jury/JuryMiscController.php
```diff
@@ -287,18 +287,44 @@ public function judgingVerifierAction(Request $request): Response
             // As we only load the needed judging, this will automatically be the first one
             /** @var Judging $judging */
             $judging         = $submission->getJudgings()->first();
-            /** @var string[] $expectedResults */
+            /** @var string[]|null $expectedResults */
```
vmcj, line 290: Is it hard to initialize `$expectedResults` with `[CORRECT]` for scoring problems? That way we don't have this `string[]null` which makes PHPStan less powerful and it also forces the extra validator check with the implied notion that this should be `CORRECT`.

.github/jobs/baseinstall.sh
meisterT: I prefer having it separate to be able to test separately.

We can of course create a command `install-pass-fail-examples` and then make `install-examples` do both (the new `install-pass-fail-examples` and `install-scoring-examples`).

configure.ac
meisterT: Done

example_scoring_problems/fibonacci/data/secret/hard/1.ans
meisterT: Done

example_scoring_problems/fibonacci/data/secret/extreme/1.ans
meisterT: Done

example_scoring_problems/fibonacci/data/secret/medium/1.ans
meisterT: Done

example_scoring_problems/README.md
meisterT: Done

.github/workflows/scoring_integration.yml
meisterT: Attempted to do this, let's see whether I break CI horribly :-)

example_scoring_problems/knapsack/output_validators/knapsack_validator/validate.py
meisterT: The spec allows to give a range, see the testdata.yaml file in `example_scoring_problems/knapsack/data/secret/main/testdata.yaml`

example_scoring_problems/knapsack/output_validators/knapsack_validator/validate.py
```diff
@@ -0,0 +1,117 @@
...
+
+    write_feedback(feedback_dir, score,
+                   f"Value: {total_value}/{optimal_value}, Weight: {total_weight}/{capacity}")
+    sys.exit(42)
+
+
+def write_feedback(feedback_dir, score, message):
```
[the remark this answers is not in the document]
meisterT: Irrelevant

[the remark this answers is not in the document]
meisterT: Irrelevant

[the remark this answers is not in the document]
meisterT: Irrelevant

example_scoring_problems/knapsack/submissions/wrong_answer/knapsack_invalid.py
meisterT: Added some C++/Java solutions

webapp/src/Controller/Jury/JuryMiscController.php
meisterT: expected result does not need to be correct for scoring, so no?

webapp/tests/Unit/Integration/ScoringScoreboardIntegrationTest.php
meisterT: First to solve? I would argue that we don't need FTS in the first version of partial/optimization scoring.

webapp/tests/Unit/Service/SubmissionServiceTest.php
meisterT: For problems with accept_score, there is no validator that does give a score. See the addone example problem for example.

webapp/tests/Unit/Service/SubmissionServiceTest.php
meisterT: I need to look into this, should be doable

meisterT:
Just sharing a judging verifier example:
image

example_scoring_problems/addone/submissions/partial/addone-tiny.py
meisterT: Added one with SCORE 20

webapp/tests/Unit/Integration/ScoringScoreboardIntegrationTest.php
meisterT: Thanks for catching. Fixed

webapp/src/Command/ImportEventFeedCommand.php
meisterT: I am not sure what function you mean? `getScoreboardType()`? What name do you prefer?

webapp/src/Controller/API/JudgehostController.php
meisterT: Probably not, added a TODO for it.

webapp/tests/Unit/Service/SubmissionServiceTest.php
meisterT: (I will do so outside of this PR)

webapp/migrations/Version20250804192300.php
meisterT: fixed

webapp/migrations/Version20250704181912.php
meisterT: Why?

.github/jobs/baseinstall.sh
meisterT: Fixed, see topmost commit

example_scoring_problems/addone/data/secret/testdata.yaml
```diff
@@ -0,0 +1,2 @@
+on_reject: continue
+grading: sum
```
nickygerritsen: We are doing the legacy format right? Shouldn't this be `grader_flags` then? https://www.kattis.com/problem-package-format/spec/legacy.html#default-grader-specification

webapp/migrations/Version20250704181912.php
nickygerritsen: If you drop the testcase_group table but still have the index in testcase, it might break?

webapp/migrations/Version20250804192300.php
nickygerritsen: What did you fix? Still a separate file and nothing in `getDescription` 😛

webapp/src/Controller/BaseController.php
```diff
@@ -14,6 +14,7 @@
 use App\Entity\ScoreCache;
 use App\Entity\Team;
 use App\Entity\TeamCategory;
+use App\Entity\TestcaseAggregationType;
```
nickygerritsen: I guess you don't... use this use?

webapp/templates/jury/problem_testcases.html.twig
```diff
@@ -39,7 +39,56 @@
             </tr>
             </thead>
             <tbody>
+            {% set lastLineage = [] %}
             {% for rank, testcase in testcases %}
+                {% set lineage = [] %}
```
nickygerritsen: Does it maybe make sense to put some of this logic in PHP and use some DTO here? I'm not sure, what do you think?

webapp/tests/Unit/Service/SubmissionServiceTest.php
```diff
@@ -44,4 +59,543 @@ public function provideRunResults(): Generator
...
+    // =========================================================================
+
+    private function createTestcaseGroup(
+        string $name,
+        TestcaseAggregationType $aggregationType,
+        ?string $acceptScore = null,
+        bool $onRejectContinue = true
+    ): TestcaseGroup {
+        $group = new TestcaseGroup();
+        $group->setTestcaseGroupId($this->nextGroupId++);
```
nickygerritsen, line 550: Isn't the ID auto generated if we don't do this?

webapp/migrations/Version20250704181912.php
meisterT: I see, I am pretty sure symfony/doctrine generated this for me, but I moved the line now.

example_scoring_problems/addone/data/secret/testdata.yaml
meisterT: Fixed

webapp/src/Controller/BaseController.php
meisterT: Fixed

webapp/tests/Unit/Service/SubmissionServiceTest.php
meisterT: Nope - I assume because we don't entity manager in this test.

webapp/migrations/Version20250804192300.php
meisterT: I merged the *other* migration in this file, so it should not be more than one migration per commit / single functional change. I have added a description now though.

webapp/templates/jury/problem_testcases.html.twig
meisterT: Check the most recent commit in this branch

meisterT:
@vmcj do you know what CI is waiting on? "integrationExpected — Waiting for status to be reported"?

vmcj:
> @vmcj do you know what CI is waiting on? "integrationExpected — Waiting for status to be reported"?

You renamed the tests:
```
[Run integration tests / integration (default) (pull_request)](https://github.com/DOMjudge/domjudge/actions/runs/21116860213/job/60723684252?pr=3337)Successful in 10m
[Run integration tests / integration (pull_request)](https://github.com/DOMjudge/domjudge/actions/runs/21116860213/job/60724024331?pr=3337)
Run integration tests / integration (pull_request)Successful in 3s
```

So the required tests should be removed for this one (to unblock) when merging and then both added after that.

webapp/src/Command/ImportEventFeedCommand.php
vmcj: The function L212 is part of, but maybe I'm the only one who thinks so.

webapp/src/Controller/Jury/JuryMiscController.php
vmcj: If you pass `EXPECTED_SCORE` you also expect `CORRECT` I would think. But you probably know better.

webapp/src/Command/ImportEventFeedCommand.php
meisterT: Ah, that name. Yeah, it makes sense to rename it. Done.

meisterT:
> > @vmcj do you know what CI is waiting on? "integrationExpected — Waiting for status to be reported"?
> 
> You renamed the tests:
> 
> ```
> [Run integration tests / integration (default) (pull_request)](https://github.com/DOMjudge/domjudge/actions/runs/21116860213/job/60723684252?pr=3337)Successful in 10m
> [Run integration tests / integration (pull_request)](https://github.com/DOMjudge/domjudge/actions/runs/21116860213/job/60724024331?pr=3337)
> Run integration tests / integration (pull_request)Successful in 3s
> ```
> 
> So the required tests should be removed for this one (to unblock) when merging and then both added after that.

Thanks for figuring that out!

webapp/src/Controller/Jury/JuryMiscController.php
meisterT: No, assume you have test groups A with children B and C.

A is set up in a way that accepts if any accepted, and sums up the results of B and C.

For B, the submission passes all cases, so it gets the score (e.g. through accept_score, or a validator).

For C, the submission times out on all cases, so the score is 0 for C.

The overall score of A is then the score of B while the overall verdict is TLE.

vmcj: Hmm, I think in that case I would prefer even more to pass the `EXPECTED_RESULTS` to add this extra information and be able to explain why a SCORE doesn't add up, knowing you expect a `TLE` with score `X` is easier to investigate than seeing you get a score `Y` where `Y>X` with `CORRECT`.

meisterT: Let me work on that in a follow up PR. See also the scoring breakdown which already exists (one of the screenshots that I've added).

meisterT: Let me work on that in a follow up PR. See also the scoring breakdown which already exists (one of the screenshots that I've added).

.github/workflows/integration.yml
```diff
@@ -9,9 +9,31 @@ on:
 jobs:
   integration:
     runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
+        mode:
+          - default
+          - scoring
```
vmcj: I think those are not needed.

configure.ac
```diff
@@ -230,6 +231,7 @@ AX_PATH(domserver_logdir,          [$domserver_root/log])
 AX_PATH(domserver_rundir,          [$domserver_root/run])
```
vmcj: Indentation.

webapp/src/Service/DOMJudgeService.php
```diff
@@ -1725,17 +1728,26 @@ private function actuallyCreateJudgetasks(int $priority, Judging $judging, int $
...
-                '(%s, :testcase_id%d, :testcase_hash%d)',
+                '(%s, :testcase_id%d, :testcase_hash%d, :compare_config%d)',
                 implode(', ', $judgetaskDefaultParamNames),
                 $testcase->getTestcaseid(),
-                $testcase->getTestcaseid()
+                $testcase->getTestcaseid(),
+                $testcase->getTestcaseid(),
             );
+            $compareConfig = $defaultCompareConfig;
+            if ($testcase->getTestcaseGroup() != null) {
```
```suggestion
            if ($testcase->getTestcaseGroup() !== null) {
```

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -183,6 +184,15 @@ public function getContestId(): string
         return $this->cachedContestData->id;
     }
 
+    public function getScoreboardType(): ScoreboardType
+    {
+        if (!$this->isValidContestSource()) {
+            throw new LogicException('The contest source is not valid');
```
vmcj:
```suggestion
            throw new LogicException('The contest source is not valid.');
```

webapp/src/Service/ImportProblemService.php
```diff
@@ -342,6 +345,45 @@ public function importZippedProblem(
...
+                            }
+                            $testCaseGroups[$dir] = $testcaseGroup;
+                        } catch (Exception $e) {
+                            $messages['warning'][] = sprintf("Could not parse test group metadata file '%s': %s", $filename, $e->getMessage());
+                        }
+                    }
+                }
+            }
+        }
+        foreach ($testCaseGroups as $dir => $testCaseGroup) {
```
vmcj, line 376: Shouldn't we stop here in case we got warnings? Inserting things in the DB at this point is probably difficult to fix with a re-upload.

webapp/src/Service/ScoreboardService.php
```diff
@@ -338,6 +357,13 @@ public function calculateScoreRow(
             $absSubmitTime = max($absSubmitTime, $contestStartTime);
             $submitTime    = $contest->getContestTime($absSubmitTime);
 
+            // For scoring problems we consider submissions after the first correct one. So we need to
+            // keep track of when the first correct submission occurred.
+            // TODO: is this actually what 'first to solve' for scoring problems means?
```
vmcj, line 362: I think FTS is solving 100% correct, so having correct output but not all points (optimization problem) would not be FTS.

.github/workflows/integration.yml
meisterT: For my own education, how does it know then to run in both modes?

webapp/src/Service/ImportProblemService.php
meisterT: Yeah, I think we should, will address this (probably tomorrow). Also `danger` vs `warning` above is not consistent.

webapp/src/Service/ScoreboardService.php
meisterT: I want to keep the debate what FTS behavior is correct (if it is needed at all, and how it is represented) out of scope of this PR.

configure.ac
meisterT: Done

webapp/src/Service/ImportProblemService.php
meisterT: Actually, took not too long to fix and test. Done.

.github/workflows/integration.yml
vmcj: You include 2 extra options, those override (well extend) the options you already had. So I think it's valid to just specify the extended ones you had. If they would have some overlap you could put it above and only use `include` to add the extra attributes as it always tries to expand first before creating complete new ones.

webapp/src/Service/ImportProblemService.php
```diff
@@ -903,9 +978,9 @@ public function importProblemFromRequest(Request $request, ?int $contestId = nul
     /**
      *
      * @param array<int, string> $zipEntries
-     * @param array{danger?: string[], info?: string[]} $messages
+     * @param array<string, string[]> $messages
```
vmcj: Why do we want less information here? Before we knew we could have {danger,info} but nothing else, now it can be everything.

webapp/src/Service/SubmissionService.php
```diff
@@ -56,6 +60,246 @@ public function __construct(
         protected readonly PaginatorInterface $paginator,
     ) {}
 
+    /**
+     * Returns a two-element array:
+     *   - The score for the testcase group, or null if not all results are ready.
```
vmcj:
```suggestion
     *   - The score for the testcase group, or null if not all scores are ready.
```

webapp/src/Service/SubmissionService.php
```diff
@@ -56,6 +60,246 @@ public function __construct(
...
+     *   - The result for the testcase group, or null if not all results are ready.
```
vmcj:
```suggestion
     * Returns a two-element array [score, result] or [null, null] f not all results are ready.
```
as alternative.

webapp/src/Service/SubmissionService.php
```diff
@@ -56,6 +60,246 @@ public function __construct(
...
+        $ignoreSample = $testcaseGroup->isIgnoreSample();
+
+        if ($testcaseGroup->getChildren()->isEmpty()) {
+            $relevantRuns = $runsByGroup[$testcaseGroup->getTestcaseGroupId()] ?? [];
+            if ($testcaseGroup->getAcceptScore() !== null) {
+                $acceptScore = $testcaseGroup->getAcceptScore();
+                foreach ($relevantRuns as $run) {
+                    if ($run->getRunresult() === null || $run->getRunresult() === '') {
+                        $allResultsReady = false;
+                    } else if ($run->getRunresult() !== 'correct') {
```
vmcj:
```suggestion
                    } elseif ($run->getRunresult() !== 'correct') {
```

PHPCS

webapp/src/Service/SubmissionService.php
```diff
@@ -56,6 +60,246 @@ public function __construct(
...
+                    } else {
+                        $results[] = '0';
+                    }
+                }
+            } else {
+                foreach ($relevantRuns as $run) {
+                    $results[] = $run->getScore();
+                    if ($run->getRunresult() === null || $run->getRunresult() === '') {
+                        $allResultsReady = false;
+                    } else if ($run->getRunresult() !== 'correct') {
```
vmcj:
```suggestion
                    } elseif ($run->getRunresult() !== 'correct') {
```
PHPCS

webapp/src/Service/SubmissionService.php
```diff
@@ -56,6 +60,246 @@ public function __construct(
...
+            $score = bcadd($score ?? '0', '0', ScoreboardService::SCALE);
+            $result = $allCorrect ? 'correct' : $firstIncorrectVerdict ?? 'judge-error';
+            return [$score, $result];
+        }
+        return [null, null];
+    }
+
+    /**
+     * Get the scoring hierarchy for the given problem and judging.
+     * @return array<string, mixed>|null
```
vmcj, line 201: It's mixed because we can have multiple level arrays there?

webapp/src/Service/SubmissionService.php
```diff
@@ -736,27 +980,42 @@ public function submitSolution(
...
-                        $message = sprintf("Found more than one @EXPECTED_RESULTS@ in file '%s'.",
-                            $file->getClientOriginalName());
-                        return null;
+                $fileAnnotation = self::parseExpectedAnnotation(
+                    file_get_contents($file->getRealPath()),
+                    $this->config->get('results_remap')
+                );
+                if ($fileAnnotation === false) {
+                    $message = sprintf(
+                        "Found more than one @EXPECTED_RESULTS@/@EXPECTED_SCORE@ in file '%s'.",
```
vmcj, line 996: Looking at the code we can now also close: https://github.com/DOMjudge/domjudge/issues/1387

webapp/src/Service/SubmissionService.php
meisterT: yes

webapp/src/Service/SubmissionService.php
meisterT: Yes, likely. I will do some more validation after this has been merged

webapp/src/Service/SubmissionService.php
meisterT: done

webapp/src/Service/SubmissionService.php
meisterT: done

webapp/src/Service/SubmissionService.php
meisterT: Used almost this

webapp/src/Service/ImportProblemService.php
meisterT: Reverted, not sure why I changed this anymore (perhaps some intermediate state)

webapp/src/Service/ImportProblemService.php
```diff
@@ -1008,6 +1083,64 @@ private function searchAndAddValidator(ZipArchive $zip, array $zipEntries, ?arra
...
+                    } catch (ValueError $e) {
+                        $messages['danger'][] = sprintf("Invalid aggregation type '%s' in test group '%s'.", $flag, $name);
+                        return null;
+                    }
+                }
+                if ($flag === 'ignore_sample') {
+                    $testcaseGroup->setIgnoreSample(true);
+                }
+                // Silently ignore currently unused flags.
+                // TODO: add support for the remaining flags and error out on unknown flags.
```
vmcj, line 1130:
```suggestion
                if (in_array($flag, ['sum', 'max', 'min', 'avg'])) {
                    try {
                        $aggregationType = TestcaseAggregationType::tryFrom($flag);
                        $testcaseGroup->setAggregationType($aggregationType);
                    } catch (ValueError $e) {
                        $messages['danger'][] = sprintf("Invalid aggregation type '%s' in test group '%s'.", $flag, $name);
                        return null;
                    }
                } elseif ($flag === 'ignore_sample') {
                    $testcaseGroup->setIgnoreSample(true);
                } else {
                    $messages['warning'][] = sprintf("Flag: '%s' not implemented.", $flag);
                // Silently ignore currently unused flags.
                // TODO: add support for the remaining flags and error out on unknown flags.
```

I agree with not erroring out, but I think adding the warning should help judges on what to expect and the change seems relatively simple.

PR #3338, opened 2026-01-18
Allow multiple classes in 1 file

vmcj:
Not sure if this is also the rule for the enums but I suspect it is. I think this is cleaner than adding the comment in the file for all enums/classes.

### DIFF
--- a/.github/jobs/data/phpruleset.xml
+++ b/.github/jobs/data/phpruleset.xml
@@ -6,6 +6,10 @@
     <rule ref="PSR2.Methods.FunctionCallSignature.Indent">
         <severity>0</severity>
     </rule>
+    <!-- We have the JudgeDaemon where we do this with intent. -->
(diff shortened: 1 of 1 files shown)


meisterT (approved):

.github/jobs/data/phpruleset.xml
```diff
@@ -6,6 +6,10 @@
     <rule ref="PSR2.Methods.FunctionCallSignature.Indent">
         <severity>0</severity>
     </rule>
+    <!-- We have the JudgeDaemon where we do this with intent. -->
```
nickygerritsen, line 9: Does this mean its also allowed in the webapp? Im not sure I'm a fan of that

vmcj: We would not flag it, as I think there are good reasons to do it (sometimes) but we still have manual review. But if you don't like it I can close it.

PR #3339, opened 2026-01-18
Add interactive judgehost timeline to overview page.

meisterT:
Examples:

image

------

image

------

The view supports pan+zoom. Colors are assigned round robin per jobid so that one visually recognizes the same submission being judged by multiple judgedaemons. A tooltip shows some more info on hover.

### DIFF
--- a/webapp/public/js/judgehost-timeline.js
+++ b/webapp/public/js/judgehost-timeline.js
@@ -1,0 +1,155 @@
+function initJudgehostTimeline(dataUrl) {
+    const margin = {top: 20, right: 30, bottom: 40, left: 150};
+    const $container = $('#judgehost-timeline');
+    const fullWidth = $container.width();
+    const fullHeight = $container.height();
+    const width = fullWidth - margin.left - margin.right;
+    const height = fullHeight - margin.top - margin.bottom;
+
+    const svg = d3.select("#judgehost-timeline svg")
+        .attr("width", fullWidth)
+        .attr("height", fullHeight);
+
+    // Clear previous content if any (for re-init)
+    svg.selectAll("*").remove();
+
+    const container = svg.append("g")
+        .attr("transform", "translate(" + margin.left + "," + margin.top + ")");
+
+    // Clip path for zooming
+    svg.append("defs").append("clipPath")
+        .attr("id", "clip")
+      .append("rect")
+        .attr("width", width)
+        .attr("height", height);
+
+    const mainContent = container.append("g")
+        .attr("clip-path", "url(#clip)");
+
+    let tooltip = d3.select(".timeline-tooltip");
+    if (tooltip.empty()) {
+        tooltip = d3.select("body").append("div")
+            .attr("class", "timeline-tooltip")
+            .style("opacity", 0);
+    }
+
+    d3.json(dataUrl, function(error, json) {
+        if (error) {
+            console.error("Error loading timeline data:", error);
+            return;
+        }
+
+        const allData = json.data;
+        const now = json.now * 1000;
+        const start = json.twenty_four_hours_ago * 1000;
+
+        allData.forEach(function(d) {
+            d.startTime = +d.startTime * 1000;
+            d.endtime = +d.endtime * 1000;
+        });
+
+        function render() {
+            const showDisabled = $('#toggle-disabled-jh').is(':checked');
+            
+            // Filter judgehosts
+            const activeJudgehosts = [];
+            for (const id in json.judgehosts) {
+                const jh = json.judgehosts[id];
+                if (showDisabled || jh.enabled) {
+                    activeJudgehosts.push({id: id, name: jh.hostname});
+                }
+            }
+            activeJudgehosts.sort(function(a, b) { return d3.ascending(a.name, b.name); });
+
+            // Filter data to only include judgehosts being shown
+            const hostNames = new Set(activeJudgehosts.map(d => d.name));
+            const data = allData.filter(d => hostNames.has(d.hostname));
+
+            container.selectAll(".timeline-axis").remove();
+            mainContent.selectAll(".timeline-rect").remove();
+
+            const minStart = d3.min(data, function(d) { return d.startTime; });
+            let displayStart = start;
+            if (minStart) {
+                displayStart = Math.max(start, minStart - 1 * 60 * 1000);
+            }
+
+            const x = d3.time.scale()
+                .domain([displayStart, now])
+                .range([0, width]);
+
+            const y = d3.scale.ordinal()
+                .domain(activeJudgehosts.map(function(d) { return d.name; }))
+                .rangeRoundBands([0, height], 0.1);
+
+            const xAxis = d3.svg.axis()
+                .scale(x)
+                .orient("bottom")
+                .tickFormat(d3.time.format("%H:%M"));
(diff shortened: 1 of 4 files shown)


meisterT:
Part of #1164

webapp/public/js/judgehost-timeline.js
```diff
@@ -0,0 +1,138 @@
+function initJudgehostTimeline(dataUrl) {
+    var margin = {top: 20, right: 30, bottom: 40, left: 150},
```
nickygerritsen: let's try to use `const` instead of `var` where possible? And if not, let's use `let` where possible?

webapp/public/js/judgehost-timeline.js
```diff
@@ -0,0 +1,138 @@
...
+            });
+
+        // Attach zoom to the top-level SVG so it captures all events
+        svg.call(zoom);
+
+        var colors = d3.scale.category20();
+
+        mainContent.selectAll(".timeline-rect")
+            .data(data)
+          .enter().append("rect")
```
nickygerritsen: Weird indent?

meisterT: Done

webapp/public/js/judgehost-timeline.js
meisterT: Lot's more `const` now. Please take another look

webapp/templates/jury/judgehosts.html.twig
```diff
@@ -44,7 +44,26 @@
...
         </div>
-
     {%- endif %}
 
+    <div class="card mb-3">
+        <div class="card-header">
+            Judgehost Activity (last 24h)
+        </div>
+        <div class="card-body">
+            <div id="judgehost-timeline" style="height: 400px; width: 100%;">
```
vmcj: can we move the style to the CSS stylesheet?

webapp/templates/jury/judgehosts.html.twig
```diff
@@ -44,7 +44,26 @@
...
+                <svg></svg>
+            </div>
+        </div>
+    </div>
+
+
+    <script src="{{ asset('js/d3.min.js') }}"></script>
```
vmcj: Can we load both in the head?

webapp/src/Controller/Jury/JudgehostController.php
```diff
@@ -227,6 +228,44 @@ public function indexAction(Request $request): Response
...
+            ->join('jt.judgehost', 'jh')
+            ->join('jr.judging', 'j')
+            ->join('j.submission', 's')
+            ->join('s.contest', 'c')
+            ->join('s.problem', 'p')
+            ->select('jh.judgehostid, jh.hostname, jr.startTime, jr.endtime, j.judgingid, s.submitid, c.shortname as contest_shortname, p.name as problem_name, jt.type, jr.runresult, jt.jobid')
+            ->andWhere('jr.startTime IS NOT NULL')
+            ->andWhere('jr.endtime IS NOT NULL')
+            ->andWhere('jr.endtime >= :twenty_four_hours_ago')
+            ->setParameter('twenty_four_hours_ago', $twenty_four_hours_ago)
```
vmcj, line 250:
```suggestion
        $query = $this->em->createQueryBuilder()
            ->select('jh.judgehostid, jh.hostname, jr.startTime, jr.endtime, j.judgingid, s.submitid, c.shortname as contest_shortname, p.name as problem_name, jt.type, jr.runresult, jt.jobid')
            ->from(JudgingRun::class, 'jr')
            ->join('jr.judgetask', 'jt')
            ->join('jt.judgehost', 'jh')
            ->join('jr.judging', 'j')
            ->join('j.submission', 's')
            ->join('s.contest', 'c')
            ->join('s.problem', 'p')
            ->andWhere('jr.startTime IS NOT NULL')
            ->andWhere('jr.endtime IS NOT NULL')
            ->andWhere('jr.endtime >= :twenty_four_hours_ago')
            ->setParameter('twenty_four_hours_ago', $twenty_four_hours_ago)
```

webapp/public/js/judgehost-timeline.js
```diff
@@ -0,0 +1,134 @@
...
+            .attr("y", function(d) { return y(d.hostname); })
+            .attr("width", function(d) { return Math.max(1, x(d.endtime) - x(d.startTime)); })
+            .attr("height", y.rangeBand())
+            .attr("fill", function(d) { return colors(d.jobid % 20); })
+            .style("pointer-events", "all")
+            .on("mouseover", function(d) {
+                tooltip.transition().duration(200).style("opacity", .9);
+                const duration = ((d.endtime - d.startTime) / 1000).toFixed(3);
+                tooltip.html(
+                    "<strong>Contest:</strong> " + d.contest_shortname + "<br/>" +
```
vmcj: I think `<strong>` is not the right tag here. a `<span>` with the right CSS is a better alternative or just the `<b>` if IRC (https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/strong)

meisterT: done

webapp/templates/jury/judgehosts.html.twig
meisterT: done

webapp/templates/jury/judgehosts.html.twig
meisterT: done

webapp/public/js/judgehost-timeline.js
```diff
@@ -0,0 +1,155 @@
...
+                })
+                .on("click", function(d) {
+                    window.location.href = domjudge_base_url + "/jury/submissions/" + d.submitid;
+                });
+        }
+
+        render();
+        $('#toggle-disabled-jh').on('change', render);
+    });
+}
```
vmcj: EOL missing.

webapp/src/Controller/Jury/JudgehostController.php
vmcj: This only moves the select higher, which I think reads more natural.

webapp/templates/jury/judgehosts.html.twig
```diff
@@ -44,7 +46,30 @@
...
+                <label class="form-check-label" for="toggle-disabled-jh">Show disabled</label>
+            </div>
+        </div>
+        <div class="card-body">
+            <div id="judgehost-timeline">
+                <svg></svg>
+            </div>
+        </div>
+    </div>
+
```
vmcj, line 65:
```suggestion
```

vmcj: Also for L50.

vmcj (approved):

webapp/templates/jury/judgehosts.html.twig
meisterT: I think removing these newlines hurts readability

webapp/src/Controller/Jury/JudgehostController.php
meisterT: done

webapp/public/js/judgehost-timeline.js
meisterT: done

PR #3340, opened 2026-01-18
Add self-service password change option.

### LINKED ISSUE #2729: Allow user change password
Introduce a feature that allows users to change their own passwords, with the option to enable or disable this functionality through settings. This will provide flexibility in cases where a fixed password is required, while allowing or prohibiting individual password changes when necessary.

The goal is to allow users to securely and conveniently manage their own passwords, while the jury can configure in the settings whether the default behavior is to allow all users to change their own passwords. It should also be possible to restrict certain users from changing their passwords individually.

### Expected behaviour
Users should be able to change their passwords via their personal settings page. Administrators should be able to enable or disable password changes for users through system settings.

DOMjudge is often used in courses on programming, data structures, and algorithms. In these contexts, student IDs or fixed passwords are typically used for long-term login throughout the semester. However, due to the recent feature in DOMjudge that allows downloading submission code, there have been incidents where some students log into others' accounts to secretly download the correct code. Given the large number of students in such classes, handling password changes individually becomes difficult. Therefore, allowing users to change their own passwords could help mitigate such issues while reducing the burden on administrators.

meisterT:
This can be enabled in two ways:
- by adding roles to the list of roles that are allowed to change their password (e.g. `jury`)
- by enabling password change for specific team categories

Fixes #2729

### DIFF
--- a/webapp/src/Controller/ProfileController.php
+++ b/webapp/src/Controller/ProfileController.php
@@ -1,0 +1,60 @@
+<?php declare(strict_types=1);
+
+namespace App\Controller;
+
+use App\Form\Type\ChangePasswordType;
+use App\Service\ConfigurationService;
+use App\Service\DOMJudgeService;
+use App\Service\EventLogService;
+use Doctrine\ORM\EntityManagerInterface;
+use Symfony\Component\DependencyInjection\Attribute\Autowire;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
+use Symfony\Component\HttpKernel\KernelInterface;
+use Symfony\Component\PasswordHasher\Hasher\UserPasswordHasherInterface;
+use Symfony\Component\Routing\Attribute\Route;
+use Symfony\Component\Security\Http\Attribute\IsGranted;
+
+#[Route(path: '/profile')]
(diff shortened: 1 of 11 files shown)


nickygerritsen (approved):

vmcj:
I rather not have this implemented but use the `REMOTE_USER` for this.

meisterT:
`REMOTE_USER` seems much harder to set up, and the change is relatively self-contained. but if you feel strongly, feel free to close this.

nickygerritsen:
> `REMOTE_USER` seems much harder to set up, and the change is relatively self-contained. but if you feel strongly, feel free to close this.

Personally i agree with Tobi on all three his remarks

meisterT:
@vmcj let's make a decision on this - any input from my side required?

vmcj:
I'm delaying this one until we have an inperson discussion.

webapp/src/Controller/ProfileController.php
```diff
@@ -0,0 +1,60 @@
...
+        EntityManagerInterface $em,
+        DOMJudgeService $dj,
+        EventLogService $eventLogService,
+        KernelInterface $kernel,
+    ) {
+        parent::__construct($em, $eventLogService, $dj, $kernel);
+    }
+
+    #[Route(path: '', name: 'profile_index')]
+    public function changePasswordAction(Request $request, UserPasswordHasherInterface $passwordHasher): Response
```
Kevinjil, line 33: If we allow these changes, they might deserve an audit log message.

PR #3341, opened 2026-01-19
Support both `@EXPECTED_SCORE@` and `@EXPECTED_RESULTS@` in the same file for scoring problems.

meisterT:


meisterT:
This PR is "stacked" on top of https://github.com/DOMjudge/domjudge/pull/3337 - I will rebase and mark this as ready for review after the other PR has been merged

meisterT:
cc @vmcj who suggested supporting both

Example:
image

webapp/src/Service/ImportProblemService.php
```diff
@@ -784,27 +784,29 @@ public function importZippedProblem(
...
+                                        "Annotated results '%s' restricted to match directory for %s",
+                                        implode(', ', $results), $path
+                                    );
+                                }
+                                $results = [$expectedResult];
                             }
-                            $results = [$expectedResult];
+                            $expectedResults = $results;
+                        } elseif ($expectedScore === null) {
+                            // This case should actually not happen if annotation is not null
```
vmcj, line 807: Isn't this expected for `pass-fail`? So I might not understand the comment.

nickygerritsen: As far as I understand this elseif says 'if the result is null (line 789) AND the score is null (this line)', which means both are null, which can't happen if we can parse annotations.

meisterT: Indeed

PR #3342, opened 2026-01-19
Add scoreboard styling for scoring problems.

meisterT:
This PR is "stacked" on top of https://github.com/DOMjudge/domjudge/pull/3337 - I will rebase and mark this as ready for review after the other PR has been merged

### DIFF
--- a/webapp/src/Utils/Scoreboard/Scoreboard.php
+++ b/webapp/src/Utils/Scoreboard/Scoreboard.php
@@ -395,4 +395,110 @@ class Scoreboard
     {
         return $this->contest->getRuntimeAsScoreTiebreaker();
     }
+
+    /**
+     * Get the maximum score achieved by any team for a given problem.
+     * Returns 0 if no team has a positive score.
+     */
+    public function getMaxScoreForProblem(ContestProblem $problem): float
+    {
+        $maxScore = 0.0;
+        $problemId = $problem->getProbid();
+
+        foreach ($this->scores as $teamScore) {
+            $teamId = $teamScore->team->getTeamid();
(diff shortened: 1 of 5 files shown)


meisterT:
Example:
image
-----
image

vmcj:
I think you can push your branches to `domjudge` instead of `meisterT` so you can use that as base? In case you want to stack more of them.

nickygerritsen:
Btw, how does scoring look in the freeze? Especially if someone already had a score before the freeze?

webapp/src/Utils/Scoreboard/Scoreboard.php
```diff
@@ -395,4 +395,110 @@ public function getRuntimeAsScoreTiebreaker(): bool
...
+            }
+        }
+
+        return $maxScore;
+    }
+
+    /**
+     * Get the CSS class for a scoreboard cell.
+     */
+    public function getCellCssClass(
```
nickygerritsen, line 421: 😍

nickygerritsen (approved):
We should think about the mobile scoreboard at some point. Ugh, I know.

meisterT:
> Btw, how does scoring look in the freeze? Especially if someone already had a score before the freeze?

Good question - is there a :cold_face:  for scoring/challenge problems?

nickygerritsen:
> > Btw, how does scoring look in the freeze? Especially if someone already had a score before the freeze?
> 
> Good question - is there a :cold_face:  for scoring/challenge problems?

Yes, at NAC and finals we did resolve the challenge so there is a freeze. Not sure how it looked like in the resolver but you could check with Tim

PR #3347, opened 2026-01-23
Make penalty time a per contest setting

nickygerritsen:
For score contests we use `0`, since the spec says "Only relevant if scoreboard_type is pass-fail.".

### DIFF
--- a/webapp/migrations/Version20260123092747.php
+++ b/webapp/migrations/Version20260123092747.php
@@ -1,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use App\Entity\ScoreboardType;
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20260123092747 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add penalty time to contest.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE contest ADD penalty_time INT UNSIGNED DEFAULT 20 NOT NULL COMMENT \'Penalty time in minutes per wrong submission (if eventually solved)\' AFTER scoreboard_type');
+
+        $penaltyTimeConfig = $this->connection->fetchAssociative('SELECT * FROM configuration WHERE name = :name', ['name' => 'penalty_time']);
+        $penaltyTime = $penaltyTimeConfig ? (int)json_decode($penaltyTimeConfig['value'], true) : 20;
+
+        $this->addSql('UPDATE contest SET penalty_time = :penalty_time WHERE scoreboard_type = :scoreboard_type', [
+            'penalty_time' => $penaltyTime,
+            'scoreboard_type' => ScoreboardType::PASS_FAIL->value,
+        ]);
+
+        $this->addSql('UPDATE contest SET penalty_time = 0 WHERE scoreboard_type = :scoreboard_type', [
+            'scoreboard_type' => ScoreboardType::SCORE->value,
+        ]);
+
+        $this->addSql('DELETE FROM configuration WHERE name = :name', ['name' => 'penalty_time']);
+    }
+
+    public function down(Schema $schema): void
+    {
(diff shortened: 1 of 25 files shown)


webapp/migrations/Version20260123092747.php
```diff
@@ -0,0 +1,59 @@
...
+        $this->addSql('UPDATE contest SET penalty_time = 0 WHERE scoreboard_type = :scoreboard_type', [
+            'scoreboard_type' => ScoreboardType::SCORE->value,
+        ]);
+
+        $this->addSql('DELETE FROM configuration WHERE name = :name', ['name' => 'penalty_time']);
+    }
+
+    public function down(Schema $schema): void
+    {
+        $contest = $this->connection->fetchAssociative('SELECT penalty_time FROM contest WHERE scoreboard_type = :scoreboard_type LIMIT 1', ['scoreboard_type' => ScoreboardType::PASS_FAIL->value]);
```
meisterT: Do we want to bail out if there are contests with different penalty times? Or at least print a warning?

nickygerritsen: I think bailing out makes sense, that leaves it to the user to decide what to do.

webapp/migrations/Version20260123092747.php
```diff
@@ -0,0 +1,67 @@
...
+final class Version20260123092747 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add penalty time to contest.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE contest ADD penalty_time INT UNSIGNED DEFAULT 20 NOT NULL COMMENT \'Penalty time in minutes per wrong submission (if eventually solved)\' AFTER scoreboard_type');
```
vmcj, line 23: You pick our default here, but someone might have set this to another value. I wonder if we should keep the global setting and just force it to 0 for Scoring contests?

This way they would need to (re)set this value each time.

nickygerritsen: I don't follow what you mean. For migrated contests we pick what they configured. For new contests we use 20 as default but allow to set it

nickygerritsen: I don't agree that allowing it makes sense. Then all of a sudden we have to take it into account on the scoring scoreboard. I don't think the branches are complicated but if others think it makes more sense to do it I can do it

meisterT:
> if so to display the extra penalty.

There is no time displayed for scoring contests anywhere in the interface

webapp/migrations/Version20260123092747.php
vmcj: > I don't follow what you mean. For migrated contests we pick what they configured. For new contests we use 20 as default but allow to set it

In case someone configured 30min or 1h (or whatever) it would be applied for all contests, both current and future.
Your code fixes the current ones but not the future ones as you set it to 20 instead of the current value. I think in case someone picks another value as the current one they have a reason for that and probably prefer it for newer ones also.

meisterT: My suggestion to improve the changelog message then to make it more clear and not carry that legacy over.

vmcj:
> > if so to display the extra penalty.
> 
> There is no time displayed for scoring contests anywhere in the interface

And the point I'm trying to make that instead of having a scoreboard per contest type we should try to generalize such things and use sane defaults. But if I'm the only one (and it seems to be I am) I'm fine with just duplicating all situations everywhere.

webapp/migrations/Version20260123092747.php
nickygerritsen: Yeah I agree

meisterT:
> And the point I'm trying to make that instead of having a scoreboard per contest type we should try to generalize such things and use sane defaults

I am with you on generalizing and using sane defaults where it makes sense. There is just nothing to generalize here since penalty time is not a thing for scoring.

nickygerritsen:
> > And the point I'm trying to make that instead of having a scoreboard per contest type we should try to generalize such things and use sane defaults
> 
> I am with you on generalizing and using sane defaults where it makes sense. There is just nothing to generalize here since penalty time is not a thing for scoring.

Yeah I don't see how we would do this. And then I think it doesn't make sense showing a value that doesn't get used.

PR #3349, opened 2026-01-23
Don't use multipart forms when not needed in the API

### LINKED ISSUE #3168: `PUT` on a clarification via api/doc doesn't work
We use `formdata` in the `api/doc` to submit the `PUT` request but this is out of spec.

```
curl -X 'PUT' \
  'http://localhost/domjudge/api/v4/contests/demo/clarifications/1?strict=false' \
  -H 'accept: application/json' \
  -H 'Content-Type: multipart/form-data' \
  -F 'text=same text as string' \
  -F 'time=2025-10-26T13:38:47.878Z' \
  -F 'id=1'
```

nickygerritsen:
Also allow JSON for clarifications

Fixes #3168

For PUT, Symfony doesn't populate the body for multipart forms. Since we only need multipart for requests with actual binary files, I changed all that don't use them to normal forms. I also added JSON for the clarification requests, since that should also be allowed. We might want to add it to other endpoints as well, but I think we can do that in a different PR.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,6 +13,8 @@ Version 10.0.0DEV
  - Make shadow mode a per contest setting.
  - Allow to specify whether to use internal or external judgings and runs in
    shadow mode.
+ - Use application/x-www-form-urlencoded instead of multipart/form-data for
+   API requests without file uploads to make them work with PUT/PATCH requests.
 
 Version 9.0.0 - 5 October 2025
 ------------------------------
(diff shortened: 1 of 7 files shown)


vmcj:
So this changes how certain tools use the API? If so we should mention this in the Changelog.

nickygerritsen:
We can mention it. I dont think many tools use the multipart format since it didn't work correctly

PR #3350, opened 2026-01-24
Properly check whether a rejudging exists.

meisterT:
Previously this would fail with a 500 further down in the code.

### DIFF
--- a/webapp/src/Controller/Jury/RejudgingController.php
+++ b/webapp/src/Controller/Jury/RejudgingController.php
@@ -227,6 +227,10 @@ class RejudgingController extends BaseController
             ->getQuery()
             ->getOneOrNullResult();
 
+        if ($rejudging === null) {
+            throw new NotFoundHttpException(sprintf('Rejudging with ID %s not found', $rejudgingId));
+        }
+
         $disabledProblems = [];
         $disabledLangs = [];
         foreach ($rejudging->getJudgings() as $judging) {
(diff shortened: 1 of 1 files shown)


webapp/src/Controller/Jury/RejudgingController.php
```diff
@@ -227,6 +227,10 @@ public function viewAction(
             ->getQuery()
             ->getOneOrNullResult();
 
+        if ($rejudging === null) {
```
vmcj, line 230: Isn't !$rejudging doing the same?

meisterT: Mostly, `!$rejudging` behaves the same unless the object is `false` or `0` (if I am not mistaken). `=== null` is the cleaner way that should be used in modern codebases.

meisterT: btw, the change is semi-accidental, as I added this without realizing we already the now removed check, until phpstan told me.

Important is that we move it here, the `!` vs `===` is less important.

vmcj: I think we removed a lot of those `=== null` checks, so if we want it we should decide on it consistently. I know the difference and I think in this case both forms would be fine.

meisterT: Did we? Let's merge this and discuss how to do it consistently separately.

PR #3351, opened 2026-01-24
Defend against '\n' at the end of score.txt.

meisterT:
This is to avoid problems with `bcadd` later on.


meisterT:
In other places (such as validator output, system error) etc, we might rather want to keep it

PR #3352, opened 2026-01-24
Add heatmap to rejudgings view for scoring problems.

meisterT:
Example:

image

----

image

### DIFF
--- a/webapp/public/js/rejudging-scores.js
+++ b/webapp/public/js/rejudging-scores.js
@@ -1,0 +1,606 @@
+'use strict';
+
+/**
+ * Score changes visualization for rejudgings.
+ * Provides a heatmap visualization, delta filtering, and sortable table.
+ */
+
+// Configuration constants
+const PRECISION = 0.0001;
+const DELTA_THRESHOLDS = [0.0001, 0.0005, 0.0025, 0.01, 0.05, 0.25, 1, 2, 5, 25, 100, 250];
+const NUM_X_BINS = 20;
+const SLIDER_DEBOUNCE_MS = 50;
+
+/**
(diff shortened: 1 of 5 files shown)


vmcj:
> image

I don't really like the contrast change from black numbers to white for the different shades of red. So I wonder if the different levels for red actually do add information or if it's only for cosmetics because I'm not sure if the cosmetic is worth the change of number color in the heatmap itself.

vmcj:
image

I think we also want the relative change, in case of poorly defined scores it would show the effect of randomness in submissions.

meisterT:
> I don't really like the contrast change from black numbers to white for the different shades of red. So I wonder if the different levels for red actually do add information or if it's only for cosmetics because I'm not sure if the cosmetic is worth the change of number color in the heatmap itself.

You are right, the color gradient doesn't add a lot of info, so I removed it. I also increase font size a little to make it more legible: 

image

meisterT:
> I think we also want the relative change, in case of poorly defined scores it would show the effect of randomness in submissions.

You mean that with a higher old score, we would tolerate a higher delta?

vmcj:
> > I think we also want the relative change, in case of poorly defined scores it would show the effect of randomness in submissions.
> 
> You mean that with a higher old score, we would tolerate a higher delta?

I meant in the table show the relative delta also. The filtering I did not consider yet (which might make this harder?).

meisterT:
Ah, I see. I tried that and it looks way to busy (and does not add anything meaningful for the random and semi-random submissions that I looked at).

meisterT:
@vmcj happy to include the relative delta in a separate PR in case we come up with an idea how to make it not too busy

PR #3353, opened 2026-01-24
Refactor scoreboard display.

meisterT:
This is to make the code cleaner and easier to understand and modify in the future. It should also make it less likely to add features just for one type (desktop vs mobile) of scoreboard.


webapp/templates/partials/scoreboard/_team_row.html.twig
```diff
@@ -0,0 +1,122 @@
+{# Detect sort order change and emit summary for previous sort order #}
+{% set previousSortOrder = prevScore ? prevScore.team.sortorder : -1 %}
+{% if prevScore and score.team.sortorder != previousSortOrder %}
+    {% include 'partials/scoreboard_summary.html.twig' with {sortOrder: previousSortOrder, isMobile: isMobile} %}
```
nickygerritsen: maybe we should also move these to the `scoreboard` folder?

webapp/templates/partials/scoreboard/_problem_cell.html.twig
```diff
@@ -0,0 +1,75 @@
...
+                    title="{% if hasScore %}Full score: {{ matrixItem.points }}{% if problem.rangeUpperBound is not null %} / {{ problem.rangeUpperBound }}{% endif %}{% if time != '' %} ({{ time }} mins){% endif %}{% endif %}"
+                {% endif %}
+            >
+                {% if scoringScoreboard %}
+                    {% if hasScore %}{{ matrixItem.points | round(2) }}{% else %}&nbsp;{% endif %}
+                {% else %}
+                    {% if matrixItem.isCorrect %}{{ time }}{% else %}&nbsp;{% endif %}
+                {% endif %}
+                <span>
+                    {% if numSubmissions is same as(1) %}
```
nickygerritsen: We probably already had this but why not `== 1`?

webapp/templates/partials/scoreboard/_team_row.html.twig
meisterT: Done

webapp/templates/partials/scoreboard/_problem_cell.html.twig
meisterT: tested and changed

PR #3356, opened 2026-01-24
Update team manual

meisterT:
Best reviewed commit by commit. I went through the change log to see what we were missing.

### DIFF
--- a/doc/manual/import.rst
+++ b/doc/manual/import.rst
@@ -18,7 +18,7 @@ the ``webapp`` directory of the DOMserver.
 External ID's
 -------------
 
-Most entities in DOMjudge have an `eternal ID`. External ID's are used to link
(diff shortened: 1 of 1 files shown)


meisterT:
(The team manual needs to be done separately.)

doc/manual/config-advanced.rst
```diff
@@ -278,6 +278,36 @@ See the ``validate.h`` file in the ``boolfind_run`` executable for some
 convenience functions you might want to use when implementing your own run
 program.
 
+Multi-pass problems
```
vmcj, line 281: I wonder if we should just link to the specification@commit_hash in the problempackage repo?

meisterT: I will find a good link for it later, but I think we should have some minimal info inline.

meisterT: done

PR #3363, opened 2026-01-25
Fix submission lookup with external IDs.

meisterT:
There is no guarantee that the IDs are unique anymore across all contests :-/

### DIFF
--- a/webapp/src/Controller/Jury/SubmissionController.php
+++ b/webapp/src/Controller/Jury/SubmissionController.php
@@ -269,6 +269,8 @@ class SubmissionController extends BaseController
             ->leftJoin('s.originalSubmission', 'os')
             ->select('s', 'os', 't', 'u', 'p', 'l', 'c', 'f', 'cp', 'ej')
             ->andWhere('s.externalid = :submitid')
+            ->andWhere('s.contest = :contest')
+            ->setParameter('contest', $this->dj->getCurrentContest())
             ->setParameter('submitid', $submitId)
             ->getQuery()
             ->getOneOrNullResult();


meisterT:
We should probably think more thoroughly what to do here but this at least gets rid of the error.

nickygerritsen:
Do we still allow people to view submissions without selecting a contest? If so indeed we need a better fix (pass contest ID?). If not we should be fine?

meisterT:
Ugh, this is annoying. 

For example, we are also using it here:
https://github.com/DOMjudge/domjudge/blob/5007a03a9225196d869f5feb1214bc5d842e8a77/webapp/src/Controller/API/JudgehostController.php#L1445

We might need to rethink whether we should really use external IDs for submissions

meisterT:
btw, the result of this currently is that we collect too many submission files (for all submissions that match this external ID), and then compilation fails, e.g.:

```
Compiling failed with exitcode 1, compiler output:
foo.cpp: In function 'int main()':
foo.cpp:27:20: warning: variable 'ss' set but not used [-Wunused-but-set-variable]
   27 |             double ss = 0;
      |                    ^~
foo.cpp: In function 'int main()':
foo.cpp:27:20: warning: variable 'ss' set but not used [-Wunused-but-set-variable]
   27 |             double ss = 0;
      |                    ^~
/usr/bin/ld: ./ccLnE6zg.o: in function `next_zz()':
foo.cpp:(.text+0x0): multiple definition of `next_zz()'; ./ccau2BIz.o:foo.cpp:(.text+0x0): first defined here
/usr/bin/ld: ./ccLnE6zg.o:(.data+0x0): multiple definition of `dir'; ./ccau2BIz.o:(.data+0x0): first defined here
/usr/bin/ld: ./ccLnE6zg.o:(.data+0x8): multiple definition of `curr_x'; ./ccau2BIz.o:(.data+0x8): first defined here
/usr/bin/ld: ./ccLnE6zg.o:(.bss+0x0): multiple definition of `k'; ./ccau2BIz.o:(.bss+0x0): first defined here
/usr/bin/ld: ./ccLnE6zg.o:(.data+0x4): multiple definition of `curr_y'; ./ccau2BIz.o:(.data+0x4): first defined here
/usr/bin/ld: ./ccLnE6zg.o: in function `main':
foo.cpp:(.text.startup+0x0): multiple definition of `main'; ./ccau2BIz.o:foo.cpp:(.text.startup+0x0): first defined here
/usr/bin/ld: ./ccLnE6zg.o:(.bss+0x4): multiple definition of `g'; ./ccau2BIz.o:(.bss+0x4): first defined here
collect2: error: ld returned 1 exit status
```

nickygerritsen:
We do need them in the API if we are shadowing, right?
I do agree it's annoying since we need to pass the contest along in all places. But using internal IDs everywhere except API is also annoying

nickygerritsen:
We could decide to use internal IDs in the judgehost API but that's also weird. Another option is the UUID

meisterT:
Discussed offline, we will try a different approach.

PR #3364, opened 2026-01-25
Include contest ID in URLs where relevant.

meisterT:
This is a follow-up on the external ID rewrite (see cf86a1849308b135c6821c96c437fbd4b91c7059).

Prior to that rewrite IDs were globally unique, now two contests can have the same IDs, e.g. for submissions, since we are using the external ID. This broke some assumptions, that we are trying to fix with this change.

We are introducing for submissions, clarifications and balloons new default URL paths which include the contest ID to make IDs unique by composition, e.g. `/jury/submissions/<submitid>` will become `/jury/contests/<contestid>/submissions/<submitid>` (similar for the other entities with contest associated data).

We have added redirects for the legacy routes if a contest has already been selected.

This continues to align the UI with the API, both in terms of IDs and also URL structure.

### DIFF
--- a/webapp/src/Controller/Jury/BalloonController.php
+++ b/webapp/src/Controller/Jury/BalloonController.php
@@ -2,6 +2,7 @@
 
 namespace App\Controller\Jury;
 
+use App\Entity\Contest;
 use App\Entity\Team;
 use App\Entity\TeamAffiliation;
 use App\Entity\TeamCategory;
@@ -20,7 +21,7 @@ use Symfony\Component\Routing\Attribute\Route;
 use Symfony\Component\Security\Http\Attribute\IsGranted;
 
 #[IsGranted(new Expression("is_granted('ROLE_ADMIN') or is_granted('ROLE_BALLOON')"))]
-#[Route(path: '/jury/balloons')]
+#[Route(path: '/jury')]
 class BalloonController extends AbstractController
 {
     public function __construct(
@@ -50,13 +51,22 @@ class BalloonController extends AbstractController
         return false;
     }
 
-    #[Route(path: '', name: 'jury_balloons')]
-    public function indexAction(BalloonService $balloonService): Response
+    // Legacy route - redirects to contest-scoped route
+    #[Route(path: '/balloons', name: 'jury_balloons_legacy')]
+    public function legacyIndexAction(): Response
     {
         $contest = $this->dj->getCurrentContest();
-        if (is_null($contest)) {
-            return $this->render('jury/balloons.html.twig');
+        if (!$contest) {
+            $this->addFlash('warning', 'Please select a contest first to view balloons.');
+            return $this->redirectToRoute('jury_index');
         }
+        return $this->redirectToRoute('jury_balloons', ['contestId' => $contest->getExternalid()]);
+    }
+
+    #[Route(path: '/contests/{contestId}/balloons', name: 'jury_balloons')]
+    public function indexAction(string $contestId, BalloonService $balloonService): Response
+    {
+        $contest = $this->dj->getContestByExternalId($contestId);
 
         $balloons_table = $balloonService->collectBalloonTable($contest);
         $teamSummary = $balloonService->collectTeamBalloonSummary($balloons_table);
@@ -72,6 +82,7 @@ class BalloonController extends AbstractController
                     'icon' => 'running',
                     'title' => 'mark balloon as done',
                     'link' => $this->generateUrl('jury_balloons_setdone', [
+                        'contestId' => $contestId,
                         'balloonId' => $element['data']['balloonid'],
                     ])]];
             }
@@ -154,9 +165,10 @@ class BalloonController extends AbstractController
         $defaultCategories = array_column($defaultCategories, "externalid");
 
         return $this->render('jury/balloons.html.twig', [
+            'contestId' => $contestId,
             'refresh' => [
                 'after' => 60,
-                'url' => $this->generateUrl('jury_balloons'),
+                'url' => $this->generateUrl('jury_balloons', ['contestId' => $contestId]),
                 'ajax' => true
             ],
             'isfrozen' => isset($contest->getState()->frozen),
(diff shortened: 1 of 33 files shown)


webapp/src/Controller/Jury/BalloonController.php
```diff
@@ -30,6 +32,19 @@ public function __construct(
         protected readonly EventLogService $eventLogService
     ) {}
 
+    /**
+     * Get a contest by its external ID, throwing NotFoundHttpException if not found.
+     */
+    private function getContestByExternalId(string $contestId): Contest
+    {
+        $contest = $this->em->getRepository(Contest::class)
```
nickygerritsen: I believe I added a ContestRepository which has `findByExternalId` as a shorthand.

nickygerritsen: Also maybe we move this to the base controller or DJ service?

webapp/src/Controller/Jury/ClarificationController.php
```diff
@@ -39,23 +39,44 @@ public function __construct(
...
     ): Response {
+        $contest = $this->getContestByExternalId($contestId);
         $categories = $this->config->get('clar_categories');
-        if ($contest = $this->dj->getCurrentContest()) {
-            $contestIds = [$contest->getExternalid()];
-        } else {
-            $contestIds = array_keys($this->dj->getCurrentContests());
-            // cid '' will never happen, but otherwise the array is empty and that is not supported.
-            if (empty($contestIds)) {
-                $contestIds = [''];
```
nickygerritsen, line 56: This means we no longer support viewing clarifications across contests. I would be fine with it, but just saying. It would also not be too hard to still support it.

webapp/src/Controller/Jury/ClarificationController.php
```diff
@@ -237,24 +280,28 @@ public function viewAction(Request $request, string $id): Response
...
-            ->setParameter('clarid', $id)
+            ->setParameter('clarid', $clarification->getClarid())
             ->getQuery()
             ->getSingleResult()['jury_member'];
 
+        $parameters['contestId'] = $contestId;
         $parameters['previousNext'] = $this->getPreviousAndNextObjectIds(
             Clarification::class,
             $clarification->getExternalid(),
+            filterOnContest: true,
```
nickygerritsen, line 284: This brings up an interesting thing I already kinda noted offline: `filterOnContest` filters on the current contest from the cookie, but you could be on a different contest from this route. We should at least fix that (have `filterOnContest` also accept a `Contest` and filter on that if set?) but that leaves the question: do we find it OK that the contest displayed in the top right differs from the contest we use for the current page if you land on a different contest object?

webapp/src/Controller/Jury/ClarificationController.php
meisterT: I actually think that's an improvement

nickygerritsen: If so we might want to drop support for it in the twig template. Same for submissions.

meisterT: Done

webapp/src/Controller/Jury/ClarificationController.php
meisterT: Fixed by updating the 🍪

webapp/src/Controller/Jury/BalloonController.php
meisterT: Done

webapp/src/EventListener/ContestCookieListener.php
```diff
@@ -0,0 +1,54 @@
...
+    public function __construct(
+        protected DOMJudgeService $dj
+    ) {}
+
+    public static function getSubscribedEvents(): array
+    {
+        return [KernelEvents::RESPONSE => 'onKernelResponse'];
+    }
+
+    public function onKernelResponse(ResponseEvent $event): void
```
nickygerritsen, line 24: I kinda prefer event listeners. You can add `#[AsEventListener]` to this method and then drop `getSubscribedEvents` and the `implements`.

webapp/src/EventListener/ContestCookieListener.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        if (!$event->isMainRequest()) {
+            return;
+        }
+
+        $request = $event->getRequest();
+        $contestId = $request->attributes->get('contestId');
```
nickygerritsen, line 42: Maybe we only want this on jury pages? Now it will also do it on the API?

webapp/src/EventListener/ContestCookieListener.php
meisterT: Done, please check

webapp/src/EventListener/ContestCookieListener.php
meisterT: Good catch, done

webapp/src/Controller/Jury/ClarificationController.php
vmcj: Does this mean that we can´t open a clarification from another contest anymore? I did use that quite a lot I think.

nickygerritsen: I believe that if you know the clarification ID (and contest ID), you can. How did you open them in the past?

vmcj: I got the URL from someone and as the clarid was internal there it worked.

nickygerritsen: Then that still works right, since that URL will now contain the contest ID.

meisterT: If you get a URL, and it worked for another jury member, it will work for you, even if you had selected another contest in the dropdown

vmcj: Ok, than I don´t understand how this thread started.

PR #3365, opened 2026-01-25
Make scoring heatmap work for shadow differences as well.

meisterT:


meisterT:
This is stacked on top of https://github.com/DOMjudge/domjudge/pull/3352 and https://github.com/DOMjudge/domjudge/pull/3364 - so no need to review yet.

vmcj:
Can you send a screenshot after you do a testrun with this?

meisterT:
Looks identical to https://github.com/DOMjudge/domjudge/pull/3352#issuecomment-3796321096, but I can add another one when I have done another round of shadowing (my poor laptop :fire: ).

vmcj:
Ok, that's fine. No need for a screenshot in that case.

PR #3366, opened 2026-01-25
Improve rejudgings table.

### LINKED ISSUE #2451: Start time of repeated rejudging is wrong on overview page
The time is picked from when the current (`i`th) rejudging started, not when the first (`0`th) rejudging started.

meisterT:
Better group repeated rejudgings, and show their grouped start time and user.

Fixes #2451

-----

Example:

image

### DIFF
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -1607,3 +1607,15 @@ if ('ResizeObserver' in window) {
         monacoObserver.observe(document.body);
     });
 }
+
+// Table group expand/collapse
+$(() => {
+    $('.data-table').on('click', '[data-group-toggle]', function(e) {
+        e.preventDefault();
+        const id = $(this).data('group-toggle');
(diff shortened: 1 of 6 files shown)


webapp/public/js/domjudge.js
```diff
@@ -1607,3 +1607,15 @@ if ('ResizeObserver' in window) {
         monacoObserver.observe(document.body);
     });
 }
+
+// Table group expand/collapse
+$(() => {
+    $('.data-table').on('click', '[data-group-toggle]', function(e) {
```
nickygerritsen, line 1613: Does this still work after the AJAX refresh of the table happened? I think it only registers it once.

meisterT: Just test, does work for me

meisterT:
This change doesn't change when/what is displayed in finishtime. But I do agree that the current state of finishtime (where it means when the rejudging has been applied or cancelled not the time when the last judging was completed) can be improved.

PR #3380, opened 2026-01-29
Account for scoring only differences in shadow mode.

meisterT:


webapp/src/Service/DOMJudgeService.php
```diff
@@ -83,6 +83,9 @@ class DOMJudgeService
     final public const EVAL_DEMAND = 3;
     final public const EVAL_ANALYST = 4;
 
+    // Minimum absolute score difference to consider as a meaningful shadow difference.
+    final public const SCORE_DIFF_EPSILON = 0.0001;
```
vmcj, line 87: I think this should be a contest variable so people can tweak it and are aware of what the setting is. Alternative is a UI hint where they can change it in case they want to change it.

PR #3384, opened 2026-01-29
Use judgements from external source by default.

meisterT:


vmcj:
What does this change? When we shadow we would trust the data from the primary over our own data?

meisterT:
For scoreboard calculation, use their data. This was the default mode before moving it to the contest level.

The idea behind it is to validate different things separately:
- we have one part that validates that verdicts/scores for submissions are identical (validate scoring/judging)
- even while we are figuring differences for verdicts/scores (and might have some unresolved or even unfixable differences) compare scoreboards (so validate scoreboard / award calculation)

vmcj:
> For scoreboard calculation, use their data. This was the default mode before moving it to the contest level.
> 
> The idea behind it is to validate different things separately:
> 
>     * we have one part that validates that verdicts/scores for submissions are identical (validate scoring/judging)
> 
>     * even while we are figuring differences for verdicts/scores (and might have some unresolved or even unfixable differences) compare scoreboards (so validate scoreboard / award calculation)

Ok, this feels wrong to me and I think I had a discussion IRL about why we should do the other thing but can´t think of the main arguments anymore.

meisterT:
As an example: you have a submission very close to the time limit. It judges AC on the primary, TLE on the shadow. You have a difference, and you have an explanation. If you were to take the shadow as the source of truth for the scoreboard calculation, you do get scoreboard differences which are already explained, so you would be making it useless.

vmcj:
I think we can still toggle to the other view in that case. But seeing the actual effect a difference has is I think important. But let's discussion IRL at one point.

nickygerritsen:
I agree we can switch but I also agree that every time *we* are shadowing (which will be the biggest usecase for sure) and we have a shadow difference we have to switch it to be able to continue doing EOC, so I think it is a good default.

meisterT:
Yeah, this is not removing the ability to toggle, it is just reducing the amount of work to bring us into the state we typically run in.

PR #3385, opened 2026-01-29
Properly display score of queued submissions as `-` instead of `0.00`.

meisterT:


nickygerritsen (approved):

webapp/templates/jury/partials/submission_list.html.twig
```diff
@@ -170,7 +170,7 @@
                 </td>
                 {% if is_scoring %}
                     <td class="{{ tdExtraClass }} right">
-                        {% if submission.score is not null %}
+                        {% if submission.result is not null and submission.score is not null %}
```
vmcj, line 173: Can we use -(a or b) here instead of -a & -b?

PR #3386, opened 2026-01-29
Only display changes for scoring problems in the heatmap / table if w…

meisterT:
…e have both internal and external results.

Note that the verdict matrix will still indicated unjudged submissions, so we won't miss them.

### DIFF
--- a/webapp/src/Controller/Jury/ShadowDifferencesController.php
+++ b/webapp/src/Controller/Jury/ShadowDifferencesController.php
@@ -151,7 +151,8 @@ class ShadowDifferencesController extends BaseController
 
             // Collect score change data for scoring problems.
             $problem = $submission->getProblem();
-            if ($problem->isScoringProblem() && $externalJudgement && $localJudging) {
+            if ($problem->isScoringProblem()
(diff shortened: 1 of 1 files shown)


webapp/src/Controller/Jury/ShadowDifferencesController.php
```diff
@@ -151,7 +151,8 @@ public function indexAction(
 
             // Collect score change data for scoring problems.
             $problem = $submission->getProblem();
-            if ($problem->isScoringProblem() && $externalJudgement && $localJudging) {
+            if ($problem->isScoringProblem() && $externalJudgement && $localJudging
+                && $externalJudgement->getResult() && $localJudging->getResult()) {
```
nickygerritsen:
```suggestion
                && $externalJudgement?->getResult() && $localJudging?->getResult()) {
```

Then you can drop the middle part of the if

nickygerritsen (approved):

vmcj (approved):
After Nicky his suggestion.

webapp/src/Controller/Jury/ShadowDifferencesController.php
meisterT: done

PR #3387, opened 2026-01-29
Add missing null check.

meisterT:


nickygerritsen (approved):

webapp/templates/jury/partials/submission_list.html.twig
```diff
@@ -291,7 +291,7 @@
                     </td>
                     {% if is_scoring %}
                         <td class="right">
-                            {% if externalJudgement.score is not null %}
+                            {% if externalJudgement is not null and externalJudgement.score is not null %}
```
vmcj: Can we use -(a or b) instead of -a & -b?

meisterT: I can change it if you prefer it, but find it less readable.

I think another alternative would be `{% if externalJudgement.score ?? null %}` (didn't try)

vmcj: I do prefer limiting the number of `not`s we use, so I do prefer it or at least we should be consistent with it.

meisterT: Changed it to `{% if externalJudgement.score ?? null %}` now

webapp/templates/jury/partials/submission_list.html.twig
```diff
@@ -291,7 +291,7 @@
                     </td>
                     {% if is_scoring %}
                         <td class="right">
-                            {% if externalJudgement.score is not null %}
+                            {% if externalJudgement.score ?? null %}
```
vmcj:
```suggestion
                            {% if externalJudgement.score ?? false %}
```

Seems to work and removes an implicit truthy conversion in our heads so it makes it easier to read?

meisterT: That sounded like a good idea, but then I realized that it doesn't work - neither the `?? null` nor the `?? false`. Since a score of `0` would be considered falsy and then not be displayed. Reverted back to the original if

PR #3388, opened 2026-01-30
Reset data structures in event feed import.

meisterT:
This is especially relevant when running in dev mode.

Where before the memory was growing to multiple GBs very quickly memory is much more stable now (~10x improvement). With that also PHP GC times improve a lot.

### DIFF
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -62,10 +62,12 @@ use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 use Symfony\Component\Serializer\SerializerInterface;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 use Symfony\Contracts\HttpClient\Exception\DecodingExceptionInterface;
 use Symfony\Contracts\HttpClient\Exception\HttpExceptionInterface;
 use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
+use Symfony\Contracts\Service\ResetInterface;
 use Traversable;
 use ZipArchive;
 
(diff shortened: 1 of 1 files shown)


webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -2158,6 +2178,36 @@ protected function warningIfUnsupported(Event $event, ?string $entityId, array $
         return true;
     }
 
+    /**
+     * Clear the entity manager to free memory and prevent the Doctrine
+     * identity map from growing unbounded during long-running imports.
+     *
+     * After clearing, $this->contest is re-fetched so it remains managed.
+     */
+    protected function clearEntityManager(): void
```
nickygerritsen: The method name is a bit weird now since it does more and doesn't actually directly call clear on the entity manager

meisterT: Indeed, the name is super misleading now. It just reflects the first thing that I tried. Let me think what to do here (and perhaps add an explicit em->clear back)

meisterT: Done, please take another look

webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -85,6 +87,15 @@ class ExternalContestSourceService
...
     protected ?string $basePath = null;
+    protected int $importedEventsCounter = 0;
+
+    /**
+     * How often to reset services to free memory.
+     * Without periodic resets, the Doctrine identity map and other service
+     * state grows unbounded during long-running imports, causing excessive
+     * memory use and GC pressure.
+     */
+    protected const SERVICES_RESET_INTERVAL = 100;
```
vmcj, line 98: Can we put this in `webapp/config/services.yaml` or as an exposed config? In case someone wants to tweak this they shouldn´t have to change the code.

meisterT: I prefer not to. This shouldn't be anything that people need to tweak. And the more config toggles we have the more confusing is it for users.

vmcj: webapp/config/services.yaml is not something people default tweak but it makes it uniform where to tweak those things.

PR #3394, opened 2026-01-31
Only allow modifying test data for admin accounts.

meisterT:
This is in line with web UI and other pages, that jury can have read-only but no write access to the data.

### DIFF
--- a/webapp/src/Controller/Jury/ProblemController.php
+++ b/webapp/src/Controller/Jury/ProblemController.php
@@ -614,6 +614,7 @@ class ProblemController extends BaseController
         }
 
         if ($request->isMethod('POST')) {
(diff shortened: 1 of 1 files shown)


webapp/src/Controller/Jury/ProblemController.php
```diff
@@ -1262,6 +1263,7 @@ public function toggleSubmitAction(
         string $probId,
         string $type
     ): Response {
+        $this->denyAccessUnlessGranted('ROLE_ADMIN');
```
vmcj: Why can´t we use the normal annotation?

meisterT: Done

PR #3396, opened 2026-01-31
Make contest edit form visually nicer and more compact.

meisterT:
A lot more fits onto the screen, and it aligns edit with the view.

Before:
image

After:
image

### DIFF
--- a/webapp/public/style_jury.css
+++ b/webapp/public/style_jury.css
@@ -173,7 +173,7 @@ div.submission-summary > span {
     font-weight: 600;
     font-size: 0.9rem;
     color: #495057;
-    padding: 0.5rem 0.75rem !important;
+    padding: 0.5rem 0.3rem !important;
     border-bottom: 2px solid #dee2e6;
     letter-spacing: 0.03em;
     text-transform: uppercase;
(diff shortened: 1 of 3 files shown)


meisterT:
I will fix tests after I have received more feedback

meisterT:
> Should the row with `contest id: 1` have the same width and be part of the same structure? Now it looks out of place.

Agreed, I will fix it. Otherwise are you in favor of the change?

vmcj:
> Agreed, I will fix it. Otherwise are you in favor of the change?

I like the extra space,

Only annoying thing is the white `textarea` is very close to the grayborder of the box above and is misaligned. But this is already an improvement if we do it for all pages.

meisterT:
> Only annoying thing is the white textarea is very close to the grayborder of the box above and is misaligned.

What do you think about this?

image

image


>  if we do it for all pages.

Yeah, the idea was to start here and after agreeing how it should look like do similar improvements to all other edit pages.

vmcj:
> > Only annoying thing is the white textarea is very close to the grayborder of the box above and is misaligned.
> 
> What do you think about this?

Instead of 50% can you make it 75%? That way long URL's or Warning messages still feet without looping. For the rest I like it.

meisterT:
I assume you mean making the text areas wider? Good idea.

vmcj:
> I assume you mean making the text areas wider? Good idea.

Yes, thatś what I meant. As first I mentioned they came close to the border, so it feels implied.

meisterT:
Actually made the input boxes full width as they look much better this way:
image


Also update the problemset page:
image


And finally did use the same headers/grouping for edit as well:
image


------

Unit tests etc are fixed as well. Once this is merged, I will apply a similar improvement to other entity edit/view pages.

vmcj:
> Actually made the input boxes full width as they look much better this way:
image

I still don´t like this.

meisterT:
good :eye: - let me see whether we can do something about it

meisterT:
Done: 
image

PR #3398, opened 2026-02-01
Avoid adding a scoring problem to a pass-fail contest and vice versa.

meisterT:


.github/workflows/integration.yml
```diff
@@ -103,6 +103,7 @@ jobs:
           for i in */problem.yaml; do
               dir="$(dirname $i)"
               (
+                  sed -i -e 's|type: scoring|type: pass-fail|' "$dir/problem.yaml"
```
vmcj, line 106: Should this not be fixed in the Kattis repo?

meisterT: No, this is a scoring problem, so it would break in a pass-fail contest. I have an in-progress change to do the right thing on our side.

PR #3405, opened 2026-02-01
Fix updating users through the API.

meisterT:
Two prior existing bugs:
- we refused updates outright since the user already existed (since the same codepath handles adding users)
- we didn't reset roles before setting the new roles


meisterT:
Indeed, didn't even remember this

PR #3407, opened 2026-02-02
Add missing null check in JudgehostController.

meisterT:
This should never happen...
...unless you do weird stuff while debugging and get your DB into an inconsistent state.

It is good to be defensive here though.


vmcj:
If we want this, shouldn´t we improve PHPstan to a higher level? I think the next level forces such checks.

meisterT:
Not sure what we need to change in phpstan. In this case I ran into the situation - probably by butchering my DB instance though.

vmcj:
> Not sure what we need to change in phpstan. In this case I ran into the situation - probably by butchering my DB instance though.

The next step would basically assume that things can be `null` which we assume to be set etc. let's discuss IRL.

PR #3419, opened 2026-02-04
Catch invalid range/accept_score specifications.

meisterT:
While we are at it, make the method static to make it easily testable.

### DIFF
--- a/webapp/public/style_jury.css
+++ b/webapp/public/style_jury.css
@@ -167,6 +167,52 @@ div.submission-summary > span {
     font-size: 22px;
 }
 
+/* Contest form styles */
+.contest-form-section-header td {
+    background-color: #e9ecef !important;
+    font-weight: 600;
+    font-size: 0.9rem;
+    color: #495057;
+    padding: 0.5rem 0.3rem !important;
+    border-bottom: 2px solid #dee2e6;
+    letter-spacing: 0.03em;
+    text-transform: uppercase;
+}
+.contest-form-section-header td i {
+    color: #6c757d;
+    width: 1.25em;
+    text-align: center;
+}
+.contest-fields-table th {
+    vertical-align: top;
+    padding-top: 0.6rem;
+    white-space: nowrap;
+    width: 1%;
+}
+.contest-fields-table td {
+    padding-right: 0 !important;
+}
+.contest-fields-table td .form-control,
+.contest-fields-table td .form-select {
+    width: 100%;
+}
+.contest-fields-table td .form-text {
+    font-size: 0.8rem;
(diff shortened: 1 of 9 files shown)


nickygerritsen (approved):

webapp/tests/Unit/Service/ImportProblemServiceTest.php
```diff
@@ -302,4 +302,101 @@ public function testKattisExample(): void
...
+
+    public function testParseTestCaseGroupMetaValidAcceptScore(): void
+    {
+        $yaml = "accept_score: 50";
+        $messages = [];
+
+        $result = ImportProblemService::parseTestCaseGroupMeta($yaml, 'test-group', $messages);
+
+        $this->assertNotNull($result);
+        $this->assertEmpty($messages['danger'] ?? []);
```
vmcj, line 314: I think we should always be consistent and output the `danger` key to make parsing easier for other tools?

webapp/tests/Unit/Service/ImportProblemServiceTest.php
```diff
@@ -302,4 +302,101 @@ public function testKattisExample(): void
...
+    {
+        $yaml = "range: 100";
+        $messages = [];
+
+        $result = ImportProblemService::parseTestCaseGroupMeta($yaml, 'test-group', $messages);
+
+        $this->assertNull($result);
+        $this->assertNotEmpty($messages['danger']);
+        $this->assertStringContainsString("Invalid range '100'", $messages['danger'][0]);
+    }
```
vmcj, line 401:
```suggestion
    }

    public function testParseTestCaseGroupMetaInvalidRangeFormatRejected(): void
    {
        $yaml = "range: 100 101 102";
        $messages = [];

        $result = ImportProblemService::parseTestCaseGroupMeta($yaml, 'test-group', $messages);

        $this->assertNull($result);
        $this->assertNotEmpty($messages['danger']);
        $this->assertStringContainsString("Invalid range '100 101 102'", $messages['danger'][0]);
    }
```

webapp/tests/Unit/Service/ImportProblemServiceTest.php
meisterT: I started this, but then the code was ugly-ish, so I didn't finish it. I think this is not too bad.

webapp/tests/Unit/Service/ImportProblemServiceTest.php
meisterT: https://github.com/DOMjudge/domjudge/pull/3475

PR #3421, opened 2026-02-04
Fix keying for disabled languages.

meisterT:


meisterT:
Otherwise we would see something like:

image

PR #3424, opened 2026-02-06
Style language edit/view similar to the contest edit/view pages.

meisterT:
While we are at it:
- add inline toggle/route for entry point (similar as we had for other boolean toggles)
- add missing data on view page (restrictions)

----- 

Example:

image

and

image


meisterT:
Some more screenshots:

image

---

image

---

image

---

image

PR #3425, opened 2026-02-07
Support time limit in `problem.yaml` -> `limits` section.

meisterT:
Also fail when it disagrees with the one in `.timelimit`.

### DIFF
--- a/webapp/src/Service/ImportProblemService.php
+++ b/webapp/src/Service/ImportProblemService.php
@@ -117,10 +117,21 @@ readonly class ImportProblemService
             unset($contestProblemProperties['short-name']);
         }
 
-        // Set timelimit from alternative source:
-        if (!isset($problemProperties['timelimit']) &&
-            ($str = $zip->getFromName($tleFile)) !== false) {
-            $problemProperties['timelimit'] = trim($str);
+        // Set timelimit from alternative source (.timelimit file):
+        $tleFileValue = null;
+        if (($str = $zip->getFromName($tleFile)) !== false) {
+            $trimmed = trim($str);
+            if (!is_numeric($trimmed)) {
+                $messages['danger'][] = sprintf(
+                    'Invalid time limit in .timelimit file: %s is not a valid number.',
+                    $trimmed
+                );
+                return null;
+            }
+            $tleFileValue = (float)$trimmed;
+            if (!isset($problemProperties['timelimit'])) {
+                $problemProperties['timelimit'] = $tleFileValue;
+            }
         }
 
         // Take problem:externalid from zip filename, and use as backup for
@@ -237,6 +248,15 @@ readonly class ImportProblemService
         // types) are set before the entity validation callbacks fire.
         $validationMode = 'default';
         if (!static::parseYaml($problemYaml, $messages, $validationMode, $propertyAccessor, $problem)) {
+            return null;
+        }
+
+        // Check for conflicting time limits between .timelimit file and problem.yaml.
+        if ($tleFileValue !== null && abs($problem->getTimelimit() - $tleFileValue) > 1e-9) {
+            $messages['danger'][] = sprintf(
+                'Conflicting time limits: .timelimit file specifies %s seconds, but problem.yaml specifies %s seconds.',
+                $tleFileValue, $problem->getTimelimit()
+            );
             return null;
         }
 
@@ -1234,6 +1254,9 @@ readonly class ImportProblemService
         }
 
         if (isset($yamlData['limits'])) {
+            if (isset($yamlData['limits']['time_limit'])) {
+                $yamlProblemProperties['timelimit'] = $yamlData['limits']['time_limit'];
+            }
             if (isset($yamlData['limits']['memory'])) {
                 $yamlProblemProperties['memlimit'] = 1024 * $yamlData['limits']['memory'];
             }
--- a/webapp/tests/Unit/Service/ImportProblemServiceTest.php
+++ b/webapp/tests/Unit/Service/ImportProblemServiceTest.php
@@ -6,6 +6,7 @@ use App\Entity\Problem;
 use App\Service\ImportProblemService;
 use App\Tests\Unit\BaseTestCase;
 use Symfony\Component\PropertyAccess\PropertyAccess;
+use ZipArchive;
 
 class ImportProblemServiceTest extends BaseTestCase
 {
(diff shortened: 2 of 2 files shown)


meisterT:
Example warning: 
image

meisterT:
> should we also have the negative/failure case?

Done

nickygerritsen (approved):

vmcj:
> Done

I don't see it (yet). The test seems to be missing.

vmcj:
> Done

I don't see it (yet). The test seems to be missing.

webapp/tests/Unit/Service/ImportProblemServiceTest.php
```diff
@@ -399,4 +438,84 @@ public function testParseTestCaseGroupMetaInvalidRangeFormatRejected(): void
...
+        $zip = new ZipArchive();
+        $zip->open($tempFile, ZipArchive::CREATE);
+        foreach ($files as $name => $content) {
+            $zip->addFromString($name, $content);
+        }
+        $zip->close();
+        return $tempFile;
+    }
+
+    public function testTimelimitConflictDetected(): void
```
meisterT, line 562: @vmcj the test is already here - can you see this file? If you can see it and meant a different test scenario, please let me know.

vmcj: I meant something like:
```
name: test
limits:
  time_limit: -5
YAML;
```

So a negative value. Thinking about it we should also see how:
```
name: test
limits:
  time_limit: 5m
YAML;
```

is handled and if we correctly warn the user.

meisterT: Understood now, added tests for both of these cases.

webapp/src/Service/ImportProblemService.php
```diff
@@ -117,10 +117,13 @@ public function importZippedProblem(
...
-        if (!isset($problemProperties['timelimit']) &&
-            ($str = $zip->getFromName($tleFile)) !== false) {
-            $problemProperties['timelimit'] = trim($str);
+        // Set timelimit from alternative source (.timelimit file):
+        $tleFileValue = null;
+        if (($str = $zip->getFromName($tleFile)) !== false) {
+            $tleFileValue = (float)trim($str);
+            if (!isset($problemProperties['timelimit'])) {
+                $problemProperties['timelimit'] = $tleFileValue;
+            }
```
vmcj, line 134: Can we add L246 here? I want halfway commenting this case was missing before I found that check.

webapp/tests/Unit/Service/ImportProblemServiceTest.php
```diff
@@ -399,4 +484,84 @@ public function testParseTestCaseGroupMetaInvalidRangeFormatRejected(): void
...
+
+        $zip->close();
+        unlink($zipFile);
+
+        $this->assertNull($result);
+        $this->assertNotEmpty($messages['danger']);
+        $this->assertStringContainsString('Conflicting time limits', $messages['danger'][0]);
+        $this->assertStringContainsString('10', $messages['danger'][0]);
+        $this->assertStringContainsString('5', $messages['danger'][0]);
+    }
```
vmcj, line 592:
```suggestion
    public function testTimelimitConflictDetected(): void
    {
        $yaml = <<<YAML
name: test
limits:
  time_limit: 5
YAML;
        $zipFile = $this->createZipWithContents([
            'problem.yaml' => $yaml,
            '.timelimit' => '10',
        ]);

        $zip = new ZipArchive();
        $zip->open($zipFile);

        /** @var ImportProblemService $service */
        $service = static::getContainer()->get(ImportProblemService::class);
        /** @var array{info: string[], warning: string[], danger: string[]} $messages */
        $messages = ['info' => [], 'warning' => [], 'danger' => []];

        $result = $service->importZippedProblem($zip, 'test-problem.zip', null, null, $messages);

        $zip->close();
        unlink($zipFile);

        $this->assertNull($result);
        $this->assertNotEmpty($messages['danger']);
        $this->assertStringContainsString('Conflicting time limits', $messages['danger'][0]);
        $this->assertStringContainsString('10', $messages['danger'][0]);
        $this->assertStringContainsString('5', $messages['danger'][0]);
    }

    public function testTimelimitConflict2(): void
    {
        $yaml = <<<YAML
name: test
YAML;
        $zipFile = $this->createZipWithContents([
            'problem.yaml' => $yaml,
            '.timelimit' => '1s',
        ]);

        $zip = new ZipArchive();
        $zip->open($zipFile);

        /** @var ImportProblemService $service */
        $service = static::getContainer()->get(ImportProblemService::class);
        /** @var array{info: string[], warning: string[], danger: string[]} $messages */
        $messages = ['info' => [], 'warning' => [], 'danger' => ['invalid type']];

        $result = $service->importZippedProblem($zip, 'test-problem.zip', null, null, $messages);

        $zip->close();
        unlink($zipFile);

        $this->assertNull($result);
        $this->assertNotEmpty($messages['danger']);
        $this->assertStringContainsString('Conflicting time limits', $messages['danger'][0]);
        $this->assertStringContainsString('10', $messages['danger'][0]);
        $this->assertStringContainsString('5', $messages['danger'][0]);
    }

public function testTimelimitConflict3(): voidExpand commentComment on line R505
    {
        $yaml = <<<YAML
name: test
YAML;
        $zipFile = $this->createZipWithContents([
            'problem.yaml' => $yaml,
            '.timelimit' => '0',
        ]);

        $zip = new ZipArchive();
        $zip->open($zipFile);

        /** @var ImportProblemService $service */
        $service = static::getContainer()->get(ImportProblemService::class);
        /** @var array{info: string[], warning: string[], danger: string[]} $messages */
        $messages = ['info' => [], 'warning' => [], 'danger' => ['should not be 0']];

        $result = $service->importZippedProblem($zip, 'test-problem.zip', null, null, $messages);

        $zip->close();
        unlink($zipFile);

        $this->assertNull($result);
        $this->assertNotEmpty($messages['danger']);
        $this->assertStringContainsString('Conflicting time limits', $messages['danger'][0]);
        $this->assertStringContainsString('10', $messages['danger'][0]);
        $this->assertStringContainsString('5', $messages['danger'][0]);
    }
```

Don´t apply directly, only to show what I think should be added also.

vmcj (approved):
2 extra tests, can be merged without those if you think they're not needed.

webapp/src/Service/ImportProblemService.php
meisterT: No, we only parse the yaml directly before L246

webapp/tests/Unit/Service/ImportProblemServiceTest.php
meisterT: Good idea, both added (and fixed the code as well :-) ).

PR #3426, opened 2026-02-07
Move scoring diff threshold to a contest level tunable value.

meisterT:


meisterT:
cc @vmcj who suggested this

PR #3427, opened 2026-02-07
PoC: Datatables scoreboard

Kevinjil:
This is not merge-able code, but demonstrates the idea to start a discussion whether or not this is a direction we want to go.

Currently, we have core logic about what is visible to users on both the API, the mobile scoreboard, and the normal scoreboard. There have been some issues that some information was visible which should not have been given configuration settings. If we render the scoreboard on the frontend, we can eliminate this complex but important logic.

The other benefit is that it enables advanced features on the frontend. If we define per scoring type a rendering function and a sorting function on the front-end, we can dynamically sort and filter without putting a huge load on the DOMjudge server.

scoreboard

For the static scoreboard, the JSONs should be included, similar to how the submission data is stored. This is not included in this PoC.

_As this is a draft, created some time ago, I have not rebased on main. If we think this is a good direction to go, I can work on a proper implementation and on current main._

### DIFF
--- a/webapp/templates/partials/scoreboard_table.html.twig
+++ b/webapp/templates/partials/scoreboard_table.html.twig
@@ -29,6 +29,254 @@
     </style>
 {% endif %}
 
+<table id="scoreboard-table" class="table scoreboard">
+	<thead>
+		<tr class="scoreheader">
+			<th class="dt-center" colspan="2">Rank</th>
+			<th class="dt-center" colspan="3">Team</th>
+			<th class="dt-center" colspan="2" data-dt-order="disable">Score</th>
+            {% for problem in problems %}
+                {% set link = null %}
+                {% set target = '_self' %}
+                {% if not static %}
+                    {% if jury %}
+                        {% set link = path('jury_problem', {'probId': problem.probid}) %}
+                    {% elseif problem.problem.problemstatementType is not empty %}
+                        {% if public %}
+                            {% set link = path('public_problem_statement', {probId: problem.probid}) %}
+                            {% set target = '_blank' %}
+                        {% else %}
+                            {% set link = path('team_problem_statement', {probId: problem.probid}) %}
+                            {% set target = '_blank' %}
+                        {% endif %}
+                    {% endif %}
+                {% endif %}
+                <th class="dt-center"
+                    title="problem {{ problem.problem.name }}"
+                    scope="col"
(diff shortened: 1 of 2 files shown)


meisterT:
In general, using our API to provide data and then doing rendering on the client side is fine and promising.

However, I dislike adding sorting functions into the client. We should keep a single source of truth in the backend, e.g. by exposing the `sort_key_public` and `sort_key_restricted` (if access).

Kevinjil:
> However, I dislike adding sorting functions into the client. We should keep a single source of truth in the backend, e.g. by exposing the `sort_key_public` and `sort_key_restricted` (if access).

Having two implementations of sorting is indeed tricky. For this PoC, I explicitly used the backend-provided sorting for the "normal" scoreboard view, and only implemented some custom sorting for a pass/fail problem type to sort on a single problem. I think that implementing a sorting function for these non-standard cases in the frontend is fine, but would love to hear other's opinions.

meisterT:
> I think that implementing a sorting function for these non-standard cases in the frontend is fine, but would love to hear other's opinions.

What's the argument against exposing the sort_key in the API so you don't have to re-implement sorting functions in multiple places?

Kevinjil:
> What's the argument against exposing the sort_key in the API so you don't have to re-implement sorting functions in multiple places?

I might have missed some changes, but as far as I am aware I could not sort on the individual problems based on merely the overall sort key. We could introduce sort-keys per-problem.

webapp/templates/partials/scoreboard_table.html.twig
```diff
@@ -29,7 +29,254 @@
...
+<script>
+$.ajax('/api/v4/contests/{{contest.externalid}}/teams').then((_teams) => {
+/* $.ajax('/teams.json').then((_teams) => { */
+    const teams = {};
+    for (team of _teams) {
+        teams[team.id] = team;
+    }
+
+    let sortCmp = (a, b) => ((a < b) ? -1 : ((a > b) ? 1 : 0));
+    let sortProblem = (a, b) => {
```
Kevinjil, line 85: Basically, to reproduce this (ugly) implementation

meisterT:
Do I understand correctly that you want to be able to filter columns and not just rows? Just rows would be trivial by the sort_key, filtering columns is indeed problematic

PR #3430, opened 2026-02-07
Expose source service loading error.

meisterT:
Previously we were saying `The contest source is not valid.`, nothing else.

### DIFF
--- a/webapp/src/Service/ExternalContestSourceService.php
+++ b/webapp/src/Service/ExternalContestSourceService.php
@@ -168,7 +168,7 @@ class ExternalContestSourceService
     public function getSourceContest(): Contest
     {
         if (!$this->isValidContestSource()) {
-            throw new LogicException('The contest source is not valid.');
+            $this->throwInvalidSourceException();
         }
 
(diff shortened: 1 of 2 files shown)


webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -189,10 +189,19 @@ public function isValidContestSource(): bool
         return ($this->cachedContestData !== null);
     }
 
+    /**
+     * Throws an exception indicating the contest source is invalid, including the actual error message.
+     * This should only be called after isValidContestSource() returns false.
+     */
+    private function throwInvalidSourceException(): never
+    {
+        throw new LogicException('The contest source is not valid: ' . ($this->loadingError ?? 'Unknown error'));
```
vmcj, line 198: Instead of the comment, can we just do the check again? Maybe PHPStan will warn about it but I don´t really like comment in code telling us we should only call something in case something is the case instead of just making sure it is the case.

We could even get a `maybeThrowInvalidSourceException` function.

PR #3431, opened 2026-02-07
Make submit errors more readable.

meisterT:


webapp/src/Service/SubmissionService.php
```diff
@@ -854,18 +854,20 @@ public function submitSolution(
             $allowedLanguages = $this->dj->getAllowedLanguagesForContest($contest);
             if (!in_array($language, $allowedLanguages, strict: true)) {
                 throw new BadRequestHttpException(
-                    sprintf("Language '%s' not allowed for contest [c%d].", $language->getLangid(), $contest->getCid()));
+                    sprintf("Language '%s' not allowed for contest [c%d / %s].",
+                        $language->getName(), $contest->getCid(), $contest->getName()));
```
vmcj: I think displaying the internal id is not needed anymore? So I think fully removing it is better.

webapp/src/Service/SubmissionService.php
```diff
@@ -854,18 +854,20 @@ public function submitSolution(
...
             }
         } else {
             $allowedLanguages = $allowedLanguages->toArray();
             if (!in_array($language, $allowedLanguages, strict: true)) {
                 throw new BadRequestHttpException(
-                    sprintf("Language '%s' not allowed for problem [p%d].", $language->getLangid(), $problem->getProbid()));
+                    sprintf("Language '%s' not allowed for problem [p%d / %s].",
```
nickygerritsen: Same here  I guess for probid

webapp/src/Service/SubmissionService.php
meisterT: good point, removed

webapp/src/Service/SubmissionService.php
meisterT: done

PR #3435, opened 2026-02-08
Improve runner/compiler checks

meisterT:
- trust on first use option
- enforcement option

-----

Example error:
image

### DIFF
--- a/webapp/src/Controller/API/JudgehostController.php
+++ b/webapp/src/Controller/API/JudgehostController.php
@@ -45,6 +45,7 @@ use Nelmio\ApiDocBundle\Attribute\Model;
 use OpenApi\Attributes as OA;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\ExpressionLanguage\Expression;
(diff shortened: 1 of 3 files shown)


webapp/src/Controller/API/JudgehostController.php
```diff
@@ -1461,12 +1471,126 @@ public function checkVersions(Request $request, string $judgetaskid): array
                 $activeVersion->setLastChangedTime(Utils::now());
                 $this->em->persist($activeVersion);
                 $this->em->flush();
+
+                // TOFU: Trust On First Use - auto-promote first reported version as canonical
```
nickygerritsen: This comment seems a bit double with the TOFU and Trust On First Use (but I like TOFU 😛 )

nickygerritsen (approved):

webapp/src/Controller/API/JudgehostController.php
meisterT: Done

PR #3436, opened 2026-02-08
Improve layout of compiler/runner version overview in team interface.

meisterT:
This is draft because we likely want to merge https://github.com/DOMjudge/domjudge/pull/3178 first (and afterwards will have merge conflicts with this one).

------

Example:

Before:

image

After:

image


vmcj:
I propose to do it the other way around, as I already have merge conflicts there which I think were non-trivial so need some extra work.

vmcj (approved):

PR #3438, opened 2026-02-09
Install DS to get set functionality for PHP

vmcj:
I needed sets for another PR and I know we discussed this in the past so looked into how hard it is to support this.

The annoying thing is the installation for fedora, adding this syntactic sugar (and maybe a performance boost) does make installation harder on other distributions.

### DIFF
--- a/doc/manual/install-domserver.rst
+++ b/doc/manual/install-domserver.rst
@@ -24,7 +24,7 @@ System requirements
 
 Software requirements
 `````````````````````
-* A web server with support for PHP >= 8.2.0 and the ``curl``, ``gd``, ``intl``,
+* A web server with support for PHP >= 8.2.0 and the ``curl``, ``ds``, ``gd``, ``intl``,
   ``json``, ``mbstring``, ``mysqli``, ``xml`` and ``zip`` extensions for PHP.
 * MySQL or MariaDB database. This can be on the same machine, but for
   advanced setups can also run on a dedicated machine.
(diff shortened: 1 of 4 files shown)


nickygerritsen:
If we do this we should also add https://github.com/php-ds/polyfill.

I'm in favor of `DS` since they types are awesome, but the Fedora thing is a bit annoying maybe,

meisterT:
Does the polyfill thingy reduce the requirement to install the `php-ds` extension from being a hard requirement to a soft requirement (just making it faster but not fail if not available/installed)?

nickygerritsen:
> Does the polyfill thingy reduce the requirement to install the `php-ds` extension from being a hard requirement to a soft requirement (just making it faster but not fail if not available/installed)?

It does.

vmcj:
> If we do this we should also add https://github.com/php-ds/polyfill.
> 
> I'm in favor of `DS` since they types are awesome, but the Fedora thing is a bit annoying maybe,

So that fixes the fedora issue right? I just want sets in PHP I don't think we need the speed (yet) so only doing the polyfill might also be fine in our setups.

nickygerritsen:
> > If we do this we should also add https://github.com/php-ds/polyfill.
> > 
> > I'm in favor of `DS` since they types are awesome, but the Fedora thing is a bit annoying maybe,
> 
> So that fixes the fedora issue right? I just want sets in PHP I don't think we need the speed (yet) so only doing the polyfill might also be fine in our setups.
> 

Yep

doc/manual/install-domserver.rst
```diff
@@ -36,22 +36,26 @@ software on the DOMjudge server as mentioned above when using Debian
...
 
 The following command can be used on Fedora, and related distributions like
-Red Hat Enterprise Linux and Rocky Linux (before V9)::
+Red Hat Enterprise Linux and Rocky Linux (before V9)[#ds]_::
 
   sudo dnf install libcgroup-devel make acl zip unzip pv mariadb-server httpd \
         php-gd php-cli php-intl php-mbstring php-mysqlnd php-fpm \
         php-xml php-zip composer chronyd python3-pyyaml php-bcmath
 
+For `php-ds` follow the instructions at: https://www.php.net/manual/en/ds.installation.php
```
nickygerritsen: This is here twice now.

vmcj: Good catch.

PR #3439, opened 2026-02-09
Split php extensions

vmcj:
Those changes were part of my other PR but as that one might need more discussion (or might even not be implemented) it's better to merge the fix I did separately


meisterT:
Note that your tests failing is likely github's problem: https://www.githubstatus.com/

PR #3440, opened 2026-02-09
Fix variable access in judgedaemon.

meisterT:


judge/judgedaemon.main.php
```diff
@@ -308,9 +308,9 @@ public function __construct()
             }
 
             $output = [];
-            exec("ps -u '$runuser' -o pid= -o comm=", $output, $retval);
+            exec("ps -u '$this->runuser' -o pid= -o comm=", $output, $retval);
```
vmcj, line 311: Why does PHPStan not detect this?

meisterT: :shrug:

vmcj: So I checked, because we don't check the judgedaemon code with PHPStan. I think we should actually do that.

meisterT: Ack, might be for historical reasons. I am merging this anyway, so that we can judge without confusing error.

PR #3442, opened 2026-02-09
Only test our minimal & maximum PHP version

vmcj:


meisterT:
Is this to speed up CI?

vmcj:
> Is this to speed up CI?

For a couple of reasons:
- Allocation of github runners is slow, we can't steer the integration job to get the first runner (it takes the longest so we want to have it start soon)
- There is not really anything which the PHP8.2 & PHP8.5 wouldn't catch which the version in between do,
- We added the different versions for the Unit/Integration tests to detect certain invocations which are deprecated or to new, it helped us for the integration job in the past (which was catched by the unit tests also for the same paths when we tested).
- During hackathons we tend to need a lot of runners, so saving everywhere where possible is good IMHO (and it might save some forests?)

So we could just pick the `8.5` which should be faster but I think testing with both of our boundaries is better, if you think we should test with all of them we should add the `8.5` but I think the `8.3` & `8.4` will not add a lot of value.

PR #3443, opened 2026-02-09
Do initial PHPStan run for judgedaemon code

vmcj:
This leave a couple of issues for next PRs as they require more individual thought. Best reviewed commit by commit so we can squash the parts together which we think are correct.

### DIFF
--- a/phpstan-baseline-judgedaemon.neon
+++ b/phpstan-baseline-judgedaemon.neon
@@ -19,6 +19,42 @@ parameters:
 			path: judge/judgedaemon.main.php
 
 		-
+			message: '#^Negated boolean expression is always false\.$#'
+			identifier: booleanNot.alwaysFalse
+			count: 1
+			path: judge/judgedaemon.main.php
+
+		-
+			message: '#^Offset ''pass_limit'' on array\{time_limit\: float, memory_limit\: int, output_limit\: int, process_limit\: int, entry_point\: string\|null, pass_limit\: int, hash\: string, overshoot\: int\} on left side of \?\? always exists and is not nullable\.$#'
+			identifier: nullCoalesce.offset
+			count: 1
+			path: judge/judgedaemon.main.php
+
+		-
+			message: '#^PHPDoc tag @var with type array\<string\> is not subtype of native type list\<array\{string, int\<0, max\>\}\|string\>\|false\.$#'
+			identifier: varTag.nativeType
+			count: 1
+			path: judge/judgedaemon.main.php
+
+		-
+			message: '#^Parameter \#3 \$data of method DOMjudge\\JudgeDaemon\:\:request\(\) expects array\<string, string\>\|string, array\<string, int\|string\> given\.$#'
+			identifier: argument.type
+			count: 1
+			path: judge/judgedaemon.main.php
+
+		-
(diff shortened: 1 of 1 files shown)


judge/judgedaemon.main.php
```diff
@@ -629,6 +629,7 @@ private function handleDebugInfoTask(array $row, ?string &$lastWorkdir, string $
             } else {
                 // Retrieving full team output for a particular testcase.
                 $testcasedir = $workdir . "/testcase" . sprintf('%05d', $judgeTask['testcase_id']);
+                // TODO: Properly fix this by sending the pass in the run_config instead of hardcoding 1
```
meisterT: which 1, I assume this was the other change you made recently but I don't see it here

judge/judgedaemon.main.php
```diff
@@ -2216,6 +2216,15 @@ private function runTestcase(
         $input = $tcfile['input'];
         $output = $tcfile['output'];
         $passLimit = $run_config['pass_limit'] ?? 1;
+        // All of those are to help PHPStan, it seems to not see that they are defined in the loop
+        // and the loop always runs as tpassLimit >= $passCnt.
+        $score = "";
+        $nextPass = false;
+        $result = Verdict::INTERNAL_ERROR;
+        $runtime = null;
+        $new_judging_run = ['runresult' => '', 'start_time' => '', 'end_time' => '', 'runtime' => '', 'output_run' => '', 'output_error' => '',
```
meisterT: can we format this nicer, e.g. over multiple lines?

meisterT: hmmm, why is this necessary at all?

judge/judgedaemon.main.php
```diff
@@ -36,6 +36,18 @@ enum Verdict
 
 /**
  * Represents program execution metadata with validated fields.
+ *
```
meisterT: Can't say I am a fan of these long docs that don't add value to humans

phpstan-baseline-judgedaemon.neon
meisterT, line 1: looks like you want to do some commit squashing in the end

judge/judgedaemon.main.php
```diff
@@ -1333,10 +1333,10 @@ private function fetchExecutableInternal(
                 }
                 chmod($execrunpath, 0755);
             }
-            if (!is_file($execrunpath) || !is_executable($execrunpath)) {
```
meisterT, line 1288: this doesn't look right to me at first glance, leaving a comment mostly to take more time to dig deeper

judge/judgedaemon.main.php
meisterT, line 2122: there are some phpstan errors here:

Image

judge/judgedaemon.main.php
vmcj: Yes, I apparantly had it documented her, forgot to extract it to a separate PR including the TODO so this just makes sure the comment is there. I'll clarify it even further.

judge/judgedaemon.main.php
vmcj: PHPStan doesn´t see that the loop always runs so complains that those variables might be unset outside of the loop. For example in https://github.com/DOMjudge/domjudge/pull/3443/changes/BASE..c48a34cd4460764597e8a84db8a475ba3a8089e2#diff-495a2bf2ac4904f9b10777c3d8643cf0de9ee5499d64a15fc1a806e350180bb6R2481

PR #3445, opened 2026-02-09
Run on our CI image.

meisterT:


meisterT:
@vmcj I assume this is what you meant?

vmcj:
> @vmcj I assume this is what you meant?

Partially, by doing this we can remove some external dependencies for the php setup action. And maybe even integrate this with the other `unit-tests.yml`.

meisterT:
Ok, I can remove deps after your PR to the same file has been merged.

I do prefer keeping it separate from the other unit tests though because it is nicer to get a high level signal of breaking the judging unit tests vs the web unit tests.

vmcj:
> I do prefer keeping it separate from the other unit tests though because it is nicer to get a high level signal of breaking the judging unit tests vs the web unit tests.

I would suggest doing it in a matrix job so it would still get its own entry (it's own testcase group) but just less duplicated code in the github folder.

meisterT:
The judge unit test do not require a full install (as the other unit tests do) and are thus much faster. If possible, we should keep that property. Do you think that is possible while moving them to a matrix job?

vmcj (approved):

PR #3448, opened 2026-02-10
Remove `->setAccessible(...)` calls from our reflective tests.

meisterT:
They are no-ops since PHP 8.1 and get flagged as deprecated by our unit tests when we enabled testing for PHP 8.5.


vmcj (approved):
Was the first PR I would have opened, so if it passes CI we can merge.

PR #3457, opened 2026-02-15
Add a progress message for when we didn't receive new events.

meisterT:


webapp/src/Service/ExternalContestSourceService.php
```diff
@@ -539,6 +539,7 @@ protected function importFromCcsApi(array $eventsToSkip, ?callable $progressRepo
                         }
                     }
                     if (!$receivedData) {
+                        $this->updateProgress('Waiting for new events (last: ' . ($this->getLastReadEventId() ?? 'none') . ')');
```
vmcj, line 542: Would this display only once or everytime? We had problems in the past with the import script filling up the disk when we didn't turn it off after shadowing.

meisterT: It would display it once, and never in the log - see the implementation of `updateProgress(...)` which does not call the logger.

PR #3461, opened 2026-02-15
Make score nullable to able to distinguish between no score and 0.

meisterT:


nickygerritsen (approved):

vmcj:
I suggest we merge most of those transactions for scoring into 1 before releasing.

PR #3463, opened 2026-02-15
Document the parsed Metadata files as return type

vmcj:
The last combines the compile & program meta files, one has the signal the other the entrypoint so those are marked as optional to not duplicate everything.

Opening this as a PR as I think we need to discuss how we want to document such cases with almost the same array, I personally think this one is the best as it's a good mix between steering PHPstan and not using up too much space but liked my initial version with all 3 split out also enough apart from the duplication.

I can't check yet if PHPstan actually parses this correctly but I seem to recall we did this before.

It seems all keys should actually be string and as mentioned by @meisterT it has a typo, this PR is mostly to discuss what shape we want to document.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -36,6 +36,12 @@ enum Verdict
 
 /**
  * Represents program execution metadata with validated fields.
+ * @phpstan-type MetaData_Program array{
+ *     exitcode: string, memory-bytes: string, time-used: string, time-result: string,
+ *     stdin-bytes: string, stdout-bytes: string, stderr-bytes: string,
+ *     cpu-time: string, sys-time: string, user-time: string, wall-time: string,
+ *     entry_point?: string, output-truncated?: string, signal?: string
+ * }
  */
 readonly class ProgramMetadata
 {
@@ -53,6 +59,7 @@ readonly class ProgramMetadata
 
     /**
      * Create from raw metadata array with validation.
+     * @param MetaData_Program $meta
      */
     public static function fromArray(array $meta): self
     {
@@ -95,6 +102,12 @@ readonly class ProgramMetadata
 
 /**
  * Represents compare script execution metadata with validated fields.
+ * @phpstan-type MetaData_Compare array{
+ *     exitcode: string, memory-bytes: string, time-used: string, time-result: string, bytes-transferred: string,
+ *     stdin-bytes: string, stdout-bytes: string, stderr-bytes: string, validator-exited-first: string,
+ *     cpu-time: string, sys-time: string, user-time: string, wall-time: string, total-duration-use: string,
+ *     output-truncated?: string
+ * }
  */
 readonly class CompareMetadata
 {
@@ -106,6 +119,7 @@ readonly class CompareMetadata
 
     /**
      * Create from raw metadata array with validation.
+     * @param MetaData_Compare $meta
      */
     public static function fromArray(array $meta): self
     {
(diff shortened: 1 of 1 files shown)


judge/judgedaemon.main.php
```diff
@@ -1322,6 +1322,17 @@ private function disable(
...
 
+    /**
+     * @return array{
+     *        exitcode: int, bytes-transferred: int, total-duration-us: int,
+     *        validator-exited-first: bool
+     *    }|array{
+     *        memory-bytes: int, exitcode: int, signal?: int, wall-time: float,
+     *        user-time: float, sys-time: float, cpu-time: float, time-used: string
+     *        time-result: string, output-truncated: string, stdin-bytes: int,
+     *        stdout-bytes: int, stderr-bytes: int, entry_point:: int
```
meisterT: What do the two `::` indicate? Or is it a typo?

judge/judgedaemon.main.php
```diff
@@ -1322,6 +1322,17 @@ private function disable(
         logmsg(LOG_ERR, "=> internal error " . $error_id);
     }
 
+    /**
+     * @return array{
+     *        exitcode: int, bytes-transferred: int, total-duration-us: int,
```
meisterT: What happens if not all fields are set?

judge/judgedaemon.main.php
vmcj: Is indeed a typo.

judge/judgedaemon.main.php
vmcj: array{foo?: int} would be an array where `foo` is optional so could be a named array with 0 or 1 elements but if it has it has to have key `foo`. So for those all have to be set, that's why there are 2 possible array shapes where the 2nd has 2 optional keys.

meisterT: I have to go through them one by one, I assume not all of these are set all the time, so we might need to sprinkle more `?`

judge/judgedaemon.main.php
meisterT: Also, is `entry_point` indeed an `int`? I would have expected it to be a `string` without looking at the code, but perhaps it is a boolean disguised as int?

vmcj: all of those should actually be string, we seem to only read the file and not guess on the type of the `value`.

vmcj: All of those seem to be strings as we don't guess/know the type, so I'll update that later.

judge/judgedaemon.main.php
vmcj: I ran `cat` on all 3 types of files, so we can have 3 array shapes where I merged the last 2 into 1 as they only differ by 2 properties. It might be we have different properties per language & problem type but I didn't find those on quick inspection.

I opened this PR mostly to get an answer on if we want to do:
`@returns array{foo: string, bar: string}|array{foo:string, alice: string, charlie:string}|array{foo:string, alice: string, bob:string}` vs `@returns array{foo:string, bar?:string, alice?: string, bob?:string, charlie?:string}` or the middle ground which I presented above.

meisterT: I don't have a strong opinion. Perhaps @nickygerritsen does?

nickygerritsen: We can have this: https://phpstan.org/writing-php-code/phpdoc-types#local-type-aliases
Maybe that's a good middle ground?

vmcj: So we would define them once very verbose on the top and use the short form in the code itself? SGTM

nickygerritsen: Yes I think that reads a bit better?

vmcj:
I wonder if the `internal-error` key should actually be near the classes, I think we should never create an instance of that class if we still have this array shape `array{internal-error: string}`, @meisterT I assume you know this answer?

meisterT:
I don't follow what the question is, can you please elaborate?

vmcj:
> I don't follow what the question is, can you please elaborate?

If I understand the code the only time where a `*.meta` file would have `internal-error` would be if runguard crashed hard. At that time we don't have any of the other values so we could read the `*.meta` file but not create an instance of `ProgramMetaData` or `CompareMetaData`. So the shape we would get in the `fromArray` of those functions would not be `array{key1: ... last_key:}|array{internal-error: string}` but only `array{key1: ... last_key:}`.

I couldn't find other paths where we seem to write the `internal-error` key into the metadata but hoped you knew the code enough to confirm that that's the case.

meisterT:
The meta data might be incomplete,  e.g. when you die in this line: https://github.com/DOMjudge/domjudge/blob/8171b16a91717193e6bd2e779b1d5331327d3f97/judge/runguard.cc#L394

So, it can contain a could of key value pairs and then internal_error, but not the other keys that would have been written later if were still alive.

vmcj:
> The meta data might be incomplete, e.g. when you die in this line:
> 
> https://github.com/DOMjudge/domjudge/blob/8171b16a91717193e6bd2e779b1d5331327d3f97/judge/runguard.cc#L394
> 
> So, it can contain a could of key value pairs and then internal_error, but not the other keys that would have been written later if were still alive.

And I assume that's for all lines where we die? So that's still very hard to properly describe.

meisterT:
Yeah, since everything is `string` typed and optional, I am no longer sure adding this annotation adds value.

vmcj:
> Yeah, since everything is `string` typed and optional, I am no longer sure adding this annotation adds value.

It's only optional when `internal-error` is set. In the future I think we should explicit check for `internal-error` and if it is set handle that and if not assume all properties are set as they normally are.

meisterT:
We at least need to add more question marks for things that are written conditionally:
- `signal`
- `output_truncated`

PR #3465, opened 2026-02-17
PHPStan fixes for the `/lib/lib.*.php` files

vmcj:
~~I left a couple open as I need to take a closer look why PHPstan would be correct.~~, PHPstan was incorrect so I massaged the code to help PHPstan a bit more.

I think I'll remove the changes to PHPstan before this gets merged so we don't need a baseline file yet and move the file structure in a separate PR.

### DIFF
--- a/lib/lib.error.php
+++ b/lib/lib.error.php
@@ -48,7 +48,7 @@ function logmsg(int $msglevel, string $string): void
     // Trim $string to reasonable length
     $string = substr($string, 0, 10000);
 
-    $msec = sprintf("%03d", (int)(explode(' ', microtime())[0]*1000));
+    $msec = sprintf("%03d", (int)(explode(' ', microtime()))[0]*1000);
     $stamp = "[" . date('M d H:i:s') . ".$msec] " . SCRIPT_ID .
         (function_exists('posix_getpid') ? "[" . posix_getpid() . "]" : "") .
         ": ";
--- a/lib/lib.misc.php
+++ b/lib/lib.misc.php
@@ -15,7 +15,6 @@ require_once('lib.wrappers.php');
  */
 function overshoot_time(float $timelimit, string $overshoot_cfg) : float
 {
-    /** @var string[] $tokens */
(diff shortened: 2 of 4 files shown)


lib/lib.error.php
```diff
@@ -32,9 +32,7 @@
 
 // Open syslog connection:
 if (defined('SYSLOG')) {
-    if (!openlog(SCRIPT_ID, LOG_NDELAY | LOG_PID | LOG_CONS, SYSLOG)) {
-        error("cannot open syslog");
```
meisterT, line 36: Why did you remove the error checking here?

lib/lib.misc.php
```diff
@@ -15,14 +15,13 @@
  */
 function overshoot_time(float $timelimit, string $overshoot_cfg) : float
 {
-    /** @var string[] $tokens */
     $tokens = preg_split('/([+&|])/', $overshoot_cfg, -1, PREG_SPLIT_DELIM_CAPTURE);
-    if (count($tokens)!=1 && count($tokens)!=3) {
+    if ($tokens === false || (count($tokens) !== 1 && count($tokens) !== 3)) {
```
meisterT: nit: my preference is to separately check for

`if ($tokens === false) { error("some good message here"); }`

and then separately check for the count, also include the count in the message?

lib/lib.error.php
nickygerritsen: See the return type here: https://www.php.net/openlog

> Return Values
> Always returns [true](https://www.php.net/manual/en/reserved.constants.php#constant.true).
>
> Changelog
> Version	Description
> 8.2.0	The function now always returns [true](https://www.php.net/manual/en/reserved.constants.php#constant.true). Previously it returned [false](https://www.php.net/manual/en/reserved.constants.php#constant.false) on failure.

meisterT: but why? :-)

lib/lib.misc.php
vmcj: Should be fixed now.

vmcj:
> I recommend squashing the commits

I think there are a couple where having a more specific commit message makes sense?

meisterT:
Your call, feel free to merge as you see fit

PR #3468, opened 2026-02-21
Use PHPstan for all PHP files

vmcj:
This means also scanning the (judgedaemon) configuration files.

The actual judgedaemon code is for another PR as it requires more work.

### DIFF
--- a/lib/lib.misc.php
+++ b/lib/lib.misc.php
@@ -120,6 +120,9 @@ function initsignals(): void
  */
 function version() : never
 {
+    if (!defined('SCRIPT_ID')) {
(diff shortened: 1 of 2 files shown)


nickygerritsen (approved):

lib/lib.misc.php
```diff
@@ -120,6 +120,9 @@ function initsignals(): void
  */
 function version() : never
 {
+    if (! defined('SCRIPT_ID')) {
```
meisterT: nit:

```suggestion
    if (!defined('SCRIPT_ID')) {
```

PR #3469, opened 2026-02-21
Webapp phpstan all files

vmcj:
Builds upon: https://github.com/DOMjudge/domjudge/pull/3468

### DIFF
--- a/webapp/config/load_db_secrets.php
+++ b/webapp/config/load_db_secrets.php
@@ -24,8 +24,7 @@ function get_db_url(): string
     $dbsecretsfile = ETCDIR . '/dbpasswords.secret';
     $db_credentials = @file($dbsecretsfile);
     if (!$db_credentials) {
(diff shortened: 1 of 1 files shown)


webapp/config/load_db_secrets.php
```diff
@@ -28,13 +28,20 @@ function get_db_url(): string
...
         if ($line[0] == '#') {
             continue;
         }
+        $found = true;
         list($_, $host, $db, $user, $pass, $port) = array_pad(explode(':', trim($line)), 6, null);
         break;
     }
+    if (!$found) {
+        # Make sure that this fails with a clear error in Symfony.
+        return 'mysql://no_dbpasswords_found_in_file@localhost:3306/';
```
meisterT: I think we should change this (and the line 28 above) to something like `throw RuntimeException(...)` with a clear error message.

PR #3473, opened 2026-02-22
Remove superfluous </div> tags

vmcj:
We already close the \<div\> tags in the main template

### DIFF
--- a/webapp/templates/jury/submission.html.twig
+++ b/webapp/templates/jury/submission.html.twig
@@ -984,6 +984,19 @@
                                 <h5>Judge message for the team</h5>
                                 <pre class="output_text">{{  runsOutput[runIdx].team_message | convertUnprintableChars }}</pre>
                             {%  endif %}
+                            {% if run.description is not null %}
+                                <p><em>
+                                    {{ run.description(true) | descriptionExpand }}
+                                </em></p>
+                            {% endif %}
+                            {% if runsOutput[runIdx].image_thumb %}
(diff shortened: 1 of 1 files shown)


webapp/templates/jury/analysis/problem.html.twig
```diff
@@ -20,63 +20,66 @@
...
-            <dd class="col-sm-6">{{ misc.num_teams_attempted }}</dd>
-
-            <dt class="col-sm-6"># Teams Correct</dt>
-            <dd class="col-sm-6">{{ misc.num_teams_correct }} <span class="text-muted">({{  misc.teams_correct_percentage | number_format(2) }}%)</span></dd>
-          </dl>
-        </div>
+    <h1>Problem {{ problem.externalid }}: <span style="font-size: 60%;">{{ contest_problem | problemBadge }}</span> {{ problem.name }}</h1>
+    {% include 'jury/partials/analysis_filter.html.twig' %}
+  {# We close the div from the base template #}
+  </div>
```
meisterT, line 25: sorry, which div are we closing? Can you link to the line?

vmcj: webapp/templates/base.html.twig@L35

I'm not sure if we should actually close that div, but this is what I found by following the indentation and the divs which get opened again.

vmcj: So I checked, and the closing is with real intent as we want to start a new row. We should consider if we like this pattern or if we should do some duplication and that we create the row inside the subtemplate so we don't have to close things from the parent template making the templates much better to read.

PR #3474, opened 2026-02-25
Properly handle all arguments when falling back to plain HTTP.

meisterT:


misc-tools/dj_utils.py
```diff
@@ -91,7 +91,7 @@ def do_api_request(name: str, method: str = 'GET', jsonData: dict = {}, decode:
                 print('Can not verify certificate chain for DOMserver.')
                 exit(1)
             else:
-                return do_api_request(name)
+                return do_api_request(name, method, jsonData, decode)
```
vmcj, line 94: I don´t see where we would generate the url with `http` but why would we want to fallback to `http` instead of doing HTTPS with an untrusted certificate?

Fix should work though, so approving.

meisterT: Argh, the commit message is wrong, let me fix that - this is of course about untrusted https not http :facepalm:

meisterT: fixed

PR #3477, opened 2026-02-26
Add more tests for results/scores and their aggregation.

meisterT:


webapp/tests/Unit/Service/SubmissionServiceTest.php
```diff
@@ -539,6 +524,411 @@ public function testParseExpectedAnnotationWithRemap(): void
         self::assertNull($result['score']);
     }
 
+    /**
+     * Test getFinalResult with an empty run results array.
+     */
+    public function testGetFinalResultEmptyArray(): void
```
vmcj: The comment is just the name of the function in other wording so doesn´t add any new information. I don´t think we need it in this case.

meisterT: Done (also for all other tests where the comment was not really necessary).

webapp/tests/Unit/Service/SubmissionServiceTest.php
```diff
@@ -539,6 +524,384 @@ public function testParseExpectedAnnotationWithRemap(): void
...
+    }
+
+    /**
+     * Test getFinalResult picks first when multiple non-null
+     * results share the same highest priority level.
+     */
+    public function testGetFinalResultSamePriorityPicksFirst(): void
+    {
+        $prios = ['timelimit' => 99, 'wrong-answer' => 99, 'correct' => 1];
+        self::assertSame('timelimit', SubmissionService::getFinalResult(['timelimit', 'wrong-answer'], $prios));
```
vmcj, line 574:
```suggestion
        self::assertSame('timelimit', SubmissionService::getFinalResult(['timelimit', 'wrong-answer'], $prios));
        self::assertSame('timelimit', SubmissionService::getFinalResult(['wrong-answer', 'timelimit'], $prios));
```

That way the order of the `$prios` doesn´t matter.

webapp/tests/Unit/Service/SubmissionServiceTest.php
```diff
@@ -539,6 +524,384 @@ public function testParseExpectedAnnotationWithRemap(): void
...
+    public function testMaybeSetScoringResultOnRejectStopPendingAfterCorrect(): void
+    {
+        $group = $this->createTestcaseGroup('group1', TestcaseAggregationType::SUM, onRejectContinue: false);
+        $problem = new Problem();
+        $problem->setTimelimit(1)->setName('test');
+        $tc1 = $this->createTestcase($problem, $group, 1);
+        $tc2 = $this->createTestcase($problem, $group, 2);
+
+        $judging = $this->createJudging();
+        $this->addJudgingRun($judging, $tc1, 'correct', '10');
```
vmcj, line 776: I can´t find the test where we show that onRejectContinue=true we go further and one where the result is not correct to show we stop.

webapp/tests/Unit/Service/SubmissionServiceTest.php
```diff
@@ -539,6 +524,384 @@ public function testParseExpectedAnnotationWithRemap(): void
...
+
+    /**
+     * Test getFinalResult returns determinate result when max priority is
+     * reached even with pending (null) runs after it.
+     */
+    public function testGetFinalResultMaxPriorityBeforeNull(): void
+    {
+        $prios = ['correct' => 1, 'wrong-answer' => 99];
+        // wrong-answer has max priority (99), so null after it doesn't matter
+        self::assertSame('wrong-answer', SubmissionService::getFinalResult(['wrong-answer', null], $prios));
```
vmcj, line 547: Maybe have one with ['correct', null] and show we don´t have the result yet?

vmcj (approved):
I think the code is correct, suggested some additional tests.

PR #3479, opened 2026-02-28
Webstandard force no 404

vmcj:
~~Opening as draft first to see what should be changed.~~

Seems to work, doing this will help us detect other issues such as when the internal ids were mapped to externalids and not all URLs were updated.


vmcj:
Merging this now, @nickygerritsen and I discussed another improvement which I'll test a bit with.

PR #3480, opened 2026-02-28
Improve look of the judging verifier page.

meisterT:
Also fix some display bugs (e.g. displaying scoring only when relevant) while we are at it.

-----

Before:

image

After:

image


vmcj:
I think the easiest way to fix this is to add a badge with the total amount of submissions. And I think that information is actually helpful as the jury probably knows the total amount of submissions so can easily verify if 1 is missing from import.

PR #3481, opened 2026-02-28
Webstandard small fixes

vmcj:
Those were already part of another PR but as that one needs another investigation I prefer to get those commits merged already.

### DIFF
--- a/webapp/templates/jury/partials/submission_diff.html.twig
+++ b/webapp/templates/jury/partials/submission_diff.html.twig
@@ -10,7 +10,7 @@
                 {% set rank = null %}
             {% endif %}
             <li class="nav-item">
-                <a id="{{ diff_id }}-link" class="nav-link {{ extra_css_classes }}" data-bs-toggle="tab" data-rank="{{ rank }}" href="#{{ diff_id }}-tab" role="tab"><i class="fas fa-fw fa-file"></i> {{ name }}</a>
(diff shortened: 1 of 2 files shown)


webapp/templates/jury/export/clarifications.html.twig
```diff
@@ -124,3 +85,42 @@
         </table>
     {% endfor %}
 {% endblock %}
+
+{% block extra_style %}
```
meisterT, line 89: Is there a reason to keep this here and not move it to the CSS file?

webapp/templates/jury/partials/submission_diff.html.twig
```diff
@@ -1,7 +1,7 @@
 <div id="{{ editor_id }}-wrapper" class="flex-pass">
     {# Mark the first tab that is shown as active. #}
     {% set extra_css_classes = "active" %}
-    <ul class="nav nav-tabs source-tab-nav align-items-end">
+    <ul class="nav nav-tabs source-tab-nav align-items-end" role="tablist">
```
meisterT: yeah, I think this deserves a comment

webapp/templates/jury/export/clarifications.html.twig
vmcj: I think for the clarification export page we don´t include a CSS file.

meisterT: ah, ok

webapp/templates/jury/partials/submission_diff.html.twig
vmcj: Investigated and it can be fully removed which makes the comment redundant.

PR #3484, opened 2026-03-01
Handle internal-errors in compare.meta

vmcj:
Left https://github.com/DOMjudge/domjudge/blob/main/judge/judgedaemon.main.php#L2256 as is as it is only used in case we already disable the judgehost. We could get missing keys here though.

Build upon: https://github.com/DOMjudge/domjudge/pull/3483 as those do related but different things.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -343,7 +343,7 @@ class JudgeDaemon
         $this->loop();
     }
 
-    private function handleProgramInternalError(?int $judgetaskid, array $metadata): void
+    private function handleMetaInternalError(?int $judgetaskid, array $metadata): void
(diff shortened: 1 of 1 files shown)


judge/judgedaemon.main.php
```diff
@@ -2268,7 +2271,11 @@ private function runTestcase(
             $metadata = $this->readMetadata($passdir . '/program.meta');
             if (isset($metadata['internal-error'])) {
                 // This should already be handled in `testcaseRunInternal`
+<<<<<<< HEAD
                 $this->handleProgramInternalError($judgetaskid, $metadata);
+=======
```
meisterT: leftover merge markers

PR #3487, opened 2026-03-03
Rerun the test each time with the violating files removed

vmcj:
Before we ignored certain issues for all files, by now removing those files before rerunning the test we can detect the same violation in new files. That way we can prevent the same mistake from happening again.

### DIFF
--- a/.github/jobs/webstandard.sh
+++ b/.github/jobs/webstandard.sh
@@ -106,6 +106,26 @@ if [ "$TEST" = "none" ]; then
     exit $NUM_ERRORS
 fi
 
+w3c_analyse () {
+    FLTR="$1"
+    LOGID="$2"
+    URL="$3"
+    TYP="$4"
+    # shellcheck disable=SC2086
+    for typ in $TYP
+    do
+        section_start "Analyse with $typ"
+        # shellcheck disable=SC2086
+        "$DIR"/vnu-runtime-image/bin/vnu --errors-only --exit-zero-always --skip-non-$typ --format json $FLTR "$URL" 2> result.json
+        # shellcheck disable=SC2086
+        NEWFOUNDERRORS=$("$DIR"/vnu-runtime-image/bin/vnu --errors-only --exit-zero-always --skip-non-$typ --format gnu $FLTR "$URL" 2>&1 | tee "$ARTIFACTS/w3c_${typ}_${URL}_${LOGID}.log" | wc -l)
+        FOUNDERR=$((NEWFOUNDERRORS+FOUNDERR))
+        python3 -m "json.tool" < result.json > "$ARTIFACTS/w3c$typ$URL${LOGID}.json"
+        trace_off; python3 .github/jobs/jsontogha.py "$ARTIFACTS/w3c$typ$URL${LOGID}.json"; trace_on
+        section_end
+    done
+}
+
 if [ "$TEST" = "w3cval" ]; then
     section_start "Remove files from upstream with problems"
     rm -rf localhost/domjudge/doc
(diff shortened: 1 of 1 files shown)


.github/jobs/webstandard.sh
```diff
@@ -125,19 +145,35 @@ if [ "$TEST" = "w3cval" ]; then
...
+    rm -rf public/localhost/domjudge/jury/by-judging-id/*.html
+    w3c_analyse "$FLTR5" "5" "public" "html"
+    rm -rf public/localhost/domjudge/jury/categories.html
+    rm -rf public/localhost/domjudge/jury/problems.html
+    rm -rf public/localhost/domjudge/jury/users.html
+    rm -rf public/localhost/domjudge/jury/teams.html
+    rm -rf public/localhost/domjudge/jury/users/reset_login_status.html
+    rm -rf public/localhost/domjudge/jury/affiliations.html
+    w3c_analyse "$FLTR6" "6" "public" "html"
+    rm -rf public/localhost/domjudge/jury/contests/demo/submissions/*.html
```
meisterT: this is already deleted above

.github/jobs/webstandard.sh
```diff
@@ -125,19 +145,35 @@ if [ "$TEST" = "w3cval" ]; then
...
-        python3 -m "json.tool" < result.json > "$ARTIFACTS/w3c$typ$URL.json"
-        trace_off; python3 .github/jobs/jsontogha.py "$ARTIFACTS/w3c$typ$URL.json"; trace_on
-        section_end
-    done
+    FLTR4='--filterpattern .*autocomplete.*|.*role=tab.*|.*descendant.*|.*Stray.*'
+    FLTR5='--filterpattern .*autocomplete.*|.*role=tab.*|.*descendant.*'
+    FLTR6='--filterpattern .*autocomplete.*|.*role=tab.*'
+    FLTR7='--filterpattern .*autocomplete.*'
+    FLTR8=''
+    rm -rf public/localhost/domjudge/jury/teams/domjudge*.html
```
meisterT, line 148: consider making multiline `rm` statements?

.github/jobs/webstandard.sh
```diff
@@ -125,19 +145,35 @@ if [ "$TEST" = "w3cval" ]; then
...
+    rm -rf public/localhost/domjudge/jury/affiliations.html
+    w3c_analyse "$FLTR6" "6" "public" "html"
+    rm -rf public/localhost/domjudge/jury/contests/demo/submissions/*.html
+    rm -rf public/localhost/domjudge/jury/contests/demo/submissions/*/edit-source*.html
+    rm -rf public/localhost/domjudge/jury/contests/demo/submissions/*/source.html
+    w3c_analyse "$FLTR7" "7" "public" "html"
+    rm -rf public/localhost/domjudge/jury/contests/demo/submissions*.html
+    rm -rf public/localhost/domjudge/jury/submissions.html
+    rm -rf public/localhost/domjudge/jury/clarifications.html
+    w3c_analyse "$FLTR8" "8" "public" "html css svg"
```
meisterT: why only do css/svg here, can't we do it earlier?

.github/jobs/webstandard.sh
```diff
@@ -125,19 +145,35 @@ if [ "$TEST" = "w3cval" ]; then
...
-        # shellcheck disable=SC2086
-        "$DIR"/vnu-runtime-image/bin/vnu --errors-only --exit-zero-always --skip-non-$typ --format json $FLTR "$URL" 2> result.json
-        # shellcheck disable=SC2086
-        NEWFOUNDERRORS=$("$DIR"/vnu-runtime-image/bin/vnu --errors-only --exit-zero-always --skip-non-$typ --format gnu $FLTR "$URL" 2>&1 | tee "$ARTIFACTS/w3c_${typ}_${URL}.log" | wc -l)
-        FOUNDERR=$((NEWFOUNDERRORS+FOUNDERR))
-        python3 -m "json.tool" < result.json > "$ARTIFACTS/w3c$typ$URL.json"
-        trace_off; python3 .github/jobs/jsontogha.py "$ARTIFACTS/w3c$typ$URL.json"; trace_on
-        section_end
-    done
+    FLTR4='--filterpattern .*autocomplete.*|.*role=tab.*|.*descendant.*|.*Stray.*'
```
meisterT: Why do we start with `FLTR4`?

Also consider moving each definition down to the line where it is actually used

.github/jobs/webstandard.sh
```diff
@@ -125,19 +145,35 @@ if [ "$TEST" = "w3cval" ]; then
...
+    FLTR5='--filterpattern .*autocomplete.*|.*role=tab.*|.*descendant.*'
+    FLTR6='--filterpattern .*autocomplete.*|.*role=tab.*'
+    FLTR7='--filterpattern .*autocomplete.*'
+    FLTR8=''
+    rm -rf public/localhost/domjudge/jury/teams/domjudge*.html
+    rm -rf public/localhost/domjudge/jury/scoreboard.html
+    w3c_analyse "$FLTR4" "4" "public" "html"
```
meisterT: consider: instead of passing numbers here (e.g. `"4"`) pass some meaningful string (e.g. `"strict"`, ...)

.github/jobs/webstandard.sh
vmcj: Because FLTR0 up to FLTR3 was already handled when I did the testing. In other words, we had around 8 exclusions and the first 3 were apparantly handled but the filter was not updated.

vmcj: > Also consider moving each definition down to the line where it is actually used

This makes it a lot harder to keep those FLTRs in line in case we fix things which are not on the right side but in the middle of the FLTR line.

As I hope that this dirty list is only needed for a couple of weeks I think I prefer listing them like this.

.github/jobs/webstandard.sh
vmcj: Sure, but we have the most violations with html so wouldn´t we want to check the thing which would fail first?

But I can easily add something like this:
w3c_analyse "" "full_test" "public" "css svg"

meisterT: ah, you are not deleting CSS / SVG files, so that's actually fine I assume

PR #3495, opened 2026-03-06
Handle scores in scientific notation.

meisterT:
It's not great that we have duplicated code for this, but there is currently no way to share code between judgedaemon and the webapp.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -2327,6 +2327,17 @@ class JudgeDaemon
                     $this->disable('compare_script', 'compare_script_id', $judgeTask['compare_script_id'], $description, $judgeTask['judgetaskid']);
                     return false;
                 }
+                // Normalize scientific notation (e.g. "1e-3") to fixed-point
+                // decimal, since bccomp does not accept scientific notation.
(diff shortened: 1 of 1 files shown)


vmcj:
Can´t we include a lib/lib.*.php again? (I think we did in the past, currently checking the gitlog)

meisterT:
I also checked history and no, we didn't have it. It's not impossible, but more ugly than a few lines of duplicated code.

meisterT:
Ah, the duplication on the webapp side is completely unnecessary, so that's removed now

judge/judgedaemon.main.php
```diff
@@ -2327,6 +2303,16 @@ private function runTestcase(
                     $this->disable('compare_script', 'compare_script_id', $judgeTask['compare_script_id'], $description, $judgeTask['judgetaskid']);
                     return false;
                 }
+                // Normalize scientific notation (e.g. "1e-3") to fixed-point
+                // decimal, since bccomp does not accept scientific notation.
```
vmcj, line 2331:
```suggestion
                // decimal, since bccomp does not accept scientific notation.
                // See: https://github.com/php/php-src/issues/17876
```

PR #3497, opened 2026-03-06
Make sure `score.txt` written by a validator has precendence.

meisterT:
... if it exists and even if `accept_score` is set.

### DIFF
--- a/webapp/migrations/Version20260306120000.php
+++ b/webapp/migrations/Version20260306120000.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20260306120000 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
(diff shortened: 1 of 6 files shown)


webapp/migrations/Version20260306120000.php
```diff
@@ -0,0 +1,32 @@
...
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE judging_run CHANGE score score NUMERIC(32, 9) DEFAULT NULL COMMENT \'Optional score for this run, e.g. for partial scoring\'');
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('UPDATE judging_run SET score = 0 WHERE score IS NULL');
+        $this->addSql('ALTER TABLE judging_run CHANGE score score NUMERIC(32, 9) DEFAULT \'0.000000000\' NOT NULL COMMENT \'Optional score for this run, e.g. for partial scoring\'');
```
vmcj, line 25: I think we should start merging the migrations before the release.

meisterT: Yes, when we are done with all the scoring changes (there will likely be more), e.g. after NAC

PR #3500, opened 2026-03-06
More compact test case box styling for submission lists

meisterT:
Before:

image

After:

image

### DIFF
--- a/webapp/public/style_jury.css
+++ b/webapp/public/style_jury.css
@@ -249,7 +249,7 @@ table.submissions-table {
 /* Sticky header for submissions table */
 .submissions-table thead.thead-light {
     position: sticky;
-    top: 3.5rem; /* below the fixed navbar */
+    top: var(--navbar-height);
     z-index: 10;
 }
 .submissions-table thead.thead-light th {
--- a/webapp/public/style_domjudge.css
+++ b/webapp/public/style_domjudge.css
@@ -2,6 +2,7 @@
     /** Specify the BG color in one place. **/
     --background-color: white;
     --text-color: black;
+    --navbar-height: 3.5rem;
 }
 
 /** WCAG fixes start **/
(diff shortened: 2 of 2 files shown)


webapp/public/style_jury.css
```diff
@@ -241,6 +242,20 @@ div.submission-summary > span {
...
     width: auto;
 }
+.submissions-table tbody td {
+    vertical-align: middle;
+}
+
+/* Sticky header for submissions table */
+.submissions-table thead.thead-light {
+    position: sticky;
+    top: 3.5rem; /* below the fixed navbar */
```
vmcj: Consider making this a --css-constant,

webapp/src/Twig/TwigExtension.php
```diff
@@ -349,42 +349,51 @@ public function testcaseResults(Submission $submission, ?bool $showExternal = fa
...
+            return '';
+        }
+
+        $segments = '';
         $lastTypeSample = true;
+
         foreach ($testcases as $key => $testcase) {
             if ($testcase['sample'] != $lastTypeSample) {
-                $results        .= ' | ';
+                $segments       .= ' | ';
```
vmcj: To me this now looks on first glance like a light gray box, in the old situation it was more clear.  Can the bar be a little higher or something?

meisterT: would something like this be better?

image

webapp/public/style_jury.css
meisterT: Done

webapp/src/Twig/TwigExtension.php
vmcj: Yes, maybe a bit thicker but I already like the extra height.

webapp/public/style_jury.css
vmcj: I think we already have that specific top for other sticky elements so I would prefer a CSS variable here. In that case it would be on the scoreboard header and although that one is sticky, I can´t see where we do this same trick so I'm probably wrong.

webapp/src/Twig/TwigExtension.php
meisterT: Done.

image


If I make it even thicker it starts to look like a box and thus stand out less

webapp/public/style_jury.css
meisterT: Ah, understood now. Done as well.

PR #3504, opened 2026-03-07
Change the jury index page to a dashboard.

meisterT:
Previously, this was a wall of text. Finding things is now easier thanks to visual clues.

------

Before:

image

After:

image

### DIFF
--- a/webapp/public/style_jury.css
+++ b/webapp/public/style_jury.css
@@ -58,6 +58,33 @@ tr.summary td { border-top: 1px solid black; }
     width: 93px; height: 196px;
 }
 
+.dj-dashboard {
+    max-width: 1000px;
+    margin: 0 auto;
+}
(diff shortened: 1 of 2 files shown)


vmcj:
I'm not sure if I find this an improvement. Did you consider the old layout with the current icons if you really think those make finding links easier?

meisterT:
Yes, that's what I tried first, and it looks still like a wall of text and doesn't help much with finding things quickly.

meisterT:
Just adding a screenshot with what Michael suggested to show that this is still like a wall of text:

image


I understand it's personal preference, but I strongly prefer the dashboard look because it allows for easy visual scan.

vmcj (approved):
I don´t want to push this right now. If i don´t get used to it I can/will raise it again.

PR #3505, opened 2026-03-07
Ignore license generation in more modern yarn versions.

meisterT:
This is only supported in Yarn 1 natively, more modern versions require a plugin, which in turn doesn't work for Yarn 1.

Follow-up on 41a194e1735448bf373be124e663f9942d9abf26


meisterT:
You did not put it in gitignore

PR #3508, opened 2026-03-07
Add concurrency groups to our github actions.

meisterT:
These should make sure that for each PR there is at most one action of the same test running, i.e. on repeated pushs for the same PR we cancel all the old runs that are no longer relevant.

This should free up some small amount of capacity.

### DIFF
--- a/.github/workflows/autoconf-check-different-distro.yml
+++ b/.github/workflows/autoconf-check-different-distro.yml
@@ -6,6 +6,10 @@ on:
       - main
       - '[0-9]+.[0-9]+'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   redhat-family:
     strategy:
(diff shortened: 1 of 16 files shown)


.github/workflows/autoconf-check-different-distro.yml
```diff
@@ -6,6 +6,10 @@ on:
       - main
       - '[0-9]+.[0-9]+'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
vmcj, line 10: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context
ctr+f `github.ref`, I think this is always defined and would work for normal push, PR & mergequeue branch?

meisterT: I think so too

vmcj: So we can remove the `github.event.pull_request.number` part.

meisterT: Not necessarily, if you do create a PR each to two different branches (e.g. `main` and `nac2026` or `9.0`) from the same source branch, `github.ref` will be the same and cancel things on the other PR.

PR #3509, opened 2026-03-07
Cache more stuff in our GitHub actions

meisterT:
stacked on top of https://github.com/DOMjudge/domjudge/pull/3508


meisterT:
Example cache hit: 
image

PR #3510, opened 2026-03-07
Remove deprecated pattern from codespell action.

meisterT:


.github/workflows/codespell.yml
```diff
@@ -15,8 +15,8 @@ jobs:
         run: awk '1;/Version 7.2.1 - 6 May 2020/{exit}' ChangeLog > latest_Changelog
       - name: Get dirs to skip
         id: list_to_csv
-        run: echo "::set-output name=SKIP::$(awk '{print $1}' .github/jobs/data/codespellignorefiles.txt | paste -s -d, -)"
-      - uses: codespell-project/actions-codespell@master
+        run: echo "SKIP=$(awk '{print $1}' .github/jobs/data/codespellignorefiles.txt | paste -s -d, -)" >> $GITHUB_OUTPUT
+      - uses: codespell-project/actions-codespell@v2
```
vmcj: That action hasn´t broken (yet) ever, so I rather pick the latest version. Mostly as we don´t have a good way of detecting if there is a newer version. This actions is very low on false positives which also means that it if we're not on the latest version it will still have a lot of false negatives.

meisterT: Reverted the pinning

vmcj (approved):

PR #3513, opened 2026-03-08
Add paths specs to CI runs.

meisterT:
There are still certain checks that we run all the time (e.g. integration) and all checks run on the merge queue.

### DIFF
--- a/.github/workflows/autoconf-check-different-distro.yml
+++ b/.github/workflows/autoconf-check-different-distro.yml
@@ -5,6 +5,15 @@ on:
     branches:
       - main
       - '[0-9]+.[0-9]+'
+    paths:
+      - 'configure.ac'
+      - 'm4/**'
+      - 'Makefile*'
+      - '**/Makefile'
+      - 'paths.mk.in'
+      - 'bootstrap'
+      - '.github/workflows/autoconf-check*.yml'
+      - '.github/jobs/configure-checks/**'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -5,6 +5,14 @@ on:
     branches:
       - main
       - '[0-9]+.[0-9]+'
+    paths:
+      - 'webapp/**'
+      - 'phpstan.dist.neon'
+      - 'Makefile*'
+      - 'configure.ac'
+      - 'm4/**'
+      - '.github/workflows/phpstan.yml'
+      - '.github/jobs/**'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -5,6 +5,15 @@ on:
     branches:
       - main
       - '[0-9]+.[0-9]+'
+    paths:
+      - 'webapp/**'
+      - 'sql/**'
+      - 'etc/**'
+      - 'Makefile*'
+      - 'configure.ac'
+      - 'm4/**'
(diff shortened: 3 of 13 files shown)


nickygerritsen (approved):

nickygerritsen:
Does this work with required checks? I.e. if a check is skipped can we merge if it would still be required?

meisterT:
> Does this work with required checks? I.e. if a check is skipped can we merge if it would still be required?

I don't know - I guess we have to experiment?

.github/workflows/autoconf-check-different-distro.yml
```diff
@@ -5,6 +5,14 @@ on:
     branches:
       - main
       - '[0-9]+.[0-9]+'
+    paths:
+      - 'configure.ac'
+      - 'm4/**'
+      - 'Makefile*'
```
vmcj, line 11: I think we want this for any Makefile so something like:
- '**/Makefile' 
but not sure if the GHA syntax understands that?

meisterT: I think it does, I will research it tonight

.github/workflows/webstandard.yml
```diff
@@ -5,6 +5,15 @@ on:
     branches:
       - main
       - '[0-9]+.[0-9]+'
+    paths:
+      - 'webapp/**'
+      - 'sql/**'
+      - 'etc/**'
```
vmcj: I think we can accept to not trigger on this and if it breaks that we need to fix main.

.github/workflows/webstandard.yml
```diff
@@ -5,6 +5,15 @@ on:
...
+      - 'Makefile*'
+      - 'configure.ac'
+      - 'm4/**'
```
vmcj: Same here

.github/workflows/unit-tests.yml
```diff
@@ -5,6 +5,15 @@ on:
     branches:
       - main
       - '[0-9]+.[0-9]+'
+    paths:
```
vmcj, line 8: I think we should always run the unit-tests, they are not that expensive and catch 90% of the mistakes.

.github/workflows/phpstan.yml
```diff
@@ -5,6 +5,14 @@ on:
     branches:
       - main
       - '[0-9]+.[0-9]+'
+    paths:
+      - 'webapp/**'
+      - 'phpstan.dist.neon'
+      - 'Makefile*'
+      - 'configure.ac'
```
vmcj, line 12: We don't use the configure scripts here I think,

.github/workflows/unit-tests.yml
meisterT: Huh, the unit tests are the **most** expensive tests:
image

(That's because a lot of them are actually webapp end to end integration tests, not unit tests)

I wouldn't want to run them if I only touch the judge directory.

vmcj: Normally the integration test took that much longer, but fine with me.

meisterT: That was before I speed up judging :-D

vmcj: And creation of the chroot, I think we all optimized a bit.

.github/workflows/webstandard.yml
meisterT: done

.github/workflows/webstandard.yml
meisterT: done

.github/workflows/phpstan.yml
meisterT: we do use them

.github/workflows/autoconf-check-different-distro.yml
meisterT: it does, added

PR #3514, opened 2026-03-08
Fix rendering submission page

vmcj:
Best reviewed commit by commit and I'll squash afterwards.

Those are small findings to get the webstandard test in the right shape.

### DIFF
--- a/webapp/templates/jury/submission.html.twig
+++ b/webapp/templates/jury/submission.html.twig
@@ -852,28 +852,28 @@
                         </div>
                     </div>
                     <div class="card-body">
-                    {% if run.firstJudgingRun is not null %}
-                        {% if run.firstJudgingRun.runresult != 'correct' and isMultiPassProblem %}
-                            <div class="alert alert-warning">Note that this is a multi-pass problem, and only data of the failing pass is presented.</div>
-                        {% endif %}
-                        {{ runsOutput[runIdx].metadata | printMetadata }}
-                        <button class="btn btn-sm btn-outline-secondary" data-bs-toggle="collapse"
-                                data-bs-target="#collapseExample-{{ runIdx }}"
-                                aria-expanded="false">
-                            show complete metadata
-                        </button>
-                        {% if runsOutput[runIdx].metadata is not null %}
-                            {% if runsOutput[runIdx].output_limit is defined and runsOutput[runIdx].output_limit %}
-                                <div class="alert alert-warning">
-                                    The submission output (<code>{{ runsOutput[runIdx].output_limit }}</code>) was
-                                    truncated because of the configured output limit.
+                        {% if run.firstJudgingRun is not null %}
+                            {% if run.firstJudgingRun.runresult != 'correct' and isMultiPassProblem %}
+                                <div class="alert alert-warning">Note that this is a multi-pass problem, and only data of the failing pass is presented.</div>
+                            {% endif %}
+                            {{ runsOutput[runIdx].metadata | printMetadata }}
+                            <button class="btn btn-sm btn-outline-secondary" data-bs-toggle="collapse"
+                                    data-bs-target="#collapseExample-{{ runIdx }}"
+                                    aria-expanded="false">
+                                show complete metadata
+                            </button>
+                            {% if runsOutput[runIdx].metadata is not null %}
+                                {% if runsOutput[runIdx].output_limit is defined and runsOutput[runIdx].output_limit %}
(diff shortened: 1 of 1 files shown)


meisterT:
What was the problem with the rendering? Can you share screenshots?

vmcj:
> What was the problem with the rendering? Can you share screenshots?

All are HTML only, so screenshots don't help.

We would get things like `<td >` in case the twig filter would not run.

meisterT:
Sorry, I am slow here to follow. So it is not rendering related? Do we have invalid HTML or what are you fixing?

webapp/templates/jury/submission.html.twig
```diff
@@ -961,8 +954,9 @@
                             {% if runsOutput[runIdx].team_message is not empty %}
                                 <h5>Judge message for the team</h5>
                                 <pre class="output_text">
-{{  runsOutput[runIdx].team_message | convertUnprintableChars }}</pre>
-                            {%  endif %}
+                                    {{  runsOutput[runIdx].team_message | convertUnprintableChars }}
```
vmcj, line 957: @meisterT this one you mentioned in the other PR, I'll check it.

vmcj:
> Sorry, I am slow here to follow. So it is not rendering related? Do we have invalid HTML or what are you fixing?

Removing some unneeded comments in the rendered HTML (<!--)
Changing the attribute checked to checked=value as the right syntax
Fixing the rendering of <td {condition}> where sometimes condition would be nothing resulting in <td > which looks ugly but does work.
merging some code which checked the same thing twice
fixing the rendering of the page where we had stray ending tags (</td> for something which was already closed)

None of those change the layout in the browser, but this should make that we don´t depend on browsers fixing the malformed HTML in the same consistent way over engines.

webapp/templates/jury/export/clarifications.html.twig
```diff
@@ -1,5 +1,46 @@
 {% extends 'jury/export/layout.html.twig' %}
 
+{% block extra_head %}
+    <style>
```
meisterT, line 4: Where is this style coming from?

The commit message says "move style", but I only see an addition.

webapp/templates/jury/submission.html.twig
```diff
@@ -312,7 +312,7 @@
             </div>
         </div>
     </div>
-    <br />
+    <br />{# TODO: Consider making this a new class="row" #}
```
meisterT, line 315: I don't understand this comment.

webapp/templates/base.html.twig
```diff
@@ -32,6 +32,7 @@
 {% block body %}
     <div class="container-fluid{% if fill_height is defined and fill_height %} flex-pass{% endif %}">
         <div class="row{% if fill_height is defined and fill_height %} flex-pass{% endif %}">
+            {# Close this in the subtemplate for multiple rows #}
```
meisterT, line 35: Is this supposed to say something like:

`This div is closed in the subtemplate`

?

webapp/templates/jury/submission.html.twig
meisterT, line 1: some of the newlines that were removed here improve readability

webapp/templates/jury/submission.html.twig
meisterT: Thinking here again, we might be able to add `-`s to remove the whitespace (I didn't try though)

webapp/templates/jury/submission.html.twig
vmcj: Instead of doing a line break to steer the divs to be on a new row we can use bootstrap to place the objects in the way how we want them.

webapp/templates/jury/submission.html.twig
vmcj: I already fixed this in another way as `<pre>{{ twig }}</pre>` which doesn´t make the line too long and removes this whole problem for the future.

webapp/templates/base.html.twig
vmcj: No, we have some templates where we close this row and create new ones, and other templates where we don´t do this. I think that's the whole reason why we had that stray </div>.

webapp/templates/jury/submission.html.twig
vmcj: Probably personal preference but which ones do you think improve the readability? As for me the opening of a block and closing of a block don´t need extra newlines and I think those were the only ones I removed.

PR #3515, opened 2026-03-09
Use correct heading

vmcj:
I'll squash this into 1 commit after approval.

Inspected and the larger heading doesn't make the page worse, we could think about the modal as it is rather large but as this is the title of the modal it kinda fits.

This gets flagged by the latest VNU/Webstandard validator, we ourselves run with an older one so this is to prepare for the upgrade.

Difference in H3 -> H2
Screenshot_2026-03-09_19-32-21

Difference in H4 -> H3
Screenshot_2026-03-09_17-31-38

h6 -> div
Screenshot_2026-03-09_19-56-02

Removing the H1, similar to the `executables` page
Screenshot_2026-03-09_20-15-11

Screenshot_2026-03-09_20-39-15

Modal, h5 -> h2
Screenshot_2026-03-09_21-05-34

Screenshot_2026-03-09_21-41-36

Screenshot_2026-03-09_22-10-01

For all of the screenshots, this is the difference, not what I committed in the PR as there the item is not copied but replaced.

### DIFF
--- a/webapp/templates/jury/clarifications.html.twig
+++ b/webapp/templates/jury/clarifications.html.twig
@@ -57,7 +57,7 @@
         {% endif %}
 
         {% if currentFilter is null or currentFilter == 'new' %}
-            <h3 id="newrequests">New requests</h3>
+            <h2 id="newrequests">New requests</h2>
             {%- if newClarifications | length == 0 %}
                 <p class="nodata">No new clarification requests.</p>
             {%- else %}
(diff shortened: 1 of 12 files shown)


meisterT:
What's the goal of these changes?

vmcj:
> What's the goal of these changes?

The validator warns us for missing levels in the headings. https://html.spec.whatwg.org/multipage/sections.html#outline

I would like our pages to be correct as it makes it also easier for alternative screenreaders.

PR #3516, opened 2026-03-10
Webstandard last items

vmcj:
Locally this is the last item for the webstandard test for admins, build upon the other PRs so those need to be merged first.

Best done commit by commit to follow the indentation, afterwards I'll squash those into larger commits.

### DIFF
--- a/webapp/templates/jury/rejudging.html.twig
+++ b/webapp/templates/jury/rejudging.html.twig
@@ -151,17 +151,17 @@
     {% if stats %}
         <h2>Statistics over repeated rejudgings</h2>
         <h3>Runtime spread</h3>
-            <table class="table table-sm table-striped">
-                <tr><th>submission/testcase</th><th>spread</th><th>#judgings</th><th>result</th></tr>
-		{% for spread in stats.runtime_spread %}
-		    <tr>
-                        <td>s{{ spread.submitid }}/tc{{ spread.rank }}</td>
-                        <td>{{ spread.spread | number_format(3) }}s</td>
-                        <td>{{ spread.count }}</td>
-                        <td>{{ spread.verdict | printResult }}</td>
-		    </tr>
-		{% endfor %}
-	    </table>
+        <table class="table table-sm table-striped">
+            <tr><th>submission/testcase</th><th>spread</th><th>#judgings</th><th>result</th></tr>
+		        {% for spread in stats.runtime_spread %}
+		            <tr>
+                    <td>s{{ spread.submitid }}/tc{{ spread.rank }}</td>
+                    <td>{{ spread.spread | number_format(3) }}s</td>
+                    <td>{{ spread.count }}</td>
+                    <td>{{ spread.verdict | printResult }}</td>
+		            </tr>
+		        {% endfor %}
+	      </table>
         {% if stats.judging_runs_differ %}
             <h4>Judging Run Differences</h4>
             <ul>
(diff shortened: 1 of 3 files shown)


webapp/templates/jury/submission.html.twig
```diff
@@ -954,7 +954,8 @@
                             {% if runsOutput[runIdx].team_message is not empty %}
                                 <h5>Judge message for the team</h5>
                                 <pre class="output_text">
-{{  runsOutput[runIdx].team_message | convertUnprintableChars }}</pre>
+                                    {{  runsOutput[runIdx].team_message | convertUnprintableChars }}
```
meisterT: this will break rendering, won't it? As in, we display all the whitespace you added (we are in a `<pre>` block

vmcj: You might be right, I didn´t check that specific block. I'll check

vmcj: Fixed now by putting everything on 1 line.

webapp/templates/base.html.twig
```diff
@@ -32,6 +32,7 @@
 {% block body %}
     <div class="container-fluid{% if fill_height is defined and fill_height %} flex-pass{% endif %}">
         <div class="row{% if fill_height is defined and fill_height %} flex-pass{% endif %}">
+            {# We have a couple of subtemplates where this <div class=row> is closed within the template and a new one is opened #}
```
meisterT: `class=row`?

webapp/templates/jury/rejudging.html.twig
meisterT, line 1: same comment as on the other PR: I think some of these newlines increased readability.

Feel free to ignore this comment if you feel strongly.

webapp/templates/jury/submission.html.twig
```diff
@@ -953,8 +953,7 @@
 
                             {% if runsOutput[runIdx].team_message is not empty %}
                                 <h5>Judge message for the team</h5>
-                                <pre class="output_text">
-{{  runsOutput[runIdx].team_message | convertUnprintableChars }}</pre>
+                                <pre class="output_text">{{ runsOutput[runIdx].team_message | convertUnprintableChars }}</pre>
```
meisterT: If you want to try what I wrote on the other PR: use `-` to avoid the whitespace instead of putting all on one line

webapp/templates/base.html.twig
vmcj: I don´t understand your remark. What is unclear about the explanation?

webapp/templates/jury/submission.html.twig
vmcj: Implemented now.

webapp/templates/base.html.twig
meisterT: I am an idiot, I looked at the div that follows, not the one in the line before and was wondering what class you are talking about.

vmcj: I'll move the comment above it so it has the same indentation. Hopefully preventing it.

webapp/templates/jury/rejudging.html.twig
```diff
@@ -237,7 +237,7 @@
                     {%- for verdict, abbreviation in verdicts %}
                         <option value="{% if verdict == '' %}judging{% else %}{{ verdict }}{% endif %}" {% if oldverdict == verdict %}selected{% endif %}>
                             {% if verdict == '' %}
-                                <i>still judging</i>
```
meisterT, line 240: I wonder whether we should move the whole `if verdict == ''` out of the option completely. But that can be done in a separate PR

PR #3524, opened 2026-03-14
Stan all judges subset

vmcj:
Extraction of https://github.com/DOMjudge/domjudge/pull/3443 to already merge some parts.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -879,7 +879,10 @@ class JudgeDaemon
         }
     }
 
-    private function request(string $url, string $verb = 'GET', $data = '', bool $failonerror = true): string|bool|null
+    /**
+     * @param array<string, string>|string $data
+     */
+    private function request(string $url, string $verb = 'GET', string|array $data = '', bool $failonerror = true): string|bool|null
     {
         // Don't flood the log with requests for new judgings every few seconds.
         if (str_starts_with($url, 'judgehosts/fetch-work') && $verb === 'POST') {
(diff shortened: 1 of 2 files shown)


judge/judgedaemon.main.php
```diff
@@ -1055,8 +1058,14 @@ private function readJudgehostLog(int $numLines = 20): string
         return trim(ob_get_clean());
     }
 
-    private function runCommandSafe(array $command_parts, &$retval = DONT_CARE, $log_nonzero_exitcode = true, $stdin_source = null, $stdout_target = null, $stderr_target = null): bool
-    {
+    /**
+     * @param string[] $command_parts
+     * @param int|DONT_CARE $retval
```
meisterT, line 1063: is `DONT_CARE` an actual type?

judge/judgedaemon.main.php
```diff
@@ -2525,6 +2543,9 @@ private function reportJudgingRun(array $judgeTask, array $new_judging_run, bool
         return $response;
     }
 
+    /**
+     * @return array{input: string, output: string}|null
```
meisterT: phpcs error complaining about trailing space

judge/judgedaemon.main.php
meisterT, line 1: do we also need to update RunCommandSafeTest.php?

judge/judgedaemon.main.php
vmcj: it's the class from here: https://github.com/DOMjudge/domjudge/blob/main/judge/judgedaemon.main.php#L21

So I think PHPStan understands enough of it that we can receive it.

judge/judgedaemon.main.php
vmcj: Yes, but that was in another set of commits, I'm still struggling with the TimeLimit there.

vmcj: Actually this PR already touches that file, so what is this comment about?

meisterT: Probably fine then

judge/judgedaemon.main.php
vmcj: Fixed.

PR #3526, opened 2026-03-14
Add option to submit via online code editor/paste.

meisterT:
The filename of the uploaded file is derived from the problem letter and language.
If the language specifies an entry point, it is derived from the derived filename.

The syntax highlighting is done according to the selected language.

The last submit method (upload vs paste/online editor) is remembered and selected for future submits.

The online editor is a full screen page, the upload stays a modal.

----

Example screenshots:

image

image

### DIFF
--- a/etc/db-config.yaml
+++ b/etc/db-config.yaml
@@ -334,6 +334,15 @@
                 2: After first submission
             regex: /^\d+$/
             error_message: A value between 0 and 2 is required.
+        -   name: submit_methods
+            type: array_val
+            default_value:
+                - upload
+            public: true
+            description: Allowed submission methods in the web interface.
+            options:
+                - upload
+                - paste
         -   name: enable_ranking
             type: bool
             default_value: true
--- a/webapp/src/Controller/Team/SubmissionController.php
+++ b/webapp/src/Controller/Team/SubmissionController.php
@@ -3,11 +3,13 @@
 namespace App\Controller\Team;
 
 use App\Controller\BaseController;
+use App\Entity\Contest;
 use App\Entity\Judging;
 use App\Entity\Language;
 use App\Entity\Problem;
 use App\Entity\Submission;
 use App\Entity\SubmissionSource;
+use App\Entity\Team;
 use App\Entity\Testcase;
 use App\Form\Type\SubmitProblemType;
 use App\Service\ConfigurationService;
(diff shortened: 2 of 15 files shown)


etc/db-config.yaml
```diff
@@ -334,6 +334,15 @@
                 2: After first submission
             regex: /^\d+$/
             error_message: A value between 0 and 2 is required.
+        -   name: submit_methods
```
vmcj, line 337: Why would we want to disable this? I don´t see a good reason to add another thing to test for.

meisterT: For people who don't want it, it clutters the interface, but mostly I want to gain some experience with it first.

webapp/src/Controller/Team/SubmissionController.php
```diff
@@ -62,57 +65,152 @@ public function createAction(Request $request, ?Problem $problem = null): Respon
...
+            }
         }
+
+        return $this->redirectToRoute('team_index');
+    }
+
+    /**
+     * Build an UploadedFile from pasted code content.
+     *
+     * @return array{UploadedFile[], string}|null [files, tmpFilePath], or null if the code content is empty
```
vmcj, line 179:
```suggestion
     * @return array{0: UploadedFile[], 1: string}|null [files, tmpFilePath], or null if the code content is empty
```
I wonder if your version is valid, if so use yours.

webapp/src/Controller/Team/SubmissionController.php
```diff
@@ -62,57 +65,152 @@ public function createAction(Request $request, ?Problem $problem = null): Respon
...
+            $filename = $entryPoint . '.' . $extension;
+        } elseif ($contestProblem !== false) {
+            $filename = $contestProblem->getShortname() . '.' . $extension;
+        } else {
+            $filename = 'submission.' . $extension;
+        }
+
+        // Sanitize filename
+        if (!preg_match(SubmissionService::FILENAME_REGEX, $filename)) {
+            $filename = preg_replace('/[^a-zA-Z0-9+_.\-]/', '_', $filename);
```
vmcj: Can we remove that sanitizing to the SubmissionService?

meisterT: Done

etc/db-config.yaml
vmcj: Ok, I get the point, but don´t really like all this extra checks. I don´t see another way of doing it so resolving.

PR #3527, opened 2026-03-14
Improve jury/languages page.

meisterT:
- Add links to compiler/runner version overview, lang stats pages
- Add more direct toggles
- Remove time factor column if all langs use the time factor of 1 (default)
- Inline entry point field description
- Make extensions code font

-------

Before:
image

After:
image

### DIFF
--- a/webapp/src/Controller/Jury/LanguageController.php
+++ b/webapp/src/Controller/Jury/LanguageController.php
@@ -50,15 +50,35 @@ class LanguageController extends BaseController
             ->from(Language::class, 'lang')
             ->orderBy('lang.name', 'ASC')
             ->getQuery()->getResult();
+        $hasNonDefaultTimefactor = false;
+        $hasEntryPoint = false;
+        foreach ($languages as $lang) {
+            if ($lang->getTimeFactor() != 1) {
+                $hasNonDefaultTimefactor = true;
+            }
+            if ($lang->getRequireEntryPoint()) {
+                $hasEntryPoint = true;
+            }
+            if ($hasNonDefaultTimefactor && $hasEntryPoint) {
+                break;
+            }
+        }
+
         $table_fields = [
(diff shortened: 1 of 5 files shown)


webapp/templates/jury/languages.html.twig
```diff
@@ -11,7 +11,12 @@
 
 {% block content %}
 
-    <h1>Enabled languages</h1>
+    <div class="d-flex align-items-center mb-2">
+        <h1 class="mb-0 me-3">Enabled languages</h1>
+        {{ button(path('jury_versions'), 'Compiler/runner versions', 'outline-secondary', 'server') }}
+        &nbsp;
+        {{ button(path('analysis_languages'), 'Language analysis', 'outline-secondary', 'chart-bar') }}
```
vmcj: The buttons look out of place IMO, I like having them on the page, but not at the location they're now.

webapp/src/Controller/Jury/LanguageController.php
```diff
@@ -106,10 +126,23 @@ public function indexAction(): Response
...
+                    ? '(' . $lang->getEntryPointDescription() . ')'
+                    : null,
+            ];
+
             // Merge in the rest of the data.
             $langdata = array_merge($langdata, [
-                'entrypoint' => ['value' => $lang->getRequireEntryPoint() ? 'yes' : 'no'],
-                'extensions' => ['value' => implode(', ', $lang->getExtensions())],
+                'allowsubmit' => $allowSubmitOptions,
+                'entrypoint' => $entryPointOptions,
```
vmcj: This looks ugly, can it be in an extra column and you do the same check as you added for the timefactor?

```
set=false
for all languages:
  if has entry_point:
    set=true

```

`if set }<td>toggle</td><td>entry_point_val</td>{ endif }`

meisterT:
Addressed both comments, see code and new screenshot:

image

PR #3529, opened 2026-03-14
Stan all judges subset 3

vmcj:


judge/judgedaemon.main.php
```diff
@@ -2484,6 +2513,16 @@ private function runTestcase(
         return $ret;
     }
 
+    /**
+     * @param JudgeTask $judgeTask
+     * @param array{
```
meisterT: Not a fan of this one, but feel free to merge if you think it adds any value

vmcj: I can extract it to the top, that way we can have a bit of testing without it taking up vertical space.

meisterT: It's not about where it is. I think it does not have any meaning for humans, so I don't see why we would add it. As I said: if you think there is some value, feel free to keep it and merge as is

vmcj: I'll keep it.

The main argument as it might not add information for humans but it does for PHPStan to warn us if we forget to set certain keys or to warn us that a key is defined as optional so we should make sure we test for it.

PR #3530, opened 2026-03-14
Remove some minor issues with judgedaemon code

vmcj:


judge/judgedaemon.main.php
```diff
@@ -1019,7 +1019,7 @@ private function restEncodeFile(string $file, bool|int $sizelimit = true): strin
             $maxsize = (int)$this->djconfigGetValue('output_storage_limit');
         } elseif ($sizelimit === false || $sizelimit == -1) {
             $maxsize = -1;
-        } elseif (is_int($sizelimit) && $sizelimit > 0) {
```
meisterT, line 1042: I think previous PHP installations complained about this, didn't they?

vmcj: ~~in that case our integration test should fail on that.~~ It seems we don´t test with php8.2 anymore.

PHPStan just mentioned this and I have to agree that there are not other boolean values, so this has to be an int.

vmcj: I checked with PHP8.2 and the judgedaemon still runs so I think our PHP versions work.

PR #3532, opened 2026-03-14
Make webstandard errors on CI more useful.

meisterT:
This helps understanding the error without downloading the logs in many cases.

### DIFF
--- a/.github/jobs/webstandard.sh
+++ b/.github/jobs/webstandard.sh
@@ -116,16 +116,18 @@ w3c_analyse () {
         # shellcheck disable=SC2086
         "$DIR"/vnu-runtime-image/bin/vnu --errors-only --exit-zero-always --skip-non-$typ --format json $FLTR "$URL" 2> result.json
 
-        # Count errors from JSON
+        # Count errors from JSON and produce gnu-format log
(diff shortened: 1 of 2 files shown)


.github/jobs/webstandard.sh
```diff
@@ -118,6 +118,12 @@ w3c_analyse () {
         # shellcheck disable=SC2086
         NEWFOUNDERRORS=$("$DIR"/vnu-runtime-image/bin/vnu --errors-only --exit-zero-always --skip-non-$typ --format gnu $FLTR "$URL" 2>&1 | tee "$ARTIFACTS/w3c_${typ}_${URL}_${LOGID}.log" | wc -l)
         FOUNDERR=$((NEWFOUNDERRORS+FOUNDERR))
+        if [ "$NEWFOUNDERRORS" -gt 0 ]; then
+            echo "::error::$typ validation ($LOGID): found $NEWFOUNDERRORS error(s)"
```
vmcj, line 129: I think these sections are collapsed by default, does the `::error` open them again? Otherwise that should also be altered.

meisterT: Good question, I have added a commit on top of this to break it and see the breakage

meisterT: Now we do:

image

PR #3533, opened 2026-03-14
Add notifications for submissions.

meisterT:
Compared to #3018, this PR has the following relevant changes:
- no filter on j.seen = 0 which has a different meaning
- scalar query to avoid hydrating full entities
- include problem name in the notification to increase usefulness


meisterT:
cc @cubercsl

PR #3534, opened 2026-03-14
Exitcode for compare is always set as otherwise its an internal-error…

vmcj:
… handled above

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -2367,10 +2367,7 @@ class JudgeDaemon
                     logmsg(LOG_ERR, "Encountered unhandled 'internal-error':" . $comparemeta['internal-error']);
                 }
                 /** @var MetaData_Compare $compareMeta */
-                $compareExitCode = 'n/a';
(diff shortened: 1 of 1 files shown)


judge/judgedaemon.main.php
```diff
@@ -2367,10 +2367,7 @@ private function runTestcase(
                     logmsg(LOG_ERR, "Encountered unhandled 'internal-error':" . $comparemeta['internal-error']);
```
meisterT, line 2367: not sure whether you already fixed this in a different change, but this should be `compareMeta` (upper case `M`)

meisterT: Also, we want to return here probably

meisterT (approved):
approving this change since it is good, but we should do what I highlighted separately

judge/judgedaemon.main.php
vmcj: I agree with the typo,

but I think we return at L2384 || L2381 (depending on your view), I added that warning for consistency to check it everywhere but in case of a compare-error we had other code to handle this.

PR #3535, opened 2026-03-14
Remove PHP code coverage.

meisterT:
We have not been uploading new coverage to codecov.io in a **very** long time and producing code coverage slows our tests down quite a bit (so much that *unit* tests are the slowest on CI right now).


meisterT:
In case we prefer to keep coverage, we should create a separate job and run it once a day or week or so.

meisterT:
saves 3-5mins per run:

before:
image

after:
image

PR #3548, opened 2026-03-15
PHPStan issues found

vmcj:


judge/judgedaemon.main.php
meisterT, line 2479: is this actually guaranteed if runguard runs into an internal error?

judge/judgedaemon.main.php
```diff
@@ -2401,7 +2401,7 @@ private function runTestcase(
             }
 
             /** @var MetaData_Program $metadata */
-            if (isset($metadata['time-used']) && array_key_exists($metadata['time-used'], $metadata)) {
```
meisterT, line 2424: same question: is this really always set, also in case of internal error?

vmcj: Here at L2389

vmcj: At L2397/L2389

PR #3559, opened 2026-03-18
Remove judge-remaining button in case there is no contest

vmcj:
This is a quick fix, if we want to display the button but have it greyed out I need to do more work.


meisterT:
We previously did run judge remaining across all contests - probably we introduced the bug that broke this behavior. The question is whether we want to go back to this or keep it constrained to the current contest (in this case your PR is the right fix)?

meisterT (approved):
I think we should merge this and see if there are cases where we think otherwise later

PR #3561, opened 2026-03-18
Fix selection of active tab in source code view.

meisterT:
Imagine a situation where we first create one submission with files A.java and B.java, and then a submission with C.java.

Previously, the active tab would be A.java (as the alphabetically first one), but in no-diff mode we would not display A.java (nor B.java), so no tab is active (not even the single tab that exists C.java). The editor would be empty. Clicking on the tab would show the source code though.

Now, we explicitly set one of the visible tabs as active.

Manually tested and verified on some example submissions.


meisterT:
cc @Kevinjil

PR #3563, opened 2026-03-18
Apply some more rector

nickygerritsen:


webapp/src/Entity/TeamAffiliation.php
```diff
@@ -37,7 +37,8 @@ class TeamAffiliation extends BaseApiEntity implements
     HasExternalIdInterface,
     AssetEntityInterface,
     ExternalIdFromInternalIdInterface,
-    PrefixedExternalIdInterface
+    PrefixedExternalIdInterface,
+    \Stringable
```
meisterT: why `\`?

webapp/src/Entity/TeamAffiliation.php
```diff
@@ -304,6 +305,6 @@ public function getLogoForApi(): array
 
     public function __toString(): string
     {
-        return $this->getName() ?? $this->getShortname();
+        return $this->getName() ?? $this->getShortname() ?? '';
```
meisterT: Should we rather fall back to id then?

webapp/src/Entity/TeamAffiliation.php
nickygerritsen: Because we implement `__tostring()` and that is what this interface says.

vmcj: Can't we include Stringable to skip the ugly `\` notation?

PR #3566, opened 2026-03-19
First start to split the autoconf testsuite

vmcj:
Will squash into 1 commit after approval, left it like this to make it easier to follow that certain tests were slightly modified and no test should be lost.

If we want we could split out every test now by making another image which has the required/default tools installed, but as that adds little benefit in speed I left it like this.


.github/workflows/autoconf-check.yml
```diff
@@ -25,18 +25,46 @@ jobs:
...
+            $(pwd)/.github/jobs/configure-checks/run_bats.sh "apache"
+          echo "Start Nginx check"
+          docker run --name nginx       -v /tmp/bats_logs/nginx:/tmp/bats_logs \
+            -v "$(pwd):$(pwd)" -w "$(pwd)" baseimage \
+            $(pwd)/.github/jobs/configure-checks/run_bats.sh "nginx"
+          echo "Start LibCgroup check"
+          docker run --name cgroup      -v /tmp/bats_logs/cgroup:/tmp/bats_logs \
+            -v "$(pwd):$(pwd)" -w "$(pwd)" baseimage \
+            $(pwd)/.github/jobs/configure-checks/run_bats.sh "cgroup"
+          echo "Start all other checks"
```
meisterT: are there other checks we should run here?

.github/workflows/autoconf-check.yml
```diff
@@ -25,18 +25,46 @@ jobs:
...
+      - name: Run check in OS containers
+        run: |
+          mkdir -p /tmp/bats_logs/{gcc,gpp,clang,apache,nginx,cgroup,all}
+          docker run --name preinstall \
+            -v "$(pwd):$(pwd)" -w "$(pwd)" ${{ matrix.os}}:${{ matrix.version }} \
+            $(pwd)/.github/jobs/configure-checks/setup_configure_image.sh 
+          docker commit preinstall baseimage
+          echo "Start GCC check"
+          docker run --name gcc         -v /tmp/bats_logs/gcc:/tmp/bats_logs \
+            -v "$(pwd):$(pwd)" -w "$(pwd)" baseimage \
```
meisterT, line 46: for readability in the output if things fail I think I would prefer each of these run_bats.sh invocations being a separate run step

.github/workflows/autoconf-check.yml
```diff
@@ -25,18 +25,46 @@ jobs:
...
-      image: ${{ matrix.os }}:${{ matrix.version }}
     steps:
-      - name: Install git so we get the .github directory
-        run: apt-get update; apt-get install -y git
       - uses: actions/checkout@v4
-      - name: Setup image and run bats tests
-        run: .github/jobs/configure-checks/setup_configure_image.sh
+      - name: Run check in OS containers
+        run: |
+          mkdir -p /tmp/bats_logs/{gcc,gpp,clang,apache,nginx,cgroup,all}
```
meisterT: I don't see where we are using `all`? Is this a leftover?

.github/jobs/configure-checks/run_bats.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+testsuite="$1"
+export testsuite
+
+files=""
```
meisterT: did you intend to use `files` somewhere?

.github/workflows/autoconf-check.yml
```diff
@@ -25,18 +25,46 @@ jobs:
             version: stable
           - os: debian
             version: testing
+          - os: fedora
+            version: latest
     runs-on: ubuntu-latest
     env:
       DEBIAN_FRONTEND: noninteractive
```
meisterT, line 32: I assume setting `DEBIAN_FRONTEND` is not problematic on fedora, but do we have to set something similar for fedora?

.github/jobs/configure-checks/setup_configure_image.sh
```diff
@@ -23,4 +23,3 @@ cp submit/assert.bash .github/jobs/configure-checks/
 
 # Run the configure tests for this usecase
 mkdir /tmp/bats_logs
-test_path="/__w/domjudge/domjudge" bats --print-output-on-failure --gather-test-outputs-in /tmp/bats_logs .github/jobs/configure-checks/all.bats
```
meisterT, line 26: `test_path` changed - I assume it is fine, but wanted to verify it is intentional?

.github/jobs/configure-checks/functions.sh
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/bash
```
meisterT: consider using `/usr/bin/env bash` here?

.github/jobs/configure-checks/cgroup.bats
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bats
+
+load 'assert'
+
+source .github/jobs/configure-checks/functions.sh
+
+@test "cgroup library needed" {
+   cgroup_init_find="checking for cgroup_init in -lcgroup... no"
```
meisterT: nit: is the indentation here different or does this just look different in how github presents it?

.github/workflows/autoconf-check.yml
vmcj: yes

.github/workflows/autoconf-check.yml
vmcj: Not that I'm aware of, it's also copied from the other test so it seems we didn't need it before.

.github/jobs/configure-checks/cgroup.bats
vmcj: Indentation was wrong.

.github/jobs/configure-checks/run_bats.sh
vmcj: See the commits for better explanation but in short:
I tried running as `bats fileA fileB fileC` with the intent of running tests from `fileA` first but it seems `bats` does its own ordering which made that we ran tests in the unintended order. `files` was the string array I used in the script. It's removed now.

.github/workflows/autoconf-check.yml
vmcj: I claim I moved all of them to dedicated files now. So this is leftover from working on it.

.github/jobs/configure-checks/setup_configure_image.sh
vmcj: Before this ran from a container as the host. I think GHA does the clone in a non standard container. So it was not fully intentional but also not something I further investigated.

.github/workflows/autoconf-check.yml
```diff
@@ -25,20 +25,53 @@ jobs:
...
+          docker run --name apache      -v /tmp/bats_logs/apache:/tmp/bats_logs \
+            -v "$(pwd):$(pwd)" -w "$(pwd)" baseimage \
+            $(pwd)/.github/jobs/configure-checks/run_bats.sh "apache"
+      - name: Check Nginx checks
+        run: |
+          docker run --name nginx       -v /tmp/bats_logs/nginx:/tmp/bats_logs \
+            -v "$(pwd):$(pwd)" -w "$(pwd)" baseimage \
+            $(pwd)/.github/jobs/configure-checks/run_bats.sh "nginx"
+          echo "Start LibCgroup check"
+      - name: Check Nginx checks
```
meisterT: this name needs to be updated

.github/jobs/configure-checks/shared_post.bats
```diff
@@ -2,166 +2,14 @@
...
-   refute_line "$discourage_root"
-}
+source .github/jobs/configure-checks/functions.sh
 
 @test "Check for missing webserver group" {
     if [ "$distro_id" != "ID=fedora" ]; then
         # Debian/Ubuntu start with a www-data group
         skip
     fi
+    repo-install gcc g++
```
meisterT, line 12: when do we need to call `setup`? Is this missing here?

.github/jobs/configure-checks/functions.sh
```diff
@@ -0,0 +1,109 @@
...
+    fi
+    if [ "$testsuite" = clang ]; then
+       repo-install libcgroup-dev composer
+    else
+       repo-install gcc g++ libcgroup-dev composer
+    fi
+}
+
+run_configure () {
+    su $u -c "./configure $*"
```
meisterT, line 56: should this be `$@`?

vmcj: It's a string, so we don't need the separate arguments here. I think shellcheck specifically commented about this (verified by changing it).

.github/jobs/configure-checks/shared_post.bats
vmcj: `setup` would install most needed tools for a normal setup. For that test we're testing the state before where `setup` would bring us. So here we explicit don't want it as it installs a webserver.

meisterT (approved):

PR #3568, opened 2026-03-21
Cleanup installed libraries after `make dist`

vmcj:
It is always hard on how to name targets, for now I went with the logic that when your run:
`make dist` or `make maintainer-conf` and `make maintainer-clean`  afterwards you still endup with a clean checkout.

From naming you would expect that `make distclean` would cleanup after `make dist` but that doesn't seem to be the convention so the next logical target without creating a new one is `make maintainer-clean`.

### DIFF
--- a/webapp/Makefile
+++ b/webapp/Makefile
@@ -117,6 +117,7 @@ maintainer-clean-l:
 			rm -rf var/$$d/$$t ; \
 		done ; \
 	done
+	-rm -rf vendor node_modules LICENSES.txt
 
 .PHONY: composer-dump-autoload composer-dump-autoload-dev copy-bundle-assets \
         composer-dependencies composer-dependencies-dev node-dependencies


webapp/Makefile
```diff
@@ -90,6 +90,7 @@ maintainer-clean-l:
 			rm -rf var/$$d/$$t ; \
 		done ; \
 	done
+	-rm -rf vendor node_modules LICENSES.txt .env.local
```
meisterT: is removing `.env.local` not a bit too much?

eldering:
> From naming you would expect that `make distclean` would cleanup after `make dist` but that doesn't seem to be the convention so the next logical target without creating a new one is `make maintainer-clean`.

`make distclean` should bring you basically back to what is generated by `make dist` as that should (roughly) be the contents for a clean tarball. But for the conventions that we should try to adhere to, see:
https://www.gnu.org/software/make/manual/html_node/Standard-Targets.html
I don't think we should take these as hard rules though, but apply where it makes sense.

webapp/Makefile
vmcj: I think the intent normally for the `*-clean` is to endup in the state before you started so I would say we could do this, if after review we decide we shouldn't that is fine by me.

meisterT: we specifically say "Adjust as needed." when creating it - I have local modifications and would be sad if they were removed in this step

vmcj:
```suggestion
	-rm -rf vendor node_modules LICENSES.txt
```

vmcj: I took it out, the rest are code generated and are targets I would want to cleanup with `make maintainer-clean`.

vmcj:
> > From naming you would expect that `make distclean` would cleanup after `make dist` but that doesn't seem to be the convention so the next logical target without creating a new one is `make maintainer-clean`.
> 
> `make distclean` should bring you basically back to what is generated by `make dist` as that should (roughly) be the contents for a clean tarball. But for the conventions that we should try to adhere to, see: https://www.gnu.org/software/make/manual/html_node/Standard-Targets.html I don't think we should take these as hard rules though, but apply where it makes sense.

I checked and as this only touches `maintainer-clean`.

I checked the link, and we seem to do the correct thing in the `distclean` target.

PR #3571, opened 2026-03-21
Show scores in popup for scoring contests

### LINKED ISSUE #3570: Show scores in submissions popup on public scoreboard for scoring contests
For scoring contests, we should show the score per submission when you click a scoreboard cell on the public scoreboard.

nickygerritsen:
Fixes #3570

### DIFF
--- a/webapp/templates/public/partials/submission_list_template.html.twig
+++ b/webapp/templates/public/partials/submission_list_template.html.twig
@@ -3,6 +3,10 @@
     <div class="alert alert-warning">No submissions</div>
 </script>
 <script type="text/template" id="submission-list">
+    {% set is_scoring = false %}
+    {% if contest is defined and contest.scoreboardTypeString == 'score' %}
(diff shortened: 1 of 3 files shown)


webapp/templates/public/partials/submission_list_template.html.twig
```diff
@@ -3,13 +3,20 @@
     <div class="alert alert-warning">No submissions</div>
 </script>
 <script type="text/template" id="submission-list">
+    {% set is_scoring = false %}
+    {% if contest is defined and contest.scoreboardTypeString == 'score' %}
```
vmcj, line 7: Shouldn't contest always be there but might be null, and shouldn't we check for `contest.scoreboardTypeString is defined`?

nickygerritsen: Contest wasnt always there and a test failed on it so i needed to add this

PR #3577, opened 2026-03-21
Add API endpoints for scoreboard.zip and results.tsv.

meisterT:
That makes result export and comparison easier after EOC.


meisterT:
Added tests

PR #3578, opened 2026-03-21
Identify submissions based on submission source.

meisterT:


tuupke:
A test would be nice though.

PR #3584, opened 2026-03-22
Display program_err in multiple columns

vmcj:
Most of those lines are short so we can use the extra horizontal space.

image

For the rest of the output the user gets a horizontal scrollbar, I think that is easy enough to follow the program without taking up the whole page.

Just as a pitch, if we want this I can apply it to the other outputs also.

### DIFF
--- a/webapp/public/style_domjudge.css
+++ b/webapp/public/style_domjudge.css
@@ -173,6 +173,10 @@ a {
     white-space: pre;
 }
 
+.output_text_truncate_columns {
+    column-rule: 1px solid #c0c0c0;
+}
+
 .clarificationform pre {
     border-top: 1px dotted #c0c0c0;
     border-bottom: 1px dotted #c0c0c0;
--- a/webapp/src/Controller/Jury/SubmissionController.php
+++ b/webapp/src/Controller/Jury/SubmissionController.php
@@ -527,6 +527,13 @@ class SubmissionController extends BaseController
                 if ($firstJudgingRun) {
                     $runResult['testcasedir'] = $firstJudgingRun->getTestcaseDir();
                 }
+                foreach(['output_run', 'output_error'] as $field) {
+                    $colwidth = 0;
+                    foreach (explode("\n", $runResult[$field] ?? '') as $line) {
+                        $colwidth = max(strlen($line), $colwidth);
+                    }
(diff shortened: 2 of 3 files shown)


meisterT:
How does this look like if you have your window width is small (e.g. tiled to the left half of the screen)?

vmcj:
> How does this look like if you have your window width is small (e.g. tiled to the left half of the screen)?

Not worse than before. if we tweak it would work the same but CSS will render the "old view" when the window doesn't have enough width.

meisterT:
> Not worse than before.

So you don't have to horizontally scroll and it starts going to the old view automatically?

vmcj:
> > Not worse than before.
> 
> So you don't have to horizontally scroll and it starts going to the old view automatically?

if we wouldn't set the `height` it would yes. I now set the height as I would prefer the horizontal scroll in case I'm on my phone and want to scroll the page. It looks like you want the opposite of it.

meisterT:
For me there is no visual distinction between columns, e.g. this looks confusing:

image

----

We might want to consider not breaking it up into multiple cols if all/most lines have the same width for these maze like outputs.

Also, perhaps it should be a toggle? Not sure whether local or global.

----

One way to make this less confusing (more code though), is to do something like this:

image

vmcj:
> For me there is no visual distinction between columns, e.g. this looks confusing:

If you zoom in you see a line there, I can make it thicker though.


> We might want to consider not breaking it up into multiple cols if all/most lines have the same width for these maze like outputs.

I had it for a submission where I had a lot output of the same width so, my preference was to have it for specifically output with almost the same width. We could do some fancy detection but that also makes it harder.

> Also, perhaps it should be a toggle? Not sure whether local or global.

I think that requires knowing it, will either not be used a lot. So in that case I would prefer it to be on by default and that it can be turned off per problem for all users? As I think it's best to compress as much as possible and leave it up to the user how such mazes should be displayed.

> One way to make this less confusing (more code though), is to do something like this:
> image

Great for a follow up, but more work than I want to do for now. I do think this is better than we had before.

meisterT:
> If you zoom in you see a line there, I can make it thicker though.

I must be blind, because I am seeing nothing.

> I do think this is better than we had before.

As user, I would want a way to turn it off in the current version, because there is no (or perhaps minimal) distinction between columns.

PR #3592, opened 2026-03-23
Add required status checks to document current settings.

meisterT:
This can be downloaded via gh API for a specific commit in the past like this:

```
SHA="16320c4d9ec24e31b16994e78ea49d6f762d469c"

gh api repos/domjudge/domjudge/commits/$SHA/check-runs \
  --paginate \
  --jq '.check_runs[] | { context: .name, app_id: .app.id }' \
  | jq -s '{ strict: true, checks: (. | unique_by(.context)) }' \
  > .github/checks.json
```

And then you can update the required checks via:
```
gh api \
  -X PATCH \
  repos/domjudge/domjudge/branches/main/protection/required_status_checks \
  --input .github/checks.json
```

### DIFF
--- a/.github/checks.json
+++ b/.github/checks.json
@@ -1,0 +1,201 @@
+{
+  "strict": true,
+  "checks": [
+    {
+      "app_id": 15368,
+      "context": "Analyze (cpp)"
+    },
+    {
+      "app_id": 15368,
+      "context": "Analyze (java)"
+    },
+    {
+      "app_id": 15368,
+      "context": "Analyze (javascript)"
+    },
+    {
+      "app_id": 15368,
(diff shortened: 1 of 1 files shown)


meisterT:
> We do have less required checks than before but I assume you didn't apply them yet.

They are now applied, I was just testing whether the update did what I expected it to do.

vmcj:
> > We do have less required checks than before but I assume you didn't apply them yet.
> 
> They are now applied, I was just testing whether the update did what I expected it to do.

image

We now have a couple which will never start and are double.

meisterT:
Thanks, removed these. Let's see whether this works.

PR #3596, opened 2026-03-23
Refine Monaco editor resizing to fix jumping cursor issue.

meisterT:
Commit b5f2352df4 introduced a `ResizeObserver` that resets the Monaco editor's height to 0 on every body resize to force a flexbox layout recompute. However, this immediate reset was too aggressive and would trigger during minor layout "changes", in particular also clicking or focusing elements, causing Monaco to miscalculate click coordinates and jump the cursor to the top of the file.

This change improves the fix in two ways:
- Adds `overflow: hidden` to the `.editor` container. This is the standard CSS way to allow a flex item to shrink even when it contains a child with a fixed height (like Monaco), addressing the root cause more natively.
- Make sure the `ResizeObserver` only fires if the editor size actually changes.

### DIFF
--- a/webapp/public/js/domjudge.js
+++ b/webapp/public/js/domjudge.js
@@ -1635,11 +1635,21 @@ function initDiffEditorTab(editorId, diffId, submissionId, models) {
 // Force a recompute of the monaco editor height when the display size changes.
 if ('ResizeObserver' in window) {
     $(() => {
-        var monacoObserver = new ResizeObserver(() => {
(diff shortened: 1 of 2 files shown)


vmcj:
Shall we merge this and @Kevinjil  can take a look when he has time again?

nickygerritsen (approved):

meisterT:
@Kevinjil did a :+1: so I am going to merge

PR #3598, opened 2026-03-27
Add some example multipass problems

vmcj:
I needed a trivial multipass problem (spoon) & one with a couple of passes (hangman). Spoon is really trivial and hangman has no optimal strategy so is only there for demonstration purposes.

Before I cleanup the python3 code of the submissions I would like to know if we want to use those problems as otherwise it's a waste of time.


meisterT:
Is this ready for review yet, a lot of tests seem to fail?

vmcj:
> Is this ready for review yet, a lot of tests seem to fail?

Some fail unrelated,
for a couple I think we should first agree that we like the problem, and if so I can fix the submissions

So yes, needs review, and no not mergable yet

.github/jobs/detect_dump.sh
```diff
@@ -9,7 +9,7 @@ OUT=$(find ./ -name ".git*" -type d -prune -o \
               -name "cache" -type d -prune -o \
               -name "ace" -type d -prune -o \
               -path "./webapp/config/reference.php" -prune -o \
-              -type f -print0 | xargs -0 grep --color "dump(" | grep -v "Yaml::dump(") || true
+              -type f -print0 | xargs -0 grep --color "dump(" | grep -v "Yaml::dump(\|json.dump(") || true
```
meisterT: Out of curiosity: why did you need to change this?

doc/examples/hangman.tex
```diff
@@ -0,0 +1,58 @@
...
+\section*{Problem description}
+
+You are to solve the game `hangman`; it should guess the right letter
+and solve the word(s).
+
+\section*{Input specification}
+
+You are given two lines;
+\begin{itemize}
+  \item the first containing the number of the pass
```
meisterT: What is this number being used for / necessary?

doc/examples/hangman.tex
```diff
@@ -0,0 +1,58 @@
...
+\begin{center}
+  \Large\bf \textsc{DOM}judge sample problem\\
+  \texttt{pass} -- `Hangman'
+\end{center}
+~\\
+
+\section*{Problem description}
+
+You are to solve the game `hangman`; it should guess the right letter
+and solve the word(s).
```
meisterT, line 25: I think we should mention this being a multi-pass problem in here?

meisterT: And state an upper bound on the number of allowed passes?

doc/examples/hangman.tex
```diff
@@ -0,0 +1,58 @@
...
+
+\section*{Input specification}
+
+You are given two lines;
+\begin{itemize}
+  \item the first containing the number of the pass
+  \item the second a line containing `\_` \& ` `, where `\_` must be guessed by your program
+\end{itemize}
+you can guess any character within the lowercase range `a-z` by outputting that one
```
meisterT, line 34: I think we need to include the previous guesses somewhere? Otherwise you wouldn't know

doc/examples/spoon.tex
```diff
@@ -0,0 +1,61 @@
...
+you can´t really think of other objects to feed him with so you start to jumble
+the original word to something Kid does not know yet.
+
+Although the table is relatively large, it is given that your jumbling of the word 
+for object $A$ will never overlap with another object known as $B$.
+
+\section*{Sample input/output}
+
+You are given a single line with either $choose$ or $feed$ followed by the string with the word $W$.
+You should jumble in case of $choose$ and dejumble in the case of $feed$.
```
meisterT: Is it allowed to just print the same word again?

doc/examples/spoon.tex
```diff
@@ -0,0 +1,61 @@
...
+  \item a kid\footnote{As this is only an example this is not further explained} named \textbf{Kid}
+\end{enumerate}
+
+You need to feed Kid; but it started to associate words with the object
+and the terrible feeling of that object. Because you feed him 42 times a day
+you can´t really think of other objects to feed him with so you start to jumble
+the original word to something Kid does not know yet.
+
+Although the table is relatively large, it is given that your jumbling of the word 
+for object $A$ will never overlap with another object known as $B$.
```
meisterT: As for the other problem, it should mention that this is a multi-pass problem, and that you are being feed you own output as input

doc/examples/spoon.tex
```diff
@@ -0,0 +1,61 @@
...
+\end{center}
+~\\
+
+\section*{Problem description}
+
+You inherited:
+\begin{enumerate}
+  \item a large house
+  \item with a relatively large table
+  \item a kid\footnote{As this is only an example this is not further explained} named \textbf{Kid}
```
meisterT: I have no clue what this footnote is supposed to tell me

example_problems/spoon/output_validator/validator.py
```diff
@@ -0,0 +1,89 @@
...
+    with open(file, 'r') as f:
+      return strip_newline(f)
+  except PermissionError:
+    exit(PD+offset)
+  except FileNotFoundError:
+    exit(FE+offset)
+
+def check_writable(path: str) -> bool:
+  try:
+    if os.path.isdir(paht):
```
meisterT: that typo is really a typo worth fixing

example_problems/spoon/output_validator/validator.py
```diff
@@ -0,0 +1,89 @@
...
+inp = inp[0].split(' ')
+if len(inp) != 2:
+  exit(IE)
+
+if inp[0] == 'choose':
+  pattern = re.compile("[a-z]+")
+  if not pattern.fullmatch(inp[1]):
+    exit(RE)
+  if pattern.fullmatch(team_output):
+    #if inp[1] == team_output:
```
meisterT: this is what I would expect

.github/jobs/detect_dump.sh
vmcj: I had a script which parsed the words file on local disk, and dump them as an array to a python3 file. Not sure if I took that part out but I think it makes sense to always ignore the python3 `json.dump` function.

doc/examples/hangman.tex
vmcj: so the solution of enumerating all numbers and translating them to letter in the alphabet just works. We could give a list of the guessed letters but there seems to be no "nice" smart solution to guess and be correct

doc/examples/hangman.tex
vmcj: Both sound good.

doc/examples/spoon.tex
vmcj: No, as the `Kid` from the story knows that word. That's why palindromes are part of the testcase as the first solution I came up with was reversing.

doc/examples/spoon.tex
vmcj: It was part of the typical joke style how some of those statements are written. As the other example problems don't have it I'll take it out.

doc/examples/hangman.tex
meisterT: I just thought that it feels more natural to play it like you would play real hangman as a human. There you know the guessed letters and the already known letters in each turn

doc/examples/spoon.tex
meisterT: I think we should be explicit about it then

.github/jobs/detect_dump.sh
vmcj: I rewrote the script so we don't need this anymore.

example_problems/problems.yaml
```diff
@@ -13,3 +13,13 @@
...
+- id: jumble
+  label: D
+  ordinal: 4
+  rgb: '#541298'
+  time_limit: 5
+- id: hangman
+  label: E
+  ordinal: 5
+  rgb: '#6f0'
+  timelimit: 1
```
meisterT: I assume `time_limit` is the correct key (see above)

example_problems/hangman/generators/create_testcase.py
```diff
@@ -0,0 +1,72 @@
...
+    print("Wordfile not found: {wordfile}")
+    exit(2)
+
+random.shuffle(acceptable_words)
+
+try:
+    for i in range(requested_number_testcases):
+        write_input_and_answer(acceptable_words[i], i, output_directory)
+except IndexError:
+    print("Requested {requested_number_testcases} but only found {len(acceptable_words)}.")
```
meisterT: this is missing a `f` prefix, isn't it?

example_problems/hangman/generators/create_testcase.py
```diff
@@ -0,0 +1,72 @@
...
+        for line in wordfile.readlines():
+            line = line.strip()
+            # If we want the time_limit_exceeded solution to know all the words
+            # we need words of max length 7 otherwise the sourcecode gets to large
+            # if len(l) > 7:
+            #  continue
+            if pattern.fullmatch(line):
+                acceptable_words.append(line)
+except FileNotFoundError:
+    print("Wordfile not found: {wordfile}")
```
meisterT: same here (`f` prefix missing)

example_problems/hangman/data/secret/with_spaces/1.ans
```diff
@@ -0,0 +1 @@
+The quick brown fox jumps over the lazy dog
```
meisterT: Isn't `T` invalid since it is upper case?

example_problems/jumble/submissions/wrong_answer/caesar.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
```
meisterT: I think it would be nice to add a comment on top why the wrong answer solutions are expected to be incorrect

example_problems/hangman/output_validator/validator.py
```diff
@@ -0,0 +1,100 @@
...
+input_file = sys.argv[1]
+answer_file = sys.argv[2]
+feedback_dir = sys.argv[3]
+
+team_output = strip_newline(sys.stdin)
+
+AC = 42
+WA = 43
+
+PERMISSION_ERROR = 50
```
meisterT, line 57: We might want to leave a comment on top of these that explains that each of them will lead to an internal error and should never happen in proper setup

example_problems/hangman/output_validator/validator.py
```diff
@@ -0,0 +1,100 @@
...
+FILE_NOT_FOUND_ERROR = 51
+ANSWER_FILE_FORMAT_ERROR = 52
+SCRIPT_ERROR = 53
+
+if len(team_output) != 1:
+    exit(WA)
```
meisterT, line 67: A good validator would also include a message why the input was rejected, so if this example is there to set a role model, we should also do this here (and in all other cases where we reject the team's submission).

example_problems/hangman/output_validator/validator.py
```diff
@@ -0,0 +1,100 @@
...
+else:
+    team_output = team_output[0]
+if len(team_output) != 1:
+    exit(WA)
+if team_output[0] not in list(chr(ord('a') + x) for x in range(26))+[' ']:
```
meisterT: Why do we allow space here?

meisterT: Also more elegant would be:
```
if team_output[0] not in string.ascii_lowercase:
...
```

example_problems/hangman/output_validator/validator.py
```diff
@@ -0,0 +1,100 @@
...
+import _io
+import random
+import tempfile
+
+
+def strip_newline(raw: _io.TextIOWrapper) -> list[str]:
+    return [x.strip() for x in raw.readlines()]
+
+
+def check_readable(file: str, offset: int) -> list[str]:
```
meisterT, line 17: Add a docstring that explains what `offset` is used for.

example_problems/jumble/output_validator/validator.py
```diff
@@ -0,0 +1,91 @@
...
+if inp[0] == 'jumble':
+    pattern = re.compile("[a-z]+")
+    if not pattern.fullmatch(inp[1]):
+        exit(REGEX_ERROR)
+    pattern = re.compile("[a-z0-9]+")
+    if pattern.fullmatch(team_output):
+        if inp[1] in team_output:
+            exit(WA)
+        with open(os.path.join(feedback_dir, 'nextpass.in'), 'w') as g:
+            g.write(f"unjumble {team_output}")
```
meisterT: should we add a newline here?

example_problems/hangman/output_validator/validator.py
vmcj: > Why do we allow space here?

Because the problem statement allows for it. Otherwise the testcase with all letters doesn´t work. I thought it would make the problem more interesting but I can take it out.

meisterT: Doesn't the problem statement say "you can guess any character within the lowercase range ‘a-z‘ by outputting that one character."?

vmcj: Yes, this check should have checked the inputfiles as there this is allowed.

example_problems/jumble/output_validator/validator.py
```diff
@@ -0,0 +1,124 @@
...
+    if pattern.fullmatch(team_output):
+        if inp[1] in team_output:
+            with open(os.path.join(feedback_dir, 'judgemessage.txt'), 'a') as judgemessage:
+                print("The original input was found in the team output", file=judgemessage)
+            exit(WA)
+        with open(os.path.join(feedback_dir, 'nextpass.in'), 'w') as nextpass:
+            print(f"unjumble {team_output}", file=nextpass)
+        exit(AC)
+    with open(os.path.join(feedback_dir, 'judgemessage.txt'), 'a') as judgemessage:
+        print("Team output doesn´t match regex [a-z09]", file=judgemessage)
```
meisterT:
```suggestion
        print("Team output doesn´t match regex [a-z0-9]", file=judgemessage)
```

example_problems/hangman/output_validator/validator.py
```diff
@@ -0,0 +1,128 @@
...
+import string
+import sys
+import tempfile
+
+
+def strip_newline(raw: _io.TextIOWrapper) -> list[str]:
+    return [x.strip() for x in raw.readlines()]
+
+
+def check_readable(file: str, offset: int) -> list[str]:
```
meisterT, line 17:
```suggestion
def check_readable(file: str, offset: int) -> list[str] | None:
```
?

example_problems/jumble/output_validator/validator.py
```diff
@@ -0,0 +1,124 @@
...
+import os
+import _io
+import tempfile
+
+
+def strip_newline(raw: _io.TextIOWrapper) -> list[str]:
+    return [x.strip() for x in raw.readlines()]
+
+
+def check_readable(file: str, offset: int) -> list[str]:
```
meisterT, line 16:
```suggestion
def check_readable(file: str, offset: int) -> list[str] | None:
```
?

meisterT: But it might crash if it is `None` below?

example_problems/hangman/submissions/time_limit_exceeded/run_frequency/words.py
```diff
@@ -0,0 +1,1925 @@
+words = [
+    "domjudge", "abcdefghijklmnopqrstuvwxyz" "belauer", "gurgele",
```
meisterT:
```suggestion
    "domjudge", "abcdefghijklmnopqrstuvwxyz", "belauer", "gurgele",
```

example_problems/hangman/output_validator/validator.py
vmcj: Yes this is not documented but I think it's better to crash in case this is a directory. It was left from earlier development.

meisterT: Why not give a proper error instead of crashing?

vmcj: Because it would crash in case you feed a directory instead of a file, I would say for 2 invocations we do ourselves handling that error feels like overcomplification. 

But as we're having the conversation i assume you want this handled so I'll just handle it.

meisterT (approved):

Issue #3284, opened 2025-12-07
Creating a rejudging for a full problem/contest sometimes doesn't complete


meisterT:
When creating a larger rejudging for a full problem/contest it sometimes gets stuck at low percent (e.g. 4%) and doesn't make any progress. No errors on the JS console or symfony log.

PR #3602, opened 2026-03-29
Reset Doctrine's DebugDataHolder during rejudging creation to prevent…

meisterT:
… memory exhaustion in dev mode.

In dev mode, Doctrine's `DebugMiddleware` accumulates all SQL queries with parameters in its `DebugDataHolder`.

Data from a large rejudging creation (> 2k judgings with ~14k queries):
  - prod mode: 34 MB used (same peak, after a9d71d7e3)
  - dev mode without fix: 651 MB used (peak 666 MB)
  - dev mode with fix: 338 MB used (peak 348 MB)

Fixes remaining issues in https://github.com/DOMjudge/domjudge/issues/3284

### DIFF
--- a/webapp/src/Service/RejudgingService.php
+++ b/webapp/src/Service/RejudgingService.php
@@ -17,6 +17,7 @@ use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\NonUniqueResultException;
 use Doctrine\ORM\NoResultException;
 use Ramsey\Uuid\Uuid;
+use Doctrine\Bundle\DoctrineBundle\Middleware\DebugMiddleware;
 
 class RejudgingService
 {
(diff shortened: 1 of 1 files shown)


webapp/src/Service/RejudgingService.php
```diff
@@ -61,6 +62,17 @@ public function createRejudging(
         // This might take a while. Make sure we do not timeout.
         set_time_limit(0);
 
+        // In dev mode, Doctrine's DebugMiddleware accumulates all SQL queries in memory.
+        // Find the DebugDataHolder so we can periodically reset it during the loop.
+        $debugDataHolder = null;
+        foreach ($this->em->getConnection()->getConfiguration()->getMiddlewares() as $middleware) {
```
nickygerritsen, line 68: I thought there was also a way to remove the debug middleware, wouldn't that be better/easier?

nickygerritsen (approved):

webapp/src/Service/RejudgingService.php
meisterT: I tried this and it didn't work, perhaps I was holding it wrong though. Do you want to give it a try?

PR #3603, opened 2026-03-29
Reduce duplicate work on problem import

### LINKED ISSUE #2019: Import of problem zip slow with many submission files

meisterT:
Speeds up importing of a large-ish problem by ~9x.

Fixes #2019.

### DIFF
--- a/webapp/src/Service/ImportProblemService.php
+++ b/webapp/src/Service/ImportProblemService.php
@@ -725,6 +725,7 @@ readonly class ImportProblemService
                 Utils::jsonDecode($submission_file_string);
 
             $numJurySolutions = 0;
+            $deferredSubmissions = [];
             foreach ($zipEntries as $j => $path) {
                 if (!Utils::startsWith($path, 'submissions/')) {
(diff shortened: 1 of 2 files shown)


vmcj (approved):
If I understand correctly we only collect all of the work and do it in 1 query (per table) instead of a query per submission? We still do this in the same request but just on a different moment.

nickygerritsen (approved):

meisterT:
> If I understand correctly we only collect all of the work and do it in 1 query (per table) instead of a query per submission? We still do this in the same request but just on a different moment.

Yes, exactly.

PR #3607, opened 2026-04-04
Allow storing input/answer validator files

vmcj:
In the next step I want to run those on the judgehost to validate the different input/answer files as extra information for the config checker. This might be controversial but:
- I think there are a couple of jury members who use DOMjudge to do problem development adding this is a relatively simple thing todo and would help them.
- It is a good usecase for the separate chroot folders and to develop this further
- It prevents us from losing this information when we export a problem.

Before I spend more time on this I would like to know if we have any objections. Currently the input validator scripts themselves are probably broken but that's a feature for the future work I proposed. If we decide to merge this I'll take them out.

### DIFF
--- a/example_problems/fltcmp/input_validator/checktest_example_fltcmp.ctd
+++ b/example_problems/fltcmp/input_validator/checktest_example_fltcmp.ctd
@@ -1,0 +1,12 @@
+# This checktestdata script tests output of the "float compare"
+# example problem from DOMjudge. The expected output is a single
+# integer between 1 and 100 on the first line, followed by that many
+# lines, each containing a floating point number, which is possibly
+# "nan" (not a number) or "Inf" (infinity).
+#
+# For up-to-date version see: edu.nl/8fbar
+INT(1,100,n) NEWLINE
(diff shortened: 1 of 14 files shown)


meisterT:
I am against this. We have always been saying that problem preparation (and thus validation) should be done outside of DOMjudge, and that we do not want to replicate problemtools or bapctools functionality in an attempt to separate responsibilities. cc @eldering who might have an opinion on this as well

eldering:
> I am against this. We have always been saying that problem preparation (and thus validation) should be done outside of DOMjudge, and that we do not want to replicate problemtools or bapctools functionality in an attempt to separate responsibilities. cc @eldering who might have an opinion on this as well

I agree in general. I didn't look through this in detail, but for the third point: I don't think we should aim to be able to export a problem again without information loss. DOMjudge as a CCS is a consumer of problem packages, not a producer, and we'll never be able to (near) perfectly be able to export what we imported.

So my main question is: what does this offer that people can't already do with problem/BAPCtools outside of DOMjudge?

PR #3610, opened 2026-04-05
Add JetBrains Qodana test in CI

vmcj:


meisterT:
Do we actually gain anything by using this on top of all the linting that we already do?

vmcj:
> Do we actually gain anything by using this on top of all the linting that we already do?

It should give us the findings which PHPStorm also provides us, I think that has some good insights in performance. So it probably doesn´t add anything in your workflow but does for the people not using PHPStorm.

I'm still checking if we get some good insights though, setting it up is a bit difficult.

vmcj:
Our license ran out.

PR #3612, opened 2026-04-12
Untoggle the current umask to make sure the 0777 is actually set

vmcj:
Also added an explicit check as we silently ignored this issue.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -1980,8 +1980,18 @@ class JudgeDaemon
             if (CREATE_WRITABLE_TEMP_DIR) {
                 putenv("TMPDIR=$prefix/write_tmp");
                 if (!is_dir("$realWorkdir/write_tmp")) {
+                    $oldumask = umask(0);
                     if (!mkdir("$realWorkdir/write_tmp", 0777, true)) {
+                        umask($oldumask);
                         logmsg(LOG_WARNING, "Could not create '$realWorkdir/write_tmp'.");
+                        return Verdict::INTERNAL_ERROR;
+                    }
+                    umask($oldumask);
+                    $permissions = substr(sprintf('%o', fileperms("$realWorkdir/write_tmp")), -4);
+                    if ($permissions !== '0777') {
+                        $message = "Incorrect permissions of newly created folder '$realWorkdir/write_tmp', '$permissions' vs '0777'.";
+                        $this->disable('judgehost', 'hostname', $this->myhost, $message);
+                        logmsg(LOG_ERR, $message);
                         return Verdict::INTERNAL_ERROR;
                     }
                 }


judge/judgedaemon.main.php
```diff
@@ -1980,10 +1980,17 @@ private function testcaseRunInternal(
...
                 putenv("TMPDIR=$prefix/write_tmp");
                 if (!is_dir("$realWorkdir/write_tmp")) {
+                    $oldumask = umask(0);
                     if (!mkdir("$realWorkdir/write_tmp", 0777, true)) {
                         logmsg(LOG_WARNING, "Could not create '$realWorkdir/write_tmp'.");
                         return Verdict::INTERNAL_ERROR;
                     }
+                    umask($oldumask);
+                    $permissions = sprintf('%o', fileperms("$realWorkdir/foo") & 0777);
+                    if ('0777' !== $permissions) {
```
nickygerritsen:
```suggestion
                    if ($permissions !== '0777') {
```

I'm not a fan of the other way around personally.

judge/judgedaemon.main.php
```diff
@@ -1980,10 +1980,17 @@ private function testcaseRunInternal(
             if (CREATE_WRITABLE_TEMP_DIR) {
                 putenv("TMPDIR=$prefix/write_tmp");
                 if (!is_dir("$realWorkdir/write_tmp")) {
+                    $oldumask = umask(0);
                     if (!mkdir("$realWorkdir/write_tmp", 0777, true)) {
                         logmsg(LOG_WARNING, "Could not create '$realWorkdir/write_tmp'.");
                         return Verdict::INTERNAL_ERROR;
                     }
+                    umask($oldumask);
+                    $permissions = sprintf('%o', fileperms("$realWorkdir/foo") & 0777);
```
```suggestion
                    $permissions = sprintf('%o', fileperms("$realWorkdir/write_tmp") & 0777);
```

judge/judgedaemon.main.php
```diff
@@ -1980,10 +1980,17 @@ private function testcaseRunInternal(
             if (CREATE_WRITABLE_TEMP_DIR) {
                 putenv("TMPDIR=$prefix/write_tmp");
                 if (!is_dir("$realWorkdir/write_tmp")) {
+                    $oldumask = umask(0);
                     if (!mkdir("$realWorkdir/write_tmp", 0777, true)) {
                         logmsg(LOG_WARNING, "Could not create '$realWorkdir/write_tmp'.");
```
meisterT, line 1986: we also need to restore the old `umask` in this case

vmcj: The intent was to do this with line 1988

meisterT: Line numbers have changed, so I'm not 100% sure what line you are referring to. 

In any case, since we return inside of the if, we also have to do the umask there

meisterT (approved):

PR #3613, opened 2026-04-13
runguard: do not attempt to get username in error handling

xry111:
This error handling code is executed as the domjudge-run-X user in the chroot, but the /etc/passwd file in the chroot does not contain an entry for the user.  Thus username() always fail and the "failed to get username" message always cover the real error, making the debugging process very frusturating.

Drop the username() thing and simply print the user ID instead.


meisterT (approved):
Thanks!

PR #3615, opened 2026-04-15
Disable the judgehost in case we can´t create new directories for sub…

vmcj:
…missions


meisterT (approved):
I guess I need to go through all these internal errors and see where we want to disable the judgehost and where it is benign to continue. But that should not stop this PR from being merged.

PR #3617, opened 2026-04-20
Add Golang language

AVM-Martin:
Provide default run script for compiling golang code


vmcj:
Hi @AVM-Martin, thank you for the PR.

We don't use semantic commits so can you squash those 2 commits?

It would be nice if you can also provide some solutions to the example problems so users can check if the language works.

AVM-Martin:
Hi, I've provided the sample solutions in golang. I tested the solution on domjudge 9.0.0 setup on ubuntu 24.04

vmcj:
LGTM, thanks for the PR.

We will not backport this to 9.0.0 but people can download the language and upload it in older releases/installations.

vmcj (approved):

PR #3619, opened 2026-04-20
Better handle disk space errors during compilation.

### LINKED ISSUE #2454: Internal error when disk size is too small during compilation
If we run out of disk space during compilation (think about `test-memstatic.cc`, we create an internal error which does not have any useful information.

We should try to check whether we have enough space ahead of time and even if we miscalculate, we should have some useful info in the internal error.

meisterT:
- Add a warning to the config checker if allowed file size > minimum free disk space.
- Look for disk space errors when compiling and handle them more explicitly.

Fixes #2454

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -1663,7 +1663,19 @@ class JudgeDaemon
             $compile_output .= "\n--------------------------------------------------------------------------------\n\n" .
                 "Internal errors reported:\n" . $internalError;
 
-            if (str_starts_with($internalError, 'compile script: ')) {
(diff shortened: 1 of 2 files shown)


vmcj:
Should this be a warning or something which is never allowed when saving? In other words, should we have a validation and a more in-your-face warning on this?

If I understand correclty this can lead to internal errors, I would say it's better to prevent that as we have edge cases where internal errors break situations and I think itś better to prevent those.

meisterT:
Probably both, let's merge this and then follow up on the in-the-face warning.

PR #3620, opened 2026-04-23
Show host run-user inside chroot.

vmcj:
I needed this for another PR, and IMHO it helps in case we need to debug something.

The last one is only to document we could revert that one in case we make this a default option.

### DIFF
--- a/judge/runguard.cc
+++ b/judge/runguard.cc
@@ -689,6 +689,19 @@ int groupid(char *name)
 	return (int) grp->gr_gid;
 }
 
+char *username()
+{
+	int saved_errno = errno;
+	errno = 0; /* per the linux GETPWNAM(3) man-page */
+
+	struct passwd *pwd;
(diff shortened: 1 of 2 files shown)


judge/runguard.cc
```diff
@@ -689,6 +689,19 @@ int groupid(char *name)
...
 
+char *username()
+{
+	int saved_errno = errno;
+	errno = 0; /* per the linux GETPWNAM(3) man-page */
+
+	struct passwd *pwd;
+	pwd = getpwuid(getuid());
+
+	if ( pwd==nullptr || errno ) die(errno,"failed to get username");
```
meisterT: perhaps handle this gracefully?

vmcj: You want to continue when this fails?

meisterT: Yes, the information here is not crucial, so I would rather go on.

meisterT (approved):

PR #3622, opened 2026-05-02
Make ScoreboardMergeCommand work again

mpsijm:
I had been floating some fixes on my machine, time to push them back upstream :smile: 

The commits should speak for themselves, let me know if something is unclear or uses a incorrect assumptions :slightly_smiling_face:

### DIFF
--- a/webapp/src/Command/ScoreboardMergeCommand.php
+++ b/webapp/src/Command/ScoreboardMergeCommand.php
@@ -242,6 +242,7 @@ readonly class ScoreboardMergeCommand
                             ->setProblem($problemObj)
                             ->setColor($baseProblem['color'])
                             ->setShortName($label);
+                        $contest->addProblem($contestProblemObj);
                         $problems[$id] = $contestProblemObj;
                         $problemNameToIdMap[$name] = $id;
                         $firstSolve[$name] = null;
--- a/webapp/src/Command/ScoreboardMergeCommand.php
+++ b/webapp/src/Command/ScoreboardMergeCommand.php
@@ -248,7 +248,8 @@ readonly class ScoreboardMergeCommand
                     } else {
                         $id = $problemNameToIdMap[$name];
                     }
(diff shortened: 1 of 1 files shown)


webapp/public/style_domjudge.css
```diff
@@ -23,12 +22,12 @@ body {
     background-color: var(--background-color);
     font-family: Roboto, sans-serif;
     padding-bottom: 4em;
-    padding-top: var(--body-padding-top);
+    padding-top: calc(var(--navbar-height) + 1rem);
```
mpsijm: Hmm, apparently the webstandards check does not like this. The variable `--navbar-height` is always defined in `:root` (and optionally overridden in `body.static`), so I don't see why the check would fail. Rather, it looks like the check crashes? I'm not familiar with this checker, any clue how to circumvent this?

The checker says:
```
public/localhost/domjudge/style_domjudge.css%3Fv=10.0.0DEV.css:26.50: error: CSS: “padding-top”: Cannot invoke "org.w3c.css.values.CssValue.getRawType()" because "this.val1" is null.
```

vmcj: I think that is not implemented in the check, I hit the same with my PR so maybe extract this to another PR so we can investigate with the (possible) newer version of the validator.

vmcj:
Can you rework your commits to not have `ScoreboardMergeCommand:` and still make sense what they do? We don't use this style of commits in general.

webapp/public/style_domjudge.css
mpsijm: Done: #3624

mpsijm:
Rewrote the commit messages! I couldn't find a style guide (only for [code style](https://github.com/DOMjudge/domjudge/blob/main/CODINGSTYLE.md), not for commit messages), so feel free to rewrite them further if you want them to be more in-style, that's probably faster :stuck_out_tongue:

vmcj (approved):

PR #3623, opened 2026-05-04
[TestcaseGroup problems] Extend testcases UI page & exporting of problems

vmcj:
Code might not be great/performant but it fixes a couple of (real) issues, so I propose to be lenient and fix performance issues when they show up.

Will squash after approval & take the commit named testing out.

@meisterT I had to change the scores, I think it makes sense as the samples should be a testgroup but maybe you did this by design and I interpret the spec wrong?

image

### DIFF
--- a/webapp/src/Controller/Jury/ProblemController.php
+++ b/webapp/src/Controller/Jury/ProblemController.php
@@ -13,6 +13,7 @@ use App\Entity\ProblemAttachmentContent;
 use App\Entity\Submission;
 use App\Entity\SubmissionFile;
 use App\Entity\Testcase;
+use App\Entity\TestcaseAggregationType;
 use App\Entity\TestcaseContent;
 use App\Form\Type\ProblemAttachmentType;
 use App\Form\Type\ProblemType;
@@ -811,7 +812,8 @@ class ProblemController extends BaseController
                         ->setimage($content);
                 }
 
-                if ($newTestcaseGroup = $request->request->get('add_testgroup') and $newTestcaseGroup !== '') {
+                $newTestcaseGroup = $request->request->get('add_testgroup');
+                if ($newTestcaseGroup !== '') {
                     $testcaseGroup = $testcaseGroups[$newTestcaseGroup] ?? null;
                     if (!$testcaseGroup) {
                         $this->addFlash('danger', sprintf('Testcase group "%s" not found', $newTestcaseGroup));
@@ -1320,7 +1322,7 @@ class ProblemController extends BaseController
             if ($testcaseGroup->getAcceptScore()) {
                 $testdataConfig['accept_score'] = $testcaseGroup->getAcceptScore();
             }
-            if ($testcaseGroup->getAggregationType()) {
+            if ($testcaseGroup->getAggregationType() !== TestcaseAggregationType::SUM) {
                 $testdataConfig['grader_flags'] = $testcaseGroup->getAggregationType()->name;
             }
             if ($testcaseGroup->getOutputValidatorFlags()) {
--- a/webapp/src/Controller/Jury/ProblemController.php
+++ b/webapp/src/Controller/Jury/ProblemController.php
@@ -635,6 +635,13 @@ class ProblemController extends BaseController
                     $testcase->setSample($newSample);
                     $messages[] = sprintf('Set testcase %d to %sbe a sample testcase', $rank, $newSample ? '' : 'not ');
                 }
+                if ($testcase->getTestcaseGroup()) {
+                    if ($testcase->getSample() && !Utils::startsWith($testcase->getTestcaseGroup()->getName(), 'data/sample')) {
+                        $this->addFlash('warning', "Testcase #" . $rank . " marked as sample but not in sample testgroup.");
+                    } elseif (!$testcase->getSample() && Utils::startsWith($testcase->getTestcaseGroup()->getName(), 'data/sample')) {
+                        $this->addFlash('warning', "Testcase #" . $rank . " not marked as sample but in sample testgroup.");
+                    }
+                }
 
                 $newDescription = $request->request->all('description')[$rank];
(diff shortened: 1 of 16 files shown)


webapp/src/Controller/Jury/ProblemController.php
```diff
@@ -1241,6 +1317,25 @@ private function addTestcasesToZip(array $testcases, ZipArchive $zip, bool $isSa
...
+                $testdataConfig['accept_score'] = $testcaseGroup->getAcceptScore();
+            }
+            if ($testcaseGroup->getAggregationType() !== TestcaseAggregationType::SUM) {
+                $testdataConfig['grader_flags'] = $testcaseGroup->getAggregationType()->name;
+            }
+            if ($testcaseGroup->getOutputValidatorFlags()) {
+                $testdataConfig['output_validator_flags'] = $testcaseGroup->getOutputValidatorFlags();
+            }
+            if ($testcaseGroup->getRangeLowerBound()) {
+                $testdataConfig['range'] = sprintf("range: %d %d", $testcaseGroup->getRangeLowerBound(), $testcaseGroup->getRangeUpperBound());
```
meisterT, line 1332: does this need the extra `'range: `? Isn't that already given by the key? I think it currently serializes as `range: range: 0 100` (or similar)

webapp/src/Controller/Jury/ProblemController.php
```diff
@@ -791,6 +812,36 @@ public function testcasesAction(Request $request, string $probId): Response
...
+                            $others[] = $testcase;
+                        }
+                    }
+
+                    foreach(array_reverse($others) as $higherRankTestcase) {
+                        $higherRankTestcase->setRank($higherRankTestcase->getRank() + 1);
+                        // TODO: This is slow but prevents us from doing a transaction
+                        $this->em->flush();
+                    }
+                    $newTestcase->setRank($last->getRank() + 1);
```
meisterT, line 842: `$last` can be `null` here, or am I missing something?

webapp/src/Controller/Jury/ProblemController.php
```diff
@@ -1241,6 +1317,25 @@ private function addTestcasesToZip(array $testcases, ZipArchive $zip, bool $isSa
                                     $testcase->getContent()->getImage());
             }
         }
+        foreach ($testcaseGroups as $testcaseDirectory => $testcaseGroup) {
+            $testdataConfig = [];
+            if ($testcaseGroup->getAcceptScore()) {
+                $testdataConfig['accept_score'] = $testcaseGroup->getAcceptScore();
+            }
+            if ($testcaseGroup->getAggregationType() !== TestcaseAggregationType::SUM) {
+                $testdataConfig['grader_flags'] = $testcaseGroup->getAggregationType()->name;
```
meisterT, line 1326: should this be `->value`?

webapp/src/Controller/Jury/ProblemController.php
```diff
@@ -1241,6 +1317,25 @@ private function addTestcasesToZip(array $testcases, ZipArchive $zip, bool $isSa
...
+            if ($testcaseGroup->getOutputValidatorFlags()) {
+                $testdataConfig['output_validator_flags'] = $testcaseGroup->getOutputValidatorFlags();
+            }
+            if ($testcaseGroup->getRangeLowerBound()) {
+                $testdataConfig['range'] = sprintf("range: %d %d", $testcaseGroup->getRangeLowerBound(), $testcaseGroup->getRangeUpperBound());
+            }
+            if ($testcaseGroup->isOnRejectContinue()) {
+                $testdataConfig['on_reject'] = 'continue';
+            }
+            $zip->addFromString($testcaseDirectory . '/testdata.yaml',  Yaml::dump($testdataConfig));
```
meisterT, line 1337: Let's only add it if `$testdataConfig` is non-empty?

webapp/src/Controller/Jury/ProblemController.php
```diff
@@ -1216,14 +1275,31 @@ private function addTestcasesToZip(array $testcases, ZipArchive $zip, bool $isSa
             $prev = $testcase->getOrigInputFilename();
         }
 
-        $formatString = sprintf('data/%%s/%%0%dd', ceil(log10(count($testcases) + 1)));
+        $formatString = sprintf('%%s/%%0%dd', ceil(log10(count($testcases) + 1)));
         $rankInGroup = 0;
```
meisterT, line 1279: This was already the case before your change, but it looks like `$rankInGroup` is actually calculating a global rank not a per group rank?

webapp/templates/jury/problem_testcases.html.twig
```diff
@@ -201,6 +204,19 @@
             </div>
         </div>
 
+        {% if testcaseGroups != [] %}
+            <div class="row mb-2">
+                <label class="col-sm-2 col-form-label" for="add_testgroup">Testgroup</label>
+                <div class="col-sm-4">
+                    <select class="form-select" name="add_testgroup" id="add_testgroup">
+                        {% for testgroup in testcaseGroups %}
```
meisterT, line 212: Do we need to add some default for no group here as well?

webapp/src/Service/ImportProblemService.php
```diff
@@ -395,6 +395,15 @@ public function importZippedProblem(
                 }
             }
         }
+        if ($testCaseGroups !== []) {
+            foreach (['data/sample', 'data/secret'] as $default_dir) {
```
meisterT, line 399: I don't think this is correct, see https://icpc.io/problem-package-format/spec/2025-09.html#samples-and-full-feedback

meisterT (commented):
there are some phpcs nits as well

webapp/src/Service/ImportProblemService.php
vmcj: Ok, that was the missing spec piece I missed.

PR #3625, opened 2026-05-05
Output visualizer ng

vmcj:
I propose to take the example visualizer out as it's very hard to visualize the `hello` problem. Not sure if it's easier to review commit by commit or just everything in one go. It should be the case that this can be tested commit by commit as I needed to debug a couple of things.

### DIFF
--- a/webapp/src/Controller/Jury/ProblemController.php
+++ b/webapp/src/Controller/Jury/ProblemController.php
@@ -316,7 +316,7 @@ class ProblemController extends BaseController
             $yaml['validator_flags'] = $problem->getSpecialCompareArgs();
         }
         if (!empty($problem->getSpecialVisualizerArgs())) {
-            $yaml['visualizer_flags'] = $problem->getVisualizerCompareArgs();
+            $yaml['visualizer_flags'] = $problem->getOutputVisualizerCompareArgs();
         }
         if (!empty($problem->getMemlimit())) {
             $yaml['limits']['memory'] = (int)round($problem->getMemlimit() / 1024);
@@ -350,7 +350,7 @@ class ProblemController extends BaseController
         } elseif ($problem->isInteractiveProblem()) {
             $compareExecutable = $problem->getRunExecutable();
         }
-        $visualizerExecutable = $problem->getVisualizerExecutable();
+        $visualizerExecutable = $problem->getOutputVisualizerExecutable();
         foreach(['validators' => $compareExecutable, 'visualizers' => $visualizerExecutable] as $type => $executable) {
             if ($executable) {
                 foreach ($executable->getImmutableExecutable()->getFiles() as $file) {
--- a/webapp/src/Controller/Jury/ProblemController.php
+++ b/webapp/src/Controller/Jury/ProblemController.php
@@ -315,6 +315,9 @@ class ProblemController extends BaseController
         if (!empty($problem->getSpecialCompareArgs())) {
             $yaml['validator_flags'] = $problem->getSpecialCompareArgs();
         }
+        if (!empty($problem->getSpecialVisualizerArgs())) {
+            $yaml['visualizer_flags'] = $problem->getVisualizerCompareArgs();
+        }
         if (!empty($problem->getMemlimit())) {
             $yaml['limits']['memory'] = (int)round($problem->getMemlimit() / 1024);
         }
@@ -347,17 +350,20 @@ class ProblemController extends BaseController
         } elseif ($problem->isInteractiveProblem()) {
             $compareExecutable = $problem->getRunExecutable();
         }
-        if ($compareExecutable) {
-            foreach ($compareExecutable->getImmutableExecutable()->getFiles() as $file) {
-                $filename = sprintf('output_validators/%s/%s', $compareExecutable->getExecid(), $file->getFilename());
-                $zip->addFromString($filename, $file->getFileContent());
-                if ($file->isExecutable()) {
-                    // 100755 = regular file, executable
-                    $zip->setExternalAttributesName(
-                        $filename,
-                        ZipArchive::OPSYS_UNIX,
-                        octdec('100755') << 16
-                    );
+        $visualizerExecutable = $problem->getVisualizerExecutable();
+        foreach(['validators' => $compareExecutable, 'visualizers' => $visualizerExecutable] as $type => $executable) {
+            if ($executable) {
+                foreach ($executable->getImmutableExecutable()->getFiles() as $file) {
+                    $filename = sprintf('output_%s/%s/%s', $type, $executable->getExecid(), $file->getFilename());
+                    $zip->addFromString($filename, $file->getFileContent());
+                    if ($file->isExecutable()) {
+                        // 100755 = regular file, executable
+                        $zip->setExternalAttributesName(
+                            $filename,
+                            ZipArchive::OPSYS_UNIX,
+                            octdec('100755') << 16
+                        );
+                    }
                 }
             }
         }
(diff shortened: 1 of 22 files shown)


webapp/src/Controller/Jury/ProblemController.php
```diff
@@ -347,17 +350,20 @@ public function exportAction(string $problemId): StreamedResponse
...
-                $zip->addFromString($filename, $file->getFileContent());
-                if ($file->isExecutable()) {
-                    // 100755 = regular file, executable
-                    $zip->setExternalAttributesName(
-                        $filename,
-                        ZipArchive::OPSYS_UNIX,
-                        octdec('100755') << 16
-                    );
+        $visualizerExecutable = $problem->getOutputVisualizerExecutable();
+        foreach(['validators' => $compareExecutable, 'visualizers' => $visualizerExecutable] as $type => $executable) {
```
vmcj:
```suggestion
        foreach (['validators' => $compareExecutable, 'visualizers' => $visualizerExecutable] as $type => $executable) {
```

vmcj:
One thing I found after discussion with @mpsijm is that we should restrict that de output_visualizer should only be allowed to write to the feedback/judgeimage.* feedback/teammessage.* which requires us to either precreate the files or to allow writing everywhere and making sure it didn´t write files it was not allowed to write. Alternative is to let runguard do additional protection.

meisterT:
Before looking at the code in detail, I want to discuss some high level facts & decisions:

- We should somehow make sure that jury understands there is a higher quality bar for output visualizers that are going to be used in this mode. Wonder what's the best way to communicate this is.
- We only run the output visualizer after incorrect verdicts. I think it is good to not run it for all correct cases, but wonder if we should restrict further, e.g. only after WA? Is there a purpose of running the visualizer for TLE, RTE scenarios where the output is likely cut off and might send the visualizer in some bad code paths? Perhaps we should allow jury to select on which verdicts to run the output visualizer?
- The output visualizer similar to the validator is run outside of the chroot - which is something we should change in the long run. We probably want to start off with the visualizer in the chroot from the start?
- The visualizer is currently folded into the `judging_run` judgetask. I wonder whether we should split it off into its own judgetask? That would allow running visualizations at lower priority.
- We don't store the full test run outputs right now in the database for storage concerns. What should be the strategy be for these visualizations?
- Should only the jury see the output or also the teams? Probably jury only.
- Should we have a mode to request the visualization of a individual judging_run after the fact?

cc @eldering who might have some opinions as well

vmcj:
> Before looking at the code in detail, I want to discuss some high level facts & decisions:
> 
>     * We should somehow make sure that jury understands there is a higher quality bar for output visualizers that are going to be used in this mode. Wonder what's the best way to communicate this is.
> 

I don´t understand what you mean with this.

>     * We only run the output visualizer after incorrect verdicts. I think it is good to not run it for all correct cases, but wonder if we should restrict further, e.g. only after WA? Is there a purpose of running the visualizer for TLE, RTE scenarios where the output is likely cut off and might send the visualizer in some bad code paths? Perhaps we should allow jury to select on which verdicts to run the output visualizer?
> 

I know the head judge of the BAPC region wanted for the FPC2026 that the output_visualizer would also show the output for correct solutions. So I think having another property in the testdata.yaml would be the best place to show for which verdicts the jury wants the image.

I do like to merge this before we have that discussion though. As to me that feels like something for another PR and maybe even a proper in person discussion and some playing around.

>     * The output visualizer similar to the validator is run outside of the chroot - which is something we should change in the long run. We probably want to start off with the visualizer in the chroot from the start?
> 

I prefer to have it run in the same way as the output_validator and agree that we should to moving both into the chroot.

>     * The visualizer is currently folded into the `judging_run` judgetask. I wonder whether we should split it off into its own judgetask? That would allow running visualizations at lower priority.
> 

This made it soo much easier to do. And as the output_validator is allowed to just make the images itself we already have that this time can get lost. I think there is also the `teamimage` in the spec so there is an argument to be made that it helps the team.

So fine with moving it to another task but for a future PR. 

>     * We don't store the full test run outputs right now in the database for storage concerns. What should be the strategy be for these visualizations?
> 

We could store them on disk, we could limit the size and could make a fetch task to fetch them when a jury member wants them or we could add a cleanup command to fix those storage concerns.

>     * Should only the jury see the output or also the teams? Probably jury only.

I think the spec allowed for a `teamimage`, so I would say we display at the same places where we display `teammessage`.
> 
>     * Should we have a mode to request the visualization of a individual judging_run after the fact?
> 

In the future, yes for now I prefer to have it running often to catch errors.

> 
> cc @eldering who might have some opinions as well

Probably we should let him discuss with the problemspec people about this. Not sure if the BAPC/NWERC jury members will have opinions on this.

mpsijm:
> * We should somehow make sure that jury understands there is a higher quality bar for output visualizers that are going to be used in this mode. Wonder what's the best way to communicate this is.
> 
> I don´t understand what you mean with this.

I agree we need a way to communicate that the jury should write "proper" output visualizer programs, for some proper definition of "proper".
For example, slow visualizers will take up valuable judge time, but note that the output _validator_ can also write images (apart from the fact that the actual validation can be written inefficiently as well), so I think communicating the quality requirements holds for both `output_v{alidat,isualiz}er` programs.
Another example, generating oversized images can clog up the network when the jury inspects many images at once, and in the current state, it makes the submission page slow to load due to 100-ish test case visualizations being fetched at once.

Some ways to communicate this would be to write this in the spec, let tooling (like BAPCtools or Problemtools) run sanity checks, show warnings when importing problems with `output_*` programs, show warnings when validating/visualizing takes longer than running the submission itself (could be in the Judging Verifier), or show a permanent warning for problems with `output_*` programs. Any of these could be a one-sentence notice with a link to some documentation or the spec, where the best practices are explained.

> > * We only run the output visualizer after incorrect verdicts. I think it is good to not run it for all correct cases, but wonder if we should restrict further, e.g. only after WA? Is there a purpose of running the visualizer for TLE, RTE scenarios where the output is likely cut off and might send the visualizer in some bad code paths? Perhaps we should allow jury to select on which verdicts to run the output visualizer?
> 
> I know the head judge of the BAPC region wanted for the FPC2026 that the output_visualizer would also show the output for correct solutions.

That would be me :stuck_out_tongue: The context here is that the output is "creative": there are many possible solutions (hence the custom output validator), and we wanted to see in a quick glance if there were any teams that submitted a non-standard solution, and if they do, we can give them a shout-out at the solutions presentation :slightly_smiling_face: 

> We could store them on disk, we could limit the size and could make a fetch task to fetch them when a jury member wants them

Only fetching the image when a jury member explicitly requests them sounds like a good idea to combat the network clogging concerns. Specifically for "creative output" problems, we may also want an easy way to fetch all images for all submissions on a single test case, but we can also script that.

meisterT:
Addressing some high-level discussion points before looking at the code:

OK to address later:
- separate judge task: I think this is the right long term design but I am willing to keep it as a TODO for a separate PR (let's make sure to document the potential risks for the jury though)
- chroot: similarly, better in the long-run, can be done in a follow-up (e.g. when moving validators as well)
- verdicts to trigger visualization generation on: we should make this configurable, and I feel quite strongly that the default should be only to generate on WAs, but also OK to defer to a follow-up
- visibility: I think this should be jury only, especially a first version, but OK to defer if you feel strongly

Should be done before merging:
- size constraints: can we store a down-scaled version in the DB (similar to jury test case images) **and** store the full size pic on disk **and** add a size restriction for what's uploaded from the judgehost? At least the first two should be done before merging, the third preferably as well, but OK to defer. (Also for a later PR: we can load the images lazily if they come into viewport as we do for some other images on the scoreboard)
- robustness: let's ensure that a misbehaving visualizer (hang, crash, OOM) does not detrimentally affect judging in any way

PR #3630, opened 2026-05-17
Silence yarn in weekly coverity run

vmcj:
As improvement on https://github.com/DOMjudge/domjudge-scripts/pull/209 by doing this closer to the problematic invocation.

PR #3631, opened 2026-05-17
Backport some fixes to 9.0

vmcj:
When merged I think we should also push a new patch release.


vmcj:
Same as here: https://github.com/DOMjudge/domjudge/pull/3630#issuecomment-4470193818

Composer seems to not be available in debian `testing` at the moment.

meisterT (approved):

PR #3637, opened 2026-05-23
The judgedaemon only uses C++ code

vmcj:
This seems to work and I can´t find another `.c` file anymore.

If we still use `C` code somewhere the `inplace-{conf,install}` targets don´t properly error out on it.

CI will fail for unrelated errors & for some errors in debian testing where some parts have updated. I rather agree we want to ditch the `C` checks before fixing the tests.

### DIFF
--- a/.github/jobs/configure-checks/all.bats
+++ b/.github/jobs/configure-checks/all.bats
@@ -72,58 +72,35 @@ repo-remove () {
     fi
 }
 
-@test "Default empty configure" {
-    # cleanup from earlier runs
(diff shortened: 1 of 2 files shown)


nickygerritsen (approved):

meisterT:
Seems to break the tests - let me know if you need help fixing them.

vmcj:
> Seems to break the tests - let me know if you need help fixing them.

Half of them because `debian/testing` didn´t have `composer` for a while (https://tracker.debian.org/pkg/composer).

PR #3651, opened 2026-05-31
Fix linking to problem clarifications on public scoreboard

vmcj:


meisterT (commented):
index_content.html.twig line 11 uses still the old path

webapp/templates/partials/clarification_content.html.twig
```diff
@@ -0,0 +1,32 @@
+<div class="container clarificationform">
+    {% include 'team/partials/clarification.html.twig' with {clarification: clarification } %}
+    {% for reply in clarification.replies | filter(reply => team.canViewClarification(reply)) %}
```
meisterT, line 3: `team` is not always defined here, is it? (for public)

vmcj: Yes, I wondered the same and I think it just worked as `team = null` so result is `false`

meisterT (approved):

PR #3654, opened 2026-06-06
Fix testcases UI for empty testcase descriptions

vmcj:
We had a couple of issues:
- Remove \r\n as we would update the description with those This can be tested with having 1.desc, the description for testcase 1. We all use unix line endings, but the http spec wants \r\n endings. So when you save the testcases we would update testcase 1 but nothing (really) changed.
- When a new testcase was added we would also update the description of the other testcases with the empty string as it was `null` before. Alternative is to always fill the description with the empty string and make it non-nullable.

This contains some commits from https://github.com/DOMjudge/domjudge/pull/3623 and an additional fix.

### DIFF
--- a/webapp/src/Controller/Jury/ProblemController.php
+++ b/webapp/src/Controller/Jury/ProblemController.php
@@ -632,6 +632,10 @@ class ProblemController extends BaseController
                 }
 
                 $newDescription = $request->request->all('description')[$rank];
+                $newDescription = str_replace("\r\n", "\n", $newDescription);
+                if ($newDescription === '') {
+                    $newDescription = null;
+                }
                 if ($newDescription !== $testcase->getDescription(true)) {
                     $testcase->setDescription($newDescription);
                     $messages[] = sprintf('Updated description of testcase %d ', $rank);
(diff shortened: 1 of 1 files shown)


webapp/src/Controller/Jury/ProblemController.php
```diff
@@ -632,6 +632,11 @@ public function testcasesAction(Request $request, string $probId): Response
                 }
 
                 $newDescription = $request->request->all('description')[$rank];
+                if ($newDescription === '') {
+                    $newDescription = null;
+                } elseif (str_contains($newDescription, "\r\n")) {
```
meisterT: there is no need for `str_contains` here, just doing the replacement is fine

webapp/src/Controller/Jury/ProblemController.php
```diff
@@ -746,12 +751,16 @@ public function testcasesAction(Request $request, string $probId): Response
...
                 $newTestcase        = new Testcase();
                 $newTestcaseContent = new TestcaseContent();
+                $newTestcaseDescription = $request->request->get('add_desc');
                 $newTestcase
                     ->setContent($newTestcaseContent)
                     ->setRank($maxrank)
                     ->setProblem($problem)
-                    ->setDescription($request->request->get('add_desc'))
                     ->setSample($request->request->has('add_sample'));
+                if ($newTestcaseDescription !== '') {
```
meisterT: can `$newTestcaseDescription` be `null` here? if so, the below `str_replace` will fail

vmcj: It's always a string even when you provide no information in that field. But I changed the code to be explicit about this in case another browser would work differently.

webapp/src/Controller/Jury/ProblemController.php
vmcj: I also had to change the order around as now `$newDescription` can be null so the string replacement would fail.

meisterT (approved):

PR #3655, opened 2026-06-08
Add default test for output validator flags in problem.yaml & testdata.yaml + display

vmcj:
image

Extracted some parts of https://github.com/DOMjudge/domjudge/pull/3623 but also extended from there with the usage of the options in the testcase group.

This adds:
- the options to a default problem
- usage in the custom validator
- unit tests for usage
- display of those options in the testcases UI.

### DIFF
--- a/example_problems/hangman/output_validator/validator.py
+++ b/example_problems/hangman/output_validator/validator.py
@@ -3,11 +3,34 @@
 # <output_validator_program> input_file answer_file feedback_dir [additional_arguments] < team_output [ > team_input ]
 
 import _io
+import argparse
 import os
 import random
 import string
 import sys
 import tempfile
+
+
+def write_message(message: str, singular = False) -> None:
+    with open(os.path.join(feedback_dir, 'judgemessage.txt'), 'a') as judgemessage:
+        print("Team" + message, file=judgemessage)
+    if not arguments.secret:
+        teammessage_file = os.path.join(feedback_dir, 'teammessage.txt')
+        # Explicit overwrite the earlier hint
+        with open(teammessage_file, 'w') as teammessage:
+            print(("You" if singular else "Your") + message, file=teammessage)
+
+
+def write_generic_message(message: str, feedback_file: str) -> None:
+    message_file_path = os.path.join(feedback_dir, feedback_file)
+    try:
+        with open(message_file_path, 'r+') as message_file:
+            if message not in message_file.read():
+                print(message, file=message_file)
+    except FileNotFoundError:
+        # Overwrite in case the file got created in between
+        with open(message_file_path, 'w') as message_file:
+            print(message, file=message_file)
 
 
 def strip_newline(raw: _io.TextIOWrapper) -> list[str]:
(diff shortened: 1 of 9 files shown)


judge/judgedaemon.main.php
```diff
@@ -2119,7 +2119,7 @@ private function testcaseRunInternal(
                 // TODO: Perhaps we should change this in the database to be an array of args?
                 $orig_compare_args = [];
                 if ($compare_args !== null && strlen($compare_args) > 0) {
-                    $orig_compare_args = explode(' ', $compare_args);
+                    $orig_compare_args = str_getcsv($compare_args, separator: ' ', escape: '');
```
meisterT, line 2122: This might still not what we want, e.g. `--flag1  --flag2 ` (double spaces in between and trailing space) is parsed into `['--flag1', '', '--flag2', '']`, isn't it?

webapp/src/Service/ImportProblemService.php
```diff
@@ -1191,8 +1191,12 @@ public static function parseTestCaseGroupMeta(string $fileContent, string $name,
                     join(', ', $unknown_flags), $name);
             }
         }
-        if (isset($yamlData['output_validator_flags'])) {
-            $testcaseGroup->setOutputValidatorFlags($yamlData['output_validator_flags']);
+        foreach(['flags', 'args'] as $postFix) {
```
meisterT, line 1194: Should we check whether the user specifies both flags and treat is as an error if so? (same in the other commit, also consider linking to the spec in the corresponding commit messages)

webapp/templates/jury/problem_testcases.html.twig
```diff
@@ -51,6 +51,9 @@
                             {% if group.acceptScore is not null %}
                                 <span class="badge ms-2 scoring-accept-badge">Accept score: {{ (group.acceptScore + 0) }}</span>
                             {% endif %}
+                            {% if group.outputValidatorFlags %}
```
meisterT, line 54: Thanks, I wonder why I didn't include this in the first place!

example_problems/hangman/output_validator/validator.py
```diff
@@ -3,13 +3,36 @@
...
+        with open(teammessage_file, 'w') as teammessage:
+            print(("You" if singular else "Your") + message, file=teammessage)
+
+
+def write_generic_message(message: str, feedback_file: str) -> None:
+    message_file_path = os.path.join(feedback_dir, feedback_file)
+    try:
+        with open(message_file_path, 'r+') as message_file:
+            if message not in message_file.read():
+                print(message, file=teammessage_file)
```
meisterT: `teammessage_file` is not defined in here, only in the above function

judge/judgedaemon.main.php
meisterT: hmmm, https://icpc.io/problem-package-format/spec/2025-09.html#test-case-configuration says this is a "sequence of strings"

vmcj: I think we don´t implement 2025-09 but https://icpc.io/problem-package-format/spec/legacy.html#test-data-groups where this is a string but only if itś the name. So I suspect we need some extra work here.

meisterT: I need to figure what spec will be used for the next challenge...

example_problems/hangman/output_validator/validator.py
vmcj: Fixed by actually writing to the opened message_file.

judge/judgedaemon.main.php
vmcj: Can we merge this for now and open another PR to fix the `string` vs `map of string` vs `sequence of strings`

I'm fine with working on that, I just prefer to not have to further extend this PR with extra features.

webapp/src/Service/ImportProblemService.php
vmcj: I added a test which checks for this edgecase,

meisterT (approved):

Issue #15, opened 2013-09-03
Limitations: Number of problems, teams, test data files


eldering:
The CCS must not impose an upper limit on the number of different contest problems. 

The CCS must not impose an upper limit on the number of teams. 

The CCS must not impose an upper limit on the number of test data files for a problem, or their sizes.

PR #3656, opened 2026-06-11
Fix #3143: Restore task routing and judgehost penalty on compilation contradiction

### LINKED ISSUE #3143: Infinite Judging Loop When Compilation Results Differ Between Judgehosts
I recently migrated directly from DOMjudge 7.3.3 to version 9, and I'm uncertain whether this issue is a bug in the current codebase or potentially related to the migration process itself. I've identified what I believe to be the root cause and have drafted a proposed fix, but I wanted to report this first to confirm whether my understanding aligns with the expected system behavior. If the development team agrees with this analysis and the proposed approach, I would be happy to submit a formal pull request with the fix. However, I want to ensure that what I'm observing is indeed a bug and not an intended behavior or a migration-specific issue before proceeding.

When multiple judgehosts process the same judging and produce different compilation results (one succeeds, another fails), the system detects the inconsistency and creates an `InternalError` record. However, it fails to properly invalidate the pending judge tasks, causing the judging to remain in an indefinite processing state.
## Expected Behavior
When compilation result inconsistencies are detected:
1. An `InternalError` should be created
2. The problematic judgehost(s) should be disabled
3. All related judge tasks should be marked as invalid
4. The judging should be marked with the internal error
5. The system should allow for proper cleanup and potential rejudging

## Actual Behavior
When compilation result inconsistencies are detected:
1. An `InternalError` is created ✓
2. The problematic judgehost(s) are disabled ✓
3. **Judge tasks remain valid and active** ✗
4. **The judging is not marked with the internal error** ✗
5. The judging continues to be processed indefinitely, creating a loop

## Technical Details
The issue occurs in in the `updateJudgingAction` method, specifically in two scenarios: `JudgehostController.php`
### Scenario 1: Lines ~420-430
When compilation succeeds but a previous result indicated compiler error:
- The code creates an `InternalError` and disables other judgehosts
- **Missing**: Invalidation of judge tasks from those judgehosts
- **Missing**: Marking the judging with `setInternalError()`

### Scenario 2: Lines ~470-480
When compilation fails but a previous result was successful:
- The code creates an `InternalError` and disables the current judgehost
- **Missing**: Invalidation of judge tasks from the current judgehost
- **Missing**: Marking the judging with `setInternalError()`

## Comparison with Similar Code
In the same file, when a compilation error occurs normally (lines ~450-460), the code properly:
``` php
// Invalidate judgetasks
$this->em->getConnection()->executeStatement(
    'UPDATE judgetask SET valid=0 WHERE jobid=:jobid',
    ['jobid' => $judging->getJudgingid()]
);
```
This pattern is missing in the "compilation results are different" cases.
## Impact
- Judgings remain in perpetual "processing" state
- Judge tasks continue to be assigned to judgehosts
- Database accumulates invalid/conflicting records
- System resources are wasted on un-finishable judgings
- Manual database intervention may be required to fix affected judgings

## Proposed Fix
Add proper cleanup in both inconsistency detection scenarios:
1. Invalidate judge tasks associated with problematic judgehosts
2. Mark the judging with the internal error using `setInternalError()`
3. Ensure database flush to persist changes

Example for Scenario 1:
``` php
// After creating InternalError
$this->em->getConnection()->executeStatement(
    'UPDATE judgetask SET valid=0'
    . ' WHERE jobid=:jobid AND judgehostid IN'
    . ' (SELECT judgehostid FROM judgehost WHERE hostname=:hostname)',
    [
        'jobid' => $judging->getJudgingid(),
        'hostname' => $hostname,
    ]
);
$judging->setInternalError($error);
$this->em->flush();
```
## Steps to Reproduce
1. Set up multiple judgehosts with different configurations
2. Submit a solution that compiles successfully on one judgehost but fails on another (e.g., due to different compiler versions or system libraries)
3. Observe that an internal error is created
4. Check that judge tasks remain valid in the database
5. Observe that the judging continues to be processed indefinitely

## Additional Notes
This appears to be a systematic issue where error handling paths don't follow the same cleanup procedures as normal compilation error handling. The missing invalidation and error marking prevents the system's normal recovery mechanisms from functioning properly.

OrneryCash:
# Fixes
_It can be manually fix by shutting down the affected judgehost and rejudge the submission. But this can takes time under busy server load and creates unfair experience for participants_

This PR fixes two issues in `JudgehostController.php`:

+ **Infinite judging loop (#3143)**: Restores compilation contradiction logic missing since the parallel judging rewrite. It now correctly links the internal error (`$judging->setInternalError()`), disables the broken judgehost (`$judgehost->setEnabled(false)`), and unassigns its tasks (`SET judgehostid=NULL`). This stops error spam and allows healthy judgehosts to finish the submission.
~~**Broken log snippets**: Uses `is_resource()` and `stream_get_contents()` to properly read compiler error streams, fixing the `Resource id #15` display bug.~~
_Edit: Use_ `$judging->getOutputCompile(true)` _instead_

**Note on Task Routing Race Condition:**
The submission's final verdict currently depends on which judgehost picks up the *first* compile task:
- **Healthy first:** Compiles successfully. The broken judgehost is later caught and disabled, tasks are re-routed, and the submission succeeds.
- **Broken first:** Fails compilation immediately. The submission is permanently marked `Compiler Error` before healthy judgehosts can attempt it.

### Containers
Containers

_Note: My screenshot shows a locally built `domserver` container used for developing this fix. To reproduce the bug on your end, simply use the official `domjudge/domserver:bleeding` image._

Make the `g++` compiler on `judgehost-1` non-existent to replicate compiler error using:

`docker exec -it judgehost-1 mv /chroot/domjudge/usr/bin/g++ /chroot/domjudge/usr/bin/g++-disabled`

Generate a new problem that has a lot of testcases to trigger parallel judging using Python

```python
import zipfile
import random

def generate_problem_archive():
    filename = "a_plus_b.zip"
    print(f"Generating DOMjudge problem archive: {filename}...")

    with zipfile.ZipFile(filename, 'w', zipfile.ZIP_DEFLATED) as zf:
        # 1. Write the mandatory problem.yaml configuration
        yaml_content = """name: A+B
validation: default
"""
        zf.writestr("problem.yaml", yaml_content)

        # 2. Write the DOMjudge-specific ini
        ini_content = """probid=aplusb
name=A+B
timelimit=5
"""
        zf.writestr("domjudge-problem.ini", ini_content)

        # 3. Generate 100 distinct testcases in the required data/secret/ directory
        for i in range(1, 101):
            a = random.randint(1, 1000)
            b = random.randint(1, 1000)
            
            # Input file: Two integers
            zf.writestr(f"data/secret/{i}.in", f"{a} {b}\n")
            # Output file: Their sum
            zf.writestr(f"data/secret/{i}.ans", f"{a + b}\n")

    print(f"Done: {filename}")

if __name__ == "__main__":
    generate_problem_archive()
```

The answer for the following problem
```cpp
#include <iostream>
using namespace std;

int main() {
    int a, b;
    if (cin >> a >> b) {
        cout << a + b << "\n";
    }
    return 0;
}
```

### Reproduce Step
+ Upload `a_plus_b.zip` through Import archive with active contest
upload a problem

+ Stop the affected compiler container `(judgehost-1)`, so that `judgehost-0` can begin judging the testcases first. This prevents the server from throwing compilation error and stop the judging.
stop a container

+ Login with a `demo` account and submit the answer through `a_plus_b` problem

submit an answer

+ While `judgehost-0` is running testcases, start the `judgehost-1` container.

start judgehost-1

And you will see that internal error have appeared

internal error

However, submission is still judging infinitely instead of terminating or passing remaining testcases to other healthy judgehosts

infinite judging

Thank you to @alessiojr for reporting the issues and your analysis

### DIFF
--- a/webapp/src/Controller/API/JudgehostController.php
+++ b/webapp/src/Controller/API/JudgehostController.php
@@ -358,7 +358,7 @@ class JudgehostController extends AbstractFOSRestController
                 } elseif ($judging->getResult() === Judging::RESULT_COMPILER_ERROR) {
                     // The new result contradicts a former one, that's not good.
                     // Since the other judgehosts were not successful, but we were, assume that the other judgehosts
-                    // are broken and disable it.
+                    // are broken and disable them.
                     $disableHostnames = [];
                     /** @var JudgingRun $run */
                     foreach ($judging->getRuns() as $run) {
@@ -371,20 +371,13 @@ class JudgehostController extends AbstractFOSRestController
                     }
 
                     foreach ($disableHostnames as $disableHostname) {
-                        $disabled = [
-                            'kind' => 'judgehost',
-                            'hostname' => $disableHostname,
-                        ];
-                        $error = new InternalError();
-                        $error
-                            ->setJudging($judging)
-                            ->setContest($judging->getContest())
-                            ->setDescription('Compilation results are different for j' . $judging->getJudgingid())
-                            ->setJudgehostlog(base64_encode('New compilation output: ' . $output_compile))
-                            ->setTime(Utils::now())
-                            ->setDisabled($disabled);
-                        $this->em->persist($error);
+                        $this->disableJudgehostForContradiction(
+                            $judging,
+                            $disableHostname,
+                            (string)$judging->getOutputCompile(true)
+                        );
                     }
+                    $this->em->flush();
                 }
             } else {
                 $compileMetadata = $request->request->get('compile_metadata');
@@ -429,19 +422,13 @@ class JudgehostController extends AbstractFOSRestController
                         // The new result contradicts a former one, that's not good.
                         // Since at least one other judgehost was successful, but we were not, assume that the
                         // current judgehost is broken and disable it.
-                        $disabled = [
-                            'kind' => 'judgehost',
-                            'hostname' => $judgehost->getHostname(),
-                        ];
-                        $error = new InternalError();
-                        $error
-                            ->setJudging($judging)
-                            ->setContest($judging->getContest())
-                            ->setDescription('Compilation results are different for j' . $judging->getJudgingid())
-                            ->setJudgehostlog(base64_encode('New compilation output: ' . $output_compile))
-                            ->setTime(Utils::now())
-                            ->setDisabled($disabled);
-                        $this->em->persist($error);
+                        $this->disableJudgehostForContradiction(
+                            $judging,
+                            $judgehost->getHostname(),
+                            (string)$output_compile
+                        );
+
+                        $this->em->flush();
                     }
 
                     $judgingId = $judging->getJudgingid();
@@ -2042,5 +2029,36 @@ class JudgehostController extends AbstractFOSRestController
             return $this->serializeJudgeTasks($judgetasks, $judgehost);
         }
         return null;
+    }
+
+    private function disableJudgehostForContradiction(Judging $judging, string $hostname, string $outputCompile): void
+    {
+        $disabled = [
+            'kind' => 'judgehost',
+            'hostname' => $hostname,
+        ];
+
+        $error = new InternalError();
+        $error
+            ->setJudging($judging)
+            ->setContest($judging->getContest())
+            ->setDescription('Compilation results are different for j' . $judging->getJudgingid())
+            ->setJudgehostlog(base64_encode($outputCompile))
+            ->setTime(Utils::now())
+            ->setDisabled($disabled);
+        $this->em->persist($error);
+
+        // Link the error to the judging so the system knows
+        // this judging needs attention.
+        $judging->setInternalError($error);
+
+        $this->dj->setInternalError($disabled, $judging->getContest(), false);
+
+        // Re-route pending tasks from the disabled judgehost back to the queue
+        // so they are not stuck forever in an infinite processing loop.
+        $judgehost = $this->em->getRepository(Judgehost::class)->findOneBy(['hostname' => $hostname]);
+        if ($judgehost) {
+            $this->giveBackJudging($judging->getJudgingid(), $judgehost);
+        }
     }
 }


webapp/src/Controller/API/JudgehostController.php
```diff
@@ -1640,6 +1698,7 @@ private function getSourceFiles(string $contestId, string $id): array
         return $result;
     }
 
+
```
vmcj:
```suggestion
```

OrneryCash: Done

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -380,11 +381,45 @@ public function updateJudgingAction(
...
                         $this->em->persist($error);
+
+                        // Link the error to the judging so the system knows
+                        // this judging needs attention.
+                        $judging->setInternalError($error);
+
+                        $this->dj->setInternalError($disabled, $judging->getContest(), false);
+                        $otherJudgehost = $this->em->getRepository(Judgehost::class)->findOneBy(['hostname' => $disableHostname]);
+                        if ($otherJudgehost) {
+                            $otherJudgehost->setEnabled(false);
```
meisterT: this seems unnecessary as `setInternalError` takes care of it, doesn't it?

see https://github.com/DOMjudge/domjudge/blob/38df0b083254c41d1a84ddc6b82548c36d93a384/webapp/src/Service/DOMJudgeService.php#L633

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -438,10 +473,31 @@ public function updateJudgingAction(
...
                             ->setTime(Utils::now())
                             ->setDisabled($disabled);
                         $this->em->persist($error);
+
+                        // Link the error to the judging so the system knows
+                        // this judging needs attention.
+                        $judging->setInternalError($error);
+
+                        $this->dj->setInternalError($disabled, $judging->getContest(), false);
+                        $judgehost->setEnabled(false);
```
meisterT: same comment as above, see https://github.com/DOMjudge/domjudge/blob/38df0b083254c41d1a84ddc6b82548c36d93a384/webapp/src/Service/DOMJudgeService.php#L633

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -380,11 +381,45 @@ public function updateJudgingAction(
                             ->setJudging($judging)
                             ->setContest($judging->getContest())
                             ->setDescription('Compilation results are different for j' . $judging->getJudgingid())
-                            ->setJudgehostlog(base64_encode('New compilation output: ' . $output_compile))
+                            ->setJudgehostlog(base64_encode(
+                                is_resource($judging->getOutputCompile()) ? stream_get_contents($judging->getOutputCompile(), -1, 0) : (string)$judging->getOutputCompile()
```
meisterT: Why not just call `$judging->getOutputCompile(true)`?

webapp/src/Controller/API/JudgehostController.php
meisterT, line 1: I wonder whether we should create a helper method for the shared code, e.g. `disableJudgehostForContradiction`

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -358,8 +358,9 @@ public function updateJudgingAction(
                 } elseif ($judging->getResult() === Judging::RESULT_COMPILER_ERROR) {
                     // The new result contradicts a former one, that's not good.
                     // Since the other judgehosts were not successful, but we were, assume that the other judgehosts
-                    // are broken and disable it.
+                    // are broken and disable them.
                     $disableHostnames = [];
+                    $errors = [];
```
meisterT: This seems unused - intentional?

meisterT (changes requested):

webapp/src/Controller/API/JudgehostController.php
OrneryCash: You're right. I have removed that line.

webapp/src/Controller/API/JudgehostController.php
OrneryCash: Done

webapp/src/Controller/API/JudgehostController.php
OrneryCash: Forgot to delete it. Done

webapp/src/Controller/API/JudgehostController.php
OrneryCash: I have created the helper function as you named it.

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -2043,4 +2030,47 @@ private function getJudgetasks(string|int|null $jobId, int $max_batchsize, Judge
...
     }
+
+    /**
+     * Disable a judgehost and unassign its tasks when a compilation contradiction occurs.
+     *
+     * @param Judging $judging
+     * @param string  $hostname
+     * @param string  $outputCompile
+     * @return void
+     * @throws \Doctrine\DBAL\Exception
```
vmcj:
```suggestion
```

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -2043,4 +2030,47 @@ private function getJudgetasks(string|int|null $jobId, int $max_batchsize, Judge
         }
         return null;
     }
+
+    /**
+     * Disable a judgehost and unassign its tasks when a compilation contradiction occurs.
```
vmcj: I would say the name of the function already hints at actually this usage, consider removing this.

OrneryCash: Done

vmcj:
@OrneryCash can you properly rebase and squash the commits so we can merge after approval?

webapp/src/Controller/API/JudgehostController.php
```diff
@@ -2043,4 +2030,38 @@ private function getJudgetasks(string|int|null $jobId, int $max_batchsize, Judge
...
+        // Link the error to the judging so the system knows
+        // this judging needs attention.
+        $judging->setInternalError($error);
+
+        $this->dj->setInternalError($disabled, $judging->getContest(), false);
+
+        // Re-route pending tasks from the disabled judgehost back to the queue
+        // so they are not stuck forever in an infinite processing loop.
+        $this->em->getConnection()->executeStatement(
+            'UPDATE judgetask SET judgehostid=NULL, starttime=NULL'
```
meisterT: This resets every valid judgetask the host ever owned, including already-finished runs and tasks from unrelated judgings, not just in-flight ones. `giveBackJudging()` deliberately skips runs where `getRunresult() !== null`.

Can we scope this to unfinished work (e.g. `starttime IS NOT NULL` and no run result), or reuse `giveBackJudging()`?

As written it wipes judgehost attribution on completed tasks and sweeps wider than the "unassign active tasks" intent.

meisterT: I just realize this is not even filtering by `type` currently

OrneryCash: Good catch! I missed that completely. I'll reuse the `giveBackJudging()`

meisterT:
Sorry for the delay, code looks good now. I have unblocked running CI, so let's hope it passes.

OrneryCash:
> Sorry for the delay

No worries! Should i update the branch right now?

meisterT (approved):

meisterT:
Not necessary, can be merged as is. Thanks!

PR #3657, opened 2026-06-13
Rewrite code to get all clarification from a dedicated ClarificationService

vmcj:
During NWERC I discussed with @nickygerritsen to extract more logic to dedicated services. This is a start to see how much work it is and if the code will look better with it.

Personally I think the code does become better as it's now clear in the Controller what selections we make (the where clauses) and all default guarding is done in the QueryBuilder. There are still some calls where we basically have the same queries which could be merged in their own functions to remove further duplication.

### DIFF
--- a/webapp/src/Controller/API/ClarificationController.php
+++ b/webapp/src/Controller/API/ClarificationController.php
@@ -7,7 +7,13 @@ use App\Entity\Clarification;
 use App\Entity\Contest;
 use App\Entity\ContestProblem;
 use App\Entity\Team;
+use App\Service\AuthorizedUserService;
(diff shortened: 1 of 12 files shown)


nickygerritsen (approved):
A lot of changes so didn't check in detail, but in general I like the idea a lot and I think the changes make sense.

The auth service is also nice, but maybe we should not squash that commit into the other OR change the commit message?

vmcj:
> The auth service is also nice, but maybe we should not squash that commit into the other OR change the commit message?

The authorizedUser service was already merged in another PR so only contained indentation fixes which were not shipped. I've now removed those.

PR #3660, opened 2026-06-26
Fix contest problemset replacement and deletion handling

### LINKED ISSUE #3633: Internal Error 500 when re-uploading problemset document

Receive a 500 when re-uploading the problemset document for a contest.

### Your environment

- DOMjudge 10.0.0DEV/069ec3200

### Steps to reproduce

Start the domserver with docker.

```bash
sudo docker network create dj-network
sudo docker run -d --net="dj-network" --hostname mariadb --name dj-mariadb -e MYSQL_ROOT_PASSWORD=dbrtpwd -e MYSQL_USER=domjudge -e MYSQL_PASSWORD=dbpwd -e MYSQL_DATABASE=domjudge mariadb:10 --max-connections=1000 --max-allowed-packet=1g --innodb-log-file-size=4g
sudo docker run -d --net="dj-network" -e MYSQL_HOST=mariadb -e MYSQL_USER=domjudge -e MYSQL_DATABASE=domjudge -e MYSQL_PASSWORD=dbpwd -e MYSQL_ROOT_PASSWORD=dbrtpwd -e CONTAINER_TIMEZONE=Asia/Shanghai -p 12345:80 --name domserver herano/domserver:bleeding
```

### Method 1

1. Upload the problemset document to demo contest. This should be OK.
2. Upload another problemset to demo contest without deleting it.

### Method 2

1. Create a new contest.
2. When creating the contest, add the problemset to the contest.

### Method 1

The problemset is replaced by the newly uploaded file.

### Method 2

The contest is created with the problemset added.

### Actual behaviour

500 Internal Error with:

```text
While adding an entity of class App\Entity\ContestProblemsetContent with an ID hash of "1" to the identity map, 
another object of class App\Entity\ContestProblemsetContent was already present for the same ID. This exception 
is a safeguard against an internal inconsistency - IDs should uniquely map to entity object instances. This problem 
may occur if: - you use application-provided IDs and reuse ID values; - database-provided IDs are reassigned after 
truncating the database without clearing the EntityManager; - you might have been using 
EntityManager#getReference() to create a reference for a nonexistent ID that was subsequently (by the RDBMS) 
assigned to another entity. Otherwise, it might be an ORM-internal inconsistency, please report it.
```

Image

The checkbox cannot be selected.

HeRaNO:
Contest problemsets and problem statements use transient form fields for uploads and deletion. Do not clear persisted content when no file is submitted or a delete checkbox is false. Otherwise, normal edits can remove the existing document.

Preserve managed Doctrine collections when replacing documents. Update the existing shared-primary-key content entity instead of creating a new one. Clear transient upload and delete state after processing so repeated lifecycle callbacks are idempotent.

Reuse existing problem statement content during ZIP re-import. Keep the current behavior that removes an old statement when an archive has none.

Rely on the existing lifecycle and saveEntity handling in Jury contest paths. Do not call processContestProblemset() explicitly. Add regressions for upload, preservation, replacement, deletion, and ZIP re-import of contest and problem documents.

Fixes #3633.

### DIFF
--- a/webapp/src/Entity/Contest.php
+++ b/webapp/src/Entity/Contest.php
@@ -1699,10 +1699,16 @@ class Contest extends BaseApiEntity implements
 
     public function setContestProblemsetContent(?ContestProblemsetContent $content): self
     {
-        $this->contestProblemsetContent = new ArrayCollection();
-        if ($content) {
-            $this->contestProblemsetContent->add($content);
+        foreach ($this->contestProblemsetContent->toArray() as $existingContent) {
+            if ($existingContent !== $content) {
+                $this->contestProblemsetContent->removeElement($existingContent);
+            }
+        }
+        if ($content !== null) {
             $content->setContest($this);
+            if (!$this->contestProblemsetContent->contains($content)) {
+                $this->contestProblemsetContent->add($content);
+            }
         }
 
         return $this;
@@ -1721,7 +1727,13 @@ class Contest extends BaseApiEntity implements
             $this
                 ->setContestProblemsetContent(null)
                 ->setContestProblemsetType(null);
-        } elseif ($this->getContestProblemsetFile()) {
+
+            $this->contestProblemsetFile = null;
+            $this->clearContestProblemset = false;
+            return;
+        }
+
+        if ($this->getContestProblemsetFile()) {
             $content               = file_get_contents($this->getContestProblemsetFile()->getRealPath());
             $clientName            = $this->getContestProblemsetFile()->getClientOriginalName();
             $contestProblemsetType = Utils::getTextType($clientName, $this->getContestProblemsetFile()->getRealPath());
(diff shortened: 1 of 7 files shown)


meisterT:
Is this also a problem for problem.pdf upload? If not, how do we solve it there? Can we/should we centralize this, e.g. in a trait?

HeRaNO:
> Is this also a problem for problem.pdf upload?

Oh, I haven't actually tested that scenario yet. I'll look into it.

HeRaNO:
I tried the following two methods:

1. Upload the problem statement when creating a new problem: OK.
2. Re-upload the statement of one problem (both with and without a statement previously): 500.

Would you like to postpone this PR and refactor the statement management? The base logic is the same for problemset uploading and problem statement uploading.

meisterT:
Yeah, I would prefer to fix both here to see how this can be centralized

webapp/src/Controller/Jury/ContestController.php
```diff
@@ -561,6 +561,7 @@ public function editAction(Request $request, string $contestId): Response
             $deletedProblems = $getDeletedEntities($contest->getProblems(), 'getProbid');
 
             $this->assetUpdater->updateAssets($contest);
+            $contest->processContestProblemset();
```
meisterT: Do we need this (and the one below)?

@nickygerritsen might be the best person to understand whether we need it in the light of `#[ORM\PrePersist]`

meisterT:
btw, you might have to rebase on top of `main` to let all expected checks run (we changed these recently)

webapp/src/Controller/Jury/ContestController.php
nickygerritsen: PrePersist should auto run this indeed, so it shouldn't be needed...

nickygerritsen (commented):
I think the change makes sense but indeed do wonder about the thing @meisterT already mentioned.

nickygerritsen (approved):
@HeRaNO did you test again with the method removed? If so, this looks good.

HeRaNO:
> @HeRaNO did you test again with the method removed? If so, this looks good.

I have tested locally on my server, and it works well in the two scenarios mentioned in the issue.

PR #3661, opened 2026-06-26
Fix previous submission testcase comparison

### LINKED ISSUE #3627: [BUG] The comparison of the team's two submissions is not functioning correctly.

The test case judging result comparison of the team's two submissions is not functioning correctly.

### Your environment

- DOMjudge 10.0.0DEV/a023fd3ec
- Debain Trixe
- Nginx

### Steps to reproduce
1. Running a contest
2. A team submit a problem twice or more
3. Jury check the submissions

Image

### Expected behaviour
When jury check the later submission, it can see the test cases run difference between the previous submission.

### Actual behaviour

It always show the test case run result for current submission.

The test cases for this problem have not undergone any editing, and the same issue occurs regardless of which submission is viewd.

HeRaNO:
Avoid relying on the partially hydrated Testcase::judging_runs collection when rendering the previous submission.

Load the previous judging runs separately, index them by testcase ID, and pass that explicit mapping to displayTestcaseResults.

Add a regression test covering different current/previous results and the missing previous-run case.

Fixes #3627

### DIFF
--- a/webapp/src/Controller/Jury/SubmissionController.php
+++ b/webapp/src/Controller/Jury/SubmissionController.php
@@ -554,6 +554,8 @@ class SubmissionController extends BaseController
         $lastJudging = null;
         /** @var Testcase[] $lastRuns */
         $lastRuns = [];
+        /** @var array<int, JudgingRun> $lastRunsByTestcaseId */
+        $lastRunsByTestcaseId = [];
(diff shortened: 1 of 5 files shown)


meisterT (approved):
Thanks, I will think a little bit how to improve the code there in general, but that shouldn't stop us from merging this fix

PR #3663, opened 2026-06-29
Fix native checkbox alignment in contest forms

HeRaNO:
The contest form removed left padding from all .form-check elements to keep bootstrap-toggle controls aligned with the table cell edge. Native Bootstrap checkboxes rely on that padding to balance the negative margin on the checkbox input, so removing it moved them out of their table cell.

This made long-label checkboxes such as "Delete contest problemset document" hard or impossible to select.

Keep the default Bootstrap layout for native checkboxes and only compensate the generated bootstrap-toggle control alignment instead.

Part of #3633.

---

Before this patch:

Image-Before

After this patch:

Image-After

### DIFF
--- a/webapp/public/style_jury.css
+++ b/webapp/public/style_jury.css
@@ -231,7 +231,13 @@ div.submission-summary > span {
     font-size: 0.8rem;
     margin-top: 0.15rem;
 }
-.contest-fields-table td .form-check { padding-left: 0; }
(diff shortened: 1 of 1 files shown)


meisterT:
Please add before/after screenshots

HeRaNO:
Screenshots updated. Would you restart the failing actions? It seems unrelated to this PR.

vmcj (approved):

PR #3665, opened 2026-07-05
Initial trivial implementation of some `draft` problempackage spec options.

vmcj:
This is some progress from https://github.com/DOMjudge/domjudge/pull/3664 in reworked format ready to be discussed.

I plan to implement some of the trivial parts of the different problemspecs to get discussion started on how much we want to implement. As we have people in the wild using `main` I added a warning for all problemspecs which I'm not reasonably sure of that we implement those. In other words, anything besides `icpc-legacy` will get a warning on imports via the API.

I did not look into uploads via the UI yet as that complicates the work quite a lot.

### DIFF
--- a/webapp/src/Entity/Problem.php
+++ b/webapp/src/Entity/Problem.php
@@ -331,7 +331,10 @@ class Problem extends BaseApiEntity implements
         return $this;
     }
 
-    public function getTypesAsString(): string
+    /**
+     * @return list<string>
+     */
+    public function getTypesAsStringArray(): array
     {
         $typeConstants = $this->getTypes();
         $typeStrings = [];
@@ -341,7 +344,12 @@ class Problem extends BaseApiEntity implements
             }
             $typeStrings[] = $this->typesToString[$type];
         }
-        return implode(', ', $typeStrings);
+        return $typeStrings;
+    }
+
+    public function getTypesAsString(): string
+    {
+        return implode(', ', $this->getTypesAsStringArray());
     }
 
     /**
--- a/webapp/src/Service/ImportProblemService.php
+++ b/webapp/src/Service/ImportProblemService.php
@@ -1206,6 +1206,27 @@ readonly class ImportProblemService
     }
 
     /**
+     * Allow to import the types both as arrays and as strings. Due
+     * to the recursion to make parsing of the strings easier this now
+     * also works for arrays inside arrays.
+     *
+     * @param string|array<mixed, mixed> $input
+     * @return string[]
+     */
+    private static function parseTypes(string|array $input): array
+    {
+        $final = [];
+        if (is_array($input)) {
+            foreach ($input as $possibleType) {
+                $final = array_merge($final, self::parseTypes($possibleType));
+            }
+        } else {
+            $final = array_merge($final, preg_split("/[\s,;]+/", $input));
+        }
+        return $final;
+    }
+
+    /**
      * Returns true iff the yaml could be parsed correctly.
      *
      * @param array{danger?: string[], info?: string[]} $messages
@@ -1241,7 +1262,7 @@ readonly class ImportProblemService
 
         $validationMode = 'default';
         if (isset($yamlData['type'])) {
-            $types = explode(' ', $yamlData['type']);
+            $types = self::parseTypes($yamlData['type']);
             // Validation happens later when we set the properties.
             $yamlProblemProperties['typesAsString'] = $types;
             if (in_array('interactive', $types)) {
--- a/webapp/src/Service/ImportProblemService.php
+++ b/webapp/src/Service/ImportProblemService.php
@@ -1293,6 +1293,20 @@ readonly class ImportProblemService
             }
         }
 
+        if (isset($yamlData['problem_format_version'])) {
+            $version = $yamlData['problem_format_version'];
+            if (in_array($version, ['legacy', '2025-09-draft', 'draft', '2025-09', '2023-07-draft'])) {
+                $messages['info'][] = sprintf("Problem format version '%s' support still experimental.", $version);
+            } elseif ($version !== 'icpc-legacy') {
+                // 2023-07-draft used in Unit tests
+                $messages['warning'][] = sprintf("Unknown problem format version '%s'.", $version);
+                return false;
+            }
+            // TODO: At the moment we only warn the user about unknown problem specifications and the experimental support.
+            // In the future we should either be more strict on only allowing properties available in the spec or warn the
+            // user on properties/values outside of the specified specification.
+        }
+
         foreach ($yamlProblemProperties as $key => $value) {
             try {
                 $propertyAccessor->setValue($problem, $key, $value);
--- a/webapp/tests/Unit/Service/ImportProblemServiceTest.php
+++ b/webapp/tests/Unit/Service/ImportProblemServiceTest.php
@@ -135,6 +135,53 @@ YAML;
         }
     }
 
+    #[DataProvider('provideAlternativeTypeNotations')]
+    public function testTypesStringWithAlternativeChars(string $separator): void
+    {
+        $expectedTypes = ['pass-fail', 'interactive'];
+        $typesAsString = implode($separator, $expectedTypes);
+        $specVersion = 'draft';
+        $yaml = <<<YAML
+name: test
+problem_format_version: $specVersion
+type: $typesAsString
+YAML;
+
+        $messages = [];
+        $validationMode = 'xxx';
+        $problem = new Problem();
+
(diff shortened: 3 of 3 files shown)


webapp/src/Service/ImportProblemService.php
```diff
@@ -1279,6 +1279,18 @@ public static function parseYaml(bool|string $problemYaml, array &$messages, str
...
+        if (isset($yamlData['problem_format_version'])) {
+            $version = $yamlData['problem_format_version'];
+            if (in_array($version, ['legacy', '2025-09-draft', 'draft', '2025-09', '2023-07-draft'])) {
+                $messages['info'][] = sprintf("Problem format version '%s' support still experimental.", $version);
+            } elseif ($version !== 'icpc-legacy') {
+                // 2023-07-draft used in Unit tests
+                $messages['warning'][] = sprintf("Unknown problem format version '%s'.", $version);
+                return false;
+            }
+            // TODO: We don´t have any differences in how we handle these specs yet.
```
meisterT: I am not sure what this TODO is saying.

webapp/tests/Unit/Service/ImportProblemServiceTest.php
```diff
@@ -54,6 +56,46 @@ public function testMinimalYamlTest(): void
         $this->assertEquals(null, $problem->getSpecialCompareArgs());
     }
 
+    /**
+     * @param array{info?: string[], warning?: string[], danger?: string[]} $messages
+     * @param string[] $expected
+     */
+    private function assertProblemSpecWarning(string $version, array $messages, array $expected = ['info']): void
+    {
+        if ($version !== 'icpc-legacy') {
```
meisterT: consider putting this into the yield below, to make the test simpler

webapp/src/Entity/Problem.php
```diff
@@ -344,6 +344,20 @@ public function getTypesAsString(): string
         return implode(', ', $typeStrings);
     }
 
+    /**
+     * @return list<string>
+     */
+    public function getTypesAsStringArray(): array
```
meisterT: It's weird to add a unused function in a commit of its own.

-----

Separately, copying the above function into this one without the `implode` and then the above function calls this one and runs `implode` on the returned array?

webapp/src/Service/ImportProblemService.php
```diff
@@ -1200,6 +1200,23 @@ public static function parseTestCaseGroupMeta(string $fileContent, string $name,
         return $testcaseGroup;
     }
 
+    /**
+     * @param string|array<mixed, mixed> $input
```
meisterT, line 1207: it might help to explain here what this can be called with - what are input examples that are parsed?

webapp/src/Service/ImportProblemService.php
vmcj: Currently you can set the problem spec, we would now parse it but there is no code at all which does something different depending on which version you have.

So you could state your problem is `icpc-legacy` and we would just try to parse the scoring properties with it, which is not according to spec but currently still what we do. Also see my question in Slack about this.

In the past we stated we only do `icpc-legacy` and now we do more, I wonder if we should be strict or try to parse everything as best as possible and therefor allow for illegal packages containing properties from different specs.

webapp/tests/Unit/Service/ImportProblemServiceTest.php
vmcj: I now moved it to the testProblemPackageFormat, I don't see how I can add this to the `yield` without still having to do a boolean check in the code.

webapp/src/Entity/Problem.php
vmcj: I added it on it's own to keep a clear distinction between the added tests and showing that there is a new function which on it's own works. When approved I'll squash some things to remove that weird commit.

webapp/src/Service/ImportProblemService.php
vmcj: I added an explanation please check if this is what you wanted.

meisterT:
Can you please rearrange the commits in a way you would like to merge them? That will make review easier

vmcj:
> Can you please rearrange the commits in a way you would like to merge them? That will make review easier

Done.

webapp/src/Service/ImportProblemService.php
```diff
@@ -1205,6 +1205,27 @@ public static function parseTestCaseGroupMeta(string $fileContent, string $name,
...
+     */
+    private static function parseTypes(string|array $input): array
+    {
+        $final = [];
+        if (is_array($input)) {
+            foreach ($input as $possibleType) {
+                $final = array_merge($final, self::parseTypes($possibleType));
+            }
+        } else {
+            $final = array_merge($final, preg_split("/[\s,;]+/", $input));
```
meisterT, line 1224: consider using `PREG_SPLIT_NO_EMPTY` to avoid tripping over `type: pass-fail,`  and similar things

webapp/src/Service/ImportProblemService.php
```diff
@@ -1293,6 +1314,20 @@ public static function parseYaml(bool|string $problemYaml, array &$messages, str
             }
         }
 
+        if (isset($yamlData['problem_format_version'])) {
+            $version = $yamlData['problem_format_version'];
+            if (in_array($version, ['legacy', '2025-09-draft', 'draft', '2025-09', '2023-07-draft'])) {
+                $messages['info'][] = sprintf("Problem format version '%s' support still experimental.", $version);
+            } elseif ($version !== 'icpc-legacy') {
+                // 2023-07-draft used in Unit tests
+                $messages['warning'][] = sprintf("Unknown problem format version '%s'.", $version);
```
meisterT, line 1323: this should be `danger`, not `warning`

webapp/src/Service/ImportProblemService.php
```diff
@@ -1293,6 +1314,20 @@ public static function parseYaml(bool|string $problemYaml, array &$messages, str
             }
         }
 
+        if (isset($yamlData['problem_format_version'])) {
```
meisterT, line 1317: I wonder whether this check should be one of the first checks (given that we are bailing out in some cases

webapp/tests/Unit/Service/ImportProblemServiceTest.php
```diff
@@ -57,6 +58,47 @@ public function testMinimalYamlTest(): void
         $this->assertEquals(null, $problem->getSpecialCompareArgs());
     }
 
+    /**
+     * @param array{info?: string[], warning?: string[], danger?: string[]} $messages
+     * @param string[] $expected
+     */
+    private function assertProblemSpecWarning(string $version, array $messages, array $expected = ['info']): void
```
meisterT, line 65: do we ever pass in the 3rd argument `$expected`?

webapp/tests/Unit/Service/ImportProblemServiceTest.php
```diff
@@ -57,6 +58,47 @@ public function testMinimalYamlTest(): void
...
+            if (!in_array($type, $expected)) {
+                if (isset($messages[$type])) {
+                    $this->assertEmpty($messages[$type]);
+                }
+            }
+        }
+        $this->assertNotEmpty($messages['info']);
+        $this->assertStringContainsString(
+            sprintf("Problem format version '%s' support still experimental.", $version),
+            $messages['info'][0]
```
meisterT, line 77: Let's make this less brittle by accepting if it is in any `$messages['info']`, not necessarily the first one?

webapp/tests/Unit/Service/ImportProblemServiceTest.php
```diff
@@ -57,6 +58,47 @@ public function testMinimalYamlTest(): void
...
+        );
+    }
+
+    #[DataProvider('problemSpecVersionProvider')]
+    public function testProblemPackageFormatTest(string $problemSpecificationVersion): void
```
meisterT, line 82: super nit: rename to `testProblemPackageFormatVersions`?

webapp/tests/Unit/Service/ImportProblemServiceTest.php
```diff
@@ -745,4 +834,60 @@ public function testParseTestCaseGroupMetaInvalidRangeTooManyValuesRejected(): v
         $this->assertNotEmpty($messages['danger']);
         $this->assertStringContainsString("Invalid range '100 101 102'", $messages['danger'][0]);
     }
+
+    public static function problemSpecVersionProvider(): Generator
+    {
+        yield ['2025-09-draft'];
```
meisterT, line 840: these are all accepted tests, should we add one negative test as well?

PR #3666, opened 2026-07-06
Allow specifying the restricted language for a problem via new `langu…

vmcj:
…ages` field

See spec: https://www.kattis.com/problem-package-format/spec/2025-09.html#languages

### DIFF
--- a/webapp/src/Service/ImportProblemService.php
+++ b/webapp/src/Service/ImportProblemService.php
@@ -5,6 +5,7 @@ namespace App\Service;
 use App\Entity\Contest;
 use App\Entity\ContestProblem;
 use App\Entity\Executable;
+use App\Entity\Language;
 use App\Entity\Problem;
 use App\Entity\ProblemAttachment;
 use App\Entity\ProblemAttachmentContent;
@@ -230,7 +231,8 @@ readonly class ImportProblemService
                 ->setSpecialCompareArgs('')
                 ->setRunExecutable()
                 ->setMemlimit(null)
-                ->setOutputlimit(null);
+                ->setOutputlimit(null)
+                ->resetLanguages();
 
             $contestProblem
                 ?->setPoints(1)
@@ -246,7 +248,14 @@ readonly class ImportProblemService
         // Parse YAML before validation so that all problem properties (including
         // types) are set before the entity validation callbacks fire.
         $validationMode = 'default';
-        if (!static::parseYaml($problemYaml, $messages, $validationMode, $propertyAccessor, $problem)) {
+        if (!static::parseYaml(
+            $problemYaml,
+            $messages,
+            $validationMode,
+            $propertyAccessor,
+            $problem,
+            fn(array $languageIds, array &$failedLanguageIds): array => $this->helperLanguages($languageIds, $failedLanguageIds)
+        )) {
             return null;
         }
 
@@ -1206,12 +1215,44 @@ readonly class ImportProblemService
     }
 
     /**
+     * @param array<string|bool|int|null> $languageIds
+     * @param string[] &$failedLanguageIds
+     * @return Language[]
+     */
+    public function helperLanguages(array $languageIds, array &$failedLanguageIds): array
+    {
+        $languages = [];
+        foreach ($languageIds as $langId) {
+            if (is_bool($langId)) {
+                $langId = $langId ? 'true' : 'false';
+            } elseif (is_null($langId)) {
+                $langId = 'null';
+            }
+            $langId = strval($langId);
+            $language = $this->em->getRepository(Language::class)->findByExternalId($langId);
+            if ($language) {
+                $languages[] = $language;
+            } else {
+                $failedLanguageIds[] = $langId;
+            }
+        }
+        return $languages;
+    }
+
+    /**
      * Returns true iff the yaml could be parsed correctly.
      *
      * @param array{danger?: string[], info?: string[]} $messages
+     * @param callable(string[], string[]&): Language[]|null $languageResolver
      */
-    public static function parseYaml(bool|string $problemYaml, array &$messages, string &$validationMode, PropertyAccessor $propertyAccessor, Problem $problem): bool
-    {
+    public static function parseYaml(
+        bool|string $problemYaml,
+        array &$messages,
+        string &$validationMode,
+        PropertyAccessor $propertyAccessor,
+        Problem $problem,
+        ?callable $languageResolver = null
+    ): bool {
         if ($problemYaml === false) {
             // While there was no problem.yaml, there was also no error in parsing.
             return true;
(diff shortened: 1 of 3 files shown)


webapp/src/Service/ImportProblemService.php
```diff
@@ -247,7 +248,14 @@ public function importZippedProblem(
...
         // types) are set before the entity validation callbacks fire.
         $validationMode = 'default';
-        if (!static::parseYaml($problemYaml, $messages, $validationMode, $propertyAccessor, $problem)) {
+        if (!static::parseYaml(
+            $problemYaml,
+            $messages,
+            $validationMode,
+            $propertyAccessor,
+            $problem,
+            fn(array $languageIds): array => $this->helperLanguages($languageIds)
```
meisterT: does this need `array|string`?

webapp/src/Service/ImportProblemService.php
```diff
@@ -1279,6 +1310,32 @@ public static function parseYaml(bool|string $problemYaml, array &$messages, str
...
+                    $languageIds = preg_split('/[\s,;]+/', $yamlData['languages']);
+                } else {
+                    $languageIds = $yamlData['languages'];
+                }
+                if (count($languageIds) === 0) {
+                    $messages['danger'][] = "'languages' field provided without content.";
+                    return false;
+                }
+                $languages = $languageResolver($languageIds);
+                if (gettype($languages) === 'string') {
```
meisterT: Ah, this is where `string` comes from - gettype use (or mixing return type) looks a bit hacky to me

webapp/src/Service/ImportProblemService.php
```diff
@@ -1200,13 +1208,36 @@ public static function parseTestCaseGroupMeta(string $fileContent, string $name,
         return $testcaseGroup;
     }
 
+    /**
+     * @param string[] $languageIds
+     * @return Language[]
```
meisterT, line 1214: depending on what we do either this or the return type need to be updated

webapp/src/Service/ImportProblemService.php
vmcj: I can do the `&` trick we do with `$messages` if you think that's better? I like to have the failing language there and don't want to make an exception out of this.

meisterT: Yes, I think that's better

webapp/src/Service/ImportProblemService.php
```diff
@@ -1200,13 +1208,37 @@ public static function parseTestCaseGroupMeta(string $fileContent, string $name,
...
+     */
+    public function helperLanguages(array $languageIds, string &$failedLangId): array
+    {
+        $languages = [];
+        foreach ($languageIds as $langId) {
+            $language = $this->em->getRepository(Language::class)->findByExternalId($langId);
+            if (!$language) {
+                $failedLangId = $langId;
```
meisterT: Can there only be one that fails? Or could there be multiple?

vmcj: In theory there can be multiple.

webapp/src/Service/ImportProblemService.php
```diff
@@ -1279,6 +1313,33 @@ public static function parseYaml(bool|string $problemYaml, array &$messages, str
             }
         }
 
+        if (isset($yamlData['languages'])) {
+            if ($yamlData['languages'] === 'all') {
+                // Explicit no restriction
```
meisterT: I think we need to reset languages somewhere before this.

Otherwise, this might not work correctly on reimport, e.g. imagine that you have a problem which is currently restricted to just one language, say Kotlin, then you reimport with `all` (or no restriction at all) and the Kotlin restriction still stands.

webapp/src/Service/ImportProblemService.php
```diff
@@ -1279,6 +1313,33 @@ public static function parseYaml(bool|string $problemYaml, array &$messages, str
...
+            } elseif ($languageResolver === null) {
+                $messages['danger'][] = 'Cannot resolve problem languages without a language resolver.';
```
meisterT, line 1335: Is there a missing `return false` after this `danger` message?

webapp/src/Service/ImportProblemService.php
```diff
@@ -1279,6 +1313,33 @@ public static function parseYaml(bool|string $problemYaml, array &$messages, str
...
+            } else {
+                if (is_string($yamlData['languages'])) {
```
meisterT: super edge case: if we ever have a language called `yes` or similar this might parse weirdly (haven't tried). Not sure whether it is worth doing something about it, since `'yes'` probably works?

webapp/src/Service/ImportProblemService.php
```diff
@@ -1279,6 +1313,33 @@ public static function parseYaml(bool|string $problemYaml, array &$messages, str
...
+                    $languageIds = preg_split('/[\s,;]+/', $yamlData['languages']);
```
meisterT: consider using `PREG_SPLIT_NO_EMPTY` here

webapp/src/Service/ImportProblemService.php
vmcj: I agree that I should do something here, not sure what I would prefer for reimport though. For now I'll set the restrictions to nothing, which means that when the other code would work with import this would correctly reset but if it doesn't this would not break it further.

webapp/src/Service/ImportProblemService.php
vmcj: I don't understand why this comment is for this line. So i can't really test it and add a test for it.

vmcj: I've now added some checks for truthy values and language names as numbers.

meisterT: I think you need to test with:
```
- languages: true
```
and
```
- languages: 1.5
```

Not likely, but this would lead to a type error still.

webapp/src/Service/ImportProblemService.php
meisterT, line 236: do we need to reset the langs here as well?

webapp/src/Service/ImportProblemService.php
```diff
@@ -1287,6 +1327,34 @@ public static function parseYaml(bool|string $problemYaml, array &$messages, str
             }
         }
 
+        if (isset($yamlData['languages'])) {
+            $problem->resetLanguages();
+            if ($languageResolver === null) {
+                $messages['danger'][] = 'Cannot resolve problem languages without a language resolver.';
+                return false;
+            } if ($yamlData['languages'] !== 'all') {
```
meisterT: Was this meant to be an `elseif` or did you mean to add a line break before the `if`?

webapp/src/Service/ImportProblemService.php
```diff
@@ -1287,6 +1327,34 @@ public static function parseYaml(bool|string $problemYaml, array &$messages, str
...
+                // When changing this also update testTruthyLanguages in importProblemServiceTest
```
meisterT: there is no `testTruthyLanguages` in the test?

webapp/src/Service/ImportProblemService.php
```diff
@@ -1287,6 +1328,38 @@ public static function parseYaml(bool|string $problemYaml, array &$messages, str
             }
         }
 
+        // We can't check for isset as `languages: null` should be handled.
+        if (key_exists('languages',$yamlData)) {
```
meisterT: phpcs nit

webapp/tests/Unit/Service/ImportProblemServiceTest.php
```diff
@@ -670,4 +673,207 @@ public function testParseTestCaseGroupMetaInvalidRangeTooManyValuesRejected(): v
         $this->assertNotEmpty($messages['danger']);
         $this->assertStringContainsString("Invalid range '100 101 102'", $messages['danger'][0]);
     }
+
+    #[DataProvider('provideLanguages')]
+    public function testLanguages(
+        array $expectedLanguages, ?array $secondUploadExpectedLanguages = null,
+        ?string $languagesString = null, ?string $secondUploadLanguagesString = null
+    ): void{
```
meisterT: phpcs nit

meisterT (approved):

PR #3668, opened 2026-07-10
runguard: Strip the trailing newline from cpu.stat values.

meisterT:
`cgroup_read_stats_*()` returns each value as read from the file, so the `cpu.stat` debug lines were logged with an embedded newline and printed with a blank line after each of them.

This only shows up with verbose/debug logging.

### DIFF
--- a/judge/runguard.cc
+++ b/judge/runguard.cc
@@ -70,6 +70,7 @@
 #include <set>
 #include <sstream>
 #include <string>
+#include <string_view>
 #include <utility>
 #include <vector>
 #include <linux/prctl.h>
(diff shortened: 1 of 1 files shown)


judge/runguard.cc
```diff
@@ -490,7 +491,12 @@ void output_cgroup_stats(double *cputime)
 	void *handle;
 	ret = cgroup_read_stats_begin("cpu", cgroupname, &handle, &stat);
 	while (ret == 0) {
-		logmsg(LOG_DEBUG, "cpu.stat: {} = {}", stat.name, stat.value);
+		/* libcgroup hands us the value from cpu.stat verbatim,
+		   including its trailing newline. */
+		std::string_view value{stat.value};
+		value = value.substr(0, value.find_first_of("\r\n"));
```
vmcj: Wouldn't we want the last in case a stat ever has `\r\n` in the middle?

meisterT: Cannot ever happen, see https://github.com/libcgroup/libcgroup/blob/main/include/libcgroup/iterators.h#L243 and also the parsing code below would break.

In order to not to have to think about this fact, I changed the code anyway to only trim the *trailing* newline chars and leave anything in the middle alone (if it were to exist).

vmcj (approved):

PR #3669, opened 2026-07-10
runguard: Do not leave SIGCHLD blocked for the command.

meisterT:
We block SIGCHLD before forking, to keep it from arriving outside the `pselect()` that waits for it. The command inherits that signal mask, and `execve()` keeps it, so every command we run so far starts with SIGCHLD blocked. That is state of ours leaking into the sandbox, and it changes how a command that waits for children of its own behaves.

Restore the signal mask in the child before executing the command.


vmcj (approved):
Explanation makes sense.

PR #3670, opened 2026-07-10
runguard: Don't close our own stderr with the output files.

meisterT:
`child_redirfd[i]` defaults to `i` when no `-o/-e` is given, so the close loop after draining the child's output shut runguard's own stdout and stderr. Every diagnostic goes to `std::cerr`, so from that point on all `logmsg()`, `warning()` and `die()` messages were silently discarded: `total memory used`, `deleted cgroup` and any warning from `output_cgroup_stats()` never reached the user, even under `-v`. Exit codes were still set, so failures were not silent, but their explanations were.

Closing fd 2 also frees the descriptor number for reuse by the next `open()`, e.g. from libcgroup during cgroup teardown.

Keep closing stdout, whose EOF signals the end of the child's output to a downstream reader such as runpipe, and only skip the close when the descriptor is our own stderr.

`test_streamsize_stderr` allowed 100 bytes on top of the stream limit for the messages runguard writes to the shared stderr. Restoring the discarded ones adds the terminating signal and the soft time limit to that, so widen the slack.

### DIFF
--- a/judge/runguard.cc
+++ b/judge/runguard.cc
@@ -1445,10 +1445,20 @@ int main(int argc, char **argv)
 			pump_pipes(&readfds, data_read, data_passed);
 		} while ( data_passed[1] + data_passed[2] > total_data );
 
-		/* Close the output files */
-		for(int i=1; i<=2; i++) {
-			ret = close(child_redirfd[i]);
-			if( ret!=0 ) die(errno,"closing output fd {}", i);
+		/* Always close stdout, also when it is our own without `-o':
+		   its EOF signals the end of the command's output to a
+		   downstream reader such as runpipe, which an interactive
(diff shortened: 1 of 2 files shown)


eldering (approved):
I agree that the change is correct and does fix the issue as mentioned.

I'm wondering if that for loop should maybe be broken into two separate statements, each with comments for clarity. The reason is that naively each of `child_redirfd[i]` should be closed IFF `redir_stdout` respectively `redir_stderr` are set. However, as you mentioned, stdout as a special case should always be closed because of the EOF detection with interactive problem handling. Two separate statements, each with their own comment would make that clearer, I think.

meisterT:
Good idea, please take another look

PR #3674, opened 2026-07-11
judgedaemon: stop forking to do `chmod -R`.

meisterT:
The feedback directory was made group/other-writable before running the compare script and read-only again afterwards by shelling out to `chmod -R go+w` / `chmod -R go-w` (PHP's `chmod()` is not recursive).

Instead, add a small `chmodRecursive()` helper that walks the tree in-process and adjusts the requested permission bits.

This speeds up a 277 testcase submission with trivial runtime by ~5%.

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -1127,6 +1127,59 @@ class JudgeDaemon
     }
 
     /**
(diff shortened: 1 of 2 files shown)


vmcj (approved):
This will most likely break some parts of my earlier PR for the output visualizers but let's merge this one first.

Issue #3417, opened 2026-02-04
500 Internal Server Error on team page, when contest delay start.


cubercsl:
### Description of the problem
500 Internal Server Error on team page, when the contest delay start.

### Your environment
DOMjudge 9.0 / Ubuntu 24.04 / Nginx

### Steps to reproduce
1. Create a test contest, set a contest start time.
2. Import problems from zip.
3. Delay contest start.
4. Jury submits another solution for testing.

### Expected behaviour
Jury can make submission without any error.

### Actual behaviour
An error occurred when jury making a submission. 
```
No negative values allowed in score key element, got -423.000000000.
```

Resuming contest start can mitigate this.

```
[2026-02-01T21:21:56.966732+08:00] request.INFO: Matched route "team_submit". {"route":"team_submit","route_parameters":{"_route":"team_submit","problem":null,"_controller":"App\\Controller\\Team\\SubmissionController::createAction"},"request_uri":"https://dj.ecf.icpc.site/team/submit","method":"POST"} []
[2026-02-01T21:21:56.969648+08:00] security.DEBUG: Read existing security token from the session. {"key":"_security_domjudge","token_class":"Symfony\\Component\\Security\\Core\\Authentication\\Token\\UsernamePasswordToken"} []
[2026-02-01T21:21:56.971273+08:00] php.INFO: User Deprecated: MySQL 5.6 support is deprecated and will be removed in DBAL 4. Consider upgrading to MySQL 5.7 or later. (AbstractMySQLDriver.php:113 called by AbstractDriverMiddleware.php:68, https://github.com/doctrine/dbal/pull/5072, package doctrine/dbal) {"exception":"[object] (ErrorException(code: 0): User Deprecated: MySQL 5.6 support is deprecated and will be removed in DBAL 4. Consider upgrading to MySQL 5.7 or later. (AbstractMySQLDriver.php:113 called by AbstractDriverMiddleware.php:68, https://github.com/doctrine/dbal/pull/5072, package doctrine/dbal) at /opt/domjudge/domserver/webapp/vendor/doctrine/deprecations/src/Deprecation.php:208)"} []
[2026-02-01T21:21:56.975873+08:00] security.DEBUG: User was reloaded from a user provider. {"provider":"Symfony\\Bridge\\Doctrine\\Security\\User\\EntityUserProvider","username":"jury"} []
[...]

Issue #2163, opened 2023-09-27
Cannot apply rejudging because `todo < 0`


eldering:
### Description of the problem
I've been running a bunch of rejudgings for a contest, and in between I sometimes ran into internal errors, e.g. due to #2162. Now I seem to have gotten in a weird state where I can't apply a (full contest) rejudging after it completed.

### Your environment
DOMjudge version 8.0.

### Steps to reproduce
Unclear exactly how to reproduce, but it likely involves running a rejudging, then hitting an internal error and then either ignoring or marking as resolved this internal error while the original rejudging was still open.

### Actual behaviour
In this state, I found
```
mysql> SELECT submitid,COUNT(*) AS cnt  FROM judging j WHERE rejudgingid = 80 AND endtime IS NOT NULL GROUP BY 1 HAVING cnt > 1;
+----------+-----+
| submitid | cnt |
+----------+-----+
|     1847 |   2 |
|     1916 |   2 |
|     1934 |   2 |
|     2022 |   2 |
+----------+-----+
4 rows in set (0.01 sec)
```
which means that due to the duplicate judgings attached to this rejudging for 4 submissions, at https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Service/RejudgingService.php#L406 `$done` becomes negative, hence the "apply" button won't show up.

meisterT:
I think this is fixed at HEAD, can you get a dump of your database and try locally at HEAD?

eldering:
Digging a bit deeper, I see that these 4 submissions have 2 valid judgings associated to them. One is from a previous, applied rejudging, the other not associated to a rejudging.
Note that I haven't manually modified the database up to this point.

eldering:
> I think this is fixed at HEAD, can you get a dump of your database and try locally at HEAD?

I'll dump the DB. Trying locally at head is impossible since I'm on CentOS 8 with PHP 7.2, but I can try it on my own computer later.

eldering:
I've added a DB dump at `domjudge@domjudge.org:domjudge-2023-09-27-debug-rejudging_upgraded_dj82.sql.gz` to reproduce it, at least on `8.2` branch, which the dump is also for. I'll post the credentials in the #maintainer channel.

eldering:
Also after migrating the data and code to current `main`, and restarting a new rejudging by doing a "Repeat rejudging" from rejudging 80, I still get at the end no apply button, only a cancel button, while there are no submissions left to rejudge.

[...]

PR #3676, opened 2026-07-12
Repair scoreboard caches of databases that predate the sort keys.

meisterT:
When we introduced sort keys in 6613b28ba, the accompanying migration only added the `sort_key_public` and `sort_key_restricted` columns, it never populated them.

Every rankcache row written before the upgrade therefore keeps the empty string default.  Since the scoreboard is ordered by that column and a team's rank is computed by counting the sort keys greater than its own, all those teams compare equal until something happens to recompute their rank cache.

Shortly after, 441070de3 disallowed negative numbers on the scoreboard: submissions made before the contest start (in practice only by the jury on problem import) are moved up to the contest start and represented as 0. While that fixed `calculateScoreRow()` going forward, but rows written by the *older* code still hold a negative contest-relative solve time. Because the rank cache sums the solve times of all problems of a team, a single such row makes every later rank cache update for that team fail with a message like this:

```
No negative values allowed in score key element, got -68825.000000000.
```

This surfaces as a 500 on `add-judging-run`, i.e. judging a perfectly ordinary pass-fail submission breaks.

This is similar to what has been reported in #3417, but not the same: a delayed contest start made the clamp degrade because `getStarttime()` returned null. ac1a42ac7 fixed that one by passing `false`.

All three above mentioned commits repaired the write path and left the already-written rows alone, so upgraded databases silently carry both defects (in general, it is likely a good idea to remind people to refresh their scoreboard caches on major upgrades, but this migration makes it more robust).

The migration deliberately duplicates the scoring logic rather than calling ScoreboardService: migrations must keep doing the same thing when that service changes.

Found and verified when looking into the pre 9.0 DB linked in #2163.

### DIFF
--- a/webapp/migrations/Version20260711120000.php
+++ b/webapp/migrations/Version20260711120000.php
@@ -1,0 +1,227 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use App\Entity\ScoreboardType;
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+use Exception;
+
+/**
+ * Repair the scoreboard caches of databases that predate the sort keys.
+ *
+ * Two defects can be present in an upgraded database:
+ *
+ * Previously, we disallowed negative numbers on the scoreboard, submissions
+ * made before the contest start (in practice only by the jury during problem
+ * import) were stored with a negative contest-relative solve time.
+ * Because the rank cache sums the solve times of all problems of a team, a
+ * single such row makes every later rank cache update of that team fail with
+ * "No negative values allowed in score key element".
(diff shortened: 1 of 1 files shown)


webapp/migrations/Version20260711120000.php
```diff
@@ -0,0 +1,227 @@
...
+use Exception;
+
+/**
+ * Repair the scoreboard caches of databases that predate the sort keys.
+ *
+ * Two defects can be present in an upgraded database:
+ *
+ * Previously, we disallowed negative numbers on the scoreboard, submissions
+ * made before the contest start (in practice only by the jury during problem
+ * import) were stored with a negative contest-relative solve time.
```
vmcj:
```suggestion
 * import) were stored with a negative contest-relative solve time.
```

vmcj (approved):

PR #3678, opened 2026-07-12
Handle multiple judgings per rejudging on the jury submission page.

meisterT:
Found and verified when debugging #2163.

Visiting a submission with `?rejudgingid=<id>` deduced the judging to show with `getOneOrNullResult()`, which assumes a submission has at most one judging within a given rejudging.

That does not always hold true: when a judging is aborted or requeued after an internal error, a second judging is created for the same (rejudging, submission) pair, resulting in

```
Doctrine\ORM\NonUniqueResultException: More than one result was found
for query although one row or none was expected.
```

Now we pick and show the most recent judging of the rejudging instead. The older ones stay reachable, as the page links to every judging of the submission anyway.

### DIFF
--- a/webapp/src/Controller/Jury/SubmissionController.php
+++ b/webapp/src/Controller/Jury/SubmissionController.php
@@ -301,6 +301,12 @@ class SubmissionController extends BaseController
                 ->setParameter('rejudgingId', $rejudgingId)
                 ->setParameter('submitId', $submitId)
                 ->setParameter('contest', $contest)
+                // A submission can have more than one judging within the same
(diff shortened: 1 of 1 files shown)


vmcj (approved):

vmcj:
What would happen if we have:
Submission:
Initial (failed due to internal error)
Rejudging Internal Error (applied) <---
Rejuding (aborted),

I would expect the last valid one to be on top, but this code seems to suggest otherwise.

meisterT:
Note that this is the query that selects per rejudging. The one that your error points to should be only judging for this rejudging, so it will be selected.

PR #3684, opened 2026-07-18
multipass: Report internal error for `nextpass.in` without exit code 42.

meisterT:
The problem package format states that it is a judge error to create the `nextpass.in` file and exit with any other code than 42. We only looked at `nextpass.in` for a correct verdict, so a validator that created it and exited with 43 silently resulted in a wrong answer instead.

Check this independently of the verdict: the validator may legitimately exit with 42 and ask for another pass while the run itself failed with for example a timelimit, in which case judging simply stops.

See https://icpc.io/problem-package-format/spec/2025-09.html?highlight=nextpass#multi-pass-validation

### DIFF
--- a/judge/judgedaemon.main.php
+++ b/judge/judgedaemon.main.php
@@ -2601,10 +2601,31 @@ class JudgeDaemon
                     . '  ...done in ' . $metadata['wall-time'] . 's (CPU: ' . $runtime . 's), result: ' . $result);
             }
 
+            $hasNextPass = file_exists($passdir . '/feedback/nextpass.in');
+
+            // The spec states we have to treat it is a judge error if
+            // `nextpass.in` is produced when exiting with any other code
+            // than 42.
+            if ($hasNextPass) {
+                $compareMeta = $this->readMetadata($passdir . '/compare.meta');
+                /** @var MetaData_Compare $compareMeta */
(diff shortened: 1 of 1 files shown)


judge/judgedaemon.main.php
```diff
@@ -2601,10 +2601,31 @@ private function runTestcase(
                     . '  ...done in ' . $metadata['wall-time'] . 's (CPU: ' . $runtime . 's), result: ' . $result);
             }
 
+            $hasNextPass = file_exists($passdir . '/feedback/nextpass.in');
+
+            // According to the spec states we have to treat it is a judge
+            // error if `nextpass.in` is produced when exiting with any other
+            // code than 42.
```
vmcj: Consider rephrasing this.

vmcj (approved):

judge/judgedaemon.main.php
meisterT: Rephrase to what? Not sure what change you would like to see

vmcj: According to the spec states we have to treat it is a judge

Feels unnatural to me, so I would have gone with either:
// According to the spec ~~states~~ we have to treat it is a judge
or
// ~~According to t~~ The spec states we have to treat it is a judge

But it might be my head not parsing that nicely.

meisterT: Thx, done

PR #3686, opened 2026-07-18
Validate `limits.validation_passes` on problem import.

meisterT:
Should be an int >=2 as per spec, see
https://icpc.io/problem-package-format/spec/2025-09.html?highlight=nextpass#multi-pass-validation

### DIFF
--- a/webapp/tests/Unit/Service/ImportProblemServiceTest.php
+++ b/webapp/tests/Unit/Service/ImportProblemServiceTest.php
@@ -5,6 +5,7 @@ namespace App\Tests\Unit\Service;
 use App\Entity\Problem;
 use App\Service\ImportProblemService;
 use App\Tests\Unit\BaseTestCase;
+use PHPUnit\Framework\Attributes\DataProvider;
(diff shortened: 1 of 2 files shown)


webapp/tests/Unit/Service/ImportProblemServiceTest.php
```diff
@@ -307,6 +308,42 @@ public function testMultipassLimit(): void
         $this->assertEquals(7, $problem->getMultipassLimit());
     }
 
+    /**
+     * @return array<string, array{string}>
```
vmcj: That shape doesn't look correct.
```suggestion
     * @return array<string, string[]>
```
or
```suggestion
     * @return array<string, array<string>>
```

vmcj (approved):
I don't see it in the linked page but do agree the spec should require the limit to be 2<=limit

meisterT:
Because I wanted to link to https://icpc.io/problem-package-format/spec/2025-09.html?highlight=nextpass#multi-pass-limit, will adjust the commit message later

PR #3689, opened 2026-07-19
runpipe: Properly pass -U to runguard.

meisterT:
runpipe tells runguard its pid with `-U`, so that runguard can raise `SIGUSR1` when the command hits its hard wall-time limit. However, this broke in a previous refactoring.

The option was appended to the argument list, but the judgedaemon already ends runguard's options with a `--` before the program path. The option therefore landed behind that separator, so runguard never received it, and `-U <pid>` were appended to the contestant's program as two stray arguments.

Consequently `runpipe_pid` stayed unset and the `SIGUSR1` was never sent, so runpipe could not distinguish a timed-out command from a normal exit. It then recorded `validator-exited-first` based on whichever process `waitpid()` happened to reap first, which feeds into verdict determination.

### DIFF
--- a/judge/runpipe.cc
+++ b/judge/runpipe.cc
@@ -236,11 +236,22 @@ struct process_t {
     std::array<int, 3> stdio = {stdin_fd, stdout_fd, FDREDIR_NONE};
 
     auto exec_args = args;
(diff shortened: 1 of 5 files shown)


vmcj (approved):
Not an expert on this, but LGTM.

PR #3691, opened 2026-07-24
Add another clarification from another team in similar style for unit tests

vmcj:
We check if we can react to our own (that was there before) and that an user can not react to the clarification of others but also can't guess that it exists.

This reuses some other fixtures to prevent the duplication of creation of teams. This way we can now easily reuse the teams from other fixtures with associated users each time extending those more and more to almost real-world usecases.

### DIFF
--- a/webapp/src/DataFixtures/Test/ClarificationFixture.php
+++ b/webapp/src/DataFixtures/Test/ClarificationFixture.php
@@ -6,9 +6,10 @@ use App\Entity\Contest;
 use App\Entity\Clarification;
 use App\Entity\Problem;
 use App\Entity\Team;
+use Doctrine\Common\DataFixtures\DependentFixtureInterface;
 use Doctrine\Persistence\ObjectManager;
 
-class ClarificationFixture extends AbstractTestDataFixture
+class ClarificationFixture extends AbstractTestDataFixture implements DependentFixtureInterface
 {
     public function load(ObjectManager $manager): void
     {
@@ -16,6 +17,8 @@ class ClarificationFixture extends AbstractTestDataFixture
         $contest = $manager->getRepository(Contest::class)->findOneBy(['shortname' => 'demo']);
         /** @var Team $team */
         $team = $manager->getRepository(Team::class)->findOneBy(['name' => 'Example teamname']);
+        /** @var Team $anotherTeam */
+        $anotherTeam = $manager->getRepository(Team::class)->findOneBy(['externalid' => 'nav-alpha']);
         /** @var Problem $problem */
         $problem = $manager->getRepository(Problem::class)->findOneBy(['externalid' => 'hello']);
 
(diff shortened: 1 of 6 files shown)


webapp/tests/Unit/Controller/API/ClarificationControllerTest.php
```diff
@@ -122,8 +145,18 @@ public function testAddNoAccess(): void
      * Test that if invalid data is supplied, the correct message is returned.
      */
     #[DataProvider('provideAddInvalidData')]
-    public function testAddInvalidData(string $user, array $dataToSend, string $expectedMessage): void
+    public function testAddInvalidData(string $user, array $dataToSend, string $expectedMessage, ?string $id = null): void
     {
+        if ($id) {
```
meisterT: does this ever run? only runs with `'0'`, right, so evaluates to false

webapp/src/DataFixtures/Test/ClarificationFixture.php
```diff
@@ -6,16 +6,19 @@
...
-class ClarificationFixture extends AbstractTestDataFixture
+class ClarificationFixture extends AbstractTestDataFixture implements DependentFixtureInterface
 {
     public function load(ObjectManager $manager): void
     {
         /** @var Contest $contest */
         $contest = $manager->getRepository(Contest::class)->findOneBy(['shortname' => 'demo']);
         /** @var Team $team */
         $team = $manager->getRepository(Team::class)->findOneBy(['name' => 'Example teamname']);
+        /** @var Team $team */
```
meisterT: s/team/anotherTeam/?

webapp/tests/Unit/Controller/Jury/ImportExportControllerTest.php
```diff
@@ -244,11 +244,11 @@ public function testResultsTsvExport(
...
     {
-        yield [0, true, true, "results	1\nexteam		Honorable	0	0	0	\n"];
-        yield [0, true, false, "results	1\nexteam		Honorable	0	0	0	\n"];
-        yield [0, false, true, "results	1\nexteam		Honorable	0	0	0	\n"];
-        yield [0, false, true, "results	1\nexteam		Honorable	0	0	0	\n"];
-        yield [1, true, true, "results	1\n"];
+        yield [0, true, true,  "results	1\nexteam		Honorable	0	0	0	\nnav-alpha		Honorable	0	0	0	\nnav-beta		Honorable	0	0	0	\nnav-gamma		Honorable	0	0	0	\n"];
+        yield [0, true, false, "results	1\nexteam		Honorable	0	0	0	\nnav-alpha		Honorable	0	0	0	\nnav-beta		Honorable	0	0	0	\nnav-gamma		Honorable	0	0	0	\n"];
+        yield [0, false, true, "results	1\nexteam		Honorable	0	0	0	\nnav-alpha		Honorable	0	0	0	\nnav-beta		Honorable	0	0	0	\nnav-gamma		Honorable	0	0	0	\n"];
+        yield [0, false, true, "results	1\nexteam		Honorable	0	0	0	\nnav-alpha		Honorable	0	0	0	\nnav-beta		Honorable	0	0	0	\nnav-gamma		Honorable	0	0	0	\n"];
```
meisterT, line 250: is the idea to have `1, false, true` here?

webapp/src/DataFixtures/Test/ClarificationFixture.php
vmcj:
```suggestion
        /** @var Team $anotherTeam */
```

webapp/tests/Unit/Controller/API/ClarificationControllerTest.php
vmcj:
```suggestion
        if (!is_null($id)) {
```

webapp/tests/Unit/Controller/Jury/ImportExportControllerTest.php
vmcj: This was only to fix the existing test. I didn't have an intention to extend on this test.

Also that case is on L253 already I think?

webapp/tests/Unit/Controller/API/ClarificationControllerTest.php
vmcj: Changed the code to check more explicit for it.

meisterT (approved):

PR #3692, opened 2026-07-24
Extract functions about authorized user from DOMJudgeService

vmcj:
Needed for https://github.com/DOMjudge/domjudge/pull/3657 but an improvement in general IMHO.

Done in multiple commits to make it easier to follow what's happening, I plan to squash them after approval.

### DIFF
--- a/webapp/src/Controller/API/AccessController.php
+++ b/webapp/src/Controller/API/AccessController.php
@@ -77,22 +77,22 @@ class AccessController extends AbstractApiController
         ];
 
         // Add files to submissions if allowed
-        if ($this->dj->checkrole('api_source_reader')) {
+        if ($this->authService->checkRole('api_source_reader')) {
             $submissionsProperties[] = 'files';
         }
 
         $capabilities = [];
 
         // Add capabilities
-        if ($this->dj->checkrole('api_writer')) {
+        if ($this->authService->checkRole('api_writer')) {
             $capabilities[] = 'contest_start';
             $capabilities[] = 'contest_thaw';
         }
-        if ($this->dj->checkrole('team') && $this->dj->getUser()->getTeam()) {
+        if ($this->authService->checkRole('team') && $this->authService->getUser()->getTeam()) {
             $capabilities[] = 'team_submit';
             $capabilities[] = 'team_clar';
         }
-        if ($this->dj->checkrole('api_writer')) {
+        if ($this->authService->checkRole('api_writer')) {
             $capabilities[] = 'proxy_submit';
             $capabilities[] = 'proxy_clar';
             $capabilities[] = 'admin_submit';
--- a/webapp/src/Controller/API/AccountController.php
+++ b/webapp/src/Controller/API/AccountController.php
@@ -94,7 +94,7 @@ class AccountController extends AbstractRestController
         // Get the contest ID to make sure the contest exists.
         $this->getContestId($request);
 
-        $user = $this->dj->getUser();
+        $user = $this->authService->getUser();
         if ($user === null) {
             throw new NotFoundHttpException();
         }
--- a/webapp/src/Controller/API/JudgementController.php
+++ b/webapp/src/Controller/API/JudgementController.php
@@ -7,6 +7,7 @@ use App\Entity\AbstractJudgement;
 use App\Entity\Contest;
 use App\Entity\ExternalJudgement;
 use App\Entity\Judging;
+use App\Service\AuthorizedUserService;
 use App\Service\ConfigurationService;
 use App\Service\DOMJudgeService;
 use App\Service\EventLogService;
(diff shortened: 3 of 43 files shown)


webapp/src/Service/DOMJudgeService.php
meisterT, line 1: What's the idea - do you want to squash commits or merge as is? Asking because this commit is breaking stuff

vmcj: Squash into a single commit. I only did it in multiple commits to make it easier to follow.

webapp/src/EventListener/ProfilerDisableListener.php
```diff
@@ -12,16 +13,17 @@
 readonly class ProfilerDisableListener
 {
     public function __construct(
-        protected KernelInterface $kernel,
-        protected DOMJudgeService $dj,
-        protected ?Profiler       $profiler
+        protected AuthorizedUserService $authService,
+        protected KernelInterface       $kernel,
+        protected DOMJudgeService       $dj,
```
meisterT: this is not used anymore, is it?

webapp/src/Service/DOMJudgeService.php
```diff
@@ -105,12 +105,12 @@ class DOMJudgeService
     ];
 
     public function __construct(
+        protected readonly AuthorizedUserService $authService,
         protected readonly EntityManagerInterface $em,
         protected readonly BalloonService $balloonService,
         protected readonly LoggerInterface $logger,
         protected readonly RequestStack $requestStack,
         protected readonly ParameterBagInterface $params,
-        protected readonly AuthorizationCheckerInterface $authorizationChecker,
```
meisterT, line 112: this was the last use, remove the import?

webapp/src/Service/DOMJudgeService.php
```diff
@@ -247,45 +247,11 @@ public function getProblem(?int $probid): ?Problem
...
-        }
-        return $this->authorizationChecker->isGranted('ROLE_' . strtoupper($rolename));
-    }
-
     public function getClientIp(): string
     {
         return $this->requestStack->getMainRequest()->getClientIp();
     }
 
-    public function getUser(): ?User
```
meisterT, line 270: this was the last use of User, remove the import?

webapp/src/Controller/API/AccessController.php
```diff
@@ -4,6 +4,10 @@
 
 use App\DataTransferObject\Access;
 use App\DataTransferObject\AccessEndpoint;
+use App\Service\ConfigurationService;
```
meisterT: Where are we using all of these new imports?

webapp/src/Controller/API/AccountController.php
```diff
@@ -3,6 +3,10 @@
 namespace App\Controller\API;
 
 use App\Entity\User;
+use App\Service\ConfigurationService;
```
meisterT: Same here, where are we using these imports?

webapp/src/Service/AuthorizedUserService.php
```diff
@@ -0,0 +1,49 @@
+<?php declare(strict_types=1);
+
+namespace App\Service;
+
+use App\Entity\User;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
+
+class AuthorizedUserService
```
meisterT, line 9: can/should be marked as readonly?

webapp/src/Controller/API/AccessController.php
vmcj: We don't anymore, I first needed a constructor but solved that differently in the end.

webapp/src/Service/AuthorizedUserService.php
vmcj: I think doing that breaks the unit tests.

vmcj: https://github.com/DOMjudge/domjudge/actions/runs/30334936555/job/90197701180?pr=3692

So how important do we find the `readonly` for this class? I can investigate how to fix the Unit tests but wonder if it's worth my time.

vmcj: So I checked, all other services in that test are also not `readonly` so if we want this we should do it for all services.

webapp/src/Controller/PublicController.php
meisterT, line 1: looks like there are no meaningful changes to this file and all changes can be reverted?

webapp/src/Controller/API/JudgementController.php
```diff
@@ -42,12 +43,13 @@ class JudgementController extends AbstractRestController implements QueryObjectT
     private bool $useExternalJudgements = false;
 
     public function __construct(
-        EntityManagerInterface $entityManager,
-        DOMJudgeService $DOMJudgeService,
+        AuthorizedUserService  $authService,
```
meisterT: nit: double space

webapp/src/Controller/API/SubmissionController.php
```diff
@@ -49,13 +50,14 @@
 class SubmissionController extends AbstractRestController
 {
     public function __construct(
-        EntityManagerInterface $entityManager,
+        AuthorizedUserService  $authService,
```
meisterT: nit: double space

meisterT (approved):
Three nits, otherwise good to be merged

Issue #645, opened 2019-10-27
Changing verified required config may lose some event-feed.


2014CAIS01:
In an official contest, at the first we want to use `verification_required` config to ensure every problem is ok and every submission is ok. But after one or two hour, some easy problems are solved by many team, and we want to judge automaticaly, so we close `verification_required`, and refresh the rank cache. But, the submissions juding finished but unverified at that moment haven't update the event, (CDS shows the related submissions pending). This issue may affect some tools like ICPC resolver or CDS using event-feed.(Domjudge 7.0.4)

In addation, I also suggest that `verification_required` can be set for a specific problem.

eldering:
Indeed, that's an edge case we have overlooked.

eldering:
For the feature request: it's better to report that as a separate issue. I think that's low priority though.

PR #3695, opened 2026-07-28
Allow verification of submissions on (contest)problem level

vmcj:
This allows for checking of the results and already allowing for some problems when enough submissions have come in.

As suggested by @mkfuron in https://domjudge-org.slack.com/archives/CHHURFUM7/p1779496255283309

That discussion also had other options by @mpsijm which I think are different ways of looking at the problem so I suspect we should support both (or neither?).

Also added some minor fixes:
image

### DIFF
--- a/webapp/src/Controller/Jury/ProblemController.php
+++ b/webapp/src/Controller/Jury/ProblemController.php
@@ -1274,7 +1274,7 @@ class ProblemController extends BaseController
     }
 
     #[IsGranted('ROLE_ADMIN')]
-    #[Route(path: '/{contestId}/{probId}/toggle/{type<judge|submit>}', name: 'jury_problem_toggle')]
+    #[Route(path: '/{contestId}/{probId}/toggle/{type<judge|submit|verification>}', name: 'jury_problem_toggle')]
     public function toggleSubmitAction(
         RouterInterface $router,
         Request $request,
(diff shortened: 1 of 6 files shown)


cubercsl:
BTW, this feature was brought up in #645 before, so I think someone might switch the related settings during the contest. So I think we need to test this kind of scenario.

mpsijm:
Thanks for pulling me in here! It looks like there are two separate concerns:
- This PR currently appears to move the global setting "Verification required – Is manual verification of judgings by jury required before publication?" to the problem level. This setting keeps submissions as "pending" as long as they are not manually verified.
- During BAPC/NWERC, we always have this setting off, but still manually verify submissions _after_ their result has already become visible (so, moving this setting to be per-problem will probably not change anything for us). However, this becomes more difficult to manage once the easier problems have been solved many times. My suggestions in https://domjudge-org.slack.com/archives/CHHURFUM7/p1779306901059579 (the message before Marc's) were ideas for improvements in this scenario :slightly_smiling_face:

vmcj:
>     * This PR currently appears to move the global setting "Verification required – Is manual verification of judgings by jury required before publication?" to the problem level. This setting keeps submissions as "pending" as long as they are not manually verified.

Yes, I think Marc handles this different, so I first wanted to implement his suggestion and address your suggestions in a future PR as it requires a bit more work to do it cleanly/performant.

And it shouldn't move it, it should add an additional option to do it either globally or for specific problems.

PR #3702, opened 2026-08-16
Fix Yarn dependency installation recipes

HeRaNO:
The node dependency recipes used `${ ... }` to group their shell commands. GNU Make treats `${foo}` as a variable reference [1].

It therefore parsed the whole block as an undefined variable and expanded it to an empty command, so both dependency targets returned successfully without running Yarn.

This left `node_modules` absent and caused symlinked frontend assets such as Monaco and jQuery to return 404 responses.

Use `@{ ... }` to execute the commands as a shell group. GNU Make uses the leading `@` to suppress echoing and discards it before passing the recipe to the shell [2]. This ensures that development and distribution dependency installs run Yarn and perform their respective license checks.

[1]: https://www.gnu.org/software/make/manual/html_node/Reference.html
[2]: https://www.gnu.org/software/make/manual/html_node/Echoing.html

### DIFF
--- a/webapp/Makefile
+++ b/webapp/Makefile
@@ -63,14 +63,14 @@ fi
 endef
 
 node-dependencies:
-	${ \
+	@{ \
 		$(detect_yarn); \
 		$$YARN_CMD install; \
 		$(yarn_licenses_warn); \
 	}
 
 node-dependencies-dist:
-	${ \
+	@{ \
 		$(detect_yarn); \
 		$$YARN_CMD install; \
 		$(yarn_licenses_strict); \


vmcj:
You're probably right but could you link to the relevant part in the docs? It worked the last time I checked so I prefer to only change this when it makes sense from the docs.

Also you commit message seems to have some duplication.

HeRaNO:
The commit message is updated.

I found this issue when I was testing my branch locally with the `domjudge-contributor` image.

```bash
sudo docker network create dj-network
sudo docker run -d --net="dj-network" --hostname mariadb --name dj-mariadb -e MYSQL_ROOT_PASSWORD=rootpw -e MYSQL_USER=domjudge -e MYSQL_PASSWORD=djpw -e MYSQL_DATABASE=domjudge mariadb:10 --max-connections=1000 --max-allowed-packet=1g --innodb-log-file-size=4g
sudo docker run -d --net="dj-network" --privileged --cgroupns=host -v [path-to-domjudge-checkout]:/domjudge -e MYSQL_HOST=mariadb -e MYSQL_USER=domjudge -e MYSQL_DATABASE=domjudge -e MYSQL_PASSWORD=djpw -e MYSQL_ROOT_PASSWORD=rootpw -p 12345:80 --name domjudge domjudge/domjudge-contributor
```

And I think it can be easily reproduced on the main branch, too.

vmcj (approved):
I did a clean checkout and agree something was not working. Based on the GNU docs this seems indeed an acceptable fix.

PR #3708, opened 2026-08-29
Restore clarification visibility lost in the QueryBuilder extraction

meisterT:
Extracting the clarification QueryBuilder into ClarificationService moved three decisions into one place, and each of them regressed a caller.

The role check only exempted `api_reader` and `judgehost`. But the jury clarification pages are guarded by `ROLE_CLARIFICATION_RW`, which is a role of its own ("Clarification handler"): it is implied by `jury`, yet it can also be granted on its own, and then it implies neither `jury` nor `api_reader`. Such a user fell through to the public-only branch, so `/jury/clarifications` showed no team requests at all -- the one thing the role exists for -- and the `jury_member` lookup, which uses `getSingleResult()`, threw a NoResultException on the detail page.

The `jury_member` lookup in the jury controller used to select on `clar.clarid`, which is globally unique. It now selects on the external id, which is only unique per `(cid, externalid)`, without passing a contest, so a second contest reusing an external id turns it into a NonUniqueResultException. External ids are derived from the internal id unless they are set explicitly, so this needs an external CCS or the API to trigger, but both do exactly that. Pass the contest at both call sites; the form type gets a `contestId` option for this, which the compose action passes as well even though it has no clarification to look up.

The two team pages dropped `recipient = :team OR recipient IS NULL` and relied on the query builder instead. That works for teams, but jury and admin users are exempt from the restriction, so a jury member with a team saw the clarifications addressed to every other team on their own team page. Rather than repeat the condition on every team facing page, the query builder grows an `onlyForRecipientTeam` argument, which is what getTwigDataForProblemsAction() was open coding already.

While here, take the clarification id and the contest id out of the form type's instance state. Form types are shared services, so options belonging to one form must not be stashed on `$this`; hand them to the validation callback through the closure instead.

Also drop the Expr\Join and Clarification imports that the extraction left unused.

### DIFF
--- a/webapp/src/Service/ClarificationService.php
+++ b/webapp/src/Service/ClarificationService.php
@@ -76,9 +76,21 @@ readonly class ClarificationService
             ->getResult();
     }
 
+    /**
+     * Build a query for the clarifications the current user is allowed to see.
+     *
+     * Team facing pages should pass `onlyForRecipientTeam` so that they keep
+     * showing what the team sees: jury and admin users are exempt from the
+     * visibility restriction below, and would otherwise see the clarifications
+     * addressed to every other team on their own team pages. A null
+     * `recipientTeamId` then limits the result to clarifications sent to
+     * everyone.
+     */
     public function getQueryBuilder(?string $externalContestId = null, ?int $internalContestId = null,
                                     ?string $externalClarificationId = null,
-                                    ?string $problem = null
+                                    ?string $problem = null,
+                                    bool $onlyForRecipientTeam = false,
+                                    ?int $recipientTeamId = null
     ): QueryBuilder {
         $queryBuilder = $this->em->createQueryBuilder()
             ->from(Clarification::class, 'clar')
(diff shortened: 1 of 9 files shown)


webapp/src/Service/ClarificationService.php
```diff
@@ -119,9 +138,9 @@ public function getQueryBuilder(?string $externalContestId = null, ?int $interna
             }
         }
 
-        if (!$this->authService->checkRole('api_reader')) {
+        if (!$isClarificationStaff) {
             $queryBuilder
-                // For non-API-reader users, only expose the problems after the contest has started.
+                // For non-staff users, only expose the problems after the contest has started.
```
vmcj, line 143: Not needed for this PR, but I think the original goal of the clarification_rw user was only for general clarifications. so it should be limited to not see the precontest clarifications about problems.

vmcj (approved):
Fine with the changes after fixing CI.

PR #3709, opened 2026-08-29
Fix clarification query semantics

meisterT:


webapp/src/Controller/API/ClarificationController.php
```diff
@@ -297,6 +298,11 @@ protected function getQueryBuilder(Request $request): QueryBuilder
         $problem = $request->query->get('problem');
         return $this->clarificationService->getQueryBuilder(internalContestId: $contestId, problem: $problem)
             ->select('clar, c, r, reply, p')
+            // `cp` is left joined on the clarification's problem, so it is only
+            // set when that problem is part of this contest. Never expose a
+            // clarification pointing at a problem the contest does not have:
+            // its `problem_id` would dangle in the feed.
+            ->andWhere('clar.problem IS NULL OR cp.problem IS NOT NULL')
```
vmcj: The whole reason of moving everything to the service is to prevent these technical where clauses. Do you have a test which shows when this is needed so we can see if we can do this in the service itself?

With this implementation we have to check this again for all controllers retrieving clarifications.

meisterT: Thanks for pushing back, I agree and moved it into the srevice, so the API controller is back to just a select and an order by.

nickygerritsen (approved):

vmcj (approved):

PR #3710, opened 2026-08-29
Render every hex notation the color parser accepts.

meisterT:
`hexColorToRGBA` pulled the channels out of the hex string by grabbing two digit pairs with `preg_match_all`, which only lines up with one of the three notations `Utils::convertToHex()` returns. That helper accepts one, two or three hex digits per channel (`#abc`, `#aabbcc`, `#aaabbbccc`), so:

*  `#abc`: one pair matched, no case in the switch, the colour was
               returned unchanged and the gradient built from it in
               scoreboard_category_color.css.twig silently did not render
 * `#aaabbbccc`   four pairs matched, so the channels were read as `aa`, `ab`
               and `bb` -- shifted by a digit -- and the leftover `cc` was
               taken for an opacity, as a 0..255 number where CSS wants
               0..1, which then also overrode the opacity the caller asked
               for

Only `#aabbcc` came out right. Use `Utils::parseHexColor()`, which already scales all three notations to 0..255 and is what the rest of the colour handling uses.

This drops the fourth channel as an opacity. Nothing could reach it as one: `convertToHex()` rejects `#RRGGBBAA` -- as its tests assert -- so a fourth pair only ever came from misreading the nine digit form. Supporting CSS alpha notations would mean widening that helper first.

### DIFF
--- a/webapp/src/Twig/TwigExtension.php
+++ b/webapp/src/Twig/TwigExtension.php
@@ -1073,26 +1073,10 @@ EOF;
         if (is_null($col)) {
             return $text;
         }
-        $ret = preg_match_all("/[0-9A-Fa-f]{2}/", $col, $m);
-        if (!$ret) {
-            return $text;
-        }
 
-        $m = current($m);
-        switch (count($m)) {
-            case 4:
-                // We also have opacity; load that and use
-                // RGB of case 3
-                $opacity = hexdec(array_pop($m));
-                // no-break
-            case 3:
-                $vals   = array_map(hexdec(...), $m);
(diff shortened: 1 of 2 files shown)


vmcj (approved):

vmcj:
> This drops the fourth channel as an opacity. Nothing could reach it as one: `convertToHex()` rejects `#RRGGBBAA` -- as its tests assert -- so a fourth pair only ever came from misreading the nine digit form. Supporting CSS alpha notations would mean widening that helper first.

The spec dropped support for alpha notations, so I would prefer not to introduce it into our code.

PR #3714, opened 2026-08-30
Allow setting webserver alias in configure step

vmcj:
This makes it easier to configure an installation from scripts without copying the webserver config files.

I needed this for a side-project, as it's trivial enough I think it makes sense to merge this and make scripted installs easier.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -265,6 +265,19 @@ else
 fi
 AC_MSG_RESULT($BASEURL)
 
+AC_MSG_CHECKING([alias])
+AC_ARG_WITH([alias], [AS_HELP_STRING([--with-alias=alias],
+[Base path of the DOMjudge web interfaces. Example default:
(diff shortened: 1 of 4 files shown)


meisterT:
Ha, I recently added something similar that I planned to upstream, but didn't have time yet. Will check tonight how compatible the two approaches are and then comment (https://github.com/DOMjudge/domjudge/compare/main...meisterT:domjudge:multi-instance-support?expand=1 if you want to have a look already).

vmcj:
> Ha, I recently added something similar that I planned to upstream, but didn't have time yet. Will check tonight how compatible the two approaches are and then comment (https://github.com/DOMjudge/domjudge/compare/main...meisterT:domjudge:multi-instance-support?expand=1 if you want to have a look already).

I think your solution is much fancier, but also starts to connect things which are not always connected IMHO. Probably your solution is better overall although it will break my usecase but I can probably workaround that.

PR #3715, opened 2026-08-30
Use RelTime for relative time fields in new API format, allow parsing it

### LINKED ISSUE #2946: Penalty time parsing from external data source in new draft spec

vmcj:
Fixes #2946

@nickygerritsen this re-orders a couple of your commits to do 1 thing in 1 commit. If you agree I can pick the next ones which probably don't need that much changes.


webapp/src/Service/EventLogService.php
```diff
@@ -851,4 +852,27 @@ public function endpointForEntity($entity): ?string
...
+     *
+     * @return array<string,mixed>
+     */
+    public function applyCcsVersionChanges(string $endpointType, array $event): array
+    {
+        /** @var CcsApiVersion $ccsApiVersion */
+        $ccsApiVersion = $this->config->get('ccs_api_version');
+
+        if ($endpointType == 'contests') {
+            $penaltyTime = $event['penalty_time'];
```
meisterT, line 867: does the `penalty_time` always exist?

webapp/src/NelmioApiDocBundle/JMSModelDescriber.php
```diff
@@ -0,0 +1,84 @@
...
+        array $propertiesToSet,
+        CcsApiVersion $ccsApiVersion,
+    ): void {
+        foreach ($schema->properties as $property) {
+            if (!in_array($property->property, $propertiesToSet, true)) {
+                continue;
+            }
+
+            if ($ccsApiVersion->useRelTimes()) {
+                $property->type = ' string';
```
meisterT, line 74: curious if the space before string (and integer below) is intentional?

webapp/src/Controller/API/ScoreboardController.php
```diff
@@ -214,7 +220,9 @@ public function getScoreboardAction(
                 } else {
                     $problem->firstToSolve = $matrixItem->isCorrect && $scoreboard->solvedFirst($teamScore->team, $contestProblem);
                     if ($matrixItem->isCorrect) {
-                        $problem->time = Utils::scoretime($matrixItem->time, $scoreIsInSeconds);
+                        $problemTime = Utils::scoretime($matrixItem->time, $scoreIsInSeconds);
+                        $problem->time = $this->formatTime($problemTime, $ccsApiVersion, $scoreIsInSeconds);
+                        $lastProblemTime = max($lastProblemTime, $problemTime);
```
meisterT, line 225: this might be problematic if anyone solves a problem in the 0th minute: `max(null, 0) => null`

webapp/src/Controller/API/ScoreboardController.php
```diff
@@ -239,4 +251,16 @@ public function getScoreboardAction(
...
         return $results;
     }
+
+    protected function formatTime(
+        int $time,
+        CcsApiVersion $ccsApiVersion,
+        bool $scoreIsInSeconds
+    ): int|string {
+        if ($ccsApiVersion->useRelTimes()) {
+            return $scoreIsInSeconds ? Utils::relTime($time) : Utils::relTime($time * 60);
```
meisterT, line 261: I wonder whether you want `, floored: true` here?

webapp/src/Controller/Jury/JuryMiscController.php
meisterT, line 55: indexAction no longer needs ConfigurationService argument as you inject it via constructor

webapp/src/DataTransferObject/Scoreboard/Score.php
```diff
@@ -5,17 +5,19 @@
 use App\Controller\API\AbstractRestController as ARC;
 use JMS\Serializer\Annotation as Serializer;
 
-readonly class Score
```
meisterT, line 8: Why is no longer `readonly`?

meisterT (commented):
Some minor comments, but probably @nickygerritsen should do the main review.

---

How do existing installations behave, do they get upgraded or marked as one specific version?

PR #3716, opened 2026-08-30
Display Sample testcase group in testcases view

vmcj:
Further extraction from https://github.com/DOMjudge/domjudge/pull/3623

Before
image
After:
image


meisterT:
I assume you need to add `ignore_sample` to `grader_flags` in the the example dirs `data/testdata.yaml` to fix the integration tests, otherwise good to go

PR #3720, opened 2026-08-30
Use the configured base URL for the REST API credentials

meisterT:
`etc/genrestapicredentials` always wrote `http://localhost/domjudge/api`,
ignoring `--with-baseurl` entirely. Every installation not reachable under
that exact URL has to hand-edit `etc/restapi.secret` after running `make`
before a judgedaemon can talk to the domserver.

Derive the API URL from the base URL instead. When no base URL was
configured the historic default is kept, so the generated file is unchanged
for that case rather than pointing at the `example.com` placeholder.

Two unrelated documentation fixes noticed along the way are included as
separate commits; happy to split them out if preferred:

- the manual instructions printed after `make maintainer-install` did not
  match what `make inplace-postinstall-nginx` actually does — the nginx
  symlink had no destination file name, and the PHP-FPM source file was
  missing its `.conf` extension;
- `webapp/.env.local` is created by `maintainer-conf`, not
  `maintainer-install`.


meisterT:
@vmcj merging this before https://github.com/DOMjudge/domjudge/pull/3714 should make it easier to discuss the differences between our approaches (and how to align them)

vmcj (approved):

PR #3721, opened 2026-08-30
Prevent metadata injection through compiler output

meisterT:
Entry point detection matched the detection message anywhere in a line and stripped everything up to the last "etected ". Since compiler output is under submitter control, a crafted line could write an arbitrary key/value pair into compile.meta, e.g. an 'internal-error' that disables the judgehost.

Anchor the match at the start of the line and only copy the detected value, always under the 'entry_point' key. The added submission fakes a compile timeout this way and was judged as COMPILER-ERROR before.

Thanks to teddyctf for finding and reporting this issue.

### DIFF
--- a/judge/compile.sh
+++ b/judge/compile.sh
@@ -181,8 +181,15 @@ fi
 
 # Check if the compile script auto-detected the entry point, and if
 # so, store it in the compile.meta for later reuse, e.g. in a replay.
-ENTRY_POINT_REGEX='[Dd]etected entry_point: '
-grep "$ENTRY_POINT_REGEX" compile.tmp | sed 's/^.*etected //' >>compile.meta
+# Only lines that start with the detection message are considered and only
+# the detected value is copied over, prefixed with the 'entry_point' key.
+# This prevents compiler output, which is under control of the submitter,
+# from injecting arbitrary keys into the metadata. When changing the regex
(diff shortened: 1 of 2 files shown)


judge/compile.sh
```diff
@@ -181,8 +181,13 @@ fi
...
 # so, store it in the compile.meta for later reuse, e.g. in a replay.
-ENTRY_POINT_REGEX='[Dd]etected entry_point: '
-grep "$ENTRY_POINT_REGEX" compile.tmp | sed 's/^.*etected //' >>compile.meta
+# Only lines that start with the detection message are considered and only
+# the detected value is copied over, prefixed with the 'entry_point' key.
+# This prevents compiler output, which is under control of the submitter,
+# from injecting arbitrary keys into the metadata.
+ENTRY_POINT_REGEX='^(Info: )?[Dd]etected entry_point: '
+grep -E "$ENTRY_POINT_REGEX" compile.tmp | \
+	sed -E "s@$ENTRY_POINT_REGEX@entry_point: @" >>compile.meta
```
vmcj, line 192: Consider adding a filereference comment to that submission to tell us to update that one when we update the detection logic.

meisterT: Done

nickygerritsen (approved):

PR #3725, opened 2026-08-31
Add multi-instance support

meisterT:


meisterT:
@vmcj this is the second half of my change

doc/manual/develop.rst
```diff
@@ -74,6 +74,82 @@ Note that these targets have to be executed *separately* and
...
+
+Running several checkouts side by side
+--------------------------------------
+When working on multiple branches at the same time, for example using
+``git worktree`` or ``jj workspace``, each checkout can have its own
+in-place installation. No extra flags are needed: ``make maintainer-conf``
+derives an *instance name* from the name of the directory the source tree
+lives in, and uses it for
+
+* the webserver and PHP-FPM configuration file names installed into
```
nickygerritsen, line 87: I might have missed it, but shouldn't we also name the nginx/apache config files something with the instance in there when installing?

i.e. instead of `/etc/nginx/snippets/domjudge-inner.conf` (or something we did) we do `/etc/nignx/snippets/domjudge-inner-{instance}.conf`?

.github/jobs/configure-checks/shared_post.bats
```diff
@@ -186,6 +186,145 @@ source .github/jobs/configure-checks/functions.sh
...
+  run run_configure
+  assert_line "checking URL path prefix... /domjudge/"
+  run run_configure "--with-baseurl=https://contest.example.org/dj/"
+  assert_line "checking URL path prefix... /dj/"
+  # A base URL without a path serves DOMjudge from the root.
+  run run_configure "--with-baseurl=https://contest.example.org/"
+  assert_line "checking URL path prefix... /"
+  # An unparseable base URL is rejected.
+  run run_configure "--with-baseurl=not-a-url"
+  assert_line "checking URL path prefix... configure: error: could not parse base URL 'not-a-url'; expected e.g. 'https://example.com/domjudge/'."
```
vmcj:
```suggestion
  assert_line "checking URL path prefix... configure: error: could not parse base URL 'not-a-url'; expected e.g. 'https://example.org/domjudge/'."
```

I like consistency.

doc/manual/develop.rst
meisterT: All installed files in `/etc` are named after the instance:
* `/etc/nginx/sites-enabled/${INSTANCE}.conf` (or in `/etc/nginx/conf.d/`)
* `/etc/apache2/conf-available/${INSTANCE}.conf`
* `/etc/php/<version>/fpm/pool.d/${INSTANCE}-fpm.conf`
* `/etc/sudoers.d/${INSTANCE}`

We don't install `nginx-conf-inner` into `/etc/nginx/snippets/` (or anywhere in `/etc`) at all.

Instead, each instance's outer `nginx-conf` includes its own inner file directly from `@domserver_etcdir@/nginx-conf-inner` (pointing to that checkout's `etc/nginx-conf-inner` for in-place installs, or `<prefix>/etc/nginx-conf-inner` for packaged installs). So each instance already has its own isolated inner snippet and there should not be sharing/conflicts in `/etc`.

PR #3729, opened 2026-08-31
Don't leak extracted LaTeX between markdown conversions

meisterT:
The LaTeX taken out of a document before converting it was stored on the shared extension instead of the call, so a placeholder lost during conversion left its LaTeX behind for the next document rendered in the same request.

### DIFF
--- a/webapp/src/Twig/TwigExtension.php
+++ b/webapp/src/Twig/TwigExtension.php
@@ -41,11 +41,6 @@ use Twig\Extra\Markdown\MarkdownRuntime;
 
 class TwigExtension
 {
-    /**
-     * @var array<string>
-     */
-    private array $latexFound;
-
     public function __construct(
         protected readonly DOMJudgeService $dj,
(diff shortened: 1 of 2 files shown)


nickygerritsen (approved):

webapp/src/Twig/TwigExtension.php
```diff
@@ -1368,7 +1365,12 @@ function (array $matches) use ($latexPlaceholder): string {
 
         return preg_replace_callback(
             "/$latexPlaceholder/",
-            fn(): string => array_shift($this->latexFound), $markdown
+            // A placeholder the conversion dropped leaves one behind, so do not assume there
```
vmcj, line 1368: I'm probably stupid but if I don't know I'll not know in the future.. which conversions do we drop but not return `''` so can still replace the `$placeholder` with?

meisterT: Update the comment in the code, that was poorly phrased.

PR #3732, opened 2026-08-31
Fix cascading delete 

meisterT:


nickygerritsen (approved):

meisterT:
Before merging I need to think though the unit tests and think whether it needs an update or points to an issue in the new code

Issue #2848, opened 2024-11-23
Compatibility broken with newer MariaDB versions / `--innodb-snapshot-isolation=true`


meisterT:
MariaDB version 11.6.2 (and perhaps other versions) set `--innodb-snapshot-isolation=true` which causes multiple queries in DOMjudge to fail. 

https://github.com/DOMjudge/domjudge/pull/2843 is likely part of a fix, perhaps https://github.com/DOMjudge/domjudge/pull/2844 also helps

Some more references:
- https://mariadb.com/resources/blog/isolation-level-violation-testing-and-debugging-in-mariadb/
- https://jira.mariadb.org/browse/MDEV-26642?orderby=created+DESC%2C+priority+DESC%2C+updated+DESC
- https://jira.mariadb.org/browse/MDEV-35124?orderby=created+DESC%2C+priority+DESC%2C+updated+DESC
- https://jepsen.io/analyses/mysql-8.0.34

meisterT:
Examples how this can manifest itself:
- `An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry \'1541319\' for key \'PRIMARY\'',`
- `An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1020 Record has changed since last read in table 'judgetask'`
- `SQLSTATE[HY000]: General error: 1205 Lock wait timeout exceeded; try restarting transaction`
- `An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1020 Record has changed since last read in table 'queuetask'`

meisterT:
To verify whether you are affected, you can run something like:
```
echo 'SHOW VARIABLES LIKE "innodb_snapshot_isolation"' | sudo mysql mysql
Variable_name	Value
innodb_snapshot_isolation	ON
```

tuupke:
The flag causes problems when 2 transactions, update a data has been 'read-locked' (not the actual term). See the following interleaving of events:

1. Start transaction `T1`
2. Select some row `R`, adding it to the read-view of `T1`.
3. Start a second, parallel, transaction `T2`.
4. Update `R` within `T2`.
5. Commit `T2`.
6. Within `T1` update `R`.
7. Depending on the value of `innodb-snapshot-isolation`  this results in:
   `0`: The update query is executed. `R` has not been 'read-locked'.
   `1`: The update query fails. `R` has been 'read-locked'.

[...]

PR #3736, opened 2026-09-02
Create judge tasks for a judging only once.

meisterT:
unblockJudgeTasks() checks for judgings without judge tasks and creates them, but two concurrent callers both pass the check. The second set fails on judging_run's unique key with the "1062 Duplicate entry" from #2848 and leaves judge tasks behind that no judging run points at. Lock the judging and re-check inside.


nickygerritsen (approved):

meisterT:
@tuupke heads-up since you have been removing transactions for #2848: this adds one, but of a different kind.

The ones you removed had the shape that snapshot isolation rejects: plain `SELECT`, someone else commits, then `UPDATE` the same row. This one starts with `SELECT ... FOR UPDATE` on the judging and then only inserts, so the row is locked instead of snapshotted and 'record has changed since last read' cannot happen. It is just a critical section around "check for judge tasks, then create them"; without it two concurrent `unblockJudgeTasks()` calls both insert, the second fails on `judging_run`'s unique key, and orphaned judgetask rows are left for a judgehost to pick up.

If you would rather have no transaction here, a `GET_LOCK` per judging like `calculateScoreRow` should give the same guarantee, but is ugly out of its own right.

PR #3742, opened 2026-09-03
Don't require yarn license generation for in-place installs

meisterT:
Commit 9e3481ae6 made dependencies-dist fail when LICENSES.txt cannot be generated, and pointed inplace-conf-common at dist. That check was silently skipped until 0878e086f fixed the recipe grouping, so maintainer-conf and inplace-conf now fail on systems whose yarn has no licenses command, such as Debian's yarnpkg 4.

These targets already install their own dependencies with the lenient check, so pass LICENSES_OPTIONAL down the in-place path and keep the strict check for release tarballs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -194,7 +194,10 @@ paths.mk:
 # Configure for running in source tree, not meant for normal use:
 maintainer-conf: inplace-conf-common dependencies-dev
 inplace-conf: inplace-conf-common dependencies
-inplace-conf-common: dist
+# In-place installs are not distributions: they install their own (dev)
+# dependencies above, and 'make domserver' builds the default data
+# archives. Only the configure script is needed here.
+inplace-conf-common: configure
 	./configure $(subst 1,-q,$(QUIET)) --prefix=$(CURDIR) \
 	            --with-domserver_root=$(CURDIR) \
 	            --with-judgehost_root=$(CURDIR) \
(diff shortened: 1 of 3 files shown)


vmcj:
Why can't we depend on `node-dependencies`? I ask because this now results in 2 hacks for the same problem and I think I like the first hack a lot more than the one introduced now. I get that we have a problem but this feels like complicating the Makefile logic even more.

meisterT:
Fair point. The in-place targets do already use the lenient `node-dependencies` via `dependencies`/`dependencies-dev`. They only hit the strict one because `inplace-conf-common` depends on `dist`, and it only needs two things from it: `configure` and the default-data zips in `sql/`.

Dropped the variable and instead took `dist` out of the in-place path and added more direct deps.

`dist` stays strict and is only used for actual distributions.

Makefile
```diff
@@ -194,7 +194,10 @@ paths.mk:
 # Configure for running in source tree, not meant for normal use:
 maintainer-conf: inplace-conf-common dependencies-dev
 inplace-conf: inplace-conf-common dependencies
-inplace-conf-common: dist
+# In-place installs are not distributions: they install their own (dev)
```
vmcj, line 197: Did not check fully but:
- This comment is correct for `maintainer-install`, there we want the (dev) dependencies
- This comment doesn't need to be correct for `inplace-install`, in WF/NWERC setting as here we shouldn't need the dev dependencies. Should we discuss making sure we don't install them here?

meisterT: Updated

PR #3744, opened 2026-09-03
Clean build artifacts between configure-check tests.

meisterT:
Commit 9b3abc2c8a split configure-checks into separate containers started from a base image. However, because all containers bind-mount the host workspace ($(pwd)), compiled artifacts leak between steps.

This broke the autoconf check on Debian testing (seen in PR #3742 and the merge queue) following Debian's transition to GCC 16 (more details below):
1. "Check G++" compiles lib/lib.error.o using GCC 16 into $(pwd).
2. "Check Clang" runs next; Clang-21 on Debian testing links against libstdc++-15-dev, but reuses the stale lib.error.o from the host.
3. Linking runguard fails with undefined references to GCC 16's internal C++20 std::format symbols missing in GCC 15 libstdc++.

More debug info, on Debian testing we have since earlier today:
  • Installing g++ gets you GCC 16.
  • Installing clang pulls in libstdc++-15-dev.

This results in the following chain reaction

    [ Container 2: gpp ]
        │
        ├─► apt installs g++ (GCC 16)
        ├─► make judgehost compiles lib/lib.error.cc using GCC 16 headers
        └─► writes lib/lib.error.o to the mounted host directory $(pwd)
             │
             ▼  (lib.error.o remains on the host disk)
    [ Container 3: clang ]
        │
        ├─► apt removes g++ (GCC 16 and libstdc++-16-dev are purged)
        ├─► apt installs clang (pulls in libstdc++-15-dev)
        ├─► make judgehost runs:
        │     - Sees lib/lib.error.o already exists on disk
        │     - Skips compiling lib.error.o!
        │     - Invokes clang++ to link runguard:
        │       clang++ -o runguard runguard.cc lib/lib.error.o ... -lstdc++
        │
        └─► Clang links against GCC 15's libstdc++.so:
              lib.error.o (compiled with GCC 16) asks for GCC 16 <format> symbols:
                - std::basic_format_arg<...>::_M_handle_unrecognized()
                - std::__format::__do_vformat_to<char, 1u>(...)
              GCC 15 libstdc++.so says: "I don't have those symbols."
              💥 LINKER ERROR

### DIFF
--- a/Makefile.global
+++ b/Makefile.global
@@ -66,8 +66,13 @@ LIBOBJECTS = $(addsuffix $(OBJEXT),$(LIBSBASE))
 CFLAGS   += -I$(TOPDIR)/lib -I$(TOPDIR)/etc
 CXXFLAGS += -I$(TOPDIR)/lib -I$(TOPDIR)/etc
 
-$(LIBOBJECTS):
+# Always delegate to lib/Makefile, which knows the real dependencies of
+# these objects; this Makefile does not, so without FORCE an existing
+# object file would never be rebuilt from here.
+$(LIBOBJECTS): FORCE
 	$(MAKE) -C $(TOPDIR)/lib $(notdir $@)
+
+FORCE:
 
 # Default recursive targets; these should always at least call the
 # TARGET-l variant:
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -8,8 +8,10 @@ OBJECTS = $(addsuffix $(OBJEXT),lib.error lib.misc)
 
 build: $(OBJECTS)
 
-lib.error$(OBJEXT): lib.error.cc lib.error.hpp
-lib.misc$(OBJEXT): lib.misc.cc lib.misc.h
+# paths.mk holds the compiler and flags chosen by configure, so a
+# reconfigure with a different compiler must rebuild these.
+lib.error$(OBJEXT): lib.error.cc lib.error.hpp $(TOPDIR)/paths.mk
+lib.misc$(OBJEXT): lib.misc.cc lib.misc.h $(TOPDIR)/paths.mk
 
 clean-l:
 	rm -f $(OBJECTS)


.github/jobs/configure-checks/run_bats.sh
```diff
@@ -5,6 +5,13 @@ set -euxo pipefail
 testsuite="$1"
 export testsuite
 
+# The workspace is bind-mounted into every per-compiler container in this
+# job, so binaries built by an earlier compiler survive into this one. Our
+# build rules only depend on sources, not on the compiler, so make would
+# happily link this suite's compiler against another compiler's leftover
+# object files. Start from a clean build every time to avoid that.
+make clean >/dev/null 2>&1 || true
```
vmcj: shouldn't we do a distclean here so the whole explanation above becomes implied?

meisterT: yes, give me a second, I am validating a better fix just now

vmcj (approved):
I like the solution in general (and would have been good for another PR).

but think the `make distclean` should also be added as we want the configure steps to have a `clean` state.

Approving as this does fix your issue, and it might just be personal preference on my side.

meisterT:
I tried `distclean` first, didn't work. It removes `$(OBJECTS)`, built from `$(OBJEXT)` in `paths.mk`, but the last test of every suit is "Build judgehost disabled", essentially skipping `AC_PROG_CXX` and leaves `OBJEXT = .`.

The next suite runs `rm -f lib.error. lib.misc.` and removes nothing. Verified in the `gcc` --> `g++` --> `clang` order in a docker container: `clang` still fails.

The real bug is that `make` can't see the objects are stale: the `$(LIBOBJECTS)` rule has no prerequisites, so it never recurses into `lib/` once the files exist (editing `lib.error.cc` + `make judgehost` doesn't rebuild today either), and `lib/Makefile` didn't depend on `paths.mk`. With those fixed, each suite's own `./configure` triggers the rebuild. No clean step needed, and no spurious relinks, since `make` still compares mtimes.

Happy to experiment with adding a `distclean` (e.g. as hygiene) or try other approaches in a separate PR.

PR #3747, opened 2026-09-04
Skip CI jobs that the changed files cannot affect.

meisterT:
Every workflow ran on every push, so a manual fix built runguard on three distributions and a judgedaemon fix ran the web accessibility suite, and the merge queue then ran it all again.

Gate the expensive jobs on a shared `paths-filter` job.

This has to be a job-level condition: a job skipped that way reports success and satisfies its required check, whereas a workflow skipped by a paths filter leaves the check pending and blocks the merge.

Since a skip counts as a pass, skip only when detection ran and found no match.

Gemini guesses from previous runs that a webapp-only PR saves about 170 job-minutes.

### DIFF
--- a/.github/paths-filters.yml
+++ b/.github/paths-filters.yml
@@ -1,0 +1,110 @@
+# Which files matter to which CI jobs.
+#
+# Each heavy workflow starts with the reusable job in
+# .github/workflows/changes.yml, which runs dorny/paths-filter with these
+# filters. The expensive jobs then carry
+#
+#     needs: changes
+#     if: ${{ !cancelled() && needs.changes.outputs.<filter> != 'false' }}
+#
+# so a PR that only touches the manual does not build runguard, and a
+# PR that only touches the judgedaemon does not run the web tests. A job
+# skipped this way still counts as a passed required status check.
+#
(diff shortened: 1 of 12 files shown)


vmcj:
We already had https://github.com/DOMjudge/domjudge-packaging/blob/main/.github/actions/helper-detect-changed-files/action.yml, I would have preferred to either remove that one and use the same thing there or to just use that one.

meisterT:
Agreed that we rather should use one vs two mechanisms.

dorny/paths-filter is widely used and documented, supports glob matching properly, e.g. `**/foo` and `foo/**` works, also the filter groups can be expressed centrally.

But I am happy to use `helper-detect-change-files` if you prefer.

vmcj:
Fine with either way, when I build our own I had an explicit reason as `action.yml` from others seemed to always get changed breaking CI in hard to debug ways.

If you believe `dorny/paths-filter` is widely used we can give it a go, it's a bit wasted time on my side but maintaining it for no good reason is also wasted effort.

It looks like this does need another step though, which costs booting another runner (I think) and with the state GHA is in currently I'm not sure if I really like that.

vmcj (approved):

PR #3754, opened 2026-09-18
Only display `scoring` categories in pre-contest scoreboard overview

### LINKED ISSUE #3753: Non-scoring team categories are shown on pre-contest scoreboard
Image

Image

In our contest, the Students team is scoring; the first-year and honours teams are not, but they still appear on the pre-contest scoreboard while I believe they should not appear.

vmcj:
Before we would display any category, also the badges/`CSS`-only categories.

I suggest to use it to close https://github.com/DOMjudge/domjudge/issues/3753


nickygerritsen (approved):

webapp/src/Service/ScoreboardService.php
```diff
@@ -815,9 +815,11 @@ public function getGroupedAffiliations(Contest $contest): array
             ->leftJoin('cat.teams', 't')
             ->leftJoin('t.affiliation', 'affil')
             ->andWhere('cat.visible = 1')
+            ->andWhere('BIT_AND(cat.types, :scoring) = :scoring')
```
meisterT, line 818: perhaps we should also add `t.enabled = 1`?

vmcj: Sure, I'll add it after CI works again and merge it to test for this weekend.

PR #3756, opened 2026-09-19
Add API action to mark balloon as undone

nickygerritsen:
If you make a mistake, you might want to undo your mistake. This allows that.


BHenkemans:
Thanks :))

vmcj (commented):
I agree with the feature but can you (or Claude) also fix it for the UI? They should align IMO.

nickygerritsen:
My Claude license expired so this feature was 100% me!

nickygerritsen:
Also updated the UI now.

vmcj (approved):

PR #3757, opened 2026-09-20
jury: add link to view existing problemset document in edit form

### LINKED ISSUE #3752: Contest problemset PDF not viewable in jury interface
In the jury interface, under the contest edit menu, you can upload a problemset document, but you can't then view it.

This is in DOMjudge 9.0.

Arslan-TR:
### Description

The jury contest edit form (`Files` section) lets you upload or replace the problemset document, but there's no way to view the currently uploaded one from that form - only the read-only contest detail page has a "view contest problemset document" link.

The route and controller action for viewing it already exist and work (`jury_contest_problemset`, used by `contest.html.twig`) - this just adds the same link to the edit form, right next to the upload field, shown only when a problemset document is already present (reusing the existing `form.offsetExists('clearContestProblemset')` check that already gates the "clear" checkbox for the same reason).

### Testing

Pure template change, no PHP/route/controller changes. Verified the Twig syntax parses correctly with the Twig engine directly (`$twig->parse($twig->tokenize(...))`) against the modified file.

Fixes #3752


vmcj:
Hi @Arslan-TR, can you post a screenshot of how it looks like on your side?

